PR #33, opened 2018-11-16
E2E: rationalize node check + kube-system check, no kms

jackfrancis:
**What this PR does / why we need it**: An active "intermittent zero byte `/etc/kubernetes/azure.json` on master nodes issue" reveals that we aren't properly (1) testing for the integrity of this cloudprovider file, and (2) our E2E tests are implemented in a way that masks this.

This PR re-arranges node readiness checks so that they more clearly communicate the state of nodes before testing, and during the E2E run.

We haven't root caused the azure.json issue, but we have repro'd it in isolation w/ KMS encryption enabled, so we're disabling that feature on standard cluster definitions.

**Testing**:
- [x] disabled rescheduler addon
- [x] disabled flexvol addons
- [ ] disabled KMS <-- this is the one :)

**Release note**:

```
E2E: rationalize node check + kube-system check, no kms
```

### DIFF
--- a/examples/e2e-tests/kubernetes/release/default/definition.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition.json
@@ -91,7 +91,7 @@
           },
           {
             "name": "blobfuse-flexvolume",
-            "enabled": true,
+            "enabled": false,
             "containers": [
               {
                 "name": "blobfuse-flexvolume",
@@ -119,7 +119,7 @@
           },
           {
             "name": "keyvault-flexvolume",
-            "enabled": true,
+            "enabled": false,
             "containers": [
               {
                 "name": "keyvault-flexvolume",
--- a/examples/e2e-tests/kubernetes/release/default/definition.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition.json
@@ -9,18 +9,19 @@
           {
             "name": "tiller",
             "enabled": true,
-            "config": {
-              "max-history": "10"
-            },
             "containers": [
               {
                 "name": "tiller",
+                "image": "gcr.io/kubernetes-helm/tiller:v2.8.1",
                 "cpuRequests": "1",
                 "memoryRequests": "1Gi",
                 "cpuLimits": "1",
                 "memoryLimits": "1Gi"
               }
-            ]
+            ],
+            "config": {
+              "max-history": "10"
+            }
           },
           {
             "name": "kubernetes-dashboard",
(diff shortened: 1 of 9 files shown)


test/e2e/config/config.go
```diff
@@ -39,6 +39,7 @@ type Config struct {
 	UseDeployCommand    bool   `envconfig:"USE_DEPLOY_COMMAND"`
 	GinkgoFocus         string `envconfig:"GINKGO_FOCUS"`
 	GinkgoSkip          string `envconfig:"GINKGO_SKIP"`
+	NewCluster          bool   `envconfig:"NEW_CLUSTER" default:"false"` // track when we create a new cluster via E2E
```
CecileRobertMichon: isn't this equivalent to `Name == ""`?

test/e2e/kubernetes/node/node.go
```diff
@@ -74,14 +74,17 @@ type List struct {
 // AreAllReady returns a bool depending on cluster state
 func AreAllReady(nodeCount int) bool {
```
CecileRobertMichon: should this function return "true" if there is an extra not-ready node? If not, we should return false when we find a node in "NotReady" state

test/e2e/config/config.go
CecileRobertMichon: nevermind, this is needed because you are using it after Name is set. I don't think we should make it an envconfig var though since its value will be set in https://github.com/Azure/aks-engine/pull/33/files#diff-881ae33659acba7c8c55e6efbf5edd3dR107 with `os.Setenv("NEW_CLUSTER", "true")`. This isn't a configurable input but a variable determined by other values.

jackfrancis: I'm pretty sure that's the only way that the env config parser works. It has a whitelist of env vars to parse and turn into a typed struct.

CecileRobertMichon: it can follow the same patter as `CurrentWorkingDir`

test/e2e/kubernetes/node/node.go
jackfrancis: This function should only return true when the exact desired number of nodes are all ready (so, an extra not ready means false).

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -142,25 +142,15 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
-					log.Printf("Error while getting nodes: %s\n", err)
-				}
-				if nodeList != nil && len(nodeList.Nodes) == eng.NodeCount() {
-					break
-				} else {
-					log.Printf("Got %d nodes, expected %d\n", len(nodeList.Nodes), eng.NodeCount())
-				}
-				// Wait a minute before proceeding to create a new job w/ the same name
-				time.Sleep(10 * time.Second)
+		It("should all nodeas in a Ready state", func() {
```
CecileRobertMichon: typo

test/e2e/kubernetes/node/node.go
```diff
@@ -74,14 +74,19 @@ type List struct {
...
 	list, _ := Get()
+	var ready int
 	if list != nil && len(list.Nodes) == nodeCount {
 		for _, node := range list.Nodes {
 			for _, condition := range node.Status.Conditions {
-				if condition.Type == "KubeletReady" && condition.Status == "false" {
+				if condition.Type == "Ready" && condition.Status == "True" {
+					ready++
+				} else {
```
CecileRobertMichon: aren't there going to be other conditions? I think this should be `if condition.Type == "Ready" && condition.Status == "False" {`

test/e2e/kubernetes/node/node.go
```diff
@@ -74,14 +74,19 @@ type List struct {
 // AreAllReady returns a bool depending on cluster state
 func AreAllReady(nodeCount int) bool {
```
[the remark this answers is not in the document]
jackfrancis: we'll have to refactor this library to allow for easy mocking of the k8s API, so let's punt on that for now

examples/e2e-tests/kubernetes/release/default/definition.json
```diff
@@ -9,31 +9,214 @@
           {
             "name": "tiller",
             "enabled": true,
-            "config": {
-              "max-history": "10"
-            },
             "containers": [
               {
                 "name": "tiller",
+                "image": "gcr.io/kubernetes-helm/tiller:v2.8.1",
```
tariq1890: Curious. Why are we doing this? Defaults should populate the exact same thing

jackfrancis: I'm testing

test/e2e/engine/template.go
```diff
@@ -150,6 +150,9 @@ func Build(cfg *config.Config, masterSubnetID string, agentSubnetID string, isVM
 	}
 
 	if config.EnableKMSEncryption && config.ClientObjectID != "" {
+		if prop.OrchestratorProfile.KubernetesConfig == nil {
```
jackfrancis: this is a nil panic vector in e2e runner if no `kubernetesConfig` object is present in input api model

jackfrancis:
Filed https://github.com/Azure/aks-engine/issues/49 to mark KMS as buggy

test/e2e/kubernetes/node/node.go
```diff
@@ -74,14 +74,22 @@ type List struct {
...
```
jackfrancis: The purpose of these changes is to make this func more scrupulous: instead of "looking for specific failure conditions and returning true otherwise" we are now "looking for specific success conditions and returning false otherwise".

PR #35, opened 2018-11-17
remove unused azconst methods

tariq1890:
We are removing unneeded methods in the azure const file. Code cleanup.


pkg/helpers/Get-AzureConstants.py
```diff
@@ -61,8 +49,10 @@ def getStorageAccountType(sizeName):
     else:
         return "Standard_LRS"
 
-def getFileContents(masterAgentMap, kubernetesAgentMap, sizeMap, locations):
-    text = r"""package helpers
+def getFileContents(kubernetesSizeMap, locations):
+    text = r"""// Copyright (c) Microsoft Corporation. All rights reserved.
```
CecileRobertMichon: why is there a copyright comment here?

tariq1890: We need it to add it here. SInce `azure_const.go` is an AUTOGENERATED file, the edits have to be done through here

tariq1890:
@mboersma @CecileRobertMichon Thanks for the review. I have updated the PR with a non-breaking changes : New tests added.

PR #62, opened 2018-11-26
scale: persist scale down in api model

jackfrancis:
**What this PR does / why we need it**: The current `aks-engine scale` implementation does not persist changes to the apimodel.json in scale down scenarios. This change fixes that.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```
E2E: node count check is flaky after scale
```

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -56,6 +56,7 @@ const (
 	scaleName             = "scale"
 	scaleShortDescription = "Scale an existing Kubernetes cluster"
 	scaleLongDescription  = "Scale an existing Kubernetes cluster by specifying increasing or decreasing the node count of an agentpool"
+	apiModelFilename      = "apimodel.json"
 )
 
 // NewScaleCmd run a command to upgrade a Kubernetes cluster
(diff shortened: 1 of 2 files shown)


cmd/scale.go
```diff
@@ -375,17 +375,19 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 		addValue(parametersJSON, fmt.Sprintf("%sOffset", sc.agentPool.Name), highestUsedIndex+1)
 	}
 
-	random := rand.New(rand.NewSource(time.Now().UnixNano()))
-	deploymentSuffix := random.Int31()
+	if needsDeployment {
```
CecileRobertMichon: instead of having a bool for the deployment, I think we could probably refactor the "save updated apimodel" part to a function and call that function right after the scale down scenario and right after the scale up scenario.

Main reason why I'm suggesting this is because not saving the apimodel right after scale down means that something could fail before we reach that part of the code and cause us to not save the apimodel even though the scale operation happened + the "save apimodel code" (starting at line 391) doesn't need any of the stuff that runs before it (it just needs to happen after the scale).

PR #80, opened 2018-12-04
Remove OpenShift support

mboersma:

The OpenShift code in aks-engine is unsupported and untested here, and was moved to https://github.com/openshift/openshift-azure. It was removed from aks-engine testing, and this PR completes the removal from this codebase. 

**Which issue this PR fixes**:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
Remove OpenShift support
```


mboersma:
I verified that this is a safe change with regard to AKS vendoring.

Issue #73, opened 2018-11-29
metrics-server is broken in 1.10.10, 1.10.11, 1.12.3, 1.13.0


jackfrancis:
Let's fix this!

jackfrancis:
```
$ kubectl describe hpa
Name:                                                  php-apache-long-running
Namespace:                                             default
Labels:                                                <none>
Annotations:                                           <none>
CreationTimestamp:                                     Wed, 28 Nov 2018 16:36:05 -0800
Reference:                                             Deployment/php-apache-long-running
Metrics:                                               ( current / target )
  resource cpu on pods  (as a percentage of request):  <unknown> / 5%
Min replicas:                                          1
Max replicas:                                          10
Conditions:
  Type           Status  Reason                   Message
  ----           ------  ------                   -------
  AbleToScale    True    SucceededGetScale        the HPA controller was able to get the target's current scale
  ScalingActive  False   FailedGetResourceMetric  the HPA was unable to compute the replica count: unable to get metrics for resource cpu: no metrics returned from resource metrics API
Events:
  Type     Reason                        Age               From                       Message
  ----     ------                        ----              ----                       -------
  Warning  FailedGetResourceMetric       6s (x11 over 5m)  horizontal-pod-autoscaler  unable to get metrics for resource cpu: no metrics returned from resource metrics API
  Warning  FailedComputeMetricsReplicas  6s (x11 over 5m)  horizontal-pod-autoscaler  failed to get cpu utilization: unable to get metrics for resource cpu: no metrics returned from resource metrics API
```

jackfrancis:
```
$ kubectl logs heapster-75f5b5c44c-blv2k -n kube-system -c heapster
I1129 00:31:20.330019       1 heapster.go:78] /heapster --source=kubernetes.summary_api:''
I1129 00:31:20.330201       1 heapster.go:79] Heapster version v1.5.1
I1129 00:31:20.330899       1 configs.go:61] Using Kubernetes client with master "https://10.0.0.1:443" and version v1
I1129 00:31:20.330923       1 configs.go:62] Using kubelet port 10255
I1129 00:31:20.332252       1 heapster.go:202] Starting with Metric Sink
I1129 00:31:20.526184       1 heapster.go:112] Starting heapster on port 8082
W1129 00:32:25.022174       1 manager.go:152] Failed to get all responses in time (got 0/3)
W1129 00:33:25.000493       1 manager.go:152] Failed to get all responses in time (got 0/3)
[...]

PR #82, opened 2018-12-04
Add support for Kubernetes 1.12.4

mboersma:

See https://github.com/kubernetes/kubernetes/releases/tag/v1.12.4

Refs #73.

TODO:

- [x] upload Windows binary

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:
```release-note
Add support for Kubernetes 1.12.4
```

### DIFF
--- a/pkg/api/k8s_versions.go
+++ b/pkg/api/k8s_versions.go
@@ -63,7 +63,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"aci-connector":                    "virtual-kubelet:latest",
 		ContainerMonitoringAddonName:       "oms:ciprod10162018-2",
 		AzureCNINetworkMonitoringAddonName: "networkmonitor:v0.0.4",
-		"cluster-autoscaler":               "cluster-autoscaler:v1.3.3",
+		"cluster-autoscaler":               "cluster-autoscaler:v1.12.1",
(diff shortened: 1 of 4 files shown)


pkg/api/k8s_versions.go
```diff
@@ -30,7 +30,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"aci-connector":                    "virtual-kubelet:latest",
 		ContainerMonitoringAddonName:       "oms:ciprod10162018-2",
 		AzureCNINetworkMonitoringAddonName: "networkmonitor:v0.0.4",
-		"cluster-autoscaler":               "cluster-autoscaler:v1.3.3",
+		"cluster-autoscaler":               "cluster-autoscaler:v1.12.1",
```
jackfrancis: Is there a good reason for including this change in this PR?

mboersma: That seems incorrect then. Should be 1.12.x for 1.12.x and 1.13.x for 1.13... I'll fix it up.

mboersma: I fixed this--thanks for noticing! `cluster-autoscaler` v1.12.1 appears to be the current 1.12.x release, so I left it there.

mboersma: See also #170.

PR #109, opened 2018-12-07
disable heapster for kubernetes versions > 1.13

### LINKED ISSUE #126: Convert heapster to type Addon
The heapster addon implementation is old and doesn't actually use the addons interface

tariq1890:
Fixes #126 

This PR does two things

    Disable Heapster for k8s clusters with version 1.13.0 or greater. As per this document , https://github.com/kubernetes/heapster/blob/master/docs/deprecation.md

    Remove sed statements to apply heapster configurations and use the new addon pre-render flow instead.

### DIFF
--- a/pkg/engine/artifacts.go
+++ b/pkg/engine/artifacts.go
@@ -21,6 +21,12 @@ type kubernetesFeatureSetting struct {
 
 func kubernetesContainerAddonSettingsInit(profile *api.Properties) map[string]kubernetesFeatureSetting {
 	return map[string]kubernetesFeatureSetting{
+		DefaultHeapsterAddonName: {
+			"kubernetesmasteraddons-heapster-deployment.yaml",
+			"kube-heapster-deployment.yaml",
+			!common.IsKubernetesVersionGe(profile.OrchestratorProfile.OrchestratorVersion, "1.12.0"),
+			profile.OrchestratorProfile.KubernetesConfig.GetAddonScript(DefaultKubeHeapsterDeploymentAddonName),
+		},
 		DefaultMetricsServerAddonName: {
 			"kubernetesmasteraddons-metrics-server-deployment.yaml",
 			"kube-metrics-server-deployment.yaml",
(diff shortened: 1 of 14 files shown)


tariq1890:
This is a clone of https://github.com/Azure/acs-engine/pull/4316

jackfrancis:
Let's do an upgrade test of a 1.12 cluster built before this change to make sure there aren't any weird side-effects.

pkg/engine/artifacts.go
```diff
@@ -21,6 +21,12 @@ type kubernetesFeatureSetting struct {
 
 func kubernetesContainerAddonSettingsInit(profile *api.Properties) map[string]kubernetesFeatureSetting {
 	return map[string]kubernetesFeatureSetting{
+		DefaultHeapsterAddonName: {
+			"kubernetesmasteraddons-heapster-deployment.yaml",
+			"kube-heapster-deployment.yaml",
+			!common.IsKubernetesVersionGe(profile.OrchestratorProfile.OrchestratorVersion, "1.12.0"),
```
jackfrancis: let's change this to 1.13.0, so that we don't change the presence of heapster in pre-existing 1.12 clusters

jackfrancis:
Do we have a unit test that proves that a 1.13 generation flow won't include the heapster addon by default?

tariq1890:
No. Right now this PR stops using heapster for 1.12.* onwards. Once the
1.13.1 version is merged I can definitely add a unit test for this :).

jackfrancis:
@tariq1890 That seems like a good plan. See my other comment, let's make this change for >= 1.13 clusters only and leave 1.12 alone.

pkg/engine/artifacts.go
tariq1890: Done

tariq1890:
@jackfrancis I just checked the coverage and the `engine_test.go` fully covers all of these lines. The 1.13 release e2e passing is another reinforcement of confidence that this change works.

This PR also increases overall code coverage :)

PR #135, opened 2018-12-13
feat: Azure DevOps PR E2E pipeline

CecileRobertMichon:
**What this PR does / why we need it**: Add E2E with Azure DevOps pipeline to eventually replace circle CI.

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


.pipelines/e2e-job-template.yaml
```diff
@@ -0,0 +1,37 @@
...
+    modulePath: '$(GOPATH)/src/github.com/$(build.repository.name)' # Path to the module's code
+    TIMEOUT: 20m
+    CLUSTER_DEFINITION: ${{ parameters.apimodel }}
+    ORCHESTRATOR_RELEASE: ${{ parameters.k8sRelease }}
+    CREATE_VNET: ${{ parameters.createVNET }}
+    CLEANUP_ON_EXIT: true
+    CLEANUP_IF_FAIL: true
+    RETAIN_SSH: false
+    ENABLE_KMS_ENCRYPTION: true
+    SUBSCRIPTION_ID: '$(SUBSCRIPTION_ID_E2E_KUBERNETES)'
```
tariq1890: Lets rename this to AZURE_SUBSCRIPTION_ID. This seems to be the env naming convention

.pipelines/e2e-step-template.yaml
```diff
@@ -0,0 +1,27 @@
...
+      shopt -s extglob
+      mv !(gopath) '$(modulePath)'
+      echo '##vso[task.prependpath]$(GOBIN)'
+      echo '##vso[task.prependpath]$(GOROOT)/bin'
+    displayName: 'Set up the Go workspace'
+  - script: make build-binary
+    displayName: Compile
+    workingDirectory: $(modulePath)
+  - script: |
+      export CLIENT_ID=$(SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES)
```
tariq1890: AZURE_CLIENT_ID?

.pipelines/e2e-job-template.yaml
CecileRobertMichon: This is defined in by our e2e tests https://github.com/Azure/aks-engine/blob/master/test/e2e/azure/cli.go#L35 and is the same as circle uses

.pipelines/e2e-step-template.yaml
CecileRobertMichon: Same here: this is defined in by our e2e tests https://github.com/Azure/aks-engine/blob/master/test/e2e/azure/cli.go#L35 and is the same as circle uses

.pipelines/e2e-job-template.yaml
tariq1890: Then let's do away this `SUBSCRIPTION_ID_E2E_KUBERNETES`. I assume that variable is being declared in vsts internally. Let's do that internal declaration for `SUBSCRIPTION_ID` itself.

CecileRobertMichon: Sure. I just copied that over from circle. I will still need to declare the var here for it to be available in the container job. I assume the idea behind e2e_kubernetes was that we have different variables subscriptions declared in circle env vars. For now this is the only one in vsts though so I'm okay with changing it.

tariq1890: Actually makes sense. We can have this PR as is for now

CecileRobertMichon:
installing sudo is failing as described in https://github.com/dotnet/dotnet-docker/issues/788... I think we need sudo in the image, or get rid of install kubectl altogether and put that in the image.

.pipelines/pr-e2e.yaml
```diff
@@ -0,0 +1,120 @@
...
+    workingDirectory: $(modulePath)
+
+- template: e2e-job-template.yaml
+  parameters:
+    name: 'default_linux_e2e'
+    apimodel: 'examples/kubernetes.json'
+
+- template: e2e-job-template.yaml
+  parameters:
+    name: 'k8s_1dot10_release_e2e'
```
tariq1890: nit: can we have an underscore instead of `dot`? More readable that way IMO

jackfrancis: If we don't like "dot" I vote for a period (`.`)

CecileRobertMichon: period isn't a valid char in the name (that was my first choice too)

I can change it to underscore

tariq1890:
```suggestion
     name: 'k8s_1_10_release_e2e'
```

CecileRobertMichon: why the extra space?

tariq1890: Sorry. Edited the suggestion

PR #138, opened 2018-12-13
add support for kubernetes 1.13.1

tariq1890:
This PR should add support for kubernetes 1.13.1

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -92,8 +92,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.12.0":         false,
 	"1.12.1":         true,
 	"1.12.2":         true,
-	"1.13.0-alpha.1": true,
+	"1.13.0-alpha.1": false,
 	"1.13.0-alpha.2": true,
+	"1.13.1":         true,
 }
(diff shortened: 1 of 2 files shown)


pkg/api/common/versions.go
```diff
@@ -92,8 +92,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.12.0":         false,
 	"1.12.1":         true,
 	"1.12.2":         true,
-	"1.13.0-alpha.1": true,
+	"1.13.0-alpha.1": false,
 	"1.13.0-alpha.2": true,
```
mboersma: I think we can mark this alpha release `false` too, because there's an official release. (We also tried the -beta and -rc and 1.13.0 builds but no joy.) 

I wonder if we should have `"1.13.0": false` here as documentation that aks-engine knows about it but it's not supported?

CecileRobertMichon: I agree with Matt we should mark pre-releases as false now that we have a supported release.

Adding `"1.13.0"` would imply that it was supported at some point in time which it wasn't but also it wouldn't have any functional impact to add it so I'm okay with either way.

PR #128, opened 2018-12-10
Add HSM support for kms encryption

ritazh:
**What this PR does / why we need it**:

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes https://github.com/Azure/acs-engine/issues/4152

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
Add HSM support for kms encryption, create Premium SKU Key Vault and RSA-HSM type key.
```

PR #139, opened 2018-12-13
Revert "Add hsm support for kms plugin (#128)"

tariq1890:
This reverts #128 since the jenkins standardLB e2e tests were failing


CecileRobertMichon:
running standardLB test on this

PR #149, opened 2018-12-14
update Calico network policy examples docs

briansan:
updates documentation around how to configure Calico policy on the acs-engine deploy


CecileRobertMichon:
Thanks @briansan! I think you have some extra diff in the readme from the acs-engine/aks-engine readme change

briansan:
nothing like a sloppy copy pasta to close off the week hahaha...
fixed!

PR #157, opened 2018-12-17
Add HSM support for kms encryption

ritazh:

**If applicable**:
- [ ] documentation
- [ ] unit tests

**Release note**:

```release-note
```


tariq1890:
@ritazh Can we add one pre-pull scripts for the kms docker image in the packer shell scripts?

Here: https://github.com/Azure/aks-engine/blob/b8f75fdff2e3d9b48e5480acc27a585efae0fed5/packer/install-dependencies.sh

ritazh:
@tariq1890 the `microsoft/k8s-azure-kms:v0.0.7` image is being pulled by the azure kms systemd service, not as part of the install-dependencies.sh script.
https://github.com/Azure/aks-engine/blob/e98468ea0c8c4a6f03ed90b18ca0c4518123b18b/parts/k8s/artifacts/kubernetesazurekms.service#L16

So I don't think this image should be added there.

tariq1890:
That's for the VHD so that the images are pre-pulled. Any VM that is created off of the VHD will already have that docker image in its local registry.

PR #162, opened 2018-12-18
cleanup pointer primitive usages

tariq1890:
Cleanup of primitive pointers. This PR imports existing functions instead of using the helper methods we had defined.


jackfrancis:
This is easy to unit test, and so we should include our own unit tests against this downstream library to make sure that the behavior we depend up on stays consistent over time.

tariq1890:
@jackfrancis Great idea. I will write the same unit tests as the ones we have for the `helpers` methods. That should give us an idea

tariq1890:
e2e's have passed !

PR #202, opened 2018-12-26
fix: add a ClusterName property and use it for resources names

### LINKED ISSUE #56: Enable user-configurable vm names
Details // TODO

Some customers wish to be able to influence the names of their node vms. Scale/upgrade would have to be taken into account in order to deliver this.

sylr:

Rebase a branch we used at my company to fix the name of the resources.

It adds a property `clusterName` which is used to prefix the name of all Azure resources, e.g. : `clusterName = k8s-prod` -> `k8s-prod-master-0, k8s-prod-master-1 ... etc`

If no  `clusterName` is provided the prefix of the resources falls back to `(k8s|aks)-<clusterId>`.

**Which issue this PR fixes** : fixes #56

**Release note**:

```release-note
Add a clusterName property which is used to prefix all Azure resources. 
```

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -56,6 +56,7 @@ type ContainerService struct {
 // Properties represents the AKS cluster definition
 type Properties struct {
 	ClusterID               string
+	ClusterName             string                   `json:"clusterName,omitempty"`
 	ProvisioningState       ProvisioningState        `json:"provisioningState,omitempty"`
 	OrchestratorProfile     *OrchestratorProfile     `json:"orchestratorProfile,omitempty"`
 	MasterProfile           *MasterProfile           `json:"masterProfile,omitempty"`
@@ -753,6 +754,16 @@ func (p *Properties) getAgentPoolIndexByName(name string) int {
 	return index
 }
 
+// GetClusterName returns resource prefix based on Properties
+func (p *Properties) GetClusterName() (ret string) {
+	if len(p.ClusterName) > 0 {
+		ret = p.ClusterName
+	} else {
+		ret = p.K8sOrchestratorName() + "-" + p.GetClusterID()
+	}
+	return
+}
+
 // GetAgentVMPrefix returns the VM prefix for an agentpool
 func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile) string {
 	index := p.getAgentPoolIndexByName(a.Name)
@@ -762,7 +773,7 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile) string {
 		if a.IsWindows() {
 			vmPrefix = nameSuffix[:4] + p.K8sOrchestratorName() + fmt.Sprintf("%02d", index)
 		} else {
-			vmPrefix = p.K8sOrchestratorName() + "-" + a.Name + "-" + nameSuffix + "-"
+			vmPrefix = p.GetClusterName() + "-" + a.Name + "-"
 			if a.IsVirtualMachineScaleSets() {
 				vmPrefix += "vmss"
 			}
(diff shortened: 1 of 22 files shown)


sylr:
Yes. I rebased this branch onto #202 to be able to run tests.

sylr:
@tariq1890 can you approve the e2e checks on circleci, I'll remove the goimports patch tomorrow.

pkg/helpers/helpers.go
```diff
@@ -140,7 +140,7 @@ func AcceleratedNetworkingSupported(sku string) bool {
 		"Standard_D64_v3", "Standard_D64s_v3", "Standard_E64_v3", "Standard_E64s_v3", "Standard_E64-16s_v3",
 		"Standard_E64-32s_v3", "Standard_F64s_v2", "Standard_F72s_v2", "Standard_M128s", "Standard_M128ms",
 		"Standard_L8s_v2", "Standard_L16s_v2", "Standard_L32s_v2", "Standard_L64s_v2", "Standard_L96s_v2",
-		"SQLGL", "SQLGLCore", "Standard_D4_v3", "Standard_D4s_v3", "Standard_D2_v2", "Standard_DS2_v2",
+		"SQLGL", "SQLGLCore", "Standard_D4_v3", "Standard_D4s_v3", "Standard_D2s", "Standard_D2_v2", "Standard_DS2_v2", "Standard_D2s_v3",
```
tariq1890: Why have we added this?

.circleci/config.yml
```diff
@@ -64,6 +64,8 @@ jobs:
           no_output_timeout: "30m"
       - store_artifacts:
           path: /go/src/github.com/Azure/aks-engine/_logs
+      - store_artifacts:
```
tariq1890: Why do we need this?  We had removed this as a part of this PR - https://github.com/Azure/acs-engine/pull/4214

pkg/api/types.go
```diff
@@ -56,6 +56,7 @@ type ContainerService struct {
 // Properties represents the AKS cluster definition
 type Properties struct {
 	ClusterID               string
+	Name                    string                   `json:"clusterName,omitempty"`
```
tariq1890: This should be `ClusterName`

pkg/api/types.go
```diff
@@ -753,6 +754,16 @@ func (p *Properties) getAgentPoolIndexByName(name string) int {
 	return index
 }
 
+// GetPrefix returns resource prefix based on Properties
+func (p *Properties) GetPrefix() (ret string) {
```
tariq1890: GetPrefix is too vague, let's have something specific. `GetClusterPrefix` would work IMO. Also open to other names :)

tariq1890:
@sylr Done!

pkg/helpers/helpers.go
sylr: That's a mistake

.circleci/config.yml
sylr: I added this for debug purposes.

Why did you remove it ?

pkg/api/types.go
sylr: All right

pkg/api/types.go
sylr: `GetClusterName` ?

test/e2e/azure/cli.go
```diff
@@ -212,7 +211,8 @@ func (a *Account) CreateDeployment(name string, e *engine.Engine) error {
 		"--name", d.Name,
 		"--resource-group", a.ResourceGroup.Name,
 		"--template-file", e.Config.GeneratedTemplatePath,
-		"--parameters", e.Config.GeneratedParametersPath)
+		"--parameters", e.Config.GeneratedParametersPath,
```
tariq1890: Assuming these were added for debugging and not meant to be included in the PR.

jackfrancis:
@sylr Can you kindly add a detailed description of the changes in the PR?

jackfrancis:
Thanks for this @sylr 

The change as-is is breaking. E.g.:

Before:

`"masterVMNamePrefix": "k8s-master-63250626-",`

After:

`"masterVMNamePrefix": "k8s-63250626-master-",`

We should make as a requirement of this change that it be non-breaking when the `clusterName` property is not used.

sylr:
@jackfrancis I disagree on this.

We shouldn't have resources with different name patterns.

.circleci/config.yml
CecileRobertMichon: let's revert adding this back - circle ci is not used anymore

sylr: ok

jackfrancis:
@sylr I'm not sure what you mean. The important thing, to preserve backwards-compatibility, is that any new change like this not effect previous usage patterns. In this specific case, we need to make sure that the vm name generation implementation produces the same output as it did before when this new feature is not enabled.

PR #204, opened 2018-12-26
fix: address goimports checks

sylr:
```
/home/sylvain/gopath/bin/goimports -w -l -srcdir ~/gopath/src/github.com/Azure/aks-engine main.go $(find . -type d ! -path ./vendor\* ! -path ./.git\* ! -path .)
```


tariq1890:
@sylr Thanks for this PR. Since `goimports` has changed significantly in how it performs it's linting, we would need to discuss with the rest of the aks-engine members if we want to go forward with this approach. We'll  leave this PR open for the time being.

mboersma:
@sylr thanks so much for taking the initiative to fix this! I think we've worked around it in #208 by updating `golangci-lint` instead, so I'm going to close this PR.

tariq1890:
@sylr Can you rebase this ?

sylr:
Ok

pkg/api/agentPoolOnlyApi/v20180331/apiloader_test.go
```diff
@@ -5,7 +5,7 @@ package v20180331_test
 
 import (
 	"github.com/Azure/aks-engine/pkg/api"
-	"github.com/Azure/aks-engine/pkg/api/agentPoolOnlyApi/v20180331"
+	v20180331 "github.com/Azure/aks-engine/pkg/api/agentPoolOnlyApi/v20180331"
```
mboersma: The rest of the changes make sense, but I don't understand why this type of change is necessary. The `v20180331` package was already imported and used, so why do we need to pin it this way?

mboersma: We're assuming it's for readability. It makes `goimports` happy, so 🤷‍♂️.

PR #218, opened 2019-01-03
feat: add pod disruption budgets

sylr:
**Reason for Change**:

This allows the cluster-autoscaler to evict the pods of the concerned deployments.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**: Add pod disruption budgets

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml
@@ -252,6 +252,19 @@ spec:
 
 ---
 
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: calico-typha
+  namespace: kube-system
+spec:
+  minAvailable: 0%
+  selector:
+    matchLabels:
+      k8s-app: calico-typha
+
+---
+
 # Typha Horizontal Autoscaler Cluster Role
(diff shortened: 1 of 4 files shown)


jackfrancis:
@sylr Thanks for this! FYI I rebased/force pushed to enable our new E2E pipeline to run.

parts/k8s/addons/kubernetesmasteraddons-calico-daemonset.yaml
```diff
@@ -252,6 +252,19 @@ spec:
 
 ---
 
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: calico-typha
+  namespace: kube-system
+spec:
+  minAvailable: 0%
```
jackfrancis: This sort of seems like a no-op configuration, but I infer that cluster-autoscaler needs *some* value here in order to engage?

sylr: Absolutely. The cluster-autoscaler will refuse to engage in a scale down if pods are non-mirrored and does not have a PDB that allows the pod to be killed.

jackfrancis:
@mboersma, could you spend a small amount of time reading up on the PodDisruptionBudget spec and sanity check the functional intent here. LGTM but I'd like a 2nd opinion. :)

PR #219, opened 2019-01-03
feat: support `make build` on Windows

bacongobbler:
This commit makes it such that `make build` will compile aks-engine.exe
and aks-engine-test.exe for testing on Windows devices.

Signed-off-by: Matthew Fisher <matt.fisher@microsoft.com>

**Reason for Change**: This allows me to start contributing to the aks-engine project from a Windows device. I have not actually tested whether it'll work on Windows yet. :)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**: 👋

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ GO              ?= go
 TAGS            :=
 LDFLAGS         :=
 BINDIR          := $(CURDIR)/bin
-BINARIES        := aks-engine
+PROJECT         := aks-engine
 VERSION         ?= $(shell git rev-parse HEAD)
 VERSION_SHORT   ?= $(shell git rev-parse --short HEAD)
 GITTAG          := $(shell git describe --exact-match --tags $(shell git log -n1 --pretty='%h') 2> /dev/null)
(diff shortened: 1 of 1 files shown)


mboersma:
Thanks @bacongobbler. I'll test this out on my Windows box.

mboersma:
This is better! But I'm running into a problem installing golangci-lint:

```
C:\Users\mboersma\go\src\github.com\Azure\aks-engine>make build
The system cannot find the path specified.
The system cannot find the path specified.
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
INFO: Could not find files for the given pattern(s).
'test' is not recognized as an internal or external command,
operable program or batch file.
The system cannot find the path specified.
The system cannot find the path specified.
go get -u github.com/golang/dep/cmd/dep
go get -u github.com/mitchellh/gox
go install ./vendor/github.com/go-bindata/go-bindata/...
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b C:\Users\mboersma\go/bin
'sh' is not recognized as an internal or external command,
operable program or batch file.
make: *** [Makefile:140: bootstrap] Error 255
```

@bacongobbler did you find a way around that curl | bash command?

bacongobbler:
Nah, we haven't found a good way to work around that in Windows yet. The change to CHECK just allows Windows devices to check for the presence of that binary.

Exec'ing out to PowerShell and using the .net web client may work, but that assumes that

A) Internet Explorer (not MS Edge) has been opened and enabled
B) it relies on IE's cache for downloading, so if you download the same object twice it fails.

So we haven't really found a great alternative to `curl` for Windows in other projects (like [Duffle](https://github.com/deislabs/duffle/blob/2af5fe49c2ab12670e3ce140f120aa57ec27601a/Makefile#L86-L92), where most of this code comes from)

PR #220, opened 2019-01-03
test: more tolerance for kube-system pod startup in E2E

jackfrancis:
Increases the tolerance for pod startup thrashing (i.e., ratio of Ready/NotReady count) during cluster startup in kube-system namespace. Also standardizes wait retry to 1 second for quicker reconciliation.

**Reason for Change**:

This intends to eliminate some E2E flakiness around kube-system pod readiness checks. E.g.:

```
"Pods from deployment (kube-addon-manager) in namespace (kube-system) have been checked out as all Ready 2 times, but NotReady 3 times. This behavior may mean it is in a crashloop"
```

By increasing the Ready count requirement to 6, we also tolerate more NotReady counts. See this code snippet for how this is currently implemented:

https://github.com/Azure/aks-engine/blob/bc250139f3f08ce7e63e42691c68282535bae216/test/e2e/kubernetes/pod/pod.go#L457

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -160,11 +160,11 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			var running bool
 			if common.IsKubernetesVersionGe(eng.ExpandedDefinition.Properties.OrchestratorProfile.OrchestratorVersion, "1.12.0") {
 				By("Ensuring that coredns is running")
-				running, err = pod.WaitOnReady("coredns", "kube-system", 3, 1*time.Second, cfg.Timeout)
+				running, err = pod.WaitOnReady("coredns", "kube-system", 6, 1*time.Second, cfg.Timeout)
 
 			} else {
 				By("Ensuring that kube-dns is running")
-				running, err = pod.WaitOnReady("kube-dns", "kube-system", 3, 1*time.Second, cfg.Timeout)
+				running, err = pod.WaitOnReady("kube-dns", "kube-system", 6, 1*time.Second, cfg.Timeout)
 			}
 			Expect(err).NotTo(HaveOccurred())
 			Expect(running).To(Equal(true))
@@ -177,7 +177,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 			for _, componentName := range coreComponents {
 				By(fmt.Sprintf("Ensuring that %s is Running", componentName))
-				running, err := pod.WaitOnReady(componentName, "kube-system", 3, 1*time.Second, cfg.Timeout)
+				running, err := pod.WaitOnReady(componentName, "kube-system", 6, 1*time.Second, cfg.Timeout)
 				Expect(err).NotTo(HaveOccurred())
 				Expect(running).To(Equal(true))
 			}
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -256,7 +257,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				Expect(err).NotTo(HaveOccurred())
 				Expect(pass).To(BeTrue())
 				By("Validating the omsagent daemonset")
-				running, err = pod.WaitOnReady("omsagent", "kube-system", 3, 1*time.Second, cfg.Timeout)
+				running, err = pod.WaitOnReady("omsagent", "kube-system", kubeSystemPodsReadinessChecks, 1*time.Second, cfg.Timeout)
```
CecileRobertMichon: all pod checks seem to be hardcoded to 3 so can we change kubeSystemPodsReadinessChecks to defaultPodsReadinessChecks and use it in all occurrences or make a separate const for readiness checks for pods in the default namespace?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -568,7 +569,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				curlDeploymentName := fmt.Sprintf("ilb-test-deployment-%s", cfg.Name)
 				curlDeploy, err := deployment.CreateLinuxDeployIfNotExist("library/nginx:latest", curlDeploymentName, "default", "")
 				Expect(err).NotTo(HaveOccurred())
-				running, err := pod.WaitOnReady(curlDeploymentName, "default", 3, 30*time.Second, cfg.Timeout)
+				running, err := pod.WaitOnReady(curlDeploymentName, "default", 3, 1*time.Second, cfg.Timeout)
```
CecileRobertMichon: why change the sleep time for these? should it also be a const?

test/e2e/kubernetes/pod/pod.go
```diff
@@ -538,7 +538,7 @@ func WaitOnSucceeded(podPrefix, namespace string, sleep, duration time.Duration)
 
 // WaitOnReady will call the static method WaitOnReady passing in p.Metadata.Name and p.Metadata.Namespace
 func (p *Pod) WaitOnReady(sleep, duration time.Duration) (bool, error) {
-	return WaitOnReady(p.Metadata.Name, p.Metadata.Namespace, 2, sleep, duration)
+	return WaitOnReady(p.Metadata.Name, p.Metadata.Namespace, 6, sleep, duration)
```
CecileRobertMichon: is this 6 another "default" for PodsReadinessChecks? It's confusing that some places call `pod.WaitOnReady(5*time.Second, 3*time.Minute)` while others call `pod.WaitOnReady(windowsDeploymentName, "default", 3, 30*time.Second, cfg.Timeout)` where one `pod` is the pkg while the other is the var of type *Pod

CecileRobertMichon: not related to this PR, I just read the code to understand what the 6 was doing and found it confusing

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: I'm only touching `kube-system` pods intentionally, as those pods have a higher likelihood of being intertwined at boot time (i.e., until a quorum of foundational componentry is up, certain other pods may not be able to come online). In other words, it's "normal" for kube-system pods to have a few restarts during initial cluster creation, and so we should be more tolerant of that normal behavior. We observe this in test flakes.

The reason not to increase everything to a higher number is that, arguably, it would slow down the test unnecessarily (as you increase the number of required readiness checks you increase the time it takes to validate that particular pod). Also, for "on-demand" workloads (e.g., nginx), we arguably want to be less tolerant if a pod is thrashing.

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Should've mentioned this in the description. There's no reason to wait so long in between "check again for ready" tests. So, basically, while I was auditiong the "wait on ready" checks in the E2E code, I came across this and rationalized it.

Basically, decreasing the retry interval should speed up test runs. (The timeout until failure is untouched, so the only practical change is more kubectl commands against the apiserver).

test/e2e/kubernetes/pod/pod.go
jackfrancis: It is confusing. Tech debt that I didn't want to address here, and for the cases that include the readiness checks as a static "black box" value, I thought it was best to more tolerant vs less tolerant.

PR #221, opened 2019-01-03
test: only check top nodes if RBAC is enabled

jackfrancis:
Only checks `kubectl top nodes` if RBAC is enabled. Also includes the convenience addition of a `IsRBACEnabled` method.

**Reason for Change**:

If RBAC is not enabled on the cluster, kubectl top nodes checks are likely to fail. We first observed this yesterday when our tests upgrade to version 1.10 of kubectl.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-controller-manager.go
+++ b/pkg/api/defaults-controller-manager.go
@@ -73,7 +73,7 @@ func (cs *ContainerService) setControllerManagerConfig() {
 		o.KubernetesConfig.ControllerManagerConfig[key] = val
 	}
 
-	if *o.KubernetesConfig.EnableRbac {
+	if o.KubernetesConfig.IsRBACEnabled() {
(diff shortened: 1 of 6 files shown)


CecileRobertMichon:
> If RBAC is not enabled on the cluster, kubectl top nodes checks are likely to fail

do we know why? 

code /lgtm, not sure I understand why we want to stop testing this though. Is the flake expected/normal or could it indicate a transient issue with RBAC clusters?

jackfrancis:
@CecileRobertMichon It is consistently broken since this went in yesterday:

https://github.com/Azure/aks-engine/commit/bc250139f3f08ce7e63e42691c68282535bae216#diff-1d37e48f9ceff6d8030570cd36286a61R6

I assume that the 1.10 kubectl version requires RBAC to connect to metrics server.

PR #224, opened 2019-01-03
chore: remove stale scripts

bacongobbler:
**Reason for Change**: These scripts/Dockerfiles are either unused by CI or are not actively being maintained, so I assume these are safe to remove.

I'm not totally sure about removing the root Dockerfile, but given that there are several outdated packages in there I assumed it was also stale. Please let me know if that is not and I'm happy to restore anything that is still in active use.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


scripts/devenv.ps1
```diff
@@ -1,7 +0,0 @@
-$pwd = (Get-Location).Path
-
-docker build --pull -t aks-engine .
```
CecileRobertMichon: I think some people might still be using this one (even though `make dev` is equivalent)

bacongobbler: Ah right you are. I'll restore this one.

jackfrancis:
Do we want to add a follow up item to standardize upon a canonical docker image for aks-engine development/CI?

bacongobbler:
Yeah that may be a nice-to-have. I'll file a ticket for that now

mboersma:
@jackfrancis you mean other than deis/go-dev? I think we use that consistently now that this cruft is removed, but we could consider alternatives.

jackfrancis:
@mboersma Not concretedly "instead of", as possibly "deis/go-dev" *is* the appropriate docker environment for aks-engine dev long-term. I think we just want to standardize that, with community input.

bingosummer:
This PR breaks the doc: https://github.com/Azure/aks-engine/blob/master/docs/tutorials/quickstart.md#docker-development-environment. Is it expected?

PR #226, opened 2019-01-03
feat: UX refresh spec via README!

### LINKED ISSUE #1: Newer, better "deploy" CLI implementation
Requirements:

- deprecate the existing `deploy` CLI verb
  - we retain the verb indefinitely, but we mark it as deprecated
- add a new CLI verb handler to implement cluster creation

jackfrancis:
This is meant to be an aspirational README-expressed specification of a UX level-up. In scope:

- Rationalized session configuration (e.g., `~/.aks-engine` with default Azure and aks-engine CLI context)
- A new "create cluster" CLI verb with an entirely refreshed, cleaner implementation
  - i.e., we will deprecate `deploy`
- A refreshed `scale` verb implementation (mainly because it would be semantically regressive to scale a cluster via the CLI using any other verb, IMO)
- A new "update cluster" CLI verb, inclusive of k8s version upgrade
  - i.e., we wil deprecate `upgrade`

**Reason for Change**:

**Issue Fixed**:

Fixes #1 

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ $ aks-engine update -v 1.11
 
 Cluster upgrade time will take between 5-20 minutes *per node* (including master node(s) running the control plane), due to cordon/drain, the addition of new vms with the desired changes, and the deletion of vms with the previous configuration. And it does this according to a rolling, one-at-a-time strategy to minimize operational side-effects. Depending on the size of your cluster, brew some coffee in the French press style, walk your dog(s), or go do that open source contribution you've been putting off for a few months.
 
-...
+```
 $ kubectl get nodes
 NAME                                           STATUS    ROLES     AGE       VERSION
 k8s-Standard_D2_v2-10678025-vmss000005         Ready     agent     18s       v1.11.6
--- a/README.md
+++ b/README.md
@@ -8,21 +8,84 @@
 
 AKS-Engine leverages ARM (Azure Resource Manager) to build Kubernetes IaaS in Azure. AKS-Engine provides convenient tooling to quickly bootstrap clusters, and implements cluster provisioning and lifecycle operations for [AKS](https://docs.microsoft.com/en-us/azure/aks), Azure's managed Kubernetes service offering.
 
-More info, including a thorough walkthrough is [here](docs/aksengine.md).
+## Install AKS-Engine
+
+To install via homebrew, run `brew tap azure/aks-engine && brew install azure/aks-engine/aks-engine` or download the binary via the [github releases page](https://github.com/Azure/aks-engine/releases)
+
+## Getting Started
+
+Building a Kubernetes cluster on Azure using AKS-Engine is as simple as:
+
+1. `aks-engine init` to set up aks-engine (after the binary has been installed in your `$PATH`)
+  - This will walk you through setting default configurations
+    - Azure subscription context, including the service principal configuration that aks-engine will use to scaffold new IaaS on behalf of your Azure subscription
+    - Preferred region(s)
+    - Preferred default cluster configuration
+2. `aks-engine create` to create a new cluster using the default Kubernetes configuration.
+
+Cluster creation time will take between 3-5 minutes, after which your console session's `KUBECONFIG` context will be automatically set to your new cluster.
+
+```
+$ kubectl get nodes
+NAME                                           STATUS    ROLES     AGE       VERSION
+k8s-Standard_D2_v2-10678025-vmss000000         Ready     agent     18s       v1.10.12
+k8s-Standard_D2_v2-10678025-vmss000001         Ready     agent     17s       v1.10.12
+k8s-Standard_D2_v2-10678025-vmss000002         Ready     agent     17s       v1.10.12
+k8s-master-10678025-0                          Ready     master    11s       v1.10.12
+```
+
+Scale out your cluster:
+
+```
+$ aks-engine scale -n 5
+```
+
+Scale out will similarly take between 3-5 minutes.
+
+```
+$ kubectl get nodes
+NAME                                           STATUS    ROLES     AGE       VERSION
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -8,21 +8,84 @@
 
 AKS-Engine leverages ARM (Azure Resource Manager) to build Kubernetes IaaS in Azure. AKS-Engine provides convenient tooling to quickly bootstrap clusters, and implements cluster provisioning and lifecycle operations for [AKS](https://docs.microsoft.com/en-us/azure/aks), Azure's managed Kubernetes service offering.
 
-More info, including a thorough walkthrough is [here](docs/aksengine.md).
+## Install AKS-Engine
 
-Please see the [FAQ](/docs/faq.md) for answers about AKS-Engine and its progenitor ACS-Engine.
+To install via homebrew, run `brew tap azure/aks-engine && brew install azure/aks-engine/aks-engine` or download the binary via the [github releases page](https://github.com/Azure/aks-engine/releases)
```
bacongobbler: Is there a ticket out there to create this homebrew tap? https://github.com/Azure/homebrew-aks-engine 404s for me, so I assume this is an action item?

bacongobbler:
I noticed that all of the listed commands don't take a cluster name as an argument. Is the idea that there can only be one aks-engine cluster created and managed at a time?

README.md
jackfrancis: Yeah, everything here is aspirational! Let's bikeshed the aspirations, and then when we have a good idea of where we want things to land, then we can start tracking TODOs.

jackfrancis:
@bacongobbler My idea is the following:

- `~/.aks-engine/` will keep track of the "current" cluster context
  - (also, the above will be authoritative for `aks-engine` setting the `KUBECONFIG` context for `kubectl` commands in the same session)
- for CLI commands that require a cluster context, if the cluster context is ommitted from the relevant command options, then it will inherit the cluster context from <whatever> @ `~/.aks-engine`

Does that make sense? (Not saying everyone agrees with me, but that's my thinking...)

README.md
CecileRobertMichon: https://github.com/Azure/aks-engine/issues/227

CecileRobertMichon: we also have [dockerhub](https://hub.docker.com/r/microsoft/acs-engine/) and [gofish](https://github.com/fishworks/fish-food/blob/master/Food/acs-engine.lua) as less-than-ideally maintained current options for installing/using acs-engine, if we want to keep those around for aks-engine we should make them automated and more officially supported.

README.md
```diff
@@ -8,21 +8,84 @@
...
 
-## User guides
+## Getting Started
+
+Building a Kubernetes cluster on Azure using AKS-Engine is as simple as:
+
+1. `aks-engine init` to set up aks-engine (after the binary has been installed in your `$PATH`)
```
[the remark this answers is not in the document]
bacongobbler: One option as well is that we could have `aks-engine init` run any time a command is run, like we do with `duffle init`: https://github.com/deislabs/duffle/blob/2af5fe49c2ab12670e3ce140f120aa57ec27601a/cmd/duffle/root.go#L30-L34

That way a user doesn't have to run `duffle init` the first time they use aks-engine; it'll start self-initializing if you forget to do that yourself.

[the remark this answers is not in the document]
jackfrancis: Cool. Is the check essentially "does the `~/.aks-engine` directory exist? if not, run init" ?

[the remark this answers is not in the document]
bacongobbler: Yep!

bacongobbler:
>     * `~/.aks-engine/` will keep track of the "current" cluster context

I *love* that experience, actually. Especially since the majority of the time I may only care about managing a single cluster's state, then I can context switch back to another cluster and manage that, etc.

README.md
PatrickLang: Should add chocolatey too to cover Windows users

README.md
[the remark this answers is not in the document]
PatrickLang: just not like `helm init` where if you forget random stuff fails without saying you forgot to init

README.md
```diff
@@ -8,21 +8,84 @@
...
+
+Upgrade your cluster:
+
+```
+$ aks-engine update -v 1.11
+```
+
+Cluster upgrade time will take between 5-20 minutes *per node* (including master node(s) running the control plane), due to cordon/drain, the addition of new vms with the desired changes, and the deletion of vms with the previous configuration. And it does this according to a rolling, one-at-a-time strategy to minimize operational side-effects. Depending on the size of your cluster, brew some coffee in the French press style, walk your dog(s), or go do that open source contribution you've been putting off for a few months.
+
+...
```
PatrickLang: should be ```

PatrickLang: breaks formatting

CecileRobertMichon: +1

README.md
```diff
@@ -8,21 +8,84 @@
...
+k8s-master-10678025-0                          Ready     master    11s       v1.10.12
+```
+
+Upgrade your cluster:
+
+```
+$ aks-engine update -v 1.11
+```
+
+Cluster upgrade time will take between 5-20 minutes *per node* (including master node(s) running the control plane), due to cordon/drain, the addition of new vms with the desired changes, and the deletion of vms with the previous configuration. And it does this according to a rolling, one-at-a-time strategy to minimize operational side-effects. Depending on the size of your cluster, brew some coffee in the French press style, walk your dog(s), or go do that open source contribution you've been putting off for a few months.
```
PatrickLang: bikeshed: can we put "improve docs" first in the list of things to do while waiting

CecileRobertMichon:
should this go into the design/ folder instead since we are not realistically going to merge it as-is until the readme reflects reality but we still don't want to loose this proposed design?

jackfrancis:
yes, will move

PR #228, opened 2019-01-03
test: pre-delete ILB stuffs before running

jackfrancis:
**Reason for Change**:

Observed some flakes in long-running cluster tests where ILB service was not deleted from a previous run. This change pre-emptively deletes the svc + deployment before running.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -556,10 +556,18 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				r := rand.New(rand.NewSource(time.Now().UnixNano()))
 				serviceName := "ingress-nginx"
 				deploymentName := fmt.Sprintf("ingress-nginx-%s-%v", cfg.Name, r.Intn(99999))
+				d, _ := deployment.Get(deploymentName, "default")
+				if d != nil {
+					_ = d.Delete(deleteResourceRetries)
+				}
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -556,10 +556,18 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				r := rand.New(rand.NewSource(time.Now().UnixNano()))
 				serviceName := "ingress-nginx"
 				deploymentName := fmt.Sprintf("ingress-nginx-%s-%v", cfg.Name, r.Intn(99999))
+				d, _ := deployment.Get(deploymentName, "default")
+				if d != nil {
+					_ = d.Delete(deleteResourceRetries)
```
CecileRobertMichon: shouldn't we catch the error here if the deployment exists but we aren't able to delete it?

jackfrancis: I'm trying to be incremental and permissive here, in the event there's a race condition between the delete operation in a prior run (soak test scenario), say the deployment exists but is in a state of being deleted.

But sure, but perhaps throwing an error here is the more incremental approach in terms of figuring out which parts of the resource scaffolding is brittle over successive E2E runs against a long-running cluster.

PR #229, opened 2019-01-04
chore: update client-go to v10.0.0 and other k8s deps to 1.13.1

tariq1890:
Update k8s deps of aks-engine


mboersma:
Let's not merge this until we have successfully vendored it into ACSRPv2, since we've had to revert the client-go change before.

CecileRobertMichon:
/hold

> Let's not merge this until we have successfully vendored it into ACSRPv2, since we've had to revert the client-go change before.

(protecting from an auto-merge)

PR #231, opened 2019-01-04
fix: change default auth method for deploy,scale and upgrade

tariq1890:
Why this PR?

1. Device auth will no longer work new users of aks-engine who wish to use `upgrade`, `scale`, and `deploy` commands. It has been deprecated by the Azure AAD team and deemed a potential security risk.

2. As device auth is deprecated, we should encourage users to stop using that method. By setting `client_secret` (now renamed to `service_principal`) as the default auth method, it would give a friendly nudge to current device auth users to use SPs directly instead :).

3. `client_secret` is poor and misleading name as an option for auth method. Renamed to `service_principal` for more clarity.

NOTE: This is a breaking change. Any automation that uses the deploy, scale, and upgrade command with the `auth-method` set to device or with no `auth-method` set will have to be changed

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -109,9 +109,9 @@ type authArgs struct {
 func addAuthFlags(authArgs *authArgs, f *flag.FlagSet) {
 	f.StringVar(&authArgs.RawAzureEnvironment, "azure-env", "AzurePublicCloud", "the target Azure cloud")
 	f.StringVarP(&authArgs.rawSubscriptionID, "subscription-id", "s", "", "azure subscription id (required)")
-	f.StringVar(&authArgs.AuthMethod, "auth-method", "service_principal", "auth method (default:`service_principal`, `device`, `client_certificate`)")
-	f.StringVar(&authArgs.rawClientID, "client-id", "", "client id (used with --auth-method=[service_principal|client_certificate])")
-	f.StringVar(&authArgs.ClientSecret, "client-secret", "", "client secret (used with --auth-mode=service_principal)")
+	f.StringVar(&authArgs.AuthMethod, "auth-method", "client_secret", "auth method (default:`client_secret`, `device`, `client_certificate`)")
+	f.StringVar(&authArgs.rawClientID, "client-id", "", "client id (used with --auth-method=[client_secret|client_certificate])")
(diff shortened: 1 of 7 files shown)


CecileRobertMichon:
> or with no auth-method set will have to be changed

can you explain this part? Is there a way to prevent this from happening by changing the default auth method? Any relevant deploy docs that should be changed/clarified to point out that a service principal is now required?

tariq1890:
@CecileRobertMichon I have changed all the docs that point to deploy-related information.

If anyone has existing automation that uses deviceAuth (the current default), it should be changed to use the SP mode auth.. We should not be defaulting to an auth method that is considered a potential security risk. That is the main reason I created this PR

cmd/root.go
```diff
@@ -109,9 +109,9 @@ type authArgs struct {
 func addAuthFlags(authArgs *authArgs, f *flag.FlagSet) {
 	f.StringVar(&authArgs.RawAzureEnvironment, "azure-env", "AzurePublicCloud", "the target Azure cloud")
 	f.StringVarP(&authArgs.rawSubscriptionID, "subscription-id", "s", "", "azure subscription id (required)")
-	f.StringVar(&authArgs.AuthMethod, "auth-method", "device", "auth method (default:`device`, `client_secret`, `client_certificate`)")
-	f.StringVar(&authArgs.rawClientID, "client-id", "", "client id (used with --auth-method=[client_secret|client_certificate])")
-	f.StringVar(&authArgs.ClientSecret, "client-secret", "", "client secret (used with --auth-mode=client_secret)")
+	f.StringVar(&authArgs.AuthMethod, "auth-method", "service_principal", "auth method (default:`service_principal`, `device`, `client_certificate`)")
```
CecileRobertMichon: is device still a valid auth method?

CecileRobertMichon:
> @CecileRobertMichon I have changed all the docs that point to deploy-related information.
> If anyone has existing automation that uses deviceAuth (the current default), it should be changed to use the SP mode auth.. We should not be defaulting to an auth method that is considered a potential security risk. That is the main reason I created this PR

I think we should add a big warning in the docs to let users know about this. Also, if they previously used `client_secret` in automation that won't work anymore correct?

PR #236, opened 2019-01-04
fix: remove VM upgrade version check for AKS

### LINKED ISSUE #235: Enable distinct versions support for aks-engine vlabs vs AKS
Currently the versions support matrix for aks-engine and AKS collide, such that we have to remove versions support in aks-engine in order for AKS to deprecate versions in its own versions support matrix.

CecileRobertMichon:
**Reason for Change**:

The upgradable() function in upgrade checks if a node is upgradable by checking if its current version is contained in a list of upgrade versions. The list of upgrade versions is determined using the list of supported Kubernetes version for aks-engine. However, this does not apply to hosted master clusters (AKS) that have their own list of supported versions and perform upgrade validation already. This allows distinct version support for AKS (as long as the versions supported in aks-engine are a superset).

**Issue Fixed**:

Fixes #235

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradecluster.go
+++ b/pkg/operations/kubernetesupgrade/upgradecluster.go
@@ -223,8 +223,11 @@ func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armh
 					uc.Logger.Infof("Master VM name: %s, orchestrator: %s (MasterVMs)\n", *vm.Name, vmOrchestratorTypeAndVersion)
 					*uc.MasterVMs = append(*uc.MasterVMs, vm)
 				} else {
-					if err := uc.upgradable(vmOrchestratorTypeAndVersion); err != nil && !uc.DataModel.Properties.IsHostedMasterProfile() {
-						return err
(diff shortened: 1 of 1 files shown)


pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -223,7 +223,7 @@ func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armh
 					uc.Logger.Infof("Master VM name: %s, orchestrator: %s (MasterVMs)\n", *vm.Name, vmOrchestratorTypeAndVersion)
 					*uc.MasterVMs = append(*uc.MasterVMs, vm)
 				} else {
-					if err := uc.upgradable(vmOrchestratorTypeAndVersion); err != nil {
+					if err := uc.upgradable(vmOrchestratorTypeAndVersion); err != nil && !uc.DataModel.Properties.IsHostedMasterProfile() {
```
CecileRobertMichon: note that I purposefully did not add this check to the upgradeable() call above for master VMs as master VMs won't  have a hosted master profile

PR #237, opened 2019-01-04
ci: check that last commit message is conformant

mboersma:
**Reason for Change**:

Ensures that commits don't slip through which can't be found by `git-chglog` or similar release tools that expect conventional commits.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -49,7 +49,7 @@ jobs:
   - script: make bootstrap
     displayName: Install dependencies
     workingDirectory: $(modulePath)
-  - script: make test-style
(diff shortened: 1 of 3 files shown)


mboersma:
This checks the most recent commit, which may not be the commit message used when we squash a PR. So maintainers still need to pay attention when merging.

I think this will be useful enforcement in most cases that will keep us from missing things in release notes. But there's a tendency not to follow all these rules when adding commits onto an existing PR, as in:
```
fix typo

add manifest i forgot in last commit
```

Those type of tacked-on commits will fail now. IDK if that's more annoying than helpful.

PR #243, opened 2019-01-07
feat: add localhost and 127.0.0.1 to certificates

sylr:
**Reason for Change**:

There are some cases where we have pods on master nodes who needs to communicate with the API server (e.g.: cluster-autoscaler) but it cannot use the `kubernetes.default` service cause it will intermittently fail due to the fact that, if the request is forwarded by the service to the apiserver on the **same node** as it originated from, the request will never come back due to `Martian source` network error.

The way around that is to use host network and send request to localhost:443 but localhost is currently not part of the default domains in the certificates generated.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/helpers/pki.go
+++ b/pkg/helpers/pki.go
@@ -14,7 +14,6 @@ import (
 	"fmt"
 	"math/big"
 	"net"
-	"sync"
 	"time"
 
 	"golang.org/x/sync/errgroup"
@@ -77,7 +76,6 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
 	var group errgroup.Group
 
 	var err error
-	var mu sync.Mutex
 	caCertificate, err = pemToCertificate(caPair.CertificatePem)
 	if err != nil {
 		return nil, nil, nil, nil, nil, nil, err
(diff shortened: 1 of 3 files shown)


jackfrancis:
@khenidak Curious for your thoughts on this one. Does this seem like a sane path we should allow?

tariq1890:
@sylr These changes may be redundant. In this [method](https://github.com/Azure/aks-engine/blob/5cadc11d6b4f149aa616b7a7ffef5be31de52f27/pkg/helpers/pki.go#L49), we are already appending the localhost IP before creating the certs

tariq1890:
@sylr We'll need to circle back a bit . You'll need to append "localhost " to this slice`ips := []net.IP{firstMasterIP}` like you did in the previous iteration of this PR. 

Once that is done. You can remove all of the append statements in `pki.go`. You can remove the mutexes too :)

pkg/helpers/pki.go
```diff
@@ -88,7 +88,11 @@ func CreatePki(extraFQDNs []string, extraIPs []net.IP, clusterDomain string, caP
 	}
 
 	group.Go(func() (err error) {
-		apiServerCertificate, apiServerPrivateKey, err = createCertificate("apiserver", caCertificate, caPrivateKey, false, true, extraFQDNs, extraIPs, nil)
+		ip := net.ParseIP("127.0.0.1").To4()
```
tariq1890: Please refer to my latest PR comment. If that is done, we won't need to do these kind of appends

PR #248, opened 2019-01-08
fix: remove IP address network configuration race condition for maste…

CecileRobertMichon:
…r nodes so using standard azure dependencies

**Reason for Change**:

**PR copied over from https://github.com/Azure/acs-engine/pull/3755**

In @fortescue 's words:

Case handled:  AzureCNI/Custom Vnet

[There is a documented race condition](https://github.com/Azure/acs-engine/blob/master/docs/kubernetes/features.md) whereby it is impossible to ensure that master ip addresses are allocated serially as REQUIRED by k8 bootstrap scripts.  I hit this problem every time i.e. the cluster fails to start. 

This PR guarantees the 'MasterCount' ip addresses are allocated serially beginning with *firstConsecutiveStaticIP* **before** dynamic ip allocation begins. 

The generated dynamic ips stanza are made dependent on the static ip (masterCount) stanzas.
  
To implement this, both the static and dynamic ips are in separate arm stanzas.
The static network interfaces reference the lb and intlb(for masterCount> 1) as before.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Strike 
```
If possible, assign to the firstConsecutiveStaticIP configuration property an IP address that is near the "end" of the available IP address space in the desired subnet.
For example, if the desired subnet is a /24, choose the "239" address in that network space
```
And change to: 
```Ensure that firstConsecutiveStaticIP + (masterCount-1) ip addresses are free within the master subnet as required by k8 installation scripts.```

### DIFF
--- a/parts/k8s/kubernetesmasterresources.t
+++ b/parts/k8s/kubernetesmasterresources.t
@@ -372,7 +372,7 @@
                     ]
                 }
           {{end}}
-          {{if or $MasterProfile.IsCustomVNET IsOpenShift}}
+          {{if $MasterProfile.IsCustomVNET}}
                   ,"networkSecurityGroup": {
                     "id": "[variables('nsgID')]"
                   }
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
@sharmasushant @nisheeth-ms could someone from the Azure CNI team please help review this change?

PR #249, opened 2019-01-08
feat: add anti affinity to coredns deployment

sylr:
**Reason for Change**:

Improve resilience of the coredns service with anti-affinity.

Those will try to spread coredns pods across `failure-domain zones` and `nodes` at scheduling.

See:

https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -100,6 +100,27 @@ spec:
       annotations:
         seccomp.security.alpha.kubernetes.io/pod: docker/default
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchExpressions:
+                - key: k8s-app
+                  operator: In
+                  values:
+                  - kube-dns
+              topologyKey: failure-domain.beta.kubernetes.io/zone
+            weight: 10
+          - podAffinityTerm:
+              labelSelector:
+                matchExpressions:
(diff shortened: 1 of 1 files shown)


parts/k8s/addons/coredns.yaml
```diff
@@ -100,6 +100,27 @@ spec:
       annotations:
         seccomp.security.alpha.kubernetes.io/pod: docker/default
     spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
```
jackfrancis: Let's try to summarize what this is expressing. Is it:

"Don't schedule coredns pods onto a node that already has a `kube-dns`-labeled pod in the same availability zone."

sylr: Yes, except it will only try, if it can't it will schedule both pods on the same node if no other is available.

jackfrancis: Indeed, thanks. Hence the `preferredDuringSchedulingIgnoredDuringExecution` instead of  `requiredDuringSchedulingIgnoredDuringExecution`.

sylr: Exactly

jackfrancis:
This looks sane to me pending E2E. @mboersma or @CecileRobertMichon, would appreciate a 2nd opinion.

PR #250, opened 2019-01-08
chore: remove commit check from Azure DevOps pipeline

mboersma:
**Reason for Change**:

`make validate-commit-msg` only looks at the most recent non-merge commit, which can be [less than optimal](https://github.com/Azure/aks-engine/pull/237#issuecomment-451593876). This removes that check from our Azure DevOps pipeline.

Instead let's try the probot [semantic pull requests](https://github.com/probot/semantic-pull-requests) GitHub service. I enabled it and will test it with this PR.

screen shot 2019-01-08 at 10 53 38 am

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -49,7 +49,7 @@ jobs:
   - script: make bootstrap
     displayName: Install dependencies
     workingDirectory: $(modulePath)
-  - script: make validate-commit-msg validate-copyright-headers test-style
+  - script: make validate-copyright-headers test-style
     displayName: Run linting rules
     workingDirectory: $(modulePath)
   - script: make validate-dependencies


.pipelines/pr-e2e.yaml
```diff
@@ -49,7 +49,7 @@ jobs:
   - script: make bootstrap
     displayName: Install dependencies
     workingDirectory: $(modulePath)
-  - script: make validate-commit-msg validate-copyright-headers test-style
```
CecileRobertMichon: should we also remove the script in the Makefile since we're not using it anymore?

CecileRobertMichon:
+1 I like this so much better as a PR check

.pipelines/pr-e2e.yaml
mboersma: I'd still like to have a way to validate commit messages locally, and IMHO this still works fine for that case. I'll probably use it, but I can rip it out if we'd rather not have the clutter.

PR #262, opened 2019-01-09
test: enable multiple subnets per agent pool

jackfrancis:
**Reason for Change**:

In practice, lots of folks segment their vnet into subnets per agent pool. The purpose of this PR is to add that configuration pattern to "custom VNET" scenarios automatically, to ensure that this known working operational scenario does not decay into an edge case over time due to lack of test coverage.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tariq1890:
I think UT's  for `CreateSubnet` and `CreateVnet` should address the codecov failures. I'll leave it you to decide @jackfrancis . Otherwise lgtm.

PR #274, opened 2019-01-10
test: print pod logs when errors occur

jackfrancis:
**Reason for Change**:

To help debug flakes (e.g., `kubectl top nodes`), we should be printing out the pod logs when errors occur)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -615,8 +615,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 						pods, err := pod.GetAllByPrefix("metrics-server", "kube-system")
 						Expect(err).NotTo(HaveOccurred())
 						for _, p := range pods {
-							err = p.Logs()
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -612,6 +612,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 					}
 					if i > 28 {
 						log.Printf("Error while running kubectl top nodes:%s\n", err)
+						pods, err := pod.GetAllByPrefix("metrics-server", "kube-system")
+						Expect(err).NotTo(HaveOccurred())
```
CecileRobertMichon: I would argue we shouldn't return an error for logging debugging data in this "if not successful" block since that will hide the important error which is `Expect(success).To(BeTrue())` on line 624. The logs are bonus, not part of the actual test.

jackfrancis: Good feedback, addressed.

PR #287, opened 2019-01-11
refactor: upgrade code refactor part 1

CecileRobertMichon:
**Reason for Change**:

This is an initial refactor of the upgrade code. More to follow.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
TODO:
- [ ] Add unit tests
- [ ] Run upgrade back compat tests


cmd/upgrade.go
```diff
@@ -178,14 +174,10 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
-		return errors.Errorf("upgrading to version %s is not supported. To see a list of available upgrades, use 'aks-engine orchestrators --orchestrator kubernetes --version %s'", uc.upgradeVersion, uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion)
+		return errors.Errorf("upgrading from Kubernetes version %s to version %s is not supported. To see a list of available upgrades, use 'aks-engine orchestrators --orchestrator kubernetes --version %s'", uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion, uc.upgradeVersion, uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion)
 	}
 
-	// Read name suffix to identify nodes in the resource group that belong
-	// to this cluster.
-	// TODO: Also update to read  namesuffix from the parameters file as
-	// user could have specified a name suffix instead of using the default
-	// value generated by AKS Engine
+	// Read the name suffix from the parameters to identify nodes in the resource group that belong to this cluster.
```
jackfrancis: s/nodes/vms ?

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -138,8 +100,8 @@ func (uc *UpgradeCluster) UpgradeCluster(subscriptionID uuid.UUID, az armhelpers
 	return nil
 }
 
-func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armhelpers.AKSEngineClient, resourceGroup, kubeConfig string) error {
-	targetOrchestratorTypeVersion := fmt.Sprintf("%s:%s", uc.DataModel.Properties.OrchestratorProfile.OrchestratorType, uc.DataModel.Properties.OrchestratorProfile.OrchestratorVersion)
+func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, resourceGroup, kubeConfig string) error {
```
CecileRobertMichon: This is a breaking change for AKS which will require the changes to be applied at vendoring, heads up

cmd/upgrade.go
CecileRobertMichon: true

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -250,16 +207,19 @@ func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armh
 	return nil
 }
 
-// getClusterNodeVersion returns a node's "orchestrator:version" via Kubernetes API or VM tag.
-func (uc *UpgradeCluster) getClusterNodeVersion(client armhelpers.KubernetesClient, name string, tags map[string]*string) string {
+// getNodeVersion returns a node's current Kubernetes version via Kubernetes API or VM tag.
+func (uc *UpgradeCluster) getNodeVersion(client armhelpers.KubernetesClient, name string, tags map[string]*string) string {
```
jackfrancis: Can we reverse the precedence order here? Shouldn't we prefer to check the k8s API first, and then fall back to the tag?

jackfrancis:
Is this an appropriate PR to include consideration of the following changes:

- implement "upgrade single node" as a composable unit
- expose above as an option for a user

Use cases:

- "I want to upgrade a single node and schedule some stuff onto it, do a/b testing with non-upgraded nodes and compare"
- "My upgrade failed halfway through and I want to cleanup my cluster by manually upgrading the remaining nodes that are running an old version"

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: This is the same behavior that was implemented in @mboersma's PR to stop requiring tags https://github.com/Azure/acs-engine/pull/4100 so I'll defer to him for choosing this approach over the other.

I think the idea was that we maintain existing behavior to limit risk and only use the alternative method if we can't find the tags. 

If we do decide we want to reverse the order, I would argue that's a behavior change and it doesn't belong in this refactor PR (which is already dangerous enough as is).

CecileRobertMichon:
@jackfrancis short answer is no.

This is a refactor PR and the first of many. I tried to break down the changes so that they are easier to review/test. Adding new features is out of the scope of this PR.

We can definitely keep that in the backlog as a future addition though.

CecileRobertMichon:
To clarify for reviewers: there are many more things I want to change/improve in the upgrade code area, this is only the beginning. I am breaking down the changes in smaller units of code changes to make it more manageable to review and limit risk of breakage since upgrade is a complicated/sensitive part of the code. Trying to change that, one step at a time :)

cmd/upgrade.go
```diff
@@ -223,12 +215,12 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 			parametersKey, nameSuffixKey, defaultValueKey, templatePath)
 	}
 
-	log.Infoln(fmt.Sprintf("Name suffix: %s", uc.nameSuffix))
+	log.Infoln(fmt.Sprintf("Upgrading cluster with name suffix: %s", uc.nameSuffix))
 
-	uc.agentPoolsToUpgrade = []string{}
-	log.Infoln(fmt.Sprintf("Gathering agent pool names..."))
+	uc.agentPoolsToUpgrade = make(map[string]bool)
+	uc.agentPoolsToUpgrade[kubernetesupgrade.MasterPoolName] = true
```
jackfrancis: Is this additive? I don't see us adding the master in the previous `agentPoolsToUpgrade` string slice pattern

CecileRobertMichon: maybe I should rename this to `poolsToUpgrade` - the `agentPoolsToUpgrade` name doesn't really make sense

cmd/upgrade.go
```diff
@@ -253,16 +245,23 @@ func (uc *upgradeCmd) run(cmd *cobra.Command, args []string) error {
 		StepTimeout: uc.timeout,
 	}
 
+	upgradeCluster.ClusterTopology = kubernetesupgrade.ClusterTopology{}
```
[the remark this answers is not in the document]
CecileRobertMichon: done

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -77,58 +75,21 @@ const MasterVMNamePrefix = "k8s-master-"
...
+func (uc *UpgradeCluster) UpgradeCluster(az armhelpers.AKSEngineClient, kubeConfig string, aksEngineVersion string) error {
 	uc.MasterVMs = &[]compute.VirtualMachine{}
 	uc.UpgradedMasterVMs = &[]compute.VirtualMachine{}
 	uc.AgentPools = make(map[string]*AgentPoolTopology)
-	uc.AgentPoolsToUpgrade = make(map[string]bool)
 
-	for _, poolName := range agentPoolsToUpgrade {
-		uc.AgentPoolsToUpgrade[poolName] = true
-	}
-	uc.AgentPoolsToUpgrade[MasterPoolName] = true
```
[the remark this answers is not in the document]
CecileRobertMichon: yeah I just copied existing behavior but I don't really understand what AgentPoolsToUpgrade is used for yet

CecileRobertMichon:
Scenarios tested so far:

✅ Windows 1.8 cluster with VMAS
✅ Linux 1.10 cluster with VMAS

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -250,16 +206,19 @@ func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armh
...
+		if len(parts) == 2 {
+			return parts[1]
+		}
 	}
-	uc.Logger.Warnf("Expected tag \"orchestrator\" not found for VM: %s", name)
+	uc.Logger.Warnf("Expected tag \"orchestrator\" not found for VM: %s. Using Kubernetes API to retrieve Kubernetes version.", name)
 	if client != nil {
 		node, err := client.GetNode(name)
 		if err == nil {
-			return api.Kubernetes + ":" + strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v")
```
CecileRobertMichon: here we return just the k8s version instead of `kubernetes:v1.x.x` since this is in the `kubernetesupgrade` package, we only care about the version - not the orchestrator.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -77,58 +75,21 @@ const MasterVMNamePrefix = "k8s-master-"
...
-	if err := uc.getClusterNodeStatus(subscriptionID, az, resourceGroup, kubeConfig); err != nil {
+	if err := uc.getClusterNodeStatus(az, uc.ResourceGroup, kubeConfig); err != nil {
 		return uc.Translator.Errorf("Error while querying ARM for resources: %+v", err)
 	}
 
 	var upgrader UpgradeWorkFlow
 	upgradeVersion := uc.DataModel.Properties.OrchestratorProfile.OrchestratorVersion
 	uc.Logger.Infof("Upgrading to Kubernetes version %s\n", upgradeVersion)
-	switch {
-	case strings.HasPrefix(upgradeVersion, "1.6."):
```
CecileRobertMichon: if anyone knows a good reason why these separate upgraders are still needed, please speak up

pkg/operations/kubernetesupgrade/upgradecluster.go
mboersma: I agree with Cécile's comment here. I think it's safer to leave it alone.

This way the existing behavior is still the default (use tags), and the k8s API is explicitly a fallback strategy. At the time, that was deemed to be less disruptive, but it has proved to be useful in failed upgrade cases: setting the tags is a safe way to "trick" the upgrade behavior to run again.

pkg/operations/kubernetesupgrade/upgradecluster.go
mboersma: :+1: as far as I can tell, these structs provide no utility over the `Upgrader` base.

Maybe someone was anticipating we would need different behavior for each version upgrade? (Don't do that in advance because [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).)

CecileRobertMichon:
VMSS upgrade is failing with ` Error upgrading cluster: Drain did not complete within 1m0s `

https://github.com/Azure/aks-engine/pull/298 might help

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -204,41 +164,37 @@ func (uc *UpgradeCluster) getClusterNodeStatus(subscriptionID uuid.UUID, az armh
 		}
 
 		for _, vm := range vmListPage.Values() {
-			vmOrchestratorTypeAndVersion := uc.getClusterNodeVersion(kubeClient, *vm.Name, vm.Tags)
-			if vmOrchestratorTypeAndVersion == "" {
+			// Windows VMs contain a substring of the name suffix
```
jackfrancis: What is the Windows-specific implications of the line(s) below?

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -394,28 +345,3 @@ func (uc *UpgradeCluster) addVMToAgentPool(vm compute.VirtualMachine, isUpgradab
 
 	return nil
 }
-
```
jackfrancis: I actually kinda think we should keep this in, uncomment it, and standardize to persisting the template that we deploy to upgrade always.

But perhaps for a later exercise.

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: `strings.Contains(*(vm.Name), uc.NameSuffix[:4]+"k8s"` is checking for VM naming specific to windows instead of looking for the whole name suffix

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: We already have logic for saving the modified template in upgrade.go https://github.com/CecileRobertMichon/aks-engine/blob/0b7a9702300ed5d033874a6ced2d2a140867fe18/cmd/upgrade.go#L281 so this commented out code is redundant.

I think we need to improve the logic for persisting templates and instead of passing them to a file, saving them in the RG or something like that but, like you said, for a later exercise.

PR #291, opened 2019-01-11
set publicIPAllocationMethod as static for master node

andyzhangx:
**Reason for Change**:

This PR sets `publicIPAllocationMethod` as static for master node, this is for doing AKS whitelist feature which requires all master nodes uses static public ip.

cc @feiskyer @ritazh 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
set `publicIPAllocationMethod` as static for master node

### DIFF
--- a/parts/k8s/kubernetesmasterresourcesvmss.t
+++ b/parts/k8s/kubernetesmasterresourcesvmss.t
@@ -238,11 +238,7 @@
     "dnsSettings": {
       "domainNameLabel": "[variables('masterFqdnPrefix')]"
     },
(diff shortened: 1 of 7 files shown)


andyzhangx:
@mboersma I have changed the commit message. there won't be breaking change if user use same version of aks-engine to do upgrade, since this PR use same config for `publicIPAllocation`.

CecileRobertMichon:
@ritazh could you please review this change?

tariq1890:
Putting the PR on hold as this needs to be incorporated in the no_template branch as well.

PR #297, opened 2019-01-11
feat: Support customized environments in apimodel.json (#276) and Support customized API version in ARM template (#283)

### LINKED ISSUE #276: Support customized environments in apimodile.json
No
---

FEATURE REQUEST
---

The latest version
---

Go-autorest pre-defined 4 environments (ChinaCloud, GermanCloud, PublicCloud, USGovernmentCloud), 
The environment will have the endpoints for Arm, Graph, etc.

In Azure stack, the endpoints are the input parameters which will be provided by customer during deploying azurestack on the the physical hardware. 

Go-autorest already support to construct the environment object from a file. 

So AKS engine should accept the endpoints as input parameter in apimodile.json and write into the master and agent node.  Kubernetes controllers will also use it to construct Environment object to communicated back to ARM on AzureStack

### LINKED ISSUE #283: Support customized  API version in ARM template
No
---

Feature request
---

The latest
---

AzureStack support subset API version which is supported on Azure. for example 

For example:

in AzureStack:

    "apiVersionCompute": "2018-06-01",

    "apiVersionStorage": "2018-07-01",

    "apiVersionKeyVault": "2018-02-14",

    "apiVersionNetwork": "2018-08-01",

on Azure:

    "apiVersionCompute": "2017-03-30",

    "apiVersionStorage": "2017-10-01",

    "apiVersionKeyVault": "2016-10-01",

    "apiVersionNetwork": "2017-10-01",

honcao:
fix:Support customized environments in apimodile.json (#276) and Support customized API version in ARM template (#283)

**Reason for Change**:

Fix issues to support AzureStack

**Issue Fixed**:

Fix #276 
Fix #283

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/types.go
```diff
@@ -669,6 +671,11 @@ type V20180331ARMManagedContainerService struct {
 	*v20180331.ManagedCluster
 }
 
+// CustomCloudProfile Represents custom cloud profile
+type CustomCloudProfile struct {
```
jackfrancis: Instead of adding a new sub-property `CustomCloudProfile` to `Properties`, have we considered adding an additional `AzureEnvironmentSpecConfig` for Azure Stack? Basically, how it is currently implemented:

- you specify which "cloud" you are in by declaring at string value for `Location`, which is a property of `ContainerService`.
- we have constructors which then return an instance of your particular `AzureEnvironmentSpecConfig`.

See how `AzureChinaCloudSpec` does this for mooncake scenarios.

I think I'd prefer to expand the `AzureEnvironmentSpecConfig` type definition, rather than add a new one to `Properties`.

Any thoughts here @CecileRobertMichon, @tariq1890, @mboersma?

honcao: It is good idea to extend the existing object.
I am planning the following changes later to support ADFS authentication, customized AzureEnvironmentSpecConfig , I will extend customeCloudProfile to the following formart:
"customeCloudProfile":{
          "environment":{},
           "azureEnvironmentSpecConfig"{}
           "identitySystem":"ADFS",
           "AuthMethod": "Certificate"
}

if we extend the existing AzureEnvironmentSpecConfig, then I will need add the following fields
          "environment":{},
        "identitySystem":"ADFS",
          "AuthMethod": "Certificate"

@jackfrancis , please let me know your thoughts.

honcao: and there will be a new sub-property `AzureEnvironmentSpecConfig` to `Properties` , since the configuration needs to be in API model.

pkg/api/types.go
```diff
@@ -669,6 +671,11 @@ type V20180331ARMManagedContainerService struct {
...
+	Enviornment *azure.Environment `json:"environment,omitempty"`
```
jackfrancis: typo: `Enviornment` should be `Environment` (you'll want to do a global search/replace on all files for this)

pkg/api/types.go
jackfrancis: Understood, thanks for clarifying

pkg/engine/template_generator.go
```diff
@@ -205,6 +206,29 @@ func (t *TemplateGenerator) getMasterCustomData(cs *api.ContainerService, textFi
...
 func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) template.FuncMap {
 	return template.FuncMap{
+		"IsAzureStackCloud": func() bool {
+			var cloudProfileName string
+			if cs.Properties.CustomCloudProfile != nil {
+				if cs.Properties.CustomCloudProfile.Enviornment != nil {
+					cloudProfileName = cs.Properties.CustomCloudProfile.Enviornment.Name
+				}
+			}
+			return strings.EqualFold(cloudProfileName, api.AzureStackCloud)
```
jackfrancis: Just curious why the `EqualFold` usage here and not a strict equality check?

pkg/engine/template_generator.go
```diff
@@ -205,6 +206,29 @@ func (t *TemplateGenerator) getMasterCustomData(cs *api.ContainerService, textFi
...
+		},
+		"GetCustomEnvironmentJSON": func() string {
+			var environmentJSON string
+			if cs.Properties.CustomCloudProfile != nil {
+				if cs.Properties.CustomCloudProfile.Enviornment != nil {
+					bytes, err := json.Marshal(cs.Properties.CustomCloudProfile.Enviornment)
```
jackfrancis: @tariq1890 should we move this "get JSON representation" logic to be a method of the `Environment` type struct?

honcao: GetCustomEnvironmentJSON() will return empty string if CustomCloudProfile field is empty.
So I put the method to struct Properties, also move IsAzureStackCloud() to struct Properties.

pkg/engine/template_generator.go
honcao: in the environment.go, 
	if strings.EqualFold(name, "AZURESTACKCLOUD") {
		return EnvironmentFromFile(os.Getenv(EnvironmentFilepathName))
	}
it load the cloud name case insensitive, so I followed the pattern.

parts/k8s/kubernetesmastervars.t
```diff
@@ -86,10 +86,18 @@
     "masterOffset": "[parameters('masterOffset')]",
     {{ end }}
 {{end}}
+{{if not IsAzureStackCloud}}
```
tariq1890: @honcao Can we flip this? I prefer using it like this
```suggestion
{{if IsAzureStackCloud}}
```

honcao: @tariq1890 , thanks for the comments, I updated the code.

CecileRobertMichon:
can we change the commit message to something more concise?

PR #298, opened 2019-01-11
fix: standardize upgrade cordon-drain timeout

jackfrancis:
**Reason for Change**:

Enable future maintenance of cordon/drain timeout by making it a discrete const.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -21,7 +21,7 @@ import (
 const (
 	interval           = time.Second * 1
 	retry              = time.Second * 5
-	cordonDrainTimeout = time.Minute * 10
+	cordonDrainTimeout = time.Minute * 1
 )
 
 // Compiler to verify QueueMessageProcessor implements OperationsProcessor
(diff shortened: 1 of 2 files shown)


pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -19,8 +19,9 @@ import (
 )
 
 const (
-	interval = time.Second * 1
-	retry    = time.Second * 5
+	interval           = time.Second * 1
+	retry              = time.Second * 5
+	cordonDrainTimeout = time.Minute * 10
```
CecileRobertMichon: why 10 minutes? if scale is 1 hour

jackfrancis: This is just a placeholder to demonstrate the surface area that needs changing. No concrete requirement yet.

CecileRobertMichon:
any blockers for moving this forward?

jackfrancis:
@CecileRobertMichon I reverted the functional change from 1 minute to 10 minutes. Now the scope of this PR is just to move the timeout value to be more maintainable as a discrete const.

jackfrancis:
I think we want to have definitive repro cases where increasing the timeout causes a demonstrable improvement (with no tradeoffs).

PR #331, opened 2019-01-17
chore: add support for Kubernetes 1.12.5

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/releases/tag/v1.12.5

**Issue Fixed**:

> Fix validation issue on AKS

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/k8s_versions.go
```diff
@@ -63,7 +63,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"aci-connector":                    "virtual-kubelet:latest",
 		ContainerMonitoringAddonName:       "oms:ciprod01092019",
 		AzureCNINetworkMonitoringAddonName: "networkmonitor:v0.0.4",
-		"cluster-autoscaler":               "cluster-autoscaler:v1.12.1",
+		"cluster-autoscaler":               "cluster-autoscaler:v1.12.2",
```
tariq1890: Can we add this to the `install-dependencies.sh` file ? More pre-pulling goodness !

mboersma:
This failed k8s 1.10 linux because of a >90 min timeout. That's unrelated to this PR and it has passed the rest of e2e.

packer/install-dependencies.sh
```diff
@@ -92,7 +92,7 @@ for TILLER_VERSION in ${TILLER_VERSIONS}; do
     pullContainerImage "docker" "gcr.io/kubernetes-helm/tiller:v${TILLER_VERSION}"
 done
 
-CLUSTER_AUTOSCALER_VERSIONS="1.13.1 1.12.1 1.3.4 1.3.3 1.2.2 1.1.2"
+CLUSTER_AUTOSCALER_VERSIONS="1.13.1 1.12.2 1.3.4 1.3.3 1.2.2 1.1.2"
```
CecileRobertMichon: is 1.12.1 not used anywhere anymore?

mboersma: cluster-autoscaler 1.12.1 was only used by k8s 1.12.x, so as far as I can tell this patch release replaces it.

PR #347, opened 2019-01-22
fix: Fix the incorrect computername parameter

tariq1890:
`computername` is the wrong name and it is not recognized by the RP schema. `computerName` should be used instead


CecileRobertMichon:
is `it is recognized by the RP schema` meant to be `it is NOT recognized by the RP schema`?

tariq1890:
@CecileRobertMichon updated the PR description.

jackfrancis:
lgtm

PR #350, opened 2019-01-23
fix: add k8s.io/client-go/.../azure missing module

sylr:
As this module is not called directly it was omitted but it is needed by the upgrade process when using AAD auth.

```
INFO[0402] Error validating upgraded master VM: k8s10-euw-sandbox-master-0
FATA[0402] Error upgrading cluster: No Auth Provider found for name "azure"
```


tariq1890:
Is this because you are dependent on the method below @sylr ?

```
func init() {
	if err := restclient.RegisterAuthProviderPlugin("azure", newAzureAuthProvider); err != nil {
		klog.Fatalf("Failed to register azure auth plugin: %v", err)
	}
}
```

tariq1890:
Not sure how I feel about this. Forcing an import because of an init-method is a huge antipattern for Go.

Gopkg.toml
```diff
@@ -96,15 +96,15 @@ required = [
 
 [[constraint]]
   name = "k8s.io/api"
-  version = "kubernetes-1.10.0"
+  version = "kubernetes-1.13.2"
```
tariq1890: We have currently put the k8s version updates on hold. @jackfrancis can provide details about this.

pkg/armhelpers/kubeclient.go
```diff
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth/azure"
```
tariq1890: :(

jackfrancis:
I'm a little confused why the AAD upgrade errors require a rev of client-go

sylr:
As far as I remember it was not compiling anymore with the current pinned versions when I forced the import of the module. 

Updating everything to 1.13.2 fixed it.

sylr:
> Is this because you are dependent on the method below @sylr ?
> 
> ```
> func init() {
> 	if err := restclient.RegisterAuthProviderPlugin("azure", newAzureAuthProvider); err != nil {
> 		klog.Fatalf("Failed to register azure auth plugin: %v", err)
> 	}
> }
> ```

Yes

sylr:
> Not sure how I feel about this. Forcing an import because of an init-method is a huge antipattern for Go.

Is it ? I'm new to Go but I've seen this quite a few times in several projects.

Anyway do you know any other way to fix that ?

sylr:
I remember now, I upgraded the libs because I needed https://github.com/kubernetes/client-go/commit/af076e5342ee3b17ac7f68b33f1b48ab42157774#diff-ccba71f00b34f4ba0c9297d2f8e5d808 to make it compile.

It is only included since kubernetes-1.13.0-beta.0.

tariq1890:
> Anyway do you know any other way to fix that ?
I saw the client-go code, looks like that pattern is baked in :(.

> I remember now, I upgraded the libs because 
Did you use a patched vendor of client-go in your aks-engine branch ?

sylr:
> Did you use a patched vendor of client-go in your aks-engine branch ?

A patched vendor ? I just updated the Gopkg.toml and ran `dep ensure`

PR #355, opened 2019-01-23
fix: Fail deployCmd when orchestrator version is invalid

### LINKED ISSUE #191: Deploy does not pre-validate Kubernetes version
No
---

ISSUE
---

0.27.0
---

**Kubernetes version**:
1.12.4

**What happened**:
1.10.9 was used by aks-engine without any errors being reported

either 1.12.4 to be used or an error when validating the cluster configuration if unsupported

run aks-engine with orchestratorRelease set to 1.12 and orchestratorVersion set to 1.12.4

serbrech:
This PR runs the orchestrator version validation logic when running deploy before it is mistakenly being reset to the default value. This is mostly a workaround until we reach a better solution

**Issue Fixed**:
Fixes #191 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
The fix is implemented with a workaround. Comment in the code describes the root cause/ideal fix.

### DIFF
--- a/pkg/api/convertertoapi.go
+++ b/pkg/api/convertertoapi.go
@@ -605,7 +605,7 @@ func convertVLabsOrchestratorProfile(vp *vlabs.Properties, api *OrchestratorProf
 		}
 		setVlabsKubernetesDefaults(vp, api)
 
-		//HACK : this validation should be done as part of the main validation, but deploy does it only after loading the container.
+		// TODO (hack): this validation should be done as part of the main validation, but deploy does it only after loading the container.
 		if err := vp.ValidateOrchestratorProfile(isUpdate); err != nil {
 			return err
 		}
--- a/pkg/api/convertertoapi.go
+++ b/pkg/api/convertertoapi.go
@@ -103,7 +103,7 @@ func ConvertV20170701ContainerService(v20170701 *v20170701.ContainerService, isU
 }
 
 // ConvertVLabsContainerService converts a vlabs ContainerService to an unversioned ContainerService
-func ConvertVLabsContainerService(vlabs *vlabs.ContainerService, isUpdate bool) *ContainerService {
+func ConvertVLabsContainerService(vlabs *vlabs.ContainerService, isUpdate bool) (*ContainerService, error) {
 	c := &ContainerService{}
 	c.ID = vlabs.ID
 	c.Location = helpers.NormalizeAzureRegion(vlabs.Location)
@@ -118,8 +118,10 @@ func ConvertVLabsContainerService(vlabs *vlabs.ContainerService, isUpdate bool) 
 	}
 	c.Type = vlabs.Type
 	c.Properties = &Properties{}
-	convertVLabsProperties(vlabs.Properties, c.Properties, isUpdate)
-	return c
+	if err := convertVLabsProperties(vlabs.Properties, c.Properties, isUpdate); err != nil {
+		return nil, err
+	}
+	return c, nil
 }
 
 // convertV20160930ResourcePurchasePlan converts a v20160930 ResourcePurchasePlan to an unversioned ResourcePurchasePlan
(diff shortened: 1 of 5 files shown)


pkg/api/convertertoapi.go
```diff
@@ -611,6 +615,21 @@ func convertVLabsOrchestratorProfile(vp *vlabs.Properties, api *OrchestratorProf
...
+	version := common.RationalizeReleaseAndVersion(
+		orchestratorProfile.OrchestratorType,
+		orchestratorProfile.OrchestratorRelease,
+		orchestratorProfile.OrchestratorVersion,
+		isUpdate,
+		hasWindows)
+	//if we didn find one, it's invalid, but we don want to get the default one, so reset to the provided version
+	//validation kicks in later for deploy and will fail.
+	if version == "" {
+		return orchestratorProfile.OrchestratorVersion
```
CecileRobertMichon: I don't think we want to simply return orchestratorProfile.OrchestratorVersion.

Here's an example where that wouldn't work:
If I set my `OrchestratorRelease` to "1.10" and my `OrchestratorVersion` to "1.11.6" -> `RationalizeReleaseAndVersion` won't be able to rationalize the release with the version since they are contradictory so it will return an empty version string. But "1.11.6"  in itself is a valid version so that won't fail validation later. Unless we call RationalizeReleaseAndVersion later on but in that case there is no need for this duplicate check all-together.

serbrech: addressed

pkg/api/convertertoapi.go
```diff
@@ -102,7 +102,7 @@ func ConvertV20170701ContainerService(v20170701 *v20170701.ContainerService, isU
 }
 
 // ConvertVLabsContainerService converts a vlabs ContainerService to an unversioned ContainerService
-func ConvertVLabsContainerService(vlabs *vlabs.ContainerService, isUpdate bool) *ContainerService {
+func ConvertVLabsContainerService(vlabs *vlabs.ContainerService, isUpdate bool) (*ContainerService, error) {
```
serbrech: Main concern is to make sure that this isn't used by projects that depend on us.  I checked but please confirm :)

pkg/api/convertertoapi.go
```diff
@@ -591,12 +597,19 @@ func convertVLabsOrchestratorProfile(vp *vlabs.Properties, api *OrchestratorProf
 			convertVLabsKubernetesConfig(vlabscs.KubernetesConfig, api.KubernetesConfig)
 		}
 		setVlabsKubernetesDefaults(vp, api)
+
+		//HACK : this validation should be done as part of the main validation, but deploy does it only after loading the container.
+		if err := vp.ValidateOrchestratorProfile(isUpdate); err != nil {
```
serbrech: if we can get the deploycmd to run with `validation` set to true just like generate, then we can drop this hack.

pkg/api/convertertoapi.go
jackfrancis: Yep, that's just us

serbrech:
rebased on master

pkg/api/convertertoapi.go
```diff
@@ -598,12 +604,19 @@ func convertVLabsOrchestratorProfile(vp *vlabs.Properties, api *OrchestratorProf
 			convertVLabsKubernetesConfig(vlabscs.KubernetesConfig, api.KubernetesConfig)
 		}
 		setVlabsKubernetesDefaults(vp, api)
+
+		//HACK : this validation should be done as part of the main validation, but deploy does it only after loading the container.
```
tariq1890: Super nit: This tools will pick up on this as a TODO item. Better for future tracking purposes
```suggestion
		// TODO (hack): this validation should be done as part of the main validation, but deploy does it only after loading the container.
```

Issue #339, opened 2019-01-21
`aks-engine orchestrators` shows DCOS/swarm orchestrators as available options


bacongobbler:
If I enter `aks-engine orchestrators`, I get Kubernetes, DCOS and Swarm as available orchestrators.

Given that the only orchestrator AKS Engine intends to support in the future is Kubernetes, `aks-engine orchestrators` is probably due for a rewrite. Given that my assumption is this command was used to inform what versions of Kubernetes are supported (as well as whether they have Windows support), should it be re-written to `aks-engine available-versions` or something similar?

Additional side note: I haven't been able to find any information on `aks-engine orchestrators` in the docs (other than an excerpt from [one of the examples outside the docs](https://github.com/Azure/aks-engine/blob/7966f8f452c621444af90f69e6bd7f43a5c1682a/examples/k8s-upgrade/README.md)), so as a newcomer I'm not sure how I'm supposed to supply that information back to the API model to get the version I choose. I'd love to see a topic guide on using `aks-engine orchestrators` to select a particular Kubernetes version for my deployment... Perhaps as a new topic guide on how to use `aks-engine upgrade`?

CecileRobertMichon:
I think the little docs with have on the topic are in the upgrade readme https://github.com/Azure/aks-engine/blob/master/examples/k8s-upgrade/README.md#L20 but totally agree these could be more comprehensive. Also agree that `orchestrators` isn't the right verb anymore. Slight preference for `get-versions` over `available-versions` since it's the same verb aks already uses.

jackfrancis:
I vote for replacing `aks-engine orchestrators --orchestrator Kubernetes` with `aks-engine --k8s-versions`.

I also vote that we add an alias `aks-engine upgrade --k8s-versions` for folks who are upgrading and who do something like `aks-engine upgrade --help` as a discovery path towards finding out "how do I figure out which versions of Kubernetes I can upgrade to?"

mboersma:
I like `aks-engine get-versions` because it's parallel to `az aks get-versions`.

PR #356, opened 2019-01-23
docs: rewrite upgrade documentation

### LINKED ISSUE #285: Day 2 Ops documentation: aks-engine upgrade
We want a narrative document that clearly outlines the present reality of `aks-engine upgrade` and recommends operational scenarios for adopting it as a production quality tool.

- What does `aks-engine upgrade` actually do under the hood?
  - expects a cluster configuration (apimodel.json) that conforms to the current state of the cluster
  - assumes Azure IaaS resources are organized/named in particular ways in accordance with their original provenance as aks-engine originating stuff
  - uses the "idemotent ARM deployment" pattern to operate operations against pre-existing resource groups
- Definitive guidance to aid customers who want to use the above tooling to their advantage
  - what can/can't they do to their Azure IaaS resources
  - what can go wrong
  - this should match up w/ current E2E test coverage to validate the long-term viability of its truth

CecileRobertMichon:
**Reason for Change**:

**Issue Fixed**:

Fixes #285 

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+  --location westus \
+  --resource-group test-upgrade \
+  --upgrade-version 1.8.7 \
+  --client-id xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx \
+  --client-secret xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+```
+
+### What can go wrong
+
+By its nature, the upgrade operation is long running and potentially could fail for various reasons, such as temporary lack of resources, etc. In this case, rerun the command. The `upgrade` command is idempotent, and will pick up execution from the point it failed on.
```
bacongobbler: Would this be something handy to put in the [troubleshooting docs](https://github.com/Azure/aks-engine/blob/master/docs/howto/troubleshooting.md) as well?

CecileRobertMichon: Possibly. Maybe I should just add a "failed upgrade" in the troubleshooting doc that redirects to this page instead of repeating this section.

docs/topics/scale.md
```diff
@@ -1,16 +1,14 @@
...
-
-This guide assumes you already have deployed a cluster using acs engine. For more details on how to do that see [deploy](./deploy.md).
+This guide assumes you already have deployed a cluster using aks-engine. For more details on how to do that see [deploy](../tutorials/deploy.md).
 
 ## Scale
 
-After a cluster has been deployed using acs engine the cluster can be interacted further by using the scale command. The scale command can add more nodes to an existing node pool or remove them. Nodes will always be added or removed from the end of the agent pool. Nodes will be cordoned and drained before deletion.
+The `aks-engine scale` command can increase or decrease the number of nodes in an existing agent pool in an `aks-engine` Kubernetes cluster. Nodes will always be added or removed from the end of the agent pool. Nodes will be cordoned and drained before deletion.
 
 This guide will assume you have a cluster deployed and the output for the deployed cluster is stored at _output/mycluster. It will also assume there is a node pool named "agentpool1" in your cluster. AKS Engine will default to storing the output at ./_output/dns-prefix from where the aks-engine command was ran.
```
jackfrancis: while we're here let's issue a grammar citation for "was ran", and correct it to "was run". :)

jackfrancis: Also, let's change `./_output/dns-prefix` to `./_output/<dnsPrefix>` to more clearly indicate we're talking about the variable `dnsPrefix` that's in the api model input.

jackfrancis: Let's modify the below paragraph. From:

```
This command will look the deployment directory to find info about the cluster currently deployed. Then it will generate and deploy a template deployment to update the cluster and add the new nodes. When it is done it will update the cluster definition in the deployment directory's apimodel.json to reflect the new node count.```

To:

```
This command will re-use the apimodel.json inside the output directory as input for a new ARM template deployment that will execute the scaling operation against the desired agent pool. When it the scaling operation is done it will update the cluster definition in that same apimodel.json file to reflect the new node count (and thus the updated, current cluster configuration.
```

Something like that

Comment (jackfrancis) [reply docs/topics/scale.md]: Also let's change the parameters table so that the params are expressed using the `--param-name` (i.e., the `--` prefix), for clarity

File: docs/topics/upgrade.md [hunk 3]
```diff
@@ -0,0 +1,77 @@
...
+
+This document provides guidance on how to upgrade the Kubernetes version for an existing `aks-engine` cluster and recommendations for adopting `aks-engine upgrade` as a production quality tool.
+
+<a name="pre-requirements"></a>
+
+### Know before you go
+
+In order to ensure that your `aks-engine upgrade` operation runs smoothly, there are a few things you should be aware of before getting started.
+
+1) You will need access to the original `_output/` directory files that were generated during cluster deployment in order to run `upgrade`.  `aks-engine` will use the `--deployment-dir` argument provided in the upgrade command to introspect the Azure Resource Manager (ARM) template and parameters to determine the cluster's current Kubernetes version and its current state.
```
jackfrancis: Strictly speaking, aks-engine is introspecting the apimodel.json, right? So we should probably clarify that.

"... to introspect the `apimodel.json` cluster configuration, in order to determine the cluster's current Kubernetes version, as well as all other cluster configuration data as defined by aks-engine during the last time that aks-engine was used to create, scale, or upgrade the cluster."

Let's leave out the determination of state, as that's not 100% correct.

docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+
+2) `aks-engine upgrade` expects a cluster configuration that conforms to the current state of the cluster. In other words, the Azure resources deployed by `aks-engine` should be in the same state that when they were originally created by `aks-engine`. If you perform manual operations on your Azure IaaS resources (other than `aks-engine scale` and `aks-engine upgrade`) you risk not being able to upgrade your cluster successfully as `aks-engine` won't be able to reconcile the cluster's current state. This includes naming of resources; `aks-engine` relies on some resources (such as VMs) to be named in accordance with the original `aks-engine` deployment.
```
jackfrancis: How about this for additional literal accuracy:

"`aks-engine upgrade` expects a cluster configuration that will produce ARM templates that can be deployed as an incremental deployment on top of the existing resource group" (<-- validate that this is true, ref: https://docs.microsoft.com/en-us/azure/azure-resource-manager/deployment-modes)

Basically, let's try to succinctly state was is actually happening. aks-engine is a convenience that is implementing an ARM template deployment against a resource group that has been created in the past as a result of a prior ARM template deployment.

jackfrancis: "...you risk not being able to upgrade your cluster successfully, as the `aks-engine`-generated ARM template won't be reconcilable against the state of the Azure resources that reside in the resource group."

Also maybe add something like this to the end:

"In summary, the set of Azure resources in the resource group are mutually reconcilable only if they have been exclusively created and managed as the result of a series of successive ARM template deployments originating from `aks-engine`."

docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+
+3) `aks-engine upgrade` allows one subsequent minor version upgrade at a time, as long as the next minor version is supported. If the next minor version is deprecated, `aks-engine` allows upgrade to the next supported minor version.
```
jackfrancis: Can we make this more comprehensive?

"`aks-engine upgrade` allows upgrading the Kubernetes version to any `aks-engine`-supported patch release in the current minor release channel that is greater than the current version on the cluster (e.g., from `1.11.4 to 1.11.5`), or to the next `aks-engine`-supported minor version (e.g., from `1.11.4` to `1.12.5`). In practice, the "next `aks-engine`-supported minor version will commonly be a single minor version ahead of the current cluster version. However, if the cluster has not been upgraded in a significant amount of time, the "next" minor version may have actually been deprecated by `aks-engine`. In such a case, your long-lived cluster will be upgradable to the nearest, supported minor version that `aks-engine` supports at the time of upgrade (e.g., from `1.7.16` to `1.9.11`).

docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+
+To get the list of all available Kubernetes versions and upgrades, run the `orchestrators` command and specify Kubernetes orchestrator type:
+
+```bash
+./bin/aks-engine orchestrators --orchestrator Kubernetes
```
jackfrancis: Do we have an issue tracking "use --orchestrator Kubernetes by default so a user doesn't ever have to pass that in"?

docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+```
+
+To get the information specific to the cluster, provide its current Kubernetes version in the `version` arg:
```
jackfrancis: s/To get the information specific to the cluster/To discover the versions of Kubernetes that your particular cluster version is upgradable to/

docs/topics/upgrade.md
```diff
@@ -0,0 +1,77 @@
...
+```bash
+./bin/aks-engine orchestrators --orchestrator Kubernetes --version 1.11.5
+```
+
+4) If using `aks-engine upgrade` in production, it is recommended to run a dry run test on an identical cluster before performing the upgrade, especially if the cluster configuration is "unusual", or in other words differs significantly from defaults. The reason for this is that `aks-engine` supports many different cluster configurations and the extent of E2E testing that the AKS-Engine team runs cannot possibly cover every single configuration out there. Therefore, it is recommended that you make sure that your specific cluster configuration works with the existing upgrade implementation before starting this long-running operation.
```
jackfrancis: s/it is recommended to run a dry run test on an identical cluster/it is recommended to stage an upgrade test on an cluster that was built to the same specifications (built with the same cluster configuration + aks-engine version) as your production cluster/

jackfrancis: s/"unusual"/"interesting"/

docs/topics/upgrade.md
CecileRobertMichon: no, for upgrade aks-engine introspects the `azuredeploy.json` and the parameters file

CecileRobertMichon: will change the state part of the sentence

jackfrancis: Hmm, that's surprising. Does it reconcile any differences that it detects between apimodel.json and azuredeploy.json?

CecileRobertMichon: correction ** it does use the apimodel to load the container service and saves the file at the end with updated k8s version but it does use the azuredeploy json and parameters for fetching the name suffix

jackfrancis: Just to be clear, upgrade is doing some local data gathering so it knows "what the resource profile in the RG looks like" <-- something like that, right?

My question is: *should* it be relying upon the template output for that data gathering pre-step? It seems to be an anti-pattern. What's preventing upgrade from referring only to the apimodel.json (i.e., refererring only to a data representation that can be coupled with an in-code, strongly typed data model)?

docs/topics/upgrade.md
CecileRobertMichon: we do

CecileRobertMichon: https://github.com/Azure/aks-engine/issues/339

docs/topics/upgrade.md
CecileRobertMichon: I don't understand your first comment "Basically, let's try to succinctly state was is actually happening"

docs/topics/upgrade.md
CecileRobertMichon: https://github.com/Azure/aks-engine/issues/361

docs/topics/upgrade.md
```diff
@@ -0,0 +1,81 @@
...
+        ./bin/aks-engine orchestrators --orchestrator Kubernetes
+        ```
+
+    To get the versions of Kubernetes that your particular cluster version is upgradable to, provide its current Kubernetes version in the `version` arg:
+
+        ```bash
+        ./bin/aks-engine orchestrators --orchestrator Kubernetes --version 1.11.5
+        ```
+
+4) If using `aks-engine upgrade` in production, it is recommended to stage an upgrade test on an cluster that was built to the same specifications (built with the same cluster configuration + `aks-engine` version) as your production cluster before performing the upgrade, especially if the cluster configuration is "interesting", or in other words differs significantly from defaults. The reason for this is that `aks-engine` supports many different cluster configurations and the extent of E2E testing that the AKS-Engine team runs cannot possibly cover every single configuration out there. Therefore, it is recommended that you make sure that your  specific cluster configuration works with the existing upgrade implementation before starting this long-running operation.
```
jackfrancis: let's s/cannot possibly cover every single configuration out there/cannot practically cover every single configuration out there

docs/topics/README.md
```diff
@@ -12,8 +12,9 @@ Introductions to all the key parts of AKS-Engine you’ll need to know.
 - [For Kubernetes Developers](kubernetes-developers.md)
 - [Kubernetes Walkthrough](kubernetes-walkthrough.md)
 - [Monitoring Kubernetes Clusters](monitoring.md)
-- [Scaling Kubernetes Clusters](scaling.md)
+- [Scaling Kubernetes Clusters](scale.md)
 - [Service Principals](service-principals.md)
+- [Upgrading Kubernetes Clusters](upgrade.md)
```
jackfrancis: Let's move this to the top of the list.

docs/topics/README.md
```diff
@@ -12,8 +12,9 @@ Introductions to all the key parts of AKS-Engine you’ll need to know.
 - [For Kubernetes Developers](kubernetes-developers.md)
 - [Kubernetes Walkthrough](kubernetes-walkthrough.md)
 - [Monitoring Kubernetes Clusters](monitoring.md)
-- [Scaling Kubernetes Clusters](scaling.md)
+- [Scaling Kubernetes Clusters](scale.md)
```
jackfrancis: And then let's move scaling below upgrading.

docs/topics/README.md
CecileRobertMichon: I think the list is meant to be alphabetical

docs/topics/README.md
CecileRobertMichon: same comment, list is in alphabetical order

docs/topics/README.md
jackfrancis: Indeed. And the result of that is that we're sort of obscuring upgrade/scale docs, which (IMO) should be emphasized more prominently. Let's give them the clickbait level up they deserve.

CecileRobertMichon: I don't agree here. One could argue `Features` or `Cluster Definition` are as if not more important then upgrade and scale. I think @bacongobbler's new organization of topics actually makes it easier for users to find what they are looking for. Keeping the list organized ensures it doesn't rapidly become a long list of unrelated items so I think alphabetical order makes sense here.

CecileRobertMichon: I also "advertised" the upgrade doc in the troubleshooting doc, we could do the same for scale

PR #359, opened 2019-01-23
fix: add getter methods for managed identity

tariq1890:
Adding helper methods to increase reuse of code.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1290,6 +1290,32 @@ func (k *KubernetesConfig) IsRBACEnabled() bool {
 	return false
 }
 
+// UserAssignedIDEnabled checks if the user assigned ID is enabled or not.
+func (k *KubernetesConfig) UserAssignedIDEnabled() bool {
+	return k.UseManagedIdentity && k.UserAssignedID != ""
+}
+
+// UserAssignedClientIDEnabled checks if the user assigned client ID is enabled or not.
+func (k *KubernetesConfig) UserAssignedClientIDEnabled() bool {
+	return k.UseManagedIdentity && k.UserAssignedClientID != ""
+}
+
+// GetUserAssignedID returns the user assigned ID if it is enabled.
+func (k *KubernetesConfig) GetUserAssignedID() string {
+	if k.UserAssignedIDEnabled() {
+		return k.UserAssignedID
+	}
+	return ""
+}
+
+// GetUserAssignedClientID returns the user assigned client ID if it is enabled.
+func (k *KubernetesConfig) GetUserAssignedClientID() string {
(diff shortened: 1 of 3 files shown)


jackfrancis:
Can we unit test these?

tariq1890:
@jackfrancis Sure thing!

pkg/api/types.go
```diff
@@ -1290,6 +1290,32 @@ func (k *KubernetesConfig) IsRBACEnabled() bool {
 	return false
 }
 
+// UserAssignedIDEnabled checks if the user assigned ID is enabled or not.
+func (k *KubernetesConfig) UserAssignedIDEnabled() bool {
```
CecileRobertMichon: Can you incorporate these new functions in existing unit tests if they aren't already getting covered?

tariq1890: Sure @CecileRobertMichon . I will put out another PR to start doing that.

CecileRobertMichon: what do you mean by "rationalize our test paths"?

pkg/api/types_test.go
```diff
@@ -2737,3 +2737,45 @@ func TestIsFeatureEnabled(t *testing.T) {
...
+		t.Errorf("expected user assigned ID to be %s, but got %s", expected, k.GetUserAssignedID())
+	}
+
+	k = KubernetesConfig{
+		UseManagedIdentity: false,
+		UserAssignedID:     "fooID",
+	}
+
+	if k.GetUserAssignedID() != "" {
+		t.Errorf("expected user assigned ID to be empty when useManagedIdentity is set to false")
```
jackfrancis: Should we file a backlog issue to add validation for these pathological scenarios?

pkg/api/types.go
jackfrancis: let's add a unit test for `UserAssignedIDEnabled`

pkg/api/types.go
```diff
@@ -1290,6 +1290,32 @@ func (k *KubernetesConfig) IsRBACEnabled() bool {
...
+	return k.UseManagedIdentity && k.UserAssignedID != ""
+}
+
+// UserAssignedClientIDEnabled checks if the user assigned client ID is enabled or not.
+func (k *KubernetesConfig) UserAssignedClientIDEnabled() bool {
```
jackfrancis: let's add a unit test for `UserAssignedClientIDEnabled`

pkg/api/types_test.go
tariq1890: Validation is already present. This test case just encompasses the behaviour of the method that we had just migrated in this PR

jackfrancis: cool, thanks for confirming!

PR #367, opened 2019-01-24
feat: enable vmss overprovision, do not run extensions on overprovisioned VMs

jackfrancis:
**Reason for Change**:

Enable overprovisioning for VMSS, and enable the feature to not invoke extensions on overprovisoned VMs; thus solving the issue where k8s nodes "disappear" from a cluster shortly after cluster creation time.

This is a configurable option via the `vmssOverProvisioningEnabled` setting on a per-agent pool basis.

Includes a bump to the API version (from `2018-06-01` to `2018-10-01`) in order to fulfill the requirements of the `doNotRunExtensionsOnOverprovisionedVMs` VMSS Overprovisioning feature. That API version is used by the following resource types:

- `"type": "Microsoft.Compute/availabilitySets"`
- `"type": "Microsoft.Compute/virtualMachines"`
- `"type": "Microsoft.Compute/virtualMachines/extensions"`
- `"type": "Microsoft.Compute/virtualMachineScaleSets"`
- `"type": "Microsoft.Compute/images"`

Note: the above API version bump will only be present if the `vmssOverProvisioningEnabled` agent pool configuration option is set to `true` for a VMSS agent pool profile.

`doNotRunExtensionsOnOverprovisionedVMs ` is currently in private preview, so users will have to opt into the `“Microsoft.Compute/OverprovisionedExtensionDelay”` feature in order to use this new AKS Engine `vmssOverProvisioningEnabled` configuration option, until this feature graduates to general availability.

Also includes updates to agent pool documentation to clarify which configs are VMSS-only.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/kubernetesmastervars.t
```diff
@@ -93,7 +93,7 @@
     "apiVersionNetwork": "2017-10-01",
     "environmentJSON": "{{GetCustomEnvironmentJSON}}",
 {{ else }}
-    "apiVersionCompute": "2018-06-01",
+    "apiVersionCompute": "2018-10-01",
```
slack: Do we know how this change interacts with the cluster autoscaler? If the VMSS was created at this API version, do all subsequent updates need to come through this API as well?

pkg/api/vlabs/validate.go
```diff
@@ -802,16 +808,6 @@ func validateVMSS(o *OrchestratorProfile, isUpdate bool, storageProfile string)
 		if sv.LT(minVersion) {
 			return errors.Errorf("VirtualMachineScaleSets are only available in Kubernetes version %s or greater. Please set \"orchestratorVersion\" to %s or above", minVersion.String(), minVersion.String())
 		}
-		// validation for instanceMetadata using VMSS with Kubernetes
```
mboersma: This looks unrelated to the rest of the PR. Is this just tech debt removal or am I missing something?

parts/k8s/kubernetesmastervars.t
jackfrancis: Good catch, this change should be wrapped inside `{{if .VMSSOverProvisioningEnabled}}` for now, to be safe.

pkg/api/vlabs/validate.go
jackfrancis: Yeah, possibly rebase bogon, addressing.

PR #381, opened 2019-01-25
docs: upgrade + cluster-autoscaler notes

jackfrancis:
**Reason for Change**:

Document current issues and limitations of `aks-engine upgrade` and `cluster-autoscaler`

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/upgrade.md
+++ b/docs/topics/upgrade.md
@@ -89,9 +89,3 @@ There are known limitations with VMSS cluster-autoscaler scenarios and upgrade. 
 ### Cluster-autoscaler + Availability Set
 
 We don't recommend using `aks-engine upgrade` on clusters that have Availability Set (non-VMSS) agent pools `cluster-autoscaler` at this time.
-
-### Cluster-autoscaler + Dynamic Agent Pools
-
-If you manually add/remove agent pools on your `aks-engine`-created cluster over time, we don't recommend using `aks-engine upgrade` at this time.
-
-
--- a/docs/topics/upgrade.md
+++ b/docs/topics/upgrade.md
@@ -84,29 +84,10 @@ The upgrade operation is long running, and for large clusters, more susceptible 
 
 ### Cluster-autoscaler + VMSS
 
-At present, the Azure cloudprovider cluster-autoscaler implementation for VMSS relies upon the original ARM template deployment specification to inform the Azure IaaS (VM, NIC, CustomScriptExtension, etc) configuration to scale out new nodes.
-
-Because `aks-engine upgrade` also employs ARM template deployments to evolve the cluster state going forward, as soon as you upgrade your cluster, the cluster-autoscaler will no longer scale out your cluster according to the latest version. You will see cluster-autoscaler scale out scenarios that look like this one:
-
-```
-$ kubectl get nodes
-NAME                                STATUS    ROLES     AGE       VERSION
-k8s-agentpool-27988949-vmss00000a   Ready     agent     1h        v1.11.5
-k8s-agentpool-27988949-vmss00000m   Ready     agent     1m        v1.10.12
-k8s-agentpool-27988949-vmss00000n   Ready     agent     1m        v1.10.12
-k8s-agentpool-27988949-vmss00000o   Ready     agent     1m        v1.10.12
-k8s-agentpool-27988949-vmss00000p   Ready     agent     1m        v1.10.12
-k8s-agentpool-27988949-vmss00000q   Ready     agent     1m        v1.10.12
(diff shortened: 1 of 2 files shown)


jackfrancis:
@feiskyer Can you kindly read through my initial description of upgrade + VMSS + cluster-autoscaler and comment if it's an accurate summary of the current situation? Thanks!

docs/topics/upgrade.md
```diff
@@ -76,6 +76,37 @@ For example,
...
+
+### Cluster-autoscaler + VMSS
+
+At present, the Azure cloudprovider cluster-autoscaler implementation for VMSS relies upon the original ARM template deployment specification to inform the Azure IaaS (VM, NIC, CustomScriptExtension, etc) configuration to scale out new nodes.
```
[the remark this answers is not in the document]
jackfrancis: In practice, this is not the case. The `kubectl get nodes` output I pasted below is from a vmss upgrade then cluster-autoscaler event. So apparently the answer is "no, `aks-engine upgrade` does not update the model of the vmss itself".

docs/topics/upgrade.md
```diff
@@ -76,6 +76,37 @@ For example,
...
 ```
 
-### What can go wrong
+## Known Limitations
+
+### Manual reconciliation
+
+The upgrade operation is long running, and for large clusters, more susceptible to single operational failures. This is based on the design principle of upgrade enumerating, one-at-a-time, through each node in the cluster. A transient Azure resource allocation error could thus interrupt the successful progression of the overall transaction. At present, the upgrade operation is implemented to "fail fast"; and so, if a well formed upgrade operation fails before completing, it can be manually retried by invoking the exact same command line arguments as were sent originally. The upgrade operation will enumerate through the cluster nodes, skipping any nodes that have already been upgraded to the desired Kubernetes version. Those nodes that match the *original* Kubernetes version will then, one-at-a-time, be cordon and drained, and upgraded to the desired version. Put another way, an upgrade command is designed to be idempotent across retry scenarios.
+
+### Cluster-autoscaler + VMSS
```
CecileRobertMichon: Let add a link/reference to this doc in the cluster-autoscaler doc examples/addons/cluster-autoscaler/README.md

feiskyer: yep, we can add one after this is merged.

docs/topics/upgrade.md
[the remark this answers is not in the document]
feiskyer: OK, then scale up for VMSS would still use the original old model. Do you have any ideas to make CA also works for upgraded clusters?

docs/topics/upgrade.md
CecileRobertMichon: why not add it in the same PR?

feiskyer: sorry, I mean add a link in cluster-autoscaler repo back to here. 

And yep, in examples, it could be added in the same PR.

docs/topics/upgrade.md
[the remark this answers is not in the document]
jackfrancis: O.K., so I have evidence that what I said is not true. Perhaps I was testing a VMAS agent pool. I will upgrade my VMSS test cluster a few more times and validate that cluster-autoscaler continues to respect the versions as they move forward.

jackfrancis:
@CecileRobertMichon this is ready for a re-review

docs/topics/upgrade.md
```diff
@@ -76,6 +76,22 @@ For example,
...
+
+There are known limitations with VMSS cluster-autoscaler scenarios and upgrade. Our current guidance is not to use `aks-engine upgrade` on clusters with `cluster-autoscaler` functionality. See [here](https://github.com/Azure/aks-engine/issues/400) to get more information and to track progress of the issues related to these limitations.
+
+### Cluster-autoscaler + Availability Set
+
+We don't recommend using `aks-engine upgrade` on clusters that have Availability Set (non-VMSS) agent pools `cluster-autoscaler` at this time.
+
+### Cluster-autoscaler + Dynamic Agent Pools
+
+If you manually add/remove agent pools on your `aks-engine`-created cluster over time, we don't recommend using `aks-engine upgrade` at this time.
```
CecileRobertMichon: Since we don't support adding/removing agent pools + upgrade in _any_ scenario (with or without autoscaler) I wouldn't point that out as a specific limitation since it might be misleading users to believe that we _do_ support "Dynamic Agent Pools" without cluster-autoscaler.

I think this point is already covered by requirement number 2 "do not upgrade your cluster if you've made changes outside of the upgrade/scale flows".

PR #386, opened 2019-01-25
feat: Support dynamic AzureEnvironmentSpecConfig

### LINKED ISSUE #278: Support dynamic AzureEnvironmentSpecConfig
No
---

Feature Request
---

the last version
---

Go-autorest pre-defined 4 environments (ChinaCloud, GermanCloud, PublicCloud, USGovernmentCloud),
The environment will have the endpoints for Arm, Graph, etc.

AKS engine pre-defined 4 corresponding AzureEnvironmentSpecConfig.
AzureEnvironmentSpecConfig will contain the endpoint to dependent docker images and binaries to set up K8s cluster.

In AzureStack, the AzureEnvironmentSpecConfig need to be customized to point to different endpoints for dependent docker images and binaries

e.g. Azure Stack will setup the docker repository On-Perm, then the docker repository need to be input from customer.

honcao:
**Reason for Change**:

Feature Change for #278 

**Issue Fixed**:

Fix #278

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


pkg/api/azenvtypes.go
```diff
@@ -7,60 +7,60 @@ import "fmt"
...
-	DCOS188BootstrapDownloadURL     string
-	DCOS190BootstrapDownloadURL     string
-	DCOS198BootstrapDownloadURL     string
-	DCOS110BootstrapDownloadURL     string
-	DCOS111BootstrapDownloadURL     string
-	DCOSWindowsBootstrapDownloadURL string
-	DcosRepositoryURL               string // For custom install, for example CI, need these three addributes
-	DcosClusterPackageListID        string // the id of the package list file
-	DcosProviderPackageID           string // the id of the dcos-provider-xxx package
+	DCOS188BootstrapDownloadURL     string `json:"dcos188BootstrapDownloadURL,omitempty"`
```
jackfrancis: Is there any way we can avoid updating the DCOS surface area here? Does the way we're composing types require this?

honcao: We will only support k8s on AzureStack, I will remove that.

honcao:
@jackfrancis Thanks to review the code. I resolve the your comments and E2E test are passed now with the latest code.

pkg/api/defaults.go
```diff
@@ -521,7 +602,7 @@ func (p *Properties) setDefaultCerts() (bool, []net.IP, error) {
 
 	var azureProdFQDNs []string
 	for _, location := range helpers.GetAzureLocations() {
-		azureProdFQDNs = append(azureProdFQDNs, FormatAzureProdFQDNByLocation(p.MasterProfile.DNSPrefix, location))
+		azureProdFQDNs = append(azureProdFQDNs, FormatAzureProdFQDNByLocation(p.MasterProfile.DNSPrefix, location, p.GetCustomCloudName()))
```
jackfrancis: Do we want to append the `p.GetCustomCloudName()` return value when we're not in an Azure Stack context? (e.g., the return value will be `""` in that case)

jackfrancis: Nevermind I see that addition has been added as the new arg to the `FormatAzureProdFQDNByLocation` func.

pkg/api/defaults.go
```diff
@@ -508,6 +512,83 @@ func (p *Properties) setHostedMasterProfileDefaults() {
...
+func (p *Properties) setCustomCloudProfileDefaults() {
+
+	if p.IsAzureStackCloud() {
+		//Set default value for ResourceManagerVMDNSSuffix
+		AzureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
+		AzureCloudSpecEnvMap[AzureStackCloud] = AzureStackCloudSpec
+
+		// Use the custom input to overwrite the default values in AzureStackCloudSpec
+		if p.CustomCloudProfile.AzureEnvironmentSpecConfig != nil {
+			if len(p.CustomCloudProfile.AzureEnvironmentSpecConfig.CloudName) != 0 {
```
jackfrancis: @tariq1890 Any thoughts on len != 0 vs string != "" ? Not having researched, I would imagine you're using the  length non-equality check because it's more efficient @honcao ? If so, why not len(string) > 0. Wouldn't the gt/lt check be more efficient compared to an non-equality check?

tariq1890: I think either are fine. We can go with what we use usually do in aks-engine for consistency's sake. I am more of a `string !=""` guy personally.

jackfrancis: Great, let's prefer `!= ""` for readability and more clear communication of intent

pkg/api/azenvtypes.go
```diff
@@ -7,18 +7,18 @@ import "fmt"
...
-	CloudName            string
-	DockerSpecConfig     DockerSpecConfig
-	KubernetesSpecConfig KubernetesSpecConfig
-	DCOSSpecConfig       DCOSSpecConfig
-	EndpointConfig       AzureEndpointConfig
-	OSImageConfig        map[Distro]AzureOSImageConfig
+	CloudName            string                        `json:"cloudName,omitempty"`
+	DockerSpecConfig     DockerSpecConfig              `json:"dockerSpecConfig,omitempty"`
+	KubernetesSpecConfig KubernetesSpecConfig          `json:"kubernetesSpecConfig,omitempty"`
+	DCOSSpecConfig       DCOSSpecConfig                `json:"-"`
```
jackfrancis: Is `json:"-"` preferable to simply leaving out the struct tag altogether?

pkg/api/vlabs/const.go
```diff
@@ -107,3 +107,36 @@ const (
...
+	//AzureChinaCloudDCOSBootstrapDownloadURL is the China specific DCOS package download url.
+	AzureChinaCloudDCOSBootstrapDownloadURL = "https://acsengine.blob.core.chinacloudapi.cn/dcos/%s.bootstrap.tar.xz"
+	//AzureEdgeDCOSWindowsBootstrapDownloadURL
+)
+
+const (
+	// AzureCniPluginVerLinux specifies version of Azure CNI plugin, which has been mirrored from
+	// https://github.com/Azure/azure-container-networking/releases/download/${AZURE_PLUGIN_VER}/azure-vnet-cni-linux-amd64-${AZURE_PLUGIN_VER}.tgz
+	// to https://acs-mirror.azureedge.net/cni
+	AzureCniPluginVerLinux = "v1.0.15"
```
jackfrancis: This should be v1.0.16

pkg/api/vlabs/const.go
```diff
@@ -107,3 +107,36 @@ const (
...
+	// AzureCniPluginVerWindows specifies version of Azure CNI plugin, which has been mirrored from
+	// https://github.com/Azure/azure-container-networking/releases/download/${AZURE_PLUGIN_VER}/azure-vnet-cni-windows-amd64-${AZURE_PLUGIN_VER}.tgz
+	// to https://acs-mirror.azureedge.net/cni
+	AzureCniPluginVerWindows = "v1.0.15"
```
jackfrancis: This should be v1.0.16

pkg/api/vlabs/azenvtypes.go
```diff
@@ -0,0 +1,287 @@
...
+
+//AzureOSImageConfig describes an Azure OS image
+type AzureOSImageConfig struct {
+	ImageOffer     string `json:"imageOffer,omitempty"`
+	ImageSku       string `json:"imageSku,omitempty"`
+	ImagePublisher string `json:"imagePublisher,omitempty"`
+	ImageVersion   string `json:"imageVersion,omitempty"`
+}
+
+var (
```
jackfrancis: Where are these vlabs defaults actually consumed?

jackfrancis:
@honcao I understand why we need to define new types for these configurations in the `vlabs` type spec, as we're now going to be accepting them as configurable input. I don't understand why we need to copy over all the default values, however, as we can re-use the existing defaults flow that happens after the `vlabs` type has been converted to the superset api type, right?

Am I missing something?

pkg/api/vlabs/const.go
honcao: in the setCustomCloudProfileDefaults() of src/github.com/Azure/aks-engine/pkg/api/defaults.go

if CustomCloudProfile is in config, but AzureEnvironmentSpecConfig is empty, 

The default AzureStackCloudSpec will be used, it will persist into the apimobel.json 

		} else {
			customCloudDefault := AzureCloudSpecEnvMap[AzureStackCloud]
			p.CustomCloudProfile.AzureEnvironmentSpecConfig = &customCloudDefault
		}

honcao:
@jackfrancis , you are correct, we do not need all the defaults in azenvtypes.go, only AzureStackCloudSpec is required. I will update the code.

pkg/api/defaults.go
```diff
@@ -508,6 +512,83 @@ func (p *Properties) setHostedMasterProfileDefaults() {
 	p.HostedMasterProfile.Subnet = DefaultKubernetesMasterSubnet
 }
 
+func (p *Properties) setCustomCloudProfileDefaults() {
+
+	if p.IsAzureStackCloud() {
+		//Set default value for ResourceManagerVMDNSSuffix
+		AzureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
```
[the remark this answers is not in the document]
honcao: good idea, I will update the code.

pkg/api/azenvtypes.go
honcao: yes,  `json:"-"` , DCOSSpecConfig will not show up in the JSON output during Marshal, 
The field is still show up in JSON output even it leave out the struct tag.

pkg/api/vlabs/const.go
honcao: On the second look, the default value of AzureEnvironmentSpecConfig under vlabs are not used. So I removed them.

pkg/api/vlabs/const.go
honcao: it is not need anymore, I removed it.

pkg/api/vlabs/azenvtypes.go
honcao: good catch, it is not used, so I removed them.

pkg/api/defaults.go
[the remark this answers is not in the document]
honcao: default value of AzureEnvironmentSpecConfig is not need under vlabs, I removed them.

pkg/api/azenvtypes.go
```diff
@@ -263,11 +263,32 @@ var (
 		},
 	}
 
+	//AzureStackCloudSpec is the default configurations for azure stack with public Azure.
```
tariq1890:
```suggestion
	//AzureStackCloudSpec is the default configuration for Azure Stack with public Azure.
```

pkg/api/defaults.go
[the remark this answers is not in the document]
jackfrancis: I don't see the change applied, it still looks like this method is mutating the "global variables" `AzureStackCloudSpec` and `AzureCloudSpecEnvMap`.

@tariq1890 do you have any thoughts/suggestions on a preferable pattern for what we're doing here?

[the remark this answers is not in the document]
honcao: AzureCloudSpecEnvMap is using by the following function.
After setting the default value, the following function will be invoked to populate parameter values and certificate FQDN. 

So until we have other pattern to retrieve the value, I would need overwrite the global variables firstly and it will be used for the following functions. 

@jackfrancis @tariq1890 , Please let me know if you have ideas for different pattern, I am happy to make the changes and veirfy them.

func (cs *ContainerService) GetCloudSpecConfig() AzureEnvironmentSpecConfig {
	targetEnv := helpers.GetCloudTargetEnv(cs.Location, cs.Properties.GetCustomCloudName())
	return AzureCloudSpecEnvMap[targetEnv]
}

func FormatAzureProdFQDNByLocation(fqdnPrefix string, location string, cloudName string) string {
	targetEnv := helpers.GetCloudTargetEnv(location, cloudName)
	FQDNFormat := AzureCloudSpecEnvMap[targetEnv].EndpointConfig.ResourceManagerVMDNSSuffix
	return fmt.Sprintf("%s.%s."+FQDNFormat, fqdnPrefix, location)
}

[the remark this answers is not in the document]
honcao: Sorry, my bad, I missed to address your comments.
Code is updated right now.

pkg/api/defaults.go
```diff
@@ -508,6 +512,101 @@ func (p *Properties) setHostedMasterProfileDefaults() {
...
+			}
+			//OSImageConfig
+			azureStackCloudSpec.OSImageConfig = make(map[Distro]AzureOSImageConfig)
+			for k, v := range p.CustomCloudProfile.AzureEnvironmentSpecConfig.OSImageConfig {
+				azureStackCloudSpec.OSImageConfig[k] = v
+			}
+		} else {
+			p.CustomCloudProfile.AzureEnvironmentSpecConfig = &azureStackCloudSpec
+		}
+		AzureCloudSpecEnvMap[AzureStackCloud] = azureStackCloudSpec
```
jackfrancis: Thanks @honcao.

This is not ideal, but I think it works. @tariq1890 @CecileRobertMichon @mboersma thoughts here?

Basically, `AzureCloudSpecEnvMap[AzureStackCloud]` acts as a kind of global configuration for the Azure Stack context. Because we've never modified it, it kind of exists statically, in a const-like form (i.e., `AzureCloudSpecEnvMap`). Now that we're making that configuration surface area user-configurable, what's a practical iterative step forward?

Should we accept this as-is? Or should we wrap all this inside a light getter/setter interface?

honcao:
@jackfrancis thanks.
Could we merge the PR as-is now? It will unblock me to send PR to support dynamic location.

And I will follow up with team to collect thoughts and refractor the code later?

pkg/api/defaults.go
CecileRobertMichon: I think I'm missing something... what's the reason AzureCloudSpecEnvMap[AzureStackCloud] can't be defined in pkg/api/azenvtypes.go? defaults.go doesn't seem like the right place to do this at first glance

jackfrancis: It is defined in `pkg/api/azenvtypes.go`. This is a "set defaults setter" func for this particular config vector, mirroring other setters for other config vectors.

The question is what to do about the fact that `AzureCloudSpecEnvMap` is now, with this PR, becoming a "mutatable" configuration? Prior to this PR it acts as a "const"-like default hash table.

tariq1890: So this points to how Azure Stack fundamentally differs as a Cloud environment when compared to the others. All the fields which are much immutable in normal Azure  Cloud Envs aren't necessarily invariant in Azure Stack environments. As each Azure  Stack environment is scoped within a customer's private cloud/data center, the fields will differ from customer to customer, unlike the Azure Public Cloud environments where fields are largely immutable and scoped within a broad domain.

pkg/api/defaults.go
```diff
@@ -508,6 +512,101 @@ func (p *Properties) setHostedMasterProfileDefaults() {
...
+		azureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
+
+		// Use the custom input to overwrite the default values in AzureStackCloudSpec
+		if p.CustomCloudProfile.AzureEnvironmentSpecConfig != nil {
+			if p.CustomCloudProfile.AzureEnvironmentSpecConfig.CloudName != "" {
+				azureStackCloudSpec.CloudName = p.CustomCloudProfile.AzureEnvironmentSpecConfig.CloudName
+			}
+
+			// DockerSpecConfig
+			if p.CustomCloudProfile.AzureEnvironmentSpecConfig.DockerSpecConfig.DockerComposeDownloadURL != "" {
```
tariq1890: Can we have a helper method which checks if a string is nil and assigns a default value then? The default value will also be passed as a parameter to the helper method. @honcao

honcao: @tariq1890 , sure, PR is updated

pkg/api/defaults.go
```diff
@@ -508,6 +512,69 @@ func (p *Properties) setHostedMasterProfileDefaults() {
...
+				AKSDockerEngine: DefaultAKSDockerEngineOSImageConfig,
+			},
+		}
+
+		//Set default value for ResourceManagerVMDNSSuffix
+		azureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
+
+		// Use the custom input to overwrite the default values in AzureStackCloudSpec
+		if p.CustomCloudProfile.AzureEnvironmentSpecConfig != nil {
+			helpers.AssignStringIfNotEmpty(&p.CustomCloudProfile.AzureEnvironmentSpecConfig.CloudName, p.CustomCloudProfile.AzureEnvironmentSpecConfig.CloudName)
```
tariq1890: A couple of Issues here

i) We have too many struct field retrievals. We can minimize that by assigning to variables

For eg:-  

i)  `azureStackCloudSpec.DockerSpecConfig` can be assigned to `dockerSpecConfig`
ii) `azureStackCloudSpec.KubernetesSpecConfig` can be assigned to  `kubernetesSpecConfig`
and so on.
and these vars can be reused throughout.

pkg/helpers/helpers.go
```diff
@@ -203,3 +208,13 @@ func GetCloudTargetEnv(location string) string {
 		return "AzurePublicCloud"
 	}
 }
+
+//AssignStringIfNotEmpty set the *dst to src if src is not empty string
+func AssignStringIfNotEmpty(dst *string, src string) {
+	if dst == nil {
```
tariq1890: Question: Is there an assumption that a nil field will never have defaults populated to it?

pkg/api/defaults.go
```diff
@@ -508,6 +512,69 @@ func (p *Properties) setHostedMasterProfileDefaults() {
...
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.ContainerdDownloadURLBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.ContainerdDownloadURLBase)
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.EtcdDownloadURLBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.EtcdDownloadURLBase)
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.KubeBinariesSASURLBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.KubeBinariesSASURLBase)
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.KubernetesImageBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.KubernetesImageBase)
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.NVIDIAImageBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.NVIDIAImageBase)
+			helpers.AssignStringIfNotEmpty(&azureStackCloudSpec.KubernetesSpecConfig.TillerImageBase, p.CustomCloudProfile.AzureEnvironmentSpecConfig.KubernetesSpecConfig.TillerImageBase)
```
tariq1890: Nice-to-have: it would be nice to have a separate private method for each struct inside of `azureStackCloudSpec`.

i.e. - a separate private method for populating `KubernetesSpecConfig` defaults, `DockerSpecConfig` defaults etc.

pkg/helpers/helpers.go
honcao: Yes, in current situation.
All the fields in AzureCloudSpecConfig are not pointer value, so they will never be true.

I add this check to avoid panic if the nil is passed unexpectedly. @tariq1890

honcao: Yes, in current situation.
All the fields in AzureCloudSpecConfig are not pointer value, so they will never be true.

I add this check to avoid panic if the nil is passed unexpectedly. @tariq1890

pkg/helpers/helpers.go
```diff
@@ -189,10 +189,15 @@ func CreateSaveSSH(username, outputDirectory string, s *i18n.Translator) (privat
 
 // GetCloudTargetEnv determines and returns whether the region is a sovereign cloud which
 // have their own data compliance regulations (China/Germany/USGov) or standard
-//  Azure public cloud
-func GetCloudTargetEnv(location string) string {
+// Azure public cloud
+// CustomCloudName is name of environment if customCloudProfile is provided, it will be empty string if customCloudProfile is empty.
+// Because customCloudProfile is empty for deployment for AzurePublicCloud, AzureChinaCloud,AzureGermanCloud,AzureUSGovernmentCloud,
+// The customCloudName value will be empty string for those clouds
+func GetCloudTargetEnv(location, customCloudName string) string {
```
tariq1890: I don't think we should change the method signature of a public method, given that we have other projects that vendor our codebase.

We should just refrain from calling this method  `GetCloudTargetEnv` altogether , if `customCloudName` is non-empty.

pkg/helpers/helpers.go
tariq1890: The title `AssignStringIfNotEmpty` is a bit confusing. The method basically assigns a default string value to a non-nil field that has an Empty string as value.

pkg/api/defaults-apiserver.go
```diff
@@ -80,7 +80,7 @@ func (cs *ContainerService) setAPIServerConfig() {
 		defaultAPIServerConfig["--oidc-groups-claim"] = "groups"
 		defaultAPIServerConfig["--oidc-client-id"] = "spn:" + cs.Properties.AADProfile.ServerAppID
 		issuerHost := "sts.windows.net"
-		if helpers.GetCloudTargetEnv(cs.Location) == "AzureChinaCloud" {
+		if helpers.GetTargetEnv(cs.Location, cs.Properties.GetCustomCloudName()) == "AzureChinaCloud" {
```
tariq1890:
```suggestion
		if helpers.GetTargetEnv(cs) == "AzureChinaCloud" {
```

It would be better to have a single param over here

pkg/helpers/helpers.go
```diff
@@ -203,3 +203,28 @@ func GetCloudTargetEnv(location string) string {
...
+		return "AzureStackCloud"
+	default:
+		return GetCloudTargetEnv(location)
+	}
+}
+
+// GetDefaultStringWithOverwrite returns an string for the default value.
+// If overwrite is not empty, return overwrite
+// If overwrite is empty, return defaultString
+func GetDefaultStringWithOverwrite(defaultString, overwrite string) string {
```
tariq1890:
```suggestion
func EnsureString(val, defaultVal string) string {
```

pkg/engine/params_k8s.go
```diff
@@ -27,8 +27,17 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
...
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
+		hyperKubeImageBase := kubernetesConfig.KubernetesImageBase
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
 
+		if properties.IsAzureStackCloud() {
+			// AzureStack will produce customized HyperKube image.
+			// The hyperKube image and other dependent images are not host on the same repository
+			// So hyperKube image repository is the value of kubernetesConfig.KubernetesImageBase and
+			// other dependent images repository is the value of cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
+			kubernetesImageBase = cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
```
tariq1890: As discussed, we can consider using the `kubernetesConfig.CustomHyperkubeImage` instead

pkg/api/defaults.go
```diff
@@ -508,6 +512,76 @@ func (p *Properties) setHostedMasterProfileDefaults() {
 	p.HostedMasterProfile.Subnet = DefaultKubernetesMasterSubnet
 }
 
+func (p *Properties) setCustomCloudProfileDefaults() {
```
tariq1890: Let's move this method declaration to a new file (like defaults-customcloudprofile.go perhaps?) as `defaults.go` is fairly big as it is.

pkg/api/types.go
```diff
@@ -685,9 +685,10 @@ type V20180331ARMManagedContainerService struct {
 	*v20180331.ManagedCluster
 }
 
-// CustomCloudProfile Represents Azure Environment
+// CustomCloudProfile Represents custom cloud profile
```
tariq1890:
```suggestion
// CustomCloudProfile represents the custom cloud profile
```

pkg/api/types_test.go
```diff
@@ -2817,3 +2880,174 @@ func TestKubernetesConfig_UserAssignedClientIDEnabled(t *testing.T) {
...
...ementEndpoint\\\":\\\"https=//management.azurestackci15.onmicrosoft.com/36f71706-54df-4305-9847-5b038a4cf189\\\",\\\"resourceManagerEndpoint\\\":\\\"https=//management.local.azurestack.external/\\\",\\\"activeDirectoryEndpoint\\\":\\\"https=//login.windows.net/\\\",\\\"galleryEndpoint\\\":\\\"https=//portal.local.azurestack.external=30015/\\\",\\\"keyVaultEndpoint\\\":\\\"https=//vault.azurestack.external/\\\",\\\"graphEndpoint\\\":\\\"https=//graph.windows.net/\\\",\\\"serviceBusEndpoint\\\":\\\"https=//servicebus.azurestack.external/\\\",\\\"batchManagementEndpoint\\\":\\\"https=//batch.azurestack.external/\\\",\\\"storageEndpointSuffix\\\":\\\"core.azurestack.external\\\",\\\"sqlDatabaseDNSSuffix\\\":\\\"database.azurestack.external\\\",\\\"trafficManagerDNSSuffix\\\":\\\"trafficmanager.cn\\\",\\\"keyVaultDNSSuffix\\\":\\\"vault.azurestack.external\\\",\\\"serviceBusEndpointSuffix\\\":\\\"servicebus.azurestack.external\\\",\\\"serviceManagementVMDNSSuffix\\\":\\\"chinacloudapp.cn\\\",\\\"resourceManagerVMDNSSuffix\\\":\\\"cloudapp.azurestack.external\\\",\\\"containerRegistryDNSSuffix\\\":\\\"azurecr.io\\\",\\\"tokenAudience\\\":\\\"https=//management.azurestack.external/\\\"}",
```
tariq1890: Let's use backtick to wrap this string.

pkg/api/types_test.go
```diff
@@ -2817,3 +2880,174 @@ func TestKubernetesConfig_UserAssignedClientIDEnabled(t *testing.T) {
...
+	for _, testcase := range testcases {
+		actual := testcase.properties.GetCustomEnvironmentJSON()
+		if testcase.expected != actual {
+			t.Errorf("Test \"%s\": expected GetCustomEnvironmentJSON() to return %s, but got %s . ", testcase.name, testcase.expected, actual)
+		}
+	}
+}
+
+func getMockPropertiesWithCustomCloudProfile(name string, hasCustomCloudProfile, hasEnvironment, hasAzureEnvironmentSpecConfig bool) Properties {
+	const (
```
tariq1890: Let's use variables instead of const here.

pkg/api/types_test.go
```diff
@@ -2817,3 +2880,174 @@ func TestKubernetesConfig_UserAssignedClientIDEnabled(t *testing.T) {
...
+		managementPortalURL          = "https=//management.local.azurestack.external/"
```
tariq1890:
```suggestion
		managementPortalURL          = "https://management.local.azurestack.external/"
```

pkg/api/defaults.go
honcao: I will work on this in following PR

PR #403, opened 2019-01-30
fix: name windows aks billing extension correctly for hosted master profile

tariq1890:
Naming should be symmetrical with that of linux cases.

### DIFF
--- a/parts/k8s/kuberneteswinagentresourcesvmas.t
+++ b/parts/k8s/kuberneteswinagentresourcesvmas.t
@@ -272,7 +272,7 @@
       ],
       "properties": {
         "publisher": "Microsoft.AKS",
-        "type": "Compute.AKS-Engine.Windows.Billing",
+        "type": {{if IsHostedMaster}}"Compute.AKS.Windows.Billing"{{else}}"Compute.AKS-Engine.Windows.Billing"{{end}},
         "typeHandlerVersion": "1.0",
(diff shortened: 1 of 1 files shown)


parts/k8s/kuberneteswinagentresourcesvmas.t
```diff
@@ -272,7 +272,7 @@
       ],
       "properties": {
         "publisher": "Microsoft.AKS",
-        "type": "Compute.AKS-Engine.Windows.Billing",
+        "type": {{if IsHostedMaster}}"Compute.AKS.Windows.Billing"{{else}}"Compute.AKS-Engine.Windows.Billing"{{end}},
```
CecileRobertMichon: I think the reason we don't need this is because hosted masters don't support Windows

(if this is helping the no template code this PR makes sense but otherwise this if statement is superfluous).

CecileRobertMichon: /hold

tariq1890:
@CecileRobertMichon ^

CecileRobertMichon:
Our templates are already complex with a lot of nested if/else statements so I don't think we should add any additional ones if they aren't needed. Besides, if this is in the name of consistency, we should add it to all template files that use the extension (for example, VMSS templates don't have the AKS case for the same reason). Not saying we should do this, just pointing out that this change doesn't really make sense to me as a standalone change.

PR #410, opened 2019-01-31
revert: JoinControllers system.conf override

jackfrancis:
**Reason for Change**:

From systemd 240 release:

> The JoinControllers= option in system.conf is no longer supported, as
> it didn't work correctly, is hard to support properly, is legacy (as
> the concept only exists on cgroupsv1) and apparently wasn't used.
**Issue Fixed**:

Reverts https://github.com/Azure/acs-engine/pull/3915

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon:
can you add the reference to the PR this is reverting?

CecileRobertMichon:
your commit message still says `fix`, you'll want to change it to `revert:` before merging

jackfrancis:
ref: https://lists.freedesktop.org/archives/systemd-devel/2018-December/041852.html

victornoel:
@jackfrancis is this already running on AKS? I have nodes I created around March 15 and the `JoinControllers` option is still enabled.

This is problematic for me because of this bug in the JVM: https://bugs.openjdk.java.net/browse/JDK-8217766

Basically the JVM can't exploit the cgroup with `JoinControllers` enabled, so all JVM based container running on AKS are using a huge amount of memory compared to the actual limits specified for the kubernetes pod.

jackfrancis:
@victornoel I believe this reversion has not yet landed in AKS. Quickest way to validate would be to build a tiny cluster and do a quick sanity check.

victornoel:
@jackfrancis maybe not the best place to ask, but is there an official place where I can see what version of aks-engine is deployed in AKS and/or information on when it happens?

CecileRobertMichon:
@victornoel each VM the AKS `MC_` resource group should have an AKS Engine Version tag that let's you know what version of aks-engine was used to deploy it.

victornoel:
@CecileRobertMichon indeed, I can see which version is deployed (it is 0.33.2 and 0.33.4 depending on the node). The only way to know which version is going to be deployed for new nodes is to create one?

CecileRobertMichon:
@victornoel I'd say that's the most reliable way. There's also notes about aks-engine minor version updates in [AKS release notes](https://github.com/Azure/AKS/releases) but it doesn't look like there is patch information there.

PR #423, opened 2019-02-01
chore: per-agent pool k8s versions and provisioning state

kkmsft:
- Add per agentpool level kubernetes version and provisioning state.
- Add calls to delete an entire vmss.

**Reason for Change**:

Required work towards multi agentpool operations.

**Issue Fixed**:

These will start getting used by AKS multi agent pool work initially and then used in aks-engine multi agentpool operations - like delete, scale, upgrade.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -486,6 +486,7 @@ type AgentPoolProfile struct {
 	DNSPrefix                           string               `json:"dnsPrefix,omitempty"`
 	OSType                              OSType               `json:"osType,omitempty"`
 	Ports                               []int                `json:"ports,omitempty"`
+	ProvisioningState                   ProvisioningState    `json:"provisioningState,omitempty"`
 	AvailabilityProfile                 string               `json:"availabilityProfile"`
 	ScaleSetPriority                    string               `json:"scaleSetPriority,omitempty"`
 	ScaleSetEvictionPolicy              string               `json:"scaleSetEvictionPolicy,omitempty"`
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
⚠️  PR does not follow commit message convention ⚠️

pkg/api/types.go
```diff
@@ -486,6 +486,7 @@ type AgentPoolProfile struct {
 	DNSPrefix                           string               `json:"dnsPrefix,omitempty"`
 	OSType                              OSType               `json:"osType,omitempty"`
 	Ports                               []int                `json:"ports,omitempty"`
+	ProvisioningState                   ProvisioningState    `json:"provisioningState,omitempty"`
```
tariq1890: For external consumption?

CecileRobertMichon: there's already a provisioningState field in Properties, how is this different?

pkg/api/types.go
```diff
@@ -503,6 +504,7 @@ type AgentPoolProfile struct {
 	PreprovisionExtension               *Extension           `json:"preProvisionExtension"`
 	Extensions                          []Extension          `json:"extensions"`
 	KubernetesConfig                    *KubernetesConfig    `json:"kubernetesConfig,omitempty"`
+	KubernetesVersion                   string               `json:"kubernetesVersion"`
```
CecileRobertMichon: why do we need this? do we expect heterogeneous k8s version support across agentpools?

pkg/api/types.go
kkmsft: you mean - for user consumption in AKS - yes.

pkg/api/types.go
jackfrancis: Just to be consistent with the pre-existing types I have a slight preference for `OrchestratorVersion` here.

jackfrancis:
lgtm

PR #433, opened 2019-02-04
feat: Add ability to enable windows ssh with open ssh install on node

vyta:
**Reason for Change**:

This allows user to have an option to enable ssh into windows nodes, instead of having to use RDP to remote into the node

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


vyta:
cc: @jsturtevant

examples/windows/kubernetes-ssh.json
```diff
@@ -0,0 +1,54 @@
...
+            {
+                "name": "winrm"
+            }
+        ]
+      }
+    ],
+    "windowsProfile": {
+      "adminUsername": "azureuser",
+      "adminPassword": "replacepassword1234$",
+      "sshEnabled": true
```
CecileRobertMichon: let's add this field description to https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md instead of adding a new example file for it

pkg/api/types.go
```diff
@@ -193,6 +193,7 @@ type WindowsProfile struct {
 	WindowsSku            string            `json:"windowsSku"`
 	WindowsDockerVersion  string            `json:"windowsDockerVersion"`
 	Secrets               []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled            bool				`json:"sshEnabled,omitempty"`
```
CecileRobertMichon: indent seems off here

pkg/engine/const.go
```diff
@@ -133,6 +133,7 @@ const (
 	kubernetesWindowsKubeletFunctionsPS1  = "k8s/windowskubeletfunc.ps1"
 	kubernetesWindowsCniFunctionsPS1      = "k8s/windowscnifunc.ps1"
 	kubernetesWindowsAzureCniFunctionsPS1 = "k8s/windowsazurecnifunc.ps1"
+	kubernetesWindowsOpenSSHFunctionPS1	  = "k8s/windowsinstallopensshfunc.ps1"
```
CecileRobertMichon: indent

pkg/api/types_test.go
```diff
@@ -811,6 +811,7 @@ func TestWindowsProfile(t *testing.T) {
 	w = WindowsProfile{
 		WindowsDockerVersion: "18.03.1-ee-3",
 		WindowsSku:           "Datacenter-Core-1809-with-Containers-smalldisk",
+		SSHEnabled:     true,
```
CecileRobertMichon: indent

pkg/api/vlabs/types.go
```diff
@@ -165,6 +165,7 @@ type WindowsProfile struct {
 	WindowsSku            string            `json:"WindowsSku"`
 	WindowsDockerVersion  string            `json:"windowsDockerVersion"`
 	Secrets               []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled            bool				`json:"sshEnabled,omitempty"`
```
CecileRobertMichon: indent

examples/windows/kubernetes-ssh.json
vyta: File removed, edits made to clusterdefinition

parts/k8s/windowsinstallopensshfunc.ps1
```diff
@@ -0,0 +1,45 @@
...
+    Add-Content $adminpath\$adminfile $SSHKey
+
+    Write-Host "Setting required permissions"
+    icacls $adminpath\$adminfile /remove "NT AUTHORITY\Authenticated Users"
+    icacls $adminpath\$adminfile /inheritance:r
+
+    Start-Service sshd
+
+    # OPTIONAL but recommended:
+    Set-Service -Name sshd -StartupType 'Automatic'
```
PatrickLang: ok good, I missed this when I was reading on my phone

parts/k8s/windowsinstallopensshfunc.ps1
```diff
@@ -0,0 +1,45 @@
+function
+Install-OpenSSH {
+    Param(
+        [Parameter(Mandatory = $true)][string] 
+        $SSHKey
+    )
+
+    $adminpath = "c:\ProgramData\ssh\"
+    $adminfile = "administrators_authorized_keys"
```
jsturtevant: It looks like the `sshd_config` file that is generated with the install of openssh has changed.  Previously there was an entry that looked like:

 ```
AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys
```

Now it is `AuthorizedKeysFile	.ssh/authorized_keys`. 

For this to work with out a password will  need to change this `adminpath` to `c:\Users\adminuser\.ssh` and `$adminfile` to `authorized_keys` and `adminuser` will need to be the user specified in the `model.json`

Comment (jsturtevant) [reply parts/k8s/windowsinstallopensshfunc.ps1]: We may want to make sure we check the content of the `sshd_config` file so we don't break if the generated file changes again?

Comment (jsturtevant) [reply parts/k8s/windowsinstallopensshfunc.ps1]: If configured properly should be able to run a command like and not be prompted for a password but be dropped directly into the windows command prompt

```
 ssh -t -i ~/.ssh/id_rsa -o "ProxyCommand ssh -W %h:%p adminuser@<controlnodeip>" adminuser@1003k8s001     
```

where `1003k8s001` is the name of a windows node from `kubectl get nodes` and `controlip` is the public ip address for the control node

jackfrancis:
general code lgtm

defer to @jsturtevant and @PatrickLang to approve powershell implementations

parts/k8s/windowsinstallopensshfunc.ps1
vyta: @jsturtevant maybe we can override the AuthorizedKeysFile enrty in the config everytime?

PatrickLang:
@jsturtevant or @vyta have you been able to get SSH agent forwarding working? I followed the suggestion James had on creating `%userprofile%\.ssh\authorized_keys` but still can't seem to get it working without a password when I connect to the Windows node.

jsturtevant:
I was able to get it working with the key in the `AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys` but not the `%userprofile%\.ssh\authorized_keys`.  

When I wrote the comment above I had the key in both places and so thought it was working the `%userprofile%\.ssh\authorized_keys` but it appears to be only working with `AuthorizedKeysFile __PROGRAMDATA__/ssh/administrators_authorized_keys` 😕 

I just worked with @vyta and I think there is a bug in the script as it is now which was also causing errors but I think we got it worked out.  Look for a code update soon.

parts/k8s/windowsinstallopensshfunc.ps1
jsturtevant: Looks like I was wrong with the `.ssh/authorized_keys`  because I had the key in both places and it was the `administrators_authorized_keys` value that is the correct place.  The script itself has a bug where it silently fails to create that file and permission properly which is what made me think the `administrators_authorized_keys` wasn't working.

vyta:
I made the fixes to the script and was able to get ssh forwarding w/o password on 1809/2019, but not on 1803. I'll make a note in the docs

PatrickLang:
Great! Just synced and built. I will give it another try.

This will be a big help because I'd like to be able to have people securely connect to a Linux node as a jumpbox for investigations. This means we don't need to share passwords just to get logs. If someone wants to use remote desktop, they'll still need the password though. That's also why I didn't enable winrm by default.

PatrickLang:
I still see `#PubkeyAuthentication yes` in sshd_config. I think that needs to be uncommented to work.

vyta:
Weird, I was able to connect w/o pw and PubkeyAuthentication being commented out.. @PatrickLang Were you unable to get it working? Not sure what exactly that configuration does.. the official documentation https://www.ssh.com/ssh/sshd_config/ is incomplete. But I found an old (2014) blog post suggesting that it's needed http://innerdot.com/windows/setting-up-sshd-on-windows-to-allow-publickey-authentication

PatrickLang:
I'll change it and check first to see if that fixes it.

Sent from Outlook
________________________________

PatrickLang:
/lgtm

Thanks so much for doing this!

PatrickLang:
My mistake was with putty. When running from the command line, you need -A.

`putty -ssh azureuser@<masternode>.westus2.cloudapp.azure.com -A`

That works and lets me log into the Windows node without a password.

FWIW this works with gpg-connect-agent running on Windows so you can keep your private key on a Yubikey. Then you can ssh to your cluster's master with agent forwarding enabled and use it as a jumpbox to the Windows node. No private keys are left in a place they could be stolen.

PatrickLang:
This is also fun
`scp -o "ProxyCommand ssh -W %h:%p azureuser@<masterNode>.westus2.cloudapp.azure.com" azureuser@2709k8s010:'c:\k\kubelet.err.log' .`

jsturtevant:
The latest updates are working for me as well.  

I have been installing ssh on my windows nodes during development of kubernetes.  It makes grabbing the logs fast as Patrick pointed out and also swapping out new versions of kubelet/azurecni:  https://github.com/jsturtevant/windows-k8-playground/blob/master/scripts/updater/update.sh

/lgtm

PatrickLang:
FYI - I'm updating the log collection script to use SSH instead so PowerShell isn't needed on the Linux nodes https://github.com/PatrickLang/logslurp/pull/9

PR #435, opened 2019-02-04
feat: enable 1.14.0-alpha.2 support

jackfrancis:
**Reason for Change**:

Adds support for Kubernetes v1.14.0-alpha.2. See:

https://github.com/kubernetes/kubernetes/releases/tag/v1.14.0-alpha.2

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -104,6 +104,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.2":         true,
 	"1.13.3":         true,
 	"1.14.0-alpha.1": true,
+	"1.14.0-alpha.2": true,
 }
 
 // GetDefaultKubernetesVersion returns the default Kubernetes version, that is the latest patch of the default release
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
let's standardize all k8s version PRs to either chore or feat so they appear in the same place in release notes. I know @mboersma said he used chore for patches and feat for new minor releases, maybe pre-releases should be chore as well? Or feat for every k8s version? I don't have a strong opinion one way or the other, other that we should be consistent.

jackfrancis:
@CecileRobertMichon Good idea. I don't see how adding a new Kubernetes version isn't a `feat`. Thinking from a user's perspective. Regardless of the fact that the actual stuff we do is largely boilerplate + CI, the addition of a Kubernetes version is (IMO) something we want to communicate to users in the "new features" section of release notes.

CecileRobertMichon:
Since `1.11.7`, `1.12.5` and `1.14.0-alpha.1`already went in as `chore`, I vote we continue this pattern until our next release then switch over to `feat` for every new component version (including k8s). I agree with @jackfrancis that new K8s are a "feature" from users' perspective.

jackfrancis:
/hold

Some interesting stuff in E2E

mboersma:
Now that #492 is merged, this might pass after a rebase.

jackfrancis:
This is blocked on our E2E kubectl version.

PR #437, opened 2019-02-04
fix: reduce etcd download retries to avoid timeouts

CecileRobertMichon:
**Reason for Change**:

I've noticed some cluster deployments fail because etcd fails to download (no outbound on the VM) but the CSE times out instead of returning an error code because the etcd retries take too long. Reducing retries to 120 and sleep to 5s between retries.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesinstalls.sh
+++ b/parts/k8s/kubernetesinstalls.sh
@@ -19,7 +19,7 @@ installEtcd() {
     if [[ "$CURRENT_VERSION" == "${ETCD_VERSION}" ]]; then
         echo "etcd version ${ETCD_VERSION} is already installed, skipping download"
     else
-        retrycmd_get_tarball 360 10 /tmp/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz ${ETCD_DOWNLOAD_URL}/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz || exit $ERR_ETCD_DOWNLOAD_TIMEOUT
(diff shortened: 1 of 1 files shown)


parts/k8s/kubernetesinstalls.sh
```diff
@@ -19,7 +19,7 @@ installEtcd() {
     if [[ "$CURRENT_VERSION" == "${ETCD_VERSION}" ]]; then
         echo "etcd version ${ETCD_VERSION} is already installed, skipping download"
     else
-        retrycmd_get_tarball 360 10 /tmp/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz ${ETCD_DOWNLOAD_URL}/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz || exit $ERR_ETCD_DOWNLOAD_TIMEOUT
+        retrycmd_get_tarball 120 5 /tmp/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz ${ETCD_DOWNLOAD_URL}/etcd-v${ETCD_VERSION}-linux-amd64.tar.gz || exit $ERR_ETCD_DOWNLOAD_TIMEOUT
```
jackfrancis: You feel good about reducing the timeout to 5 seconds here?

CecileRobertMichon: that number is sleep between retries https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesprovisionsource.sh#L85

timeout of each retry is hardcoded in retrycmd_get_tarball as 60 seconds

5 second sleep is the same number we use for other retrycmd_get_tarball calls (eg, Azure CNI). I don't think sleeping 10s between failures is necessary.

jackfrancis: Got it, thanks!

PR #447, opened 2019-02-06
fix: prevent race condition with docker/moby install

CecileRobertMichon:
**Reason for Change**:

Cleaning up docker images in the background before installing docker can cause a race condition with docker install if we need to change the version of moby installed.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -46,7 +46,6 @@ if [ -f /var/log/azure/golden-image-install.complete ]; then
     echo "detected golden image pre-install"
     FULL_INSTALL_REQUIRED=false
     rm -rf /home/packer
-    cleanUpContainerImages
 else
     FULL_INSTALL_REQUIRED=true
 fi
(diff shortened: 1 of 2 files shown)


parts/k8s/kubernetesinstalls.sh
```diff
@@ -15,7 +15,7 @@ removeEtcd() {
 }
 
 removeMoby() {
-    sudo apt-get purge moby-engine moby-cli
+    sudo apt-get purge -y moby-engine moby-cli
```
CecileRobertMichon: forgot to commit this in #407, needed to avoid force purge without prompt

parts/k8s/kubernetescustomscript.sh
```diff
@@ -165,6 +164,10 @@ echo `date`,`hostname`, endcustomscript>>/opt/m
 mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
 ps auxfww > /opt/azure/provision-ps.log &
 
+if ! $FULL_INSTALL_REQUIRED; then
+    cleanUpContainerImages
```
CecileRobertMichon: moved this down to after container runtime is installed to avoid cleanup colliding with docker uninstall (if new version is needed)

jackfrancis:
lgtm

PR #449, opened 2019-02-06
build: commit generated code files

### LINKED ISSUE #286: Check In Code Generation Files to Allow for Vendoring
**Is this a request for help?**: No

---

---

**What version of aks-engine?**:  Latest code in the repo

---

**Kubernetes version**: NA

**What happened**: When vendoring aks-engine using dep or another dependency management tool, it is required to run "go generate" to have to code compile correctly. 

**What you expected to happen**:  Ideally, the generated code is already checked into the repo allowing to directly vendor it into other applications.

mboersma:
**Reason for Change**:

Keeping these generated files in the repository will help with vendoring and should make aks-engine "go get"-able. Since `make build` already calls `make generate` as a subtask, I don't think keeping them up to date will be onerous.

**Issue Fixed**:

Fixes #286

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tariq1890:
If we are doing this, we need to add a CI check to make sure the CI-runtime generated code doesn't differ from the committed generated code.

mboersma:
> need to add a CI check to make sure the CI-runtime generated code doesn't differ...

Good point--I'll work on adding that.

tariq1890:
@CecileRobertMichon Is there a way to exclude the autogenerated files from the `codecov` check ?

CecileRobertMichon:
> @CecileRobertMichon Is there a way to exclude the autogenerated files from the `codecov` check ?
Yes, we can add any pkg we want skipped in https://github.com/Azure/aks-engine/blob/master/Makefile#L122

mboersma:
You can only skip entire packages with go cover tools, so I took the approach of giving the generated files a naming pattern and then stripping their results from the coverage file based on that. Seemed less disruptive than putting them in separate packages or working with `+build` flags.

Makefile
```diff
@@ -125,6 +125,11 @@ test: generate
 test-style:
 	@scripts/validate-go.sh
 
+.PHONY: validate-generated
+validate-generated: generate
+	@echo "==> Checking generated files <=="
+	! git diff --name-only pkg/ | grep _generated.go
```
tariq1890:
```suggestion
	git diff --name-only --exit-code pkg/ | grep _generated.go
```

.pipelines/pr-e2e.yaml
```diff
@@ -52,6 +52,9 @@ jobs:
   - script: make validate-copyright-headers test-style
     displayName: Run linting rules
     workingDirectory: $(modulePath)
+  - script: make validate-generated
+    displayName: Check if generated code is up to date
```
tariq1890:
```suggestion
    displayName: Check if the generated code is in sync with the static assets.
```

Makefile
mboersma: I don't think that will work actually. I'm using the `!` to invert the return code of the entire statement--so really it affects `grep`. I want to return 1 if `grep` matches rather than the default of 0.

tariq1890: Got it thanks!

.pipelines/pr-e2e.yaml
mboersma: I prefer the original wording, especially in context at VSTS. @Azure/aks-engine opinions?

screen shot 2019-02-08 at 12 41 56 pm

tariq1890: Sure!

Issue #398, opened 2019-01-30
Upgrade from 1.11 to 1.12 results in both kube-dns and coredns


jackfrancis:
Needs validation, but I built a 1.11 cluster then upgraded to 1.12 and observed that kube-dns componentry was not replaced by coredns, but we simply added coredns to the cluster.

mboersma:
Verified that this bug happens using current master of aks-engine:

```
...
INFO[0705] Successfully deleted VM k8s-agentpool1-88093162-vmss000000 in VMSS k8s-agentpool1-88093162-vmss 
INFO[0705] Completed upgrading VMSS k8s-agentpool1-88093162-vmss 
INFO[0705] Completed upgrading all VMSS                 
INFO[0705] Cluster upgraded successfully to Kubernetes version 1.12.4 
$ k get deploy --all-namespaces
NAMESPACE     NAME                   DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
kube-system   coredns                1         1         1            1           7m15s
kube-system   heapster               1         1         1            1           25m
kube-system   kube-dns               1         1         1            1           24m
kube-system   kubernetes-dashboard   1         1         1            1           24m
kube-system   metrics-server         1         1         1            1           24m
kube-system   tiller-deploy          1         1         1            1           25m
```

jackfrancis:
is the kube-dns spec present in /etc/kubernetes/addons after upgrade?

mboersma:
It is not: 

```
azureuser@k8s-master-88093162-0:~$ ls /etc/kubernetes/addons/
audit-policy.yaml                     coredns.yaml                         kube-proxy-daemonset.yaml
azure-cloud-provider-deployment.yaml  ip-masq-agent.yaml                   kubernetes-dashboard-deployment.yaml
azure-cni-networkmonitor.yaml         keyvault-flexvolume-installer.yaml   kube-tiller-deployment.yaml
azure-storage-classes.yaml            kube-heapster-deployment.yaml
blobfuse-flexvolume-installer.yaml    kube-metrics-server-deployment.yaml
```

jackfrancis:
10-4. So, I think the immediate next step is to publish a simple bash script that will manually clean up the kube-dns artifacts. And validate that container resolution is in good working order w/ coredns.

That will help folks now running upgraded 1.12 clusters w/ both kube-dns and coredns.

oscarwest:
1.11.5 -> 1.12.5
[...]

PR #454, opened 2019-02-07
chore: Update calico to 3.5 and allow calico to work with azure CNI

song-jiang:
**Reason for Change**:

- Upgrade calico to v3.5
- Allow calico to work with azure CNI by setting 
```NetworkPolicy: "calico" , NetworkPlugin: "azure"```
In this case, calico per-host daemon works with azure CNI plugin and azure IPAM to support network policy. 

- Continue support kubenet network plugin by setting
```NetworkPolicy: "calico" , NetworkPlugin: "kubenet"```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -191,9 +191,10 @@ configureCNI() {
     retrycmd_if_failure 120 5 25 modprobe br_netfilter || exit $ERR_MODPROBE_FAIL
     echo -n "br_netfilter" > /etc/modules-load.d/br_netfilter.conf
     if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
-        if [[ "${NETWORK_POLICY}" != "calico" ]]; then
-            mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
-            chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
+        mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
+        chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
+        if [[ "${NETWORK_POLICY}" == "calico" ]]; then
+          sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
         fi
         /sbin/ebtables -t nat --list
     fi
--- a/parts/k8s/addons/1.10/kubernetesmasteraddons-calico-daemonset.yaml
+++ b/parts/k8s/addons/1.10/kubernetesmasteraddons-calico-daemonset.yaml
@@ -1,108 +1,8 @@
-apiVersion: v1
-kind: ServiceAccount
-metadata:
-  name: calico-node
-  namespace: kube-system
-  labels:
-    addonmanager.kubernetes.io/mode: "EnsureExists"
----
-
-kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
-metadata:
-  name: calico-node
-  labels:
-    addonmanager.kubernetes.io/mode: "EnsureExists"
-rules:
-  - apiGroups: [""]
-    resources:
-      - namespaces
-      - serviceaccounts
-    verbs:
-      - get
-      - list
-      - watch
-  - apiGroups: [""]
-    resources:
-      - pods/status
-    verbs:
-      - patch
-  - apiGroups: [""]
-    resources:
-      - pods
-    verbs:
-      - get
-      - list
-      - watch
-  - apiGroups: [""]
-    resources:
-      - services
-    verbs:
-      - get
-  - apiGroups: [""]
-    resources:
-      - endpoints
-    verbs:
-      - get
-  - apiGroups: [""]
-    resources:
-      - nodes
-    verbs:
-      - get
-      - list
-      - update
-      - watch
-  - apiGroups: ["extensions"]
-    resources:
-      - networkpolicies
-    verbs:
-      - get
-      - list
-      - watch
-  - apiGroups: ["networking.k8s.io"]
-    resources:
-      - networkpolicies
-    verbs:
-      - watch
-      - list
-  - apiGroups: ["crd.projectcalico.org"]
-    resources:
-      - globalfelixconfigs
-      - felixconfigurations
-      - bgppeers
-      - globalbgpconfigs
-      - bgpconfigurations
-      - ippools
-      - globalnetworkpolicies
-      - globalnetworksets
-      - networkpolicies
-      - clusterinformations
-      - hostendpoints
-    verbs:
-      - create
-      - get
-      - list
-      - update
-      - watch
-
----
-
-apiVersion: rbac.authorization.k8s.io/v1beta1
-kind: ClusterRoleBinding
-metadata:
-  name: calico-node
-  labels:
-    addonmanager.kubernetes.io/mode: "EnsureExists"
-roleRef:
-  apiGroup: rbac.authorization.k8s.io
-  kind: ClusterRole
-  name: calico-node
-subjects:
-- kind: ServiceAccount
-  name: calico-node
-  namespace: kube-system
-
----
+# Calico Version v3.5.0
+# https://docs.projectcalico.org/v3.5/releases#v3.5.0
+# This manifest includes the following component versions:
+#   calico/node:v3.5.0
+#   calico/cni:v3.5.0
 
 # This ConfigMap is used to configure a self-hosted Calico installation.
 kind: ConfigMap
@@ -113,9 +13,11 @@ metadata:
   labels:
     addonmanager.kubernetes.io/mode: "EnsureExists"
 data:
+  # You must set a non-zero value for Typha replicas below.
   typha_service_name: "calico-typha"
 
-  # The CNI network configuration to install on each node.
+  # The CNI network configuration to install on each node.  The special
+  # values in this config will be automatically populated.
   cni_network_config: |-
     {
       "name": "k8s-pod-network",
@@ -158,13 +60,12 @@ metadata:
     addonmanager.kubernetes.io/mode: "EnsureExists"
 spec:
   ports:
-    - port: 5473
-      protocol: TCP
-      targetPort: calico-typha
-      name: calico-typha
+  - port: 5473
+    protocol: TCP
+    targetPort: calico-typha
+    name: calico-typha
   selector:
     k8s-app: calico-typha
-
 ---
 
 # This manifest creates a Deployment of Typha to back the above service.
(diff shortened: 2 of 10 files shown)

### CODE REVIEW

**Comment by CecileRobertMichon** [comment]
@jackfrancis should we hold this until we cut the release?

/hold

**Comment by jackfrancis** [comment]
Yes. We want to run this through the appropriate back-channel cluster configuration tests appropriate for this surface area.

**Comment by jackfrancis** [comment]
I mean to say that this doesn't need to go in ASAP with the next aks-engine release, so we're on the same page. :)

**Comment by jackfrancis** [comment]
@song-jiang Can you add more description about the differences between `NetworkPlugin: "calico+azureipam"` and `NetworkPlugin: "azure"` in a calico context?

File: parts/k8s/kubernetesconfigs.sh [hunk 1]
```diff
@@ -196,9 +196,10 @@ configureCNI() {
     retrycmd_if_failure 120 5 25 modprobe br_netfilter || exit $ERR_MODPROBE_FAIL
     echo -n "br_netfilter" > /etc/modules-load.d/br_netfilter.conf
     if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
-        if [[ "${NETWORK_POLICY}" != "calico" ]]; then
-            mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
-            chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
+        mv $CNI_BIN_DIR/10-azure.conflist $CNI_CONFIG_DIR/
+        chmod 600 $CNI_CONFIG_DIR/10-azure.conflist
+        if [[ "${NETWORK_POLICY}" == "calico" ]] || [[ "${NETWORK_POLICY}" == "" ]]; then
+          sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist
```
jackfrancis: Do we really want to change the Azure CNI configuration here? This bridge mode setting will now apply to non-calico scenarios given the `|| [[ "${NETWORK_POLICY}" == "" ]]` condition

song-jiang: The purpose for this change is to make life easier for users who want to apply calico for network policy on clusters previously installed without specify a network policy provider. 
My understanding is it should not affect any user experience if they choose a non-calico network policy provider. I am not sure if user really care about using a `bridge` mode network or a `transparent` mode network. Would love to hear your input on this. Thanks.

jackfrancis: @sharmasushant @tamilmani1989 can you comment on this change?

jackfrancis:
@sharmasushant @tamilmani1989 Could you kindly read through this PR description and comment on @song-jiang's description of the difference between `NetworkPolicy: "calico" , NetworkPlugin: "azure"` and `NetworkPolicy: "calico" , NetworkPlugin: "calico+azureipam"`.

Based on the reading through the changeset, the `"calico+azureipam"` flavor is the one that represents the current implementation of calico + Azure CNI. `NetworkPolicy: "calico" , NetworkPlugin: "azure"` appears to be a novel configuration/implementation.

Any thoughts?

parts/k8s/kubernetesconfigs.sh
tamilmani1989: By default we should have bridge mode and not transparent mode. To my understanding, the config should be set to transparent mode only if user speicifes calico policy. @sharmasushant  can comment on this.

sharmasushant:
@jackfrancis 
We should not change the existing defaults in this PR.
The newly added `"transparent"` mode in azure cni is only supported with calico policy, and should not be default.

The supported default mode for azure cni is `"bridge"`, and default policy should be `"azure"` as well.

song-jiang:
@jackfrancis @sharmasushant @tamilmani1989 Thanks for your input and clarifications. Apologies for my misunderstanding. I will make changes to my PR accordingly.

song-jiang:
@jackfrancis Could you take another look?

parts/k8s/kubernetesinstalls.sh
```diff
@@ -169,7 +169,7 @@ installClearContainersRuntime() {
 }
 
 installNetworkPlugin() {
-    if [[ "${NETWORK_PLUGIN}" = "azure" ]]; then
+    if [[ "${NETWORK_PLUGIN}" = "azure" ]] || [[ "${NETWORK_PLUGIN}" = "calico+azureipam"  ]]; then
```
tamilmani1989: what is calico+azureipam? I may be missing something here. Can you please explain us which scenario this is used?

feiskyer:
> @sharmasushant @tamilmani1989 Could you kindly read through this PR description and comment on @song-jiang's description of the difference between NetworkPolicy: "calico" , NetworkPlugin: "azure" and NetworkPolicy: "calico" , NetworkPlugin: "calico+azureipam".

Same questions here.

song-jiang:
@tamilmani1989 "NetworkPlugin: calico+azureipam, NetworkPolicy: calico" is the replacement of current implementation of "NetworkPlugin: azure, NetworkPolicy: calico" which uses calico CNI plugin and azure IPAM. 

We leave that as an option for anyone who might be using it and continue wish to do so. I will check with PM on email conversations made earlier.

jackfrancis:
@song-jiang I have a preference for simply moving the implementation forward. If we have good reason to change the implementation, we should stand by that opinion, rather than offering two implementations that basically do the same thing, which I would argue is confusing for users.

song-jiang:
@jackfrancis Agree it is less confusing to have just one option. Let me check with PM and come back to you.

ahrkrak:
@song-jiang I think we have agreed on this with PM - can you confirm and give @jackfrancis an idea of when you're likely to get this update done?

song-jiang:
@ahrkrak Yes. It has been confirmed with PM we will drop option `calico+azureipam`.  Another changes I need to put in is to disable `azure-cni-networkmonitor` daemonset if azure CNI set to `transparent mode`.  I expect to get PR updated by Wednesday. cc @jackfrancis

song-jiang:
@jackfrancis Could you please review this PR again? Changes made since last time
- Set `transparent mode` for calico network policy only.
- Removed network plugin option `calico+azureipam`.
- Disabled `azure-cni-networkmonitor` addon with "networkPlugin: azure, networkPolicy: Calico".

jackfrancis:
Thanks @song-jiang! FYI I rebased/force-pushed to fix the unit test coverage failures

jackfrancis:
Validated that pre-existing, known-working scenaros work against this change.

Still validating back-compat for upgrade scenarios.

@song-jiang, to clarify: calico + Azure CNI was already supported. Can you clarify what exactly the functional delta is between this changeset and what is already present for the calico + Azure CNI scenario?

Thanks again!

song-jiang:
@jackfrancis Thanks for fixing UT coverage. 

Current support for "networkPolicy: calico, networkPlugin: azure" is implemented as "networkPolicy: calico, networkPlugin: calico + azure ipam".  It actually uses calico CNI under the hood instead of azure CNI. Because at the time when we did integration, azure CNI did not support `transparent mode` yet. 
https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesconfigs.sh#L199 This line disables azure CNI if calico is enabled.

This PR replaced calico CNI with azure CNI plugin in transparent mode. 
- Make sure azure CNI config is installed and set to `transparent mode`.
- Remove `initContainer` section from `calico-node` daemonset hence no calico CNI is installed. 
- Disable `azure-cni-networkmonitor` addon.

jackfrancis:
@song-jiang I haven't been able to confirm backwards-compatibility with previous clusters built w/ calico. Our E2E test when run against an upgraded cluster against this branch says:

```
/go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:87
  with NetworkPolicy enabled
  /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:933
    should apply various network policies and enforce access to nginx pod [It]
    /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:934

    Expected an error to have occurred.  Got:
        <nil>: nil

    /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1015
------------------------------
SSSSSSS

Summarizing 1 Failure:

[Fail] Azure Container Cluster using the Kubernetes Orchestrator with NetworkPolicy enabled [It] should apply various network policies and enforce access to nginx pod 
/go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1015
```

You don't have to worry about debugging our particular E2E test runner, but: is this change intended to work against pre-existing clusters?

song-jiang:
@jackfrancis I am afraid this is a break change which does not support backwards-compatibility with previous clusters. During upgrade process, the cluster would have nodes with calico CNI (pre-exist) and nodes with azure CNI (new configuration). These two plugins are not compatible with each other. AFAIU, network traffic and existing network policy should continue to function once upgrade process is complete.

Could you tell me if E2E failed during upgrade process? Or it happens after cluster has been upgrade.

 cc @ahrkrak

jackfrancis:
@song-jiang Understood. Could you kindly repro the upgrade scenario and document the required steps to remove the pre-existing Azure CNI configuration from the cluster? Based on your description, once we remove the prior componentry, I would expect the new CNI implementation to take over and apply NetworkPolicy correctly.

We will then add those manual steps to our upgrade documentation so folks are aware: if you have an pre-existing AKS Engine-enabled calico cluster and wish to upgrade to the newer 3.5 AKS Engine-enabled implementation, here is how you can manually achieve that.

Make sense?

song-jiang:
> remove the pre-existing Azure CNI configuration from the cluster

I suppose you mean `remove the pre-existing calico CNI configuration from the cluster`?

Yes. I need to test upgrade process and document the procedures. I am not able to find documents about upgrading cluster with same kubernetes version but different cluster configurations. My understanding is I need to:
- Generate new template by executing `aks-engine generate template/new-cluster-config.json`
- Upgrade cluster by `aks-engine upgrade --deployment-dir generated-template-dir`.

Am I right?  @jackfrancis

jackfrancis:
blech, sorry, yes calico :)

You should be able to repro by simply:

1) use an recently released version of aks-engine
2) create a cluster w/ networkPolicy: calico
3) change to this branch/build aks-engine binary
4) with newly built aks-engine binary, upgrade the cluster to the next patch or minor version (shouldn't matter)
5) manually remove the "old" calico config and document those steps, until the "new" calico is the only one present, and NetworkPolicy has been confirmed to work

Here's a recent example of how someone documented those steps for cleaning up kube-dns/coredns co-existence:

https://github.com/Azure/aks-engine/issues/398#issuecomment-467890973

Thanks so much!

song-jiang:
@jackfrancis Awesome. That's really helpful. Thanks! 

I expect to test upgrade process early next week. Will let you know once I am done.

feiskyer:
> @jackfrancis I am afraid this is a break change which does not support backwards-compatibility with previous clusters. During upgrade process, the cluster would have nodes with calico CNI (pre-exist) and nodes with azure CNI (new configuration). These two plugins are not compatible with each other. AFAIU, network traffic and existing network policy should continue to function once upgrade process is complete.

This may introduce issues for AKS scenario. Do we have some way to clean up the calico CNI first and then configure Azure CNI later?

sauryadas:
> > @jackfrancis I am afraid this is a break change which does not support backwards-compatibility with previous clusters. During upgrade process, the cluster would have nodes with calico CNI (pre-exist) and nodes with azure CNI (new configuration). These two plugins are not compatible with each other. AFAIU, network traffic and existing network policy should continue to function once upgrade process is complete.
> 
> This may introduce issues for AKS scenario. Do we have some way to clean up the calico CNI first and then configure Azure CNI later?

@song-jiang We need to ensure that existing customers using CNI + Calico in AKS arent broken..

song-jiang:
@feiskyer @sauryadas @jackfrancis Double checked with my colleague, in theory azure CNI in transparent mode would work with calico CNI side by side. Originally I thought veth prefixes are different which could have an impact on how calico renders network policy. So my previous statement about backwards-compatibility may be wrong. 

I will start testing upgrade and let you know the result.

song-jiang:
> You should be able to repro by simply:
> 
> 1. use an recently released version of aks-engine
> 2. create a cluster w/ networkPolicy: calico
> 3. change to this branch/build aks-engine binary
> 4. with newly built aks-engine binary, upgrade the cluster to the next patch or minor version (shouldn't matter)
> 5. manually remove the "old" calico config and document those steps, until the "new" calico is the only one present, and NetworkPolicy has been confirmed to work
> 

Hit an issue while following above steps to upgrade a cluster created by `aks-engine v0.32.0`
```STATUS  InternalServerError
STATUS MESSAGE  {
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "Encountered internal server error. Diagnostic information: timestamp '20190312T180205Z', subscription id '058e3078-8970-434e-a2f0-98a72d053ea9', tracking id 'ee14adaf-909f-46f0-a46e-f8eacda4540e', request correlation id 'c1b51db0-3a27-47ce-9172-298314ab1fde'."
  }
}
PROVISIONING STATE  Failed
TIMESTAMP  12/03/2019, 18:02:05
DURATION  3 minutes 34 seconds
TYPE  Microsoft.Compute/virtualMachines/extensions
RESOURCE ID  /subscriptions/058e3078-8970-434e-a2f0-98a72d053ea9/resourceGroups/song-calico-azure/providers/Microsoft.Compute/virtualMachines/k8s-master-72678590-0/extensions/computeAksLinuxBilling
RESOURCE  k8s-master-72678590-0/computeAksLinuxBilling 
@jackfrancis

**Comment by tamilmani1989** [comment]
> Originally I thought veth prefixes are different which could have an impact on how calico renders network policy.

@song-jiang  - Calico creates veth with "cali" prefix while azure cni creates with "azv" prefix. So in your previous comment you mentioned that should not be a problem. In upgrade scenario, How are we making sure pods created using calico cni (cali prefix) works with new calico policy that supports azv prefix?

**Comment by song-jiang** [comment]
@tamilmani1989 Let me clarify on this. It should not be a problem if they are consistent within the same node.  Which means network policy will be honoured if a node has either old `calico-node` (prefix `cali`) + calico CNI or new `calico-node` (prefix `azv`) + azure CNI.

Network connection/policy will function on mixed old/new nodes as long as the prefix is consistent for each node.

**Comment by song-jiang** [comment]
@jackfrancis Could not repo the upgrade issue I posted earlier. Upgrade works fine for me now.

**Comment by sauryadas** [comment]
@palma21 please prioritize merging this PR. It is dependent on taking NP - Calico to GA

**Comment by tamilmani1989** [comment]
Hi @song-jiang  Thanks for responding. My question here is do we end up in situation where we have a node with both "cali" prefix and "azv" prefix veth interfaces (during upgrade). What will happen in that case?

**Comment by CecileRobertMichon** [comment]
@song-jiang please rebase on latest master to get up to date unit test coverage stats (the last rebase seems to be from 8 days ago).

File: pkg/api/vlabs/validate_test.go [hunk 3]
```diff
@@ -704,6 +704,10 @@ func Test_Properties_ValidateNetworkPluginPlusPolicy(t *testing.T) {
 			networkPlugin: "azure",
 			networkPolicy: "flannel",
 		},
+		{
+			networkPlugin: "azure",
```
CecileRobertMichon: this combination is already in `networkPluginPlusPolicyAllowed ` in [validate.go](https://github.com/Azure/aks-engine/blob/master/pkg/api/vlabs/validate.go#L71) while this is adding it to the unit test expecting it to not be valid. Unless I'm missing something, it seems contradictory.

song-jiang:
@tamilmani1989  It is unlikely pods on same host have mixed veth prefixes. During upgrade process, either calico CNI (old node) or azure CNI (new node) will be used. However, what I found is we could have a scenario where new node gets old version of `calico-node` ( prefix cali) and azure CNI (prefix azv). This would break network policy.

So current status on upgrading from existing (AKS or bare-metal) cluster with my PR
- network connection functions as normal.
- network policy is broken. 

I am working on possible solutions to fix this. cc @jackfrancis @sauryadas

dtzar:
Don't want to delay this PR!  Would be nice to see Calico at 3.5.3 or 3.6.0 to have the latest fixes/functionality, ensure all works, and avoid another PR to update the daemonset yamls.

feiskyer:
> network policy is broken.

Is this fixed now?

song-jiang:
@feiskyer Not yet. It is still in progress.

ahrkrak:
We have agreed an upgrade strategy that will work, but needs to be documented. @jackfrancis is picking up ownership of the docs changes.
@jackfrancis will you add those docs changes into this PR or a separate one?

jackfrancis:
@ahrkrak filed this as a follow-up:

https://github.com/Azure/aks-engine/issues/908

parts/k8s/kubernetesconfigs.sh
jackfrancis: @tamilmani1989 I'm running this through some final smoke tests. Please comment if we still have open concerns about this change before merging.

jackfrancis:
smoke tests checked out

lgtm

@tamilmani1989 for sign off on changes as well

tamilmani1989:
@song-jiang @ahrkrak is this fixed ? 
>network policy is broken

ahrkrak:
Correct. Currently #908 just includes a reference to the yaml that you have to kubectl delete. That removes any older calico-nodes, which will then get updated by the addon manager to the newer version. @jackfrancis will update the actual docs to include the howto instructions for doing that.

PR #183, opened 2018-12-19
use containerd 1.1.5, enable w/ non-vhd flows

jackfrancis:
**What this PR does / why we need it**: Update containerd to 1.1.5:

https://github.com/containerd/containerd/releases/tag/v1.1.5

We should implement containerd 1.2 for k8s 1.11 and above at some point, as it appears to be the preferred (and breaking for k8s.1.10 and below) containerd version for modern k8s.

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: 

fixes #184

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```
use containerd 1.1.5, enable w/ non-vhd flows
```

PR #457, opened 2019-02-07
fix: do not re-download containerd if pre-downloaded in VHD

CecileRobertMichon:
**Reason for Change**:

Follow up on #183: currently if the VHD script installs containerd, which deletes the download directory after it's done. At provisioning time, we check for the downloaded binary and re-download it if it is already present. This is causing containerd to always download, even if it was pre-installed.
Instead, follow the Azure CNI pattern: download containerd in the base image and check for the download at provisioning. Only download again if the version we want is missing. This also will allow us to pre-bake several versions in the future. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+CONTAINERD_DOWNLOAD_URL_BASE="https://storage.googleapis.com/cri-containerd-release/"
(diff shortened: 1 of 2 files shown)


jackfrancis:
See #516

PR #459, opened 2019-02-07
WIP fix: ip-masq-agent has different addon URL compared to VHD

jackfrancis:
**Reason for Change**:

We pre-pull `gcr.io/google-containers/ip-masq-agent-amd64:v2.0.0` onto our VHDs. As long as we do so, we should also reference the same image URL in our default addon spec for ip-masq-agent.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -221,7 +221,7 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 				MemoryRequests: "50Mi",
 				CPULimits:      "50m",
 				MemoryLimits:   "250Mi",
-				Image:          specConfig.KubernetesImageBase + "ip-masq-agent-amd64:v2.0.0",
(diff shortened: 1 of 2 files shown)


pkg/api/addons.go
```diff
@@ -221,7 +221,7 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 				MemoryRequests: "50Mi",
 				CPULimits:      "50m",
 				MemoryLimits:   "250Mi",
-				Image:          specConfig.KubernetesImageBase + "ip-masq-agent-amd64:v2.0.0",
```
CecileRobertMichon: isn't this to allow for a different base for Mooncake? Hardcoding it means mooncake won't be able to pull it anymore

PR #466, opened 2019-02-08
refactor: swap satori's uuid package to gofrs

mdanzinger:
**Reason for Change**:
Satori's go.uuid is effectively unmaintained and contained at least one critical security flaw which took half a year to be merged as the owner isn't active. See [satori/go.uuid#75](https://github.com/satori/go.uuid/pull/75)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tariq1890:
This is great!! Thank you @mdanzinger

mboersma:
Some import errors were caught by `make test-style`.

tariq1890:
@ultimateboy Are you okay with this Matt? IMO we should take this change as soon as possible. If you vendor aks-engine after this is merged, similar changes will have to be made in the projects that dep on aks-engine.

tariq1890:
@amanohar @sozercan FYI. This should go into openshift as well.

cmd/root.go
```diff
@@ -15,7 +15,7 @@ import (
 	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/pkg/errors"
-	uuid "github.com/satori/go.uuid"
+	"github.com/gofrs/uuid"
```
tariq1890:
```suggestion
	uuid "github.com/gofrs/uuid"
```

cmd/root_test.go
```diff
@@ -6,7 +6,7 @@ package cmd
 import (
 	"testing"
 
-	uuid "github.com/satori/go.uuid"
+	"github.com/gofrs/uuid"
```
tariq1890:
```suggestion
	uuid "github.com/gofrs/uuid"
```

pkg/api/mocks.go
```diff
@@ -5,13 +5,13 @@ package api
 
 import (
 	"github.com/Azure/go-autorest/autorest/to"
-	uuid "github.com/satori/go.uuid"
+	"github.com/gofrs/uuid"
```
tariq1890:
```suggestion
	uuid "github.com/gofrs/uuid"
```

pkg/armhelpers/graph.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest"
 	"github.com/Azure/go-autorest/autorest/date"
 	"github.com/Azure/go-autorest/autorest/to"
-	uuid "github.com/satori/go.uuid"
+	"github.com/gofrs/uuid"
```
tariq1890:
```suggestion
	uuid "github.com/gofrs/uuid"
```

mdanzinger:
@mboersma  The only new errors I receive are :

``` cmd/root.go:17: File is not `gofmt`-ed with `-s` (gofmt)
	"github.com/pkg/errors"
cmd/deploy_test.go:15: File is not `gofmt`-ed with `-s` (gofmt)
	"github.com/pkg/errors"
pkg/api/vlabs/validate.go:20: File is not `gofmt`-ed with `-s` (gofmt)
	"github.com/pkg/errors"```

Is that what you're referring to?

File: Gopkg.lock [hunk 5]
```diff
@@ -2,15 +2,12 @@
 
 
 [[projects]]
-  digest = "1:f323f98930459f65f4699b5bfba563743680e2633d96e72d61a9732648e2d07c"
   name = "contrib.go.opencensus.io/exporter/ocagent"
```
tariq1890: These changes should not happen. Please use `dep 0.5.0` when running `dep ensure`

tariq1890:
@mdanzinger Can you make sure you `go install` the goimports installed in `go 1.11.5` and then run `goimports` to address the goimports issues?

Gopkg.lock
mdanzinger: I was using an outdating version of dep. I've updated it and pushed a commit.

tariq1890:
@mdanzinger The codecov patch seems to fail. Could you rebase this PR?

mdanzinger:
@tariq1890 I've rebased, squashed my previous commits, and pushed (I had to force push due to me accidentally merging two diverging branches), but it seems like the codecov patch is still failing, any ideas why?

tariq1890:
@mdanzinger The codecov patch diff is failing because you have changed a line which has had no coverage previously. Normally, we'd request the contributor to add tests to fill that gap. In this case, adding test cases for the said method is less straightforward and would involve too much work which would be well out of the bounds of this PR's scope.

So this PR LGTM. I would defer to the other maintainers on assessing the impact of this dep change as some projects vendor in aks-engine as a dependency.

tariq1890:
@jackfrancis @mboersma I have tested vendoring this fork against other repos that utilize aks-engine and I've found that it works with no issues.

tariq1890:
Thanks @mdanzinger :)

PR #473, opened 2019-02-11
fix: specify "shared" for unmanaged disk class

### LINKED ISSUE #419: After update to aks-engine 0.29.1, PVCs are created as managed disks, even if the cluster is using StorageAccount
**Is this a request for help?**: Yup.

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): Issue.

---

**What version of aks-engine?**: 0.29.1

---

**Kubernetes version**: 1.13.2

**What happened**: After the update from acs-engine (0.22.1 or something) to aks-engine 0.29.1, the `default` storage class for our cluster, which is `StorageAccount` based, seems to be managed disks. I.e., creating a `pvc` with the `storageClassName` set to `default` creates a managed disk. As the VMs are storage account based, these managed disks cannot be mounted to the agents.

What's worse though is that setting the `storageClassName` to `unmanaged-standard` **also** creates a managed disk.

**What you expected to happen**: (A) For a cluster which uses `StorageAccount` as storage model, the default storage class for dynamically provisioned disks should be unmanaged disks, and (B) if I specify that I need an unmanaged disk via the `storageClassName: unmanaged-standard` property, it really should create an unmanaged disk.

**How to reproduce it** (as minimally and precisely as possible): (A) Create a 1.13.2 cluster with storage type set to `StorageAccount`. (B) Create a PVC with either the `default` or `unmanaged-standard` storage class name. (C) Watch the cluster create a managed disk instead.

**Anything else we need to know**: I bet I missed this somewhere in the docs... The reason why we still have stuck to the storage account mode is that those disks are, at least by the specs, a lot faster, plus they are a lot cheaper.

andyzhangx:
**Reason for Change**:

Due to [k8s upstream PR](https://github.com/kubernetes/kubernetes/pull/67483), default azure disk storage class has been changed to managed disk, this PR explicitly specify unmanaged disk class kind as `shared`, otherwise it will always create managed disk which is incorrect.

**Issue Fixed**:

Fixes #419

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

@jackfrancis 

cc @DonMartin76

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-unmanaged-azure-storage-classes.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-unmanaged-azure-storage-classes.yaml
@@ -19,6 +19,7 @@ metadata:
     kubernetes.io/cluster-service: "true"
 provisioner: kubernetes.io/azure-disk
 parameters:
+  kind: shared
   storageaccounttype: Premium_LRS
   cachingmode: ReadOnly
 ---
@@ -31,6 +32,7 @@ metadata:
     kubernetes.io/cluster-service: "true"
 provisioner: kubernetes.io/azure-disk
 parameters:
+  kind: shared
   storageaccounttype: Standard_LRS
   cachingmode: ReadOnly
 ---


mboersma:
lgtm

serbrech:
@tariq1890 Take note of this too for your no_template branch

tariq1890:
@serbrech Thanks

/approve

PR #476, opened 2019-02-11
set caching mode of data disks as ReadOnly by default

### LINKED ISSUE #418: change data disk cache to ReadOnly
No
---

ISSUE
---

0.29.1
---

**Kubernetes version**:
1.13.2

**What happened**:
k8s changed azure disk host cache to ReadOnly in 1.13.2, see https://github.com/kubernetes/kubernetes/issues/72228.
The cache mode of attached data disk by aks-engine is still None. 

change data disk cache to ReadOnly

Build a cluster with data disks on Windows nodes, check cache mode of data disk in Azure Portal.

andyzhangx:
**Reason for Change**:

change azure disk host cache to ReadOnly
The default value `None` would lead to slow perf. And according to below doc: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/premium-storage-performance#disk-caching, default host cache should be `ReadOnly`

**Issue Fixed**:

Fixes #418

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -553,6 +553,7 @@ func getDataDisks(a *api.AgentPoolProfile) string {
               "createOption": "Empty",
               "diskSizeGB": "%d",
               "lun": %d,
+              "caching": "ReadOnly",
               "name": "[concat(variables('%sVMNamePrefix'), copyIndex(),'-datadisk%d')]",
               "vhd": {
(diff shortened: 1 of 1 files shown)


andyzhangx:
This is the first step, next I will set host cache of etcd disk as `ReadOnly`, which would improve the read perf a lot compared to `None`

serbrech:
@andyzhangx Are we sure about changing the `etcd` host cache to `ReadOnly`? 
I think that `etcd` is write-intensive to provide consistent distributed log replication, so I'm not sure we would benefit from this in reality.
For example, it is explicitly recommended to set the host caching to `None` for Sql Server log disks for similar reasons. See https://stackoverflow.com/a/39653131

andyzhangx:
@serbrech thanks for the comments, I will check the etcd setting later. While for this PR, change host cache of data disk on agent node to `ReadOnly` is the correct way.

mboersma:
@andyzhangx could you `git commit --amend` this so the commit message starts with `fix: `? We use automated tools to generate the changelog and they fail if there isn't a category prefix like that.

PR #491, opened 2019-02-13
ci: use diff instead of git diff for the CI validation checks of the generated files

tariq1890:
We are using the old pattern of validating the generated files (circle back to acs-engine openshift dayz, a time I never knew).

This seems to be the most effective way to handle this CI check


tariq1890:
Thank you so much @serbrech :)

PR #502, opened 2019-02-13
docs: adding godoc package comments

### LINKED ISSUE #29: Add missing Go package comments
We link to generated go docs at https://godoc.org/github.com/Azure/aks-engine but several important packages lack a package-level comment.  See https://blog.golang.org/godoc-documenting-go-code

(I like the standalone `package.go` file convention but YMMV.)

serbrech:
**Reason for Change**:

Add missing godoc comments on packages where missing

**Issue Fixed**:

fix #29

### DIFF
--- a/cmd/doc.go
+++ b/cmd/doc.go
@@ -1,0 +1,5 @@
+// Package cmd stores the command line interface implementation.
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+package cmd
--- a/cmd/doc.go
+++ b/cmd/doc.go
@@ -1,5 +1,5 @@
-// Package cmd stores the command line interface implementation.
-//
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
+
+// Package cmd stores the command line interface implementation.
 package cmd
--- a/pkg/api/agentPoolOnlyApi/v20170831/doc.go
+++ b/pkg/api/agentPoolOnlyApi/v20170831/doc.go
@@ -1,5 +1,5 @@
-// Package v20170831 stores the API model for version "2017-08-31."
-//
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
+
+// Package v20170831 stores the API model for version "2017-08-31."
 package v20170831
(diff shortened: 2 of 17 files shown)


serbrech:
FYI the generated files break the linting unfortunately : 
```
golint $(go list ./... | grep -v /vendor/) | grep comment
/go/src/github.com/Azure/aks-engine/pkg/engine/templates_generated.go:1:1: package comment should be of the form "Package engine ..."
/go/src/github.com/Azure/aks-engine/pkg/i18n/translations_generated.go:1:1: package comment should be of the form "Package i18n ..."
```

cmd/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package cmd stores the command line interface implementation.
+//
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+package cmd
```
tariq1890: It should be of the following order

i) Copyright License Header
ii) Package documentation

tariq1890:
Thanks @serbrech ! I've provided my comment

cmd/doc.go
serbrech: erg, right, I took example on an existing package in aks-engine, which was wrong to begin with :)

mboersma:
Does `godoc` pick up all these comments? I remember in [4324](https://github.com/Azure/aks-engine/commit/3e57d2887a4b0c624942a378afa7a714223e2811) I had to make the `// Package` comment on the first line to satisfy both `godoc` and our license checking script...

cmd/doc.go
mboersma: I'm not sure that's true. We should run `godoc` as a server and make sure that the comment is picked up, because I had a reason for committing the initial comments that way.

tariq1890:
I believe it should not a problem. Kubernetes and Prometheus open source follow the pattern of license header first and then Package doc. Their godocs seem to populate as expected.

mboersma:
I ran `godoc -http=6060` on this branch and I don't see package comments for some of those where we've put the copyright header first. `cmd` for example:
screen shot 2019-02-13 at 2 32 42 pm

mboersma:
This looks like my local problem somehow. (The doc comment rendering doesn't get updated for me even when I make local modifications.) @serbrech showed me his renders correctly and I think @tariq1890's guidance is correct.

PR #503, opened 2019-02-13
fix: set nodes to panic and reboot on oom

cchildress:
**Reason for Change**:

This PR sets nodes to panic rather than running the oom-killer during out of memory conditions outside of a memory cgroup. After the panic the node reboots 30 seconds later.
The node should boot up in a known good state and re-join the Kubernetes cluster instead of continuing to run in an unknown state with some process gone.

**Issue Fixed**:

N/A

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Is there any way to deliver these systemd configs in a single file? (it would optimize the cloud-init payload delivery to the vm)

cchildress:
Sure, sysctl would have no problem with these being a single file. I can squash them together if that makes things better.

jackfrancis:
Yeah, let's do that

cchildress:
Done in [4bee4c7](https://github.com/Azure/aks-engine/pull/503/commits/4bee4c71ee676d4224a389dfa640f1066e113ae0)

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -130,6 +130,16 @@ write_files:
...
 
+- path: /etc/sysctl.d/10-oom-panic.conf
+  permissions: "0644"
+  owner: root
+  content: |
+    # These settings, and other Node settings are put in place by AKS. Do not attempt to change these values as any alterations will be removed by the service
+    # This setting disabled the linux OOM killer and sets panic on OOM to prevent partial node failures in out of memory conditions
+    vm.panic_on_oom = 1
+    # This setting instructs the kernel to reboot 30 seconds after a panic event to allow the node to automatically recover
+    kernel.panic = 30
```
jackfrancis: Just to play devil's advocate: why wait this long?

cchildress: We can go shorter if we want to. The recommended value for a watchdog reset (which is basically what this is) is 60 seconds. I thought that was a little too long for a node to be dead and not recovering, but I didn't want to swing too far either.

jackfrancis: Fair enough.

jackfrancis:
lgtm

PR #506, opened 2019-02-14
fix: Update client-go library to allow upgrade & scale operations to succeed in AAD scenarios

skinny:
**Reason for Change**:
While `deploy` commands work with an `aadProfile` in the API model, `upgrade` and `scale` commands fail with the following error message: 

```FATA[0377] Error upgrading cluster: No Auth Provider found for name "azure"```

I have tried to include the azure plugin in the current 7.0.0 version (in engine.go) but that fails too : 
```

# github.com/Azure/aks-engine/vendor/k8s.io/client-go/plugin/pkg/client/auth/azure
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:246:4: cannot use expiresIn (type string) as type json.Number in field value
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:247:4: cannot use expiresOn (type string) as type json.Number in field value
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:248:4: cannot use expiresOn (type string) as type json.Number in field value
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:265:23: cannot use token.token.ExpiresIn (type json.Number) as type string in assignment
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:266:23: cannot use token.token.ExpiresOn (type json.Number) as type string in assignment
vendor\k8s.io\client-go\plugin\pkg\client\auth\azure\azure.go:300:3: cannot use spt.Token (type func() adal.Token) as type adal.Token in field value
exit status 2

```

So this PR upgrades the client-go to the latest 10.0 version and includes the `azure` plugin in engine.go.  Now the upgrade and scale commands work as expected.

I could not find any specific reason why the update to v9 was reverted in december (https://github.com/Azure/aks-engine/pull/85) but as a user of the AAD functionality in aks-engine I would really like to upgrade my clusters too ;-)

### CODE REVIEW

**Comment by mboersma** [comment]
@skinny after #546 merged to master, this needs to be rebased and force-pushed. It's no longer necessary to commit the generated files. Sorry about that!

**Comment by skinny** [comment]
@mboersma done !

**Comment by skinny** [comment]
@mboersma any chance this can get merged any time soon? Let me know if it needs any changes

**Comment by mboersma** [comment]
@skinny is it possible to update client-go without updating k8s.io/api or k8s.io/apimachinery? Changing the k8s API version has significant ripple effects for us into Azure's AKS and OpenShift services, so if it needs to change we'll have more communication and testing to do before this could be merged.

**Comment by skinny** [comment]
@mboersma Unfortunately the 'scale' operation starts to fail again if I only upgrade the `client-go` dependency in `Gopkg.toml`. Running `dep ensure` results in multiple errors like :

```
Solving failure: No versions of k8s.io/api met constraints:
        kubernetes-1.10.0: Could not introduce k8s.io/api@kubernetes-1.10.0 due to multiple problematic subpackages:
        Subpackage k8s.io/api/auditregistration/v1alpha1 is missing. (Package is required by k8s.io/client-go@v10.0.0.) Subpackage k8s.io/api/autoscaling/v2beta2 is missing. (Package is required by k8s.io/client-go@v10.0.0.)      Subpackage k8s.io/api/coordination/v1beta1 is missing. (Package is required by k8s.io/client-go@v10.0.0.)     Subpackage k8s.io/api/scheduling/v1beta1 is missing. (Package is required by k8s.io/client-go@v10.0.0.)
        master: Could not introduce k8s.io/api@master, as it is not allowed by constraint kubernetes-1.10.0 from project github.com/Azure/aks-engine.
        release-1.10: Could not introduce k8s.io/api@release-1.10, as it is not allowed by constraint kubernetes-1.10.0 from project github.com/Azure/aks-engine.
        release-1.11: Could not introduce k8s.io/api@release-1.11, as it is not allowed by constraint kubernetes-1.10.0 from project github.com/Azure/aks-engine.

```

I will try to narrow this down tomorrow but it looks like we need to upgrade the other packages too.

mboersma:
> it looks like we need to upgrade the other packages too.

Ok, thanks for trying! I'll just need to check to make sure that updating the API to k8s 1.13 won't cause problems for the big projects that vendor in aks-engine.

PR #516, opened 2019-02-15
chore: install specific containerd version, make version configurable

### LINKED ISSUE #512: containerd + flannel CSE 4 failures
Initial symptoms suggest the new VHD image broke flannel + containerd.

jackfrancis:
**Reason for Change**:

This PR does specific version checking against any pre-installed versions of containerd and compares to the desired version. If a different version is present it cleans up the non-desired version and installs the desired version.

Also adds a `containerdVersion` configuration option to `kuberntesConfig` to enable user-configurable containerd versions. Ships with the following versions support:

- 1.1.5
- 1.1.6
- 1.2.4

No change made to the existing default 1.1.5 version, that change will happen in a follow-up PR.

WIP assumes moby provides any needed containerd deps, so skips install for normal docker flows.

**Issue Fixed**:

Fixes #512 

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -19,6 +19,7 @@ if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
     DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
     installDockerEngine
     installGPUDrivers
+    installContainerd
```
jackfrancis: Just to be conservative/iterative, let's keep the `installContainerd` invocation in the VHD for non-moby scenario.

(See below that it used to be here.)

I don't think we strictly need to be doing this installation, but the safest way is to make that change distinctly.

The reason we're removing it from moby is that the moby installation implementation currently installs containerd stuffs *itself*.

packer/install-dependencies.sh
```diff
@@ -39,7 +40,10 @@ for CNI_PLUGIN_VERSION in $CNI_PLUGIN_VERSIONS; do
     downloadCNI
 done
 
-installContainerd
+for CONTAINERD_VERSION in $CONTAINERD_VERSIONS; do
```
jackfrancis: Instead of actually doing the install, we're just downloading the artifacts in the VHD.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -71,7 +71,9 @@ fi
 
 installContainerRuntime
 installNetworkPlugin
-installContainerd
+if [[ "$CONTAINER_RUNTIME" == "clear-containers" ]] || [[ "$CONTAINER_RUNTIME" == "kata-containers" ]] || [[ "$CONTAINER_RUNTIME" == "containerd" ]]; then
```
jackfrancis: We only actually install the containerd artifacts for required scenarios at provisioning time.

serbrech:
Reviewed 
lgtm

pkg/api/converterfromapi.go
```diff
@@ -706,9 +706,11 @@ func convertKubernetesConfigToVLabs(apiCfg *KubernetesConfig, vlabsCfg *vlabs.Ku
 	vlabsCfg.ServiceCidr = apiCfg.ServiceCIDR
 	vlabsCfg.NetworkPolicy = apiCfg.NetworkPolicy
 	vlabsCfg.NetworkPlugin = apiCfg.NetworkPlugin
+	vlabsCfg.ContainerRuntime = apiCfg.ContainerRuntime
```
jackfrancis: Whoops, found a bug while doing this work :(

PR #523, opened 2019-02-15
feat: add support for private registry

### LINKED ISSUE #271: Allow for pulling customHyperkubeImage from private repository
---
no

---
feature request

**What version of acs-engine?**:
---
any

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
kuberentes 

**What happened**:
Would like to be able to pull `customHyperkubeImage` as describe in https://github.com/Azure/acs-engine/blob/master/docs/kubernetes/k8s-developers.md from a private Azure Container Registry. 

Would it be possible to pass Service Principle credentials or use MSI to pull the image?  What would be the effort for making this possible?

Looks like this happens in https://github.com/Azure/acs-engine/blob/8b616722debcddaf0d0d9d2baa605bb7aa77eddb/parts/k8s/kubernetesinstalls.sh#L178:1

cc: @PatrickLang and @ritazh.

ritazh:
**Reason for Change**:

Add support to pull custom hyperkube image from private registry

**Issue Fixed**:

Fixes #271 
This should also fix the broken cloud-provider-azure prow job as that uses ACR.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

cc @feiskyer


pkg/engine/params_k8s.go
```diff
@@ -46,6 +46,7 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 
 			addValue(parametersMap, "kubeDNSServiceIP", kubernetesConfig.DNSServiceIP)
 			addValue(parametersMap, "kubernetesHyperkubeSpec", kubernetesHyperkubeSpec)
+			addValue(parametersMap, "privateRegistryServer", kubernetesConfig.PrivateRegistryServer)
```
jackfrancis: Should we only add the param if it's not empty string?

parts/k8s/kubernetesinstalls.sh
```diff
@@ -273,6 +273,11 @@ installKubeletAndKubectl() {
 pullContainerImage() {
     CLI_TOOL=$1
     DOCKER_IMAGE_URL=$2
+    if [ ! -z "${PRIVATE_REGISTRY_SERVER}" ]; then
+        if [[ "$CLI_TOOL" == "docker" ]]; then
+            $CLI_TOOL login -u $SERVICE_PRINCIPAL_CLIENT_ID -p $SERVICE_PRINCIPAL_CLIENT_SECRET $PRIVATE_REGISTRY_SERVER
```
jackfrancis: Does this in practice not work w/ `img`?

jackfrancis: Seems like it should:

`https://github.com/genuinetools/img`

jackfrancis: In any event, if `img login` doesn't work in this context, we'll want to document something like the following:

"Private registry support is for default docker (moby) runtime only. It is not supported for non-docker cluster configuration scenarios, e.g., containerd, clear-containers, kata-containers."

ritazh: Added support for `img` as well.

docs/topics/kubernetes-developers.md
```diff
@@ -43,6 +43,16 @@ For VERSION, we recommend that you provide a value which would help you identify
 }
 ```
 
+* If the container registry is private, for example Azure Container Registry, then provide the name of the private registry along with the custom hyperkube image like this:
+
+```
+"kubernetesConfig": {
+    "customHyperkubeImage": "<your-private-registry>/hyperkube-amd64:<your-custom-version>",
+    "privateRegistryServer": "<your-private-registry>"
```
jackfrancis: Does privateRegistryServer make any sense in a context when customHyperkubeImage is not provided? If not, we should probably add validation.

jackfrancis:
Final high-level comment: because `privateRegistryServer` seems to be implemented in a MS-specific context (i.e., service principals are accepted auth), should be call this `privateMSRegistryServer`, or something like that? I don't see this implementation working with any old private image repository, or am I missing something?

ritazh:
@jackfrancis In theory this code can be expanded to work for other private container repositories as well, but the real reason is that only azure credentials can actually pull this image from ACR if you are using the Kubernetes Azure cloud provider. So I will update this variable to reflect that.

Should we change it to `privateMSRegistryServer` or `privateAzureRegistryServer`?

jackfrancis:
How about just `azureRegistryServer`?

ritazh:
Can we keep `private` as I think there are plans to enable public ACR in the future? WDYT?

pkg/api/vlabs/validate.go
```diff
@@ -1122,6 +1122,28 @@ func (k *KubernetesConfig) Validate(k8sVersion string, hasWindows bool) error {
...
+
+	// Check PrivateAzureRegistryServer has a valid value.
+	valid := false
+	if k.PrivateAzureRegistryServer != "" {
+		if k.CustomHyperkubeImage != "" {
+			valid = true
+		}
+	} else {
+		valid = true
+	}
```
jsturtevant: I found this logic hard to read.  would something like the following be more more readable?  I think that would remove the need for the `valid` check

```go
k.PrivateAzureRegistryServer != "" and k.CustomHyperkubeImage == "" {
    return errors.Errorf("customHyperkubeImage must be provided when privateAzureRegistryServer is provided")
}
```

parts/k8s/kubernetesinstalls.sh
```diff
@@ -278,6 +278,9 @@ installKubeletAndKubectl() {
 pullContainerImage() {
     CLI_TOOL=$1
     DOCKER_IMAGE_URL=$2
+    if [ ! -z "${PRIVATE_AZURE_REGISTRY_SERVER}" ]; then
+        $CLI_TOOL login -u $SERVICE_PRINCIPAL_CLIENT_ID -p $SERVICE_PRINCIPAL_CLIENT_SECRET $PRIVATE_AZURE_REGISTRY_SERVER
```
jsturtevant: should this use `--password-stdin` here?  https://docs.docker.com/engine/reference/commandline/login/#provide-a-password-using-stdin

parts/k8s/kubernetesmastervars.t
```diff
@@ -145,7 +145,7 @@
...
...g').cloudProviderBackoffDuration,' CLOUDPROVIDER_BACKOFF_JITTER=',parameters('cloudproviderConfig').cloudProviderBackoffJitter,' CLOUDPROVIDER_RATELIMIT=',toLower(string(parameters('cloudproviderConfig').cloudProviderRatelimit)),' CLOUDPROVIDER_RATELIMIT_QPS=',parameters('cloudproviderConfig').cloudProviderRatelimitQPS,' CLOUDPROVIDER_RATELIMIT_BUCKET=',parameters('cloudproviderConfig').cloudProviderRatelimitBucket,' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension'),' USER_ASSIGNED_IDENTITY_ID=',variables('userAssignedClientID'),' USE_INSTANCE_METADATA=',variables('useInstanceMetadata'),' LOAD_BALANCER_SKU=',variables('loadBalancerSku'),' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB'),' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount'),' CONTAINER_RUNTIME=',parameters('containerRuntime'),' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase'),' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec'),' KMS_PROVIDER_VAULT_NAME=',variables('clusterKeyVaultName'),' IS_HOSTED_MASTER={{IsHostedMaster}} PRIVATE_AZURE_REGISTRY_SERVER=',parameters('privateAzureRegistryServer'))]",
```
tariq1890: This should have been separated by a comma. I'll fix this in another PR.

Issue #522, opened 2019-02-15
Proposal: force upgrade to same k8s version


serbrech:
**Is this a request for help?**:  
No

---
**Is this an ISSUE or FEATURE REQUEST?** (choose one):
Feature Proposal

---

**description**:
We want to be able to force the upgrade process on a cluster without changing the kubernetes version. This will ensure that all components of the cluster are healthy and using the latest validated version for a given kubernetes version.

**proposed cli**: 
For an existing kubernetes `1.12.5` cluster, we want the following command to go through each nodes of the cluster and "refresh" them if necessary (update to latest valid version).

```bash
> aks-engine upgrade --force --location westus --resource-group rg --subsctiption $sub --upgrade-version 1.12.5
```

**Open questions**: 
- What should the behavior of the `--force` flag be for a usual next k8s version upgrade?
- Should the --force flag allow other arbitrary upgrade (not valid per our `get-versions` matrix)? 

`--force` makes it seem like you can _force_ the upgrade no matter what. However, a forbidden upgrade path (not in the `get-versions` mattrix) should still be forbidden I suppose?

**Related**
https://github.com/Azure/acs-engine/pull/3810

jackfrancis:
I suppose the most graceful way to do this is to support "downgrade" as well. In other words, any version that is supported for create can be the target version, including versions "lower" than the current version on the cluster.

I think that would in fact offer more help to folks, in scenarios where an upgrade brought undesired breakage.

mboersma:
I think ideally `--force` would allow you to bypass our normal upgrade rules and target any supported version (allowing "downgrade" as well). We should print some warning in that case, however, because the chances of a downgrade breaking deployed apps are much higher.

serbrech:
@mboersma this is what we are going for

CecileRobertMichon:
fixed in #525

PR #525, opened 2019-02-15
feat: add force parameter to the upgrade command

serbrech:
No

---
Feature Proposal

---

**description**:
We want to be able to force the upgrade process on a cluster without changing the kubernetes version. This will ensure that all components of the cluster are healthy and using the latest validated version for a given kubernetes version.

**proposed cli**: 
For an existing kubernetes `1.12.5` cluster, we want the following command to go through each nodes of the cluster and "refresh" them if necessary (update to latest valid version).

```bash
> aks-engine upgrade --force --location westus --resource-group rg --subsctiption $sub --upgrade-version 1.12.5
```

**Open questions**: 
- What should the behavior of the `--force` flag be for a usual next k8s version upgrade?
Same behavior

- What happens for same-version upgrade?
A: Currently, when a node is already on the requested version, the upgrade either fail validation, or the node upgrade ends up being a no-op. `--force` runs the upgrade process on these nodes as well

- Should the --force flag allow other arbitrary upgrade (not valid per our `get-versions` matrix)? 
A : yes, using --force will allow any arbitrary version jump, or even downgrade. You're on your own.

**Related**
implements #522 
https://github.com/Azure/acs-engine/pull/3810

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -277,6 +277,7 @@ func (uc *upgradeCmd) run(cmd *cobra.Command, args []string) error {
 	upgradeCluster.DataModel = uc.containerService
 	upgradeCluster.NameSuffix = uc.nameSuffix
 	upgradeCluster.AgentPoolsToUpgrade = uc.agentPoolsToUpgrade
+	upgradeCluster.Force = uc.force
 
 	kubeConfig, err := engine.GenerateKubeConfig(uc.containerService.Properties, uc.location)
 	if err != nil {
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -219,9 +219,9 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 }
 
 func readNameSuffixFromARMTemplate(armTemplateHandle io.Reader) (string, error) {
-	var template *map[string]interface{}
+	var azureDeployTemplate *map[string]interface{}
 	decoder := json.NewDecoder(armTemplateHandle)
-	if err := decoder.Decode(&template); err != nil {
+	if err := decoder.Decode(&azureDeployTemplate); err != nil {
 		return "", err
 	}
 
@@ -234,7 +234,7 @@ func readNameSuffixFromARMTemplate(armTemplateHandle io.Reader) (string, error) 
 		defaultValueKey = "defaultValue"
 	)
 
-	if templateParameters, okType = (*template)[parametersKey].(map[string]interface{}); !okType {
+	if templateParameters, okType = (*azureDeployTemplate)[parametersKey].(map[string]interface{}); !okType {
 		return "", errors.Errorf("error asserting data from key \"%s\" in file %q",
 			parametersKey, "azuredeploy.json")
 	}
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -201,7 +201,7 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 	}
 
 	//allows to identify VMs in the resource group that belong to this cluster.
-	if nameSuffix, err := readNameSuffixFromArmTemplate(armTemplateHandle); err == nil {
+	if nameSuffix, err := readNameSuffixFromARMTemplate(armTemplateHandle); err == nil {
 		uc.nameSuffix = nameSuffix
 	} else {
 		return errors.Wrap(err, "Failed to read nameSuffix from azuredeploy.json")
@@ -217,7 +217,7 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 	return nil
 }
 
-func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
+func readNameSuffixFromARMTemplate(armTemplateHandle io.Reader) (string, error) {
 	var template *map[string]interface{}
 	decoder := json.NewDecoder(armTemplateHandle)
 	if err := decoder.Decode(&template); err != nil {
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -204,7 +204,8 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 	if nameSuffix, err := readNameSuffixFromARMTemplate(armTemplateHandle); err == nil {
 		uc.nameSuffix = nameSuffix
 	} else {
-		return errors.Wrap(err, "Failed to read nameSuffix from azuredeploy.json")
+		templatePath := path.Join(uc.deploymentDirectory, "azuredeploy.json")
+		return errors.Wrapf(err, "Failed to read nameSuffix from %s", templatePath)
 	}
 
 	log.Infoln(fmt.Sprintf("Upgrading cluster with name suffix: %s", uc.nameSuffix))
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -7,7 +7,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"os"
 	"path"
 	"time"
@@ -151,20 +151,20 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
-	err = uc.validateCurrentState()
+	templatePath := path.Join(uc.deploymentDirectory, "azuredeploy.json")
+	armTemplateHandle, err := os.Open(templatePath)
+	if err != nil {
+		return errors.Wrap(err, "error reading ARM file")
+	}
+	defer armTemplateHandle.Close()
+	err = uc.validateCurrentLocalState(armTemplateHandle)
 	if err != nil {
 		return errors.Wrap(err, "error validating the api model")
 	}
 	return nil
 }
 
-func (uc *upgradeCmd) validateCurrentState() error {
-	if uc.containerService.Location == "" {
-		uc.containerService.Location = uc.location
-	} else if uc.containerService.Location != uc.location {
-		return errors.New("--location does not match api model location")
-	}
-
+func (uc *upgradeCmd) validateTargetVersion() error {
 	// Get available upgrades for container service.
 	orchestratorInfo, err := api.GetOrchestratorVersionProfile(uc.containerService.Properties.OrchestratorProfile, uc.containerService.Properties.HasWindows())
 	if err != nil {
@@ -183,43 +183,28 @@ func (uc *upgradeCmd) validateCurrentState() error {
 	if !found {
 		return errors.Errorf("upgrading from Kubernetes version %s to version %s is not supported. To see a list of available upgrades, use 'aks-engine get-versions --version %s'", uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion, uc.upgradeVersion, uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion)
 	}
+	return nil
+}
 
-	// Read the name suffix from the parameters to identify VMs in the resource group that belong to this cluster.
-	templatePath := path.Join(uc.deploymentDirectory, "azuredeploy.json")
-	contents, err := ioutil.ReadFile(templatePath)
-	if err != nil {
-		return errors.Wrap(err, "error reading ARM file")
+func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) error {
+	if uc.containerService.Location == "" {
+		uc.containerService.Location = uc.location
+	} else if uc.containerService.Location != uc.location {
+		return errors.New("--location does not match api model location")
 	}
 
-	var template interface{}
-	json.Unmarshal(contents, &template)
-
-	var templateMap, templateParameters, nameSuffixParam map[string]interface{}
-	var okType bool
-
-	if templateMap, okType = template.(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from file %q", templatePath)
+	if !uc.force {
+		err := uc.validateTargetVersion()
+		if err != nil {
+			return errors.Wrap(err, "Invalid upgrade target version. Consider using --force if you really want to proceed")
+		}
 	}
 
-	const (
-		parametersKey   = "parameters"
-		nameSuffixKey   = "nameSuffix"
-		defaultValueKey = "defaultValue"
-	)
-
-	if templateParameters, okType = templateMap[parametersKey].(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from key \"%s\" in file %q",
-			parametersKey, templatePath)
-	}
-
-	if nameSuffixParam, okType = templateParameters[nameSuffixKey].(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from key \"%s.%s\" in file %q",
-			parametersKey, nameSuffixKey, templatePath)
-	}
-
-	if uc.nameSuffix, okType = nameSuffixParam[defaultValueKey].(string); !okType {
-		return errors.Errorf("error asserting data from key \"%s.%s.%s\" in file %q",
-			parametersKey, nameSuffixKey, defaultValueKey, templatePath)
+	//allows to identify VMs in the resource group that belong to this cluster.
+	if nameSuffix, err := readNameSuffixFromArmTemplate(armTemplateHandle); err == nil {
+		uc.nameSuffix = nameSuffix
+	} else {
+		return errors.Wrap(err, "Failed to read nameSuffix from azuredeploy.json")
 	}
 
 	log.Infoln(fmt.Sprintf("Upgrading cluster with name suffix: %s", uc.nameSuffix))
@@ -230,6 +215,40 @@ func (uc *upgradeCmd) validateCurrentState() error {
 		uc.agentPoolsToUpgrade[agentPool.Name] = true
 	}
 	return nil
+}
+
+func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
+	var template *map[string]interface{}
+	decoder := json.NewDecoder(armTemplateHandle)
+	if err := decoder.Decode(&template); err != nil || template == nil {
+		return "", err
+	}
+
+	var templateParameters, nameSuffixParam map[string]interface{}
+	var okType bool
+
+	const (
+		parametersKey   = "parameters"
+		nameSuffixKey   = "nameSuffix"
+		defaultValueKey = "defaultValue"
+	)
+
+	if templateParameters, okType = (*template)[parametersKey].(map[string]interface{}); !okType {
+		return "", errors.Errorf("error asserting data from key \"%s\" in file %q",
+			parametersKey, "azuredeploy.json")
+	}
+
+	if nameSuffixParam, okType = templateParameters[nameSuffixKey].(map[string]interface{}); !okType {
+		return "", errors.Errorf("error asserting data from key \"%s.%s\" in file %q",
+			parametersKey, nameSuffixKey, "azuredeploy.json")
+	}
+
+	var nameSuffix string
+	if nameSuffix, okType = nameSuffixParam[defaultValueKey].(string); !okType {
+		return "", errors.Errorf("error asserting data from key \"%s.%s.%s\" in file %q",
+			parametersKey, nameSuffixKey, defaultValueKey, "azuredeploy.json")
+	}
+	return nameSuffix, nil
 }
 
 func (uc *upgradeCmd) run(cmd *cobra.Command, args []string) error {
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -220,7 +220,7 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
 	var template *map[string]interface{}
 	decoder := json.NewDecoder(armTemplateHandle)
-	if err := decoder.Decode(&template); err != nil || template == nil {
+	if err := decoder.Decode(&template); err != nil {
 		return "", err
 	}
 
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -40,6 +40,7 @@ type upgradeCmd struct {
 	upgradeVersion      string
 	location            string
 	timeoutInMinutes    int
+	force               bool
 
 	// derived
 	containerService    *api.ContainerService
@@ -69,6 +70,7 @@ func newUpgradeCmd() *cobra.Command {
 	f.StringVar(&uc.deploymentDirectory, "deployment-dir", "", "the location of the output from `generate` (required)")
 	f.StringVarP(&uc.upgradeVersion, "upgrade-version", "k", "", "desired kubernetes version (required)")
 	f.IntVar(&uc.timeoutInMinutes, "vm-timeout", -1, "how long to wait for each vm to be upgraded in minutes")
+	f.BoolVarP(&uc.force, "force", "f", false, "force upgrading the cluster to desired version. Allows same version upgrades and downgrades.")
 	addAuthFlags(&uc.authArgs, f)
 
 	return upgradeCmd
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -32,7 +32,7 @@ const (
 )
 
 type upgradeCmd struct {
-	authArgs
+	authProvider
 
 	// user input
 	resourceGroupName   string
@@ -53,7 +53,9 @@ type upgradeCmd struct {
 }
 
 func newUpgradeCmd() *cobra.Command {
-	uc := upgradeCmd{}
+	uc := upgradeCmd{
+		authProvider: &authArgs{},
+	}
 
 	upgradeCmd := &cobra.Command{
 		Use:   upgradeName,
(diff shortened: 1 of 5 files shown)


cmd/upgrade.go
```diff
@@ -52,7 +53,9 @@ type upgradeCmd struct {
 }
 
 func newUpgradeCmd() *cobra.Command {
-	uc := upgradeCmd{}
+	uc := upgradeCmd{
```
serbrech: This follows the same pattern as in deploy and abstract away the auth and client during tests.

cmd/upgrade.go
```diff
@@ -147,6 +151,14 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
```
serbrech: separating the validation logic

cmd/upgrade.go
```diff
@@ -159,11 +171,6 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error getting list of available upgrades")
 	}
```
serbrech: This was adding the current version to the allowed upgrade matrix, making the Cmd validation pass. the upgrade process however was re validating the same thing, disallowing same-version upgrades, and skipping validation alltogether for HostedMaster. I chose to move the validation in the cmd and let the underlying library (operations/upgradecluster) do the upgrades as it pleases.

cmd/upgrade_test.go
```diff
@@ -4,104 +4,139 @@
...
+
+	g.Expect(output.Use).Should(Equal(upgradeName))
+	g.Expect(output.Short).Should(Equal(upgradeShortDescription))
+	g.Expect(output.Long).Should(Equal(upgradeLongDescription))
+	g.Expect(output.Flags().Lookup("location")).NotTo(BeNil())
+	g.Expect(output.Flags().Lookup("resource-group")).NotTo(BeNil())
+	g.Expect(output.Flags().Lookup("deployment-dir")).NotTo(BeNil())
+	g.Expect(output.Flags().Lookup("upgrade-version")).NotTo(BeNil())
+}
+
```
serbrech: it is clearer now that we will authorize same version upgrade only when the --force parameter is provided

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -177,13 +176,6 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 				continue
 			}
 
-			// Skip the VM upgrade validation for managed clusters as it only applies to aks-engine version support.
```
serbrech: removeing this validation as it is redundant with the cmd validation, and ignored for hosted master clusters

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -197,29 +195,6 @@ var _ = Describe("Upgrade Kubernetes cluster tests", func() {
 		Expect(err.Error()).To(Equal("DeleteNetworkInterface failed"))
 	})
```
serbrech: This was testing for aks-engine downgrade validation, which is now handled at the cmd layer.

jackfrancis:
This is going in the right direction. The business logic implementation should be focussed on doing actual operations, and should be decomposed in such a way so as to be general and composable. Validation should happen at a higher layer, and should inform particular implementation composition patterns. (As opposed to the validation being statically coupled to the business logic itself.)

pkg/armhelpers/mockclients.go
```diff
@@ -27,26 +27,28 @@ import (
...
+	FailGetStorageClient                    bool
+	FailDeleteNetworkInterface              bool
+	FailGetKubernetesClient                 bool
+	FailListProviders                       bool
+	ShouldSupportVMIdentity                 bool
+	FailDeleteRoleAssignment                bool
+	MockKubernetesClient                    *MockKubernetesClient
+	FakeListVirtualMachineScaleSetsResult   func() []compute.VirtualMachineScaleSet
+	FakeListVirtualMachineScaleSetVMsResult func() []compute.VirtualMachineScaleSetVM
 }
```
serbrech: introduce ways to stub out Azure client VMSS responses.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -84,21 +85,26 @@ func (uc *UpgradeCluster) UpgradeCluster(az armhelpers.AKSEngineClient, kubeConf
...
-	upgrader = u
 
-	if err := upgrader.RunUpgrade(); err != nil {
+	if err := uc.getUpgradeWorkflow(kubeConfig, aksEngineVersion).RunUpgrade(); err != nil {
 		return err
 	}
 
 	uc.Logger.Infof("Cluster upgraded successfully to Kubernetes version %s\n", upgradeVersion)
 	return nil
 }
```
serbrech: getUpgradeWorkflow encapsulates the default initialization of the UpgradeWorkflow.

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -221,6 +241,49 @@ var _ = Describe("Upgrade Kubernetes cluster tests", func() {
 		Expect(err.Error()).To(Equal("DeleteRoleAssignmentByID failed"))
 	})
```
serbrech: probably could use some cleanup/refactoring to make the test more readable

pkg/armhelpers/interfaces.go
```diff
@@ -26,6 +26,24 @@ type VirtualMachineListResultPage interface {
 	Values() []compute.VirtualMachine
 }
```
serbrech: the interface for AKSEngineClient did not add the VMSS func, so they we couldn't fake them.

cmd/upgrade.go
```diff
@@ -147,23 +151,26 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
-	if uc.containerService.Location == "" {
```
serbrech: this validation was pushed down in the validate func

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
+	uc.agentPoolsToUpgrade = make(map[string]bool)
+	uc.agentPoolsToUpgrade[kubernetesupgrade.MasterPoolName] = true
+	for _, agentPool := range uc.containerService.Properties.AgentPoolProfiles {
+		uc.agentPoolsToUpgrade[agentPool.Name] = true
+	}
+	return nil
+}
 
-	if templateMap, okType = template.(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from file %q", templatePath)
```
serbrech: Below, passing the handle in the function instead of the string allows for easy testing without touching the fileSystem

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
+func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
+	var template *map[string]interface{}
```
serbrech: Below, passing the handle in the function instead of the string allows for easy testing without touching the fileSystem

cmd/upgrade.go
```diff
@@ -147,23 +151,26 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
 
+func (uc *upgradeCmd) validateTargetVersion() error {
 	// Get available upgrades for container service.
 	orchestratorInfo, err := api.GetOrchestratorVersionProfile(uc.containerService.Properties.OrchestratorProfile, uc.containerService.Properties.HasWindows())
 	if err != nil {
 		return errors.Wrap(err, "error getting list of available upgrades")
 	}
 
-	// Add the current version to account for failed upgrades.
-	orchestratorInfo.Upgrades = append(orchestratorInfo.Upgrades, &api.OrchestratorProfile{
```
CecileRobertMichon: the primary purpose of this was to allow upgrade operation retries on previously failed upgrades. Does this change mean it is not possible to retry a failed upgrade anymore without passing in the --force flag? we need to update the upgrade documentation to reflect this if that's the case.

I have a slight preference for not requiring the --force flag for retries. It might break CI for some users and/or encourage the use of the --force flag by default which is dangerous since it skips validation all together...

serbrech: The behaviour of the cmd validation is to validate against local state when it comes to version.
the behaviour of the upgrade process is unchanged when you don't pass --force, if the validation of the cmd goes through.
Which means that as long as your local apimodel passes the cmd validation, you get the same behaviour as before (skip upgraded nodes, upgrade nodes that are behind)

serbrech: I spent a good amount of time writing tests around operations/upgradecluster.go to not break existing behaviour.

CecileRobertMichon:
Did an online review with Stephane. To summarize:

1) We need to make sure that the "retry failed upgrade" functionality still works without the current version in the list of available upgrades in cmd validation.

2) Separating the PR in smaller chunks to permit easier review and faster revert if necessary might be a good idea but will leave it up to Stephane to decide if the trade off is worth the additional work to separate.

3) Removing the `upgradable` validation in upgradeCluster reduces the strength of validation:
- the cmd validation only checks if the upgrade is valid based on the version string in the apimodel which does not necessarily reflect the actual state of things and can be easily modified by the user
- if an upgrade fails, we don't have in place validation to check that the user retries upgrade to the same version which means that the following could happen:
User starts upgrade of 1.9.5 cluster to 1.10.7. Upgrade fails leaving some nodes in 1.10.7 and some in 1.9.5. User upgrades to 1.9.6: this will pass validation because the apimodel will still have version 1.9.5 (and 1.9.5 -> 1.9.6 is a valid upgrade) but will cause a downgrade of two of the nodes. Per-node validation would have previously caught this invalid downgrade operation but removing per-node validation makes us vulnerable in such scenarios.

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
+	return nil
+}
 
-	// Read the name suffix from the parameters to identify VMs in the resource group that belong to this cluster.
-	templatePath := path.Join(uc.deploymentDirectory, "azuredeploy.json")
-	contents, err := ioutil.ReadFile(templatePath)
-	if err != nil {
-		return errors.Wrap(err, "error reading ARM file")
+func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) error {
+	if uc.containerService.Location == "" {
```
tariq1890: Let's put `uc.containerService.Location` to a variable.

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
+	}
+	return nil
+}
 
-	if templateMap, okType = template.(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from file %q", templatePath)
+func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
+	var template *map[string]interface{}
+	decoder := json.NewDecoder(armTemplateHandle)
+	if err := decoder.Decode(&template); err != nil || template == nil {
```
tariq1890: I don't think the `template == nil` check is required. If this is to handle a case where `armTemplateHandle` is nil or empty, then `readNameSuffixFromArmTemplate` should never be called in those cases.

```suggestion
	if err := decoder.Decode(&template); err != nil {
```

cmd/upgrade_test.go
```diff
@@ -4,104 +4,149 @@
...
+	containerServiceMock := api.CreateMockContainerService("testcluster", "1.10.13", 3, 2, false)
+	containerServiceMock.Location = "centralus"
+	upgradeCmd.containerService = containerServiceMock
+	err := upgradeCmd.validateCurrentLocalState(fakeArmTemplateHandle)
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("upgrading from Kubernetes version 1.10.13 to version 1.10.13 is not supported"))
+}
+
+func TestUpgradeForceSameVersionShouldSucceed(t *testing.T) {
+	fakeArmTemplateHandle := strings.NewReader("{\"parameters\" : { \"nameSuffix\" : {\"defaultValue\" : \"test\"}}}")
```
tariq1890:
```suggestion
	fakeArmTemplateHandle := strings.NewReader(`{"parameters" : { "nameSuffix" : {"defaultValue" : "test"}}}`)
```

pkg/armhelpers/mockclients.go
```diff
@@ -371,6 +462,57 @@ func (mc *MockAKSEngineClient) ListVirtualMachines(ctx context.Context, resource
...
+//GetVirtualMachine mock
+func (mc *MockAKSEngineClient) GetVirtualMachine(ctx context.Context, resourceGroup, name string) (compute.VirtualMachine, error) {
+	if mc.FailGetVirtualMachine {
+		return compute.VirtualMachine{}, errors.New("GetVirtualMachine failed")
+	}
+	return mc.MakeFakeVirtualMachine(DefaultFakeVMName, defaultK8sVersionForFakeVMs), nil
+}
+
+// MakeFakeVirtualMachineScaleSetVM creates a fake vmss vm
+func (mc *MockAKSEngineClient) MakeFakeVirtualMachineScaleSetVM(orchestratorTag string) compute.VirtualMachineScaleSetVM {
```
tariq1890: `MockAKSEngineClient` already communicates that it is going to be a Fake VM 

```suggestion
func (mc *MockAKSEngineClient) CreateVirtualMachineScaleSetVM(orchestratorTag string) compute.VirtualMachineScaleSetVM {
```

pkg/armhelpers/mockclients.go
```diff
@@ -412,42 +559,19 @@ func (mc *MockAKSEngineClient) ListVirtualMachines(ctx context.Context, resource
...
 
-//GetVirtualMachine mock
-func (mc *MockAKSEngineClient) GetVirtualMachine(ctx context.Context, resourceGroup, name string) (compute.VirtualMachine, error) {
-	if mc.FailGetVirtualMachine {
-		return compute.VirtualMachine{}, errors.New("GetVirtualMachine failed")
-	}
-
-	vm1Name := "k8s-agentpool1-12345678-0"
+//MakeFakeVirtualMachine returns a fake compute.VirtualMachine
+func (mc *MockAKSEngineClient) MakeFakeVirtualMachine(vmName string, orchestratorVersion string) compute.VirtualMachine {
```
tariq1890: Same as previous comment.

pkg/armhelpers/mockclients.go
serbrech: I think that looks like it's an actual method of what we are faking, versus a helper to provide you with an instance of a faked VM, so I'll keep it as is

cmd/upgrade_test.go
serbrech: that's better, thanks

cmd/upgrade.go
serbrech: That's correct, it was defensive programming because I wrote test directly on validateCurrentLocalState, but it's not public outside the package so not needed as we error out earlier in that case.

cmd/upgrade.go
serbrech: Yeah, I just didn't touch that code as there is no tests around it. I just extracted it to a func.
I'll leave it as is for this PR if that's ok.

serbrech:
@CecileRobertMichon 
> 2. Separating the PR in smaller chunks to permit easier review and faster revert if necessary might be a good idea but will leave it up to Stephane to decide if the trade off is worth the additional work to separate.

Waiting for #590 to be merged

serbrech:
@CecileRobertMichon 
> 3\. Removing the `upgradable` validation in upgradeCluster reduces the strength of validation:
Looking at the diff to reintroduce it, note that this validation was only done for AvailibilitySets and not for VMSS vms

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -176,34 +182,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
+					continue
 				}
-			} else if currentVersion == goalVersion {
-				if strings.Contains(*(vm.Name), MasterVMNamePrefix) {
-					uc.Logger.Infof("Master VM name: %s, orchestrator: %s (UpgradedMasterVMs)", *vm.Name, currentVersion)
-					*uc.UpgradedMasterVMs = append(*uc.UpgradedMasterVMs, vm)
-				} else {
-					uc.addVMToAgentPool(vm, false)
+				// If the current version is different than the desired version then we add the VM to the list of VMs to upgrade.
+				if currentVersion != goalVersion {
```
CecileRobertMichon: where is the `|| uc.Force` for VMAS?

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: this comment isn't relevant anymore is it?

serbrech: nope resolving

cmd/upgrade.go
```diff
@@ -147,23 +151,26 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
-	if uc.containerService.Location == "" {
-		uc.containerService.Location = uc.location
-	} else if uc.containerService.Location != uc.location {
-		return errors.New("--location does not match api model location")
+	templatePath := path.Join(uc.deploymentDirectory, "azuredeploy.json")
+	armTemplateHandle, err := os.Open(templatePath)
```
CecileRobertMichon: s/arm/ARM

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
+	uc.agentPoolsToUpgrade[kubernetesupgrade.MasterPoolName] = true
+	for _, agentPool := range uc.containerService.Properties.AgentPoolProfiles {
+		uc.agentPoolsToUpgrade[agentPool.Name] = true
+	}
+	return nil
+}
 
-	if templateMap, okType = template.(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from file %q", templatePath)
+func readNameSuffixFromArmTemplate(armTemplateHandle io.Reader) (string, error) {
```
CecileRobertMichon: s/Arm/ARM

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -176,34 +182,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
-			if currentVersion == "" {
-				uc.Logger.Infof("Skipping VM: %s for upgrade as the orchestrator version could not be determined.", *vm.Name)
-				continue
-			}
 
-			// Skip the VM upgrade validation for managed clusters as it only applies to aks-engine version support.
-			if !uc.DataModel.Properties.IsHostedMasterProfile() {
-				if err := uc.upgradable(currentVersion); err != nil {
-					return err
+			if uc.Force {
```
serbrech: the if Force is above for VMAS, we don't need to check anything we just add it to the ToUpgrade list.

pkg/operations/kubernetesupgrade/upgradecluster.go
serbrech: added comment above

cmd/upgrade.go
```diff
@@ -176,53 +183,72 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
...
 		nameSuffixKey   = "nameSuffix"
 		defaultValueKey = "defaultValue"
 	)
 
-	if templateParameters, okType = templateMap[parametersKey].(map[string]interface{}); !okType {
-		return errors.Errorf("error asserting data from key \"%s\" in file %q",
-			parametersKey, templatePath)
+	if templateParameters, okType = (*template)[parametersKey].(map[string]interface{}); !okType {
+		return "", errors.Errorf("error asserting data from key \"%s\" in file %q",
+			parametersKey, "azuredeploy.json")
```
CecileRobertMichon: why not print the path of the file here? (same question for below)

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: makes sense. I didn't see the `uc.addVMToUpgradeSets` below the deleted lines...

cmd/upgrade.go
serbrech: we don't have it because we don't rely on file path anymore but on io.Reader for testability.
the error is wrapped from the caller. it's not outputing the whole path though, that can be changed

serbrech: To be clear, is that good enough or do I need to reconstruct the path to the azuredeploy.json?

CecileRobertMichon: I think it's fine like that, was just curious.

cmd/upgrade.go
tariq1890: `template` is very generic and leads to ambiguous naming since its commonly used. It could mean `text/template` or `html/template` and it also leads to a name clash with that of an import name. Can we use so mething more specific?

cmd/upgrade.go
serbrech: added and tested

cmd/upgrade.go
serbrech: yep

pkg/operations/kubernetesupgrade/upgradecluster.go
serbrech: is that ACK @CecileRobertMichon ?

cmd/upgrade.go
serbrech: done

cmd/upgrade_test.go
```diff
@@ -4,104 +4,169 @@
...
+	g.Expect(err).To(HaveOccurred())
+	g.Expect(err.Error()).To(ContainSubstring("_output/myspecialtestfolder/azuredeploy.json"))
+}
+func TestUpgradeForceSameVersionShouldSucceed(t *testing.T) {
+	fakeARMTemplateHandle := strings.NewReader(`{"parameters" : { "nameSuffix" : {"defaultValue" : "test"}}}`)
+	g := NewGomegaWithT(t)
+	upgradeCmd := &upgradeCmd{
+		resourceGroupName:   "rg",
+		deploymentDirectory: "_output/test",
+		upgradeVersion:      "1.10.13",
```
CecileRobertMichon: Does this test (and any other tests) require this k8s patch version to be supported? If so you could use `common.GetLatestPatchVersion` to get a valid 1.10 version and avoid having to update this UT every time we deprecate a patch version.

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -266,47 +241,168 @@ var _ = Describe("Upgrade Kubernetes cluster tests", func() {
...
-		err := uc.UpgradeCluster(&mockClient, "kubeConfig", TestAKSEngineVersion)
-		Expect(err).NotTo(HaveOccurred())
-		Expect(uc.AgentPoolScaleSetsToUpgrade[0].VMsToUpgrade).To(HaveLen(2))
+	Context("When upgrading a cluster with AvailibilitySets VMs", func() {
+		var cs *api.ContainerService
+		var uc UpgradeCluster
+		var mockClient armhelpers.MockAKSEngineClient
+		BeforeEach(func() {
+			mockClient = armhelpers.MockAKSEngineClient{}
+			cs = api.CreateMockContainerService("testcluster", "1.9.10", 3, 3, false)
```
CecileRobertMichon: same question here about version support for 1.9.10

serbrech:
Good point, I considered it but didn't fix it.
For --force, this has no incidence as it should just always pass. for not
--force, the base data we rely on for the test will be changing. The only
way to fix this is to fake the version/valid upgrade table.
Can we add it as a //TODO in the backlog, I can look into it before the end
of the week.

CecileRobertMichon:
why not use `
common.GetLatestPatchVersion` to dynamically set the version?

serbrech:
> why not use ` common.GetLatestPatchVersion` to dynamically set the version?

I think there is an easier way : 
`common.AllKubernetesSupportedVersions` is public, so I can simply set it to a minimal map that makes sense for the test case. :)

cmd/upgrade_test.go
```diff
@@ -4,104 +4,239 @@
...
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
-var _ = Describe("the upgrade command", func() {
+var validVersionsBackup map[string]bool
 
-	It("should create an upgrade command", func() {
-		output := newUpgradeCmd()
+func SetupValidVersions(validVersions map[string]bool) {
```
CecileRobertMichon: does function need to be exported?

cmd/upgrade_test.go
```diff
@@ -4,104 +4,239 @@
...
 
-		Expect(output.Use).Should(Equal(upgradeName))
-		Expect(output.Short).Should(Equal(upgradeShortDescription))
-		Expect(output.Long).Should(Equal(upgradeLongDescription))
-		Expect(output.Flags().Lookup("location")).NotTo(BeNil())
-		Expect(output.Flags().Lookup("resource-group")).NotTo(BeNil())
-		Expect(output.Flags().Lookup("deployment-dir")).NotTo(BeNil())
-		Expect(output.Flags().Lookup("upgrade-version")).NotTo(BeNil())
-	})
+func ResetValidVersions() {
```
CecileRobertMichon: ditto

cmd/upgrade_test.go
serbrech: indeed no 👍

CecileRobertMichon:
merging manually because windows 1.13 test failure was a flake.

PR #529, opened 2019-02-16
fix: Allow to create only masters so that agents can be created at a later…

awesomenix:
… point, this can only be possible only if Standard loadbalancer would be used, regardless nothing should stop from aks-engine creating masters only

**Reason for Change**:

Aks engine should have ability to create masters only, but later can be updated to have agent pools, this PR fixes ability to do that by making sure aks engine doesnt panic

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


tariq1890:
Looks like Windows is not able to handle this too well.  My guess is that it's depending on `primaryAvailabilitySetName` or `primaryScaleSetName`.

tariq1890:
Yup. No AgentPoolProfiles lead to not setting of the aforesaid parameters. So it's a no go for windows. See [here](https://github.com/Azure/aks-engine/blob/master/parts/k8s/windowskubeletfunc.ps1#L46)

parts/k8s/kubernetesmastervars.t
```diff
@@ -209,11 +211,19 @@
 {{end}}
     "nsgID": "[resourceId('Microsoft.Network/networkSecurityGroups',variables('nsgName'))]",
 {{if AnyAgentUsesVirtualMachineScaleSets}}
+{{if .AgentPoolProfiles}}
     "primaryScaleSetName": "[concat(parameters('orchestratorName'), '-{{ (index .AgentPoolProfiles 0).Name }}-',parameters('nameSuffix'), '-vmss')]",
+{{else}}
+    "primaryScaleSetName": "",
```
tariq1890: @awesomenix As mentioned earlier, this is an issue for Windows. @PatrickLang @jsturtevant  Is it feasible to allow for the Windows flows to acknowledge the absence of AgentPoolProfiles and have a MastersOnly setup?

(Note: Apologies for tagging you all on a holiday. This is by no means urgent :) )

awesomenix: Hey! Tariq, thank you for taking time to review this PR on holiday, I am looking at reason for windows failures, its because of a flaky test

span>• Failure [1525.715 seconds]‌
/__w/1/s/gopath/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:84‌
  with a windows agent pool
  ‌/__w/1/s/gopath/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1052‌
    ‌should be able to deploy an iis webserver [It]‌
    ‌/__w/1/s/gopath/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1053‌

    ‌Expected‌
        <bool>: false
    to be true‌

    /__w/1/s/gopath/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1077

Its unrelated to this change.

But i completely agree with you this is not a valid scenario, but the script you tagged doesnt even run in during deployment of Linux Master VM. That script runs only during execution of deployment of windows agent VMs in which case we would have specified the necessary variables like primaryScaleSetName and  primaryAvailabilitySetName. So the scenario is something like

1. Generate template for Master Profiles only
2. Deploy
3. Generate template for Agent profiles only
4. Deploy

tariq1890: It's very unlikely that all the windows tests would fail with the same error (which seems to be the case right now) if this is a flake.

I am triggering the build again to confirm.

awesomenix: Seems like windows e2e never passed looking at the history https://msazure.visualstudio.com/One/_build?definitionId=50661&_a=summary
Can @PatrickLang @jsturtevant look into these failures?

jackfrancis: Indeed, there is a known issue w/ the IIS tests on all Windows E2E configs right now.

jsturtevant: Do we know what the issue is? Looks like the pods are starting but external communication is failing. @PatrickLang Is this related to Azure CNI1.17?  

The logs in the point are not super helpful:

```
2019/02/17 05:10:35 #### $ kubectl expose deployment iis-kubernetes-southeastasia-23972-97695 --type LoadBalancer -n default --target-port 80 --port 80 completed in 1.249608524s
STEP‌: Verifying that the service is reachable and returns the default IIS start page‌
2019/02/17 05:18:23 Unable to validate URL http://13.76.182.186 after 3h40m0s, err: &url.Error{Op:"Get", URL:"http://13.76.182.186", Err:(*net.OpError)(0xc0000aa050)}
```

parts/k8s/kubernetesbase.t
```diff
@@ -50,25 +54,29 @@
...
+          {{else}}
+            {{template "k8s/kubernetesagentresourcesvmas.t" .}}
+          {{end}}
         {{end}}
       {{end}}
+      ,
     {{end}}
     {{if IsHostedMaster}}
       {{if not IsCustomVNET}}
-        ,{
```
jackfrancis: Just to make this changeset as clean as possible, can we use the original `,{` usage, even if it's arguably less readable?

parts/k8s/kubernetesmastervars.t
```diff
@@ -166,7 +166,9 @@
     "vnetResourceGroupNameResourceSegmentIndex": 4,
 {{if IsHostedMaster}}
   {{if IsCustomVNET}}
+{{if .AgentPoolProfiles}}
```
jackfrancis: nit: indent this `{{if .AgentPoolProfiles}}` (and the `{{end}}`) so that it is 2 chars indented inside the `{{if IsCustomVNET}}` block.

jackfrancis:
Added some nits. Overall looks fine.

parts/k8s/kubernetesbase.t
awesomenix: Assuming that there isnt any agentpool profiles, then this statement would start with a "," which arm doesnt seem to like :( had tested with different combo and had to settle with this change.

jackfrancis:
@awesomenix this PR just needs one more `make build` to produce (and commit) updated go template output

awesomenix:
> @awesomenix this PR just needs one more `make build` to produce (and commit) updated go template output

Oops! thanks for catching, updated the PR

jackfrancis:
lgtm

tariq1890:
Deferring to @jackfrancis on deciding If we are to wait until the windows e2e fixes to merge this PR or not.

Issue #277, opened 2019-01-10
Support ADFS Identity System


honcao:
**Is this a request for help?**:
No
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
Feature Request
---

**What version of aks-engine?**:
The latest version
---

Azure Stack support ADFS as identity system. 
Currently aks- engine only support AzureAD.

So Aks engine will also need support ADFS as identity system.

honcao:
Fixed.

PR #531, opened 2019-02-19
feat: support ADFS for azurestack

honcao:
**Reason for Change**:

feat: support ADFS for azurestack

**Issue Fixed**:

#277 

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
Azure Stack support ADFS as identity system.
Currently aks- engine only support AzureAD.
This PR is for support ADFS on Azure Stack.


pkg/api/const.go
```diff
@@ -419,3 +419,17 @@ const (
...
+const (
+	//AzureAD Identity System
+	AzureAD = "azure_ad"
+	//ADFS Identity System
+	ADFS = "adfs"
+)
+
+const (
+	//ClientSecret Authentication method
+	ClientSecret = "client_secret"
```
jackfrancis: let's change this to `ClientSecretAuthMethod` and `ClientCertificateAuthMethod`, just to make the reference more explicit

pkg/api/vlabs/const.go
```diff
@@ -120,3 +120,17 @@ const (
...
+const (
+	//AzureAD Identity System
+	AzureAD = "azure_ad"
+	//ADFS Identity System
+	ADFS = "adfs"
+)
+
+const (
+	//ClientSecret Authentication method
+	ClientSecret = "client_secret"
```
jackfrancis: ditto above

pkg/api/types.go
```diff
@@ -1428,6 +1430,26 @@ func (cs *ContainerService) GetLocations() []string {
 	return allLocations
 }
 
+// GetAuthenticationMethod returns authentication method which k8s azure cloud provider will use
```
jackfrancis: Is this function ever used in a non-Azure Stack context? If not, it would make more sense to name it something like `GetCustomCloudAuthenticationMethod `.

parts/k8s/kubernetesconfigs.sh
```diff
@@ -183,6 +183,42 @@ configureK8s() {
     "providerKeyVersion": ""
 }
 EOF
+    if [[ "${TARGET_ENVIRONMENT,,}" == "azurestackcloud"  ]]; then 
+        echo "AUTHENTICATION_METHOD: ${AUTHENTICATION_METHOD}"
```
jackfrancis: Do we need these echo statements? There is an upper bound on the CSE payloads in an ARM template, so let's omit them if we can. If we want to display this in the provision log, let's enable `set -x` for any commands that are not private. Make sure we put `set +x` before any output that should not be printed to stdout and captured in logs.

honcao: @jackfrancis  Thanks for the instructions about set -x and set +x
I removed the echo command here since I could find the value in the ARM template

and set +x already is present at line 144 to protect the secrets. set -x is executed at line 220 to enable the logs.

PR #535, opened 2019-02-19
feat: enable configurable --enforce-node-allocatable

### LINKED ISSUE #507: Allow system-reserved and kube-reserved to be set
No
---

FEATURE REQUEST
---

0.31.0
---

**Kubernetes version**:
n/a

**What happened**:
Processes running on the agents and masters outside of kubernetes are being starved of resources.

I tried to specify the kubelet flags for system-reserved and kube-reserved but they took no effect because enforce-node-allocatable cannot be overridden to include them as per the docs at enforce-node-allocatable.

The only related flag that can be configured is the eviction-hard flag.

I'd expect to aks-engine to set sensible defaults for system-reserved and kube-reserved for the reasons described at https://github.com/kubernetes/kops/blob/09808ecadae35c7b78e0dbe0a48a8a597938cd78/docs/node_resource_handling.md#so-what-could-possibly-go-wrong or at least make it possible for them to be applied.

jackfrancis:
**Reason for Change**:

Enable configurable `--enforce-node-allocatable` kubelet flag to unblock scenarios such as described in #507.

**Issue Fixed**:

Fixes #507 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@mboersma I think the open-endedness of it makes sense from a maintenance perspective.

PR #538, opened 2019-02-20
fix: move the definition of inboundNatRules inside the loadbalancer specification

tariq1890:
Using a separate definition of inboundNatRules makes it less testable as it is a SubResource. This PR deletes the SubResource definition and adds it to the LoadBalancer Resource definition instead.


jackfrancis:
Is this change backwards-compatible w/ upgrade?

tariq1890:
I don't think this change would pose risks in that surface area. We are
just changing the way of defining the same resources as before.

tariq1890:
@jackfrancis I have tested upgrading a cluster (that was created from master) with this branch and it has worked fine.

jackfrancis:
lgtm

ritazh:
lgtm

PR #546, opened 2019-02-20
revert: commit generated code files

mboersma:
**Reason for Change**:

The change in #449 really didn't work out--there has been constant contention over templates_generated.go which requires every PR to be rebased.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Let's revisit the idea of aks-engine being purely `go get`-table soon. For now we don't want the confusion + churn that comes with committing the generated files.


mboersma:
I'm going to update this to leave the `make ensure-generated` target intact, but not required in CI.

PR #548, opened 2019-02-20
Update windowskubeletfunc.ps1

DerekMeek:
Fix typo for kubeletwin/pause instead of kubletwin/pause

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/windowskubeletfunc.ps1
+++ b/parts/k8s/windowskubeletfunc.ps1
@@ -153,7 +153,7 @@ New-InfraContainer {
 
     "FROM $($windowsBase)" | Out-File -encoding ascii -FilePath Dockerfile
     "CMD cmd /c ping -t localhost" | Out-File -encoding ascii -FilePath Dockerfile -Append
-    docker build -t kubletwin/pause .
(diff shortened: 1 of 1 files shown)


mboersma:
@DerekMeek this is the same docker image referred to [here](https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults-kubelet.go#L39) by the mistyped name. So changing it only in the PowerShell script will break things.

(If you do a `git grep kublet` you'll see a few other places where this typo exists.

mboersma:
@DerekMeek could you `git commit --amend` at least the first commit so it starts with a category and a colon? If it's not a [semantic commit](https://www.conventionalcommits.org/en/v1.0.0-beta.3/) it won't show up in the CHANGELOG. Something like:

```
chore: fix "kubletwin/pause" typo
```

DerekMeek:
I’ll do that when I get the time to sit down and check if it’s really a typo like I thought.  My setup required “kubeletwin/pause” but whatever file referenced that may have been the typo and all the others might be correct

PR #551, opened 2019-02-20
chore: disable ip-masq-agent when using cilium networkPlugin

jackfrancis:
**Reason for Change**:

`cilium` network plugin does its own masquerading, so `ip-masq-agent` should never be used.

Also commiting several const conveniences so we don't ever accidentally mistype the "cilium" val.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1325,7 +1325,7 @@ func (k *KubernetesConfig) IsDashboardEnabled() bool {
 
 // IsIPMasqAgentEnabled checks if the ip-masq-agent addon is enabled
 func (k *KubernetesConfig) IsIPMasqAgentEnabled() bool {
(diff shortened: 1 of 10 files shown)


pkg/api/types.go
```diff
@@ -1325,7 +1325,7 @@ func (k *KubernetesConfig) IsDashboardEnabled() bool {
 
 // IsIPMasqAgentEnabled checks if the ip-masq-agent addon is enabled
 func (k *KubernetesConfig) IsIPMasqAgentEnabled() bool {
-	return k.isAddonEnabled(IPMASQAgentAddonName, IPMasqAgentAddonEnabled)
+	return k.isAddonEnabled(IPMASQAgentAddonName, (k.NetworkPlugin != NetworkPluginCilium && IPMasqAgentAddonEnabled))
```
jackfrancis: This is the functional change that will no longer enable `ip-masq-agent` by default if `"networkPlugin"` is `"cilium"`

PR #554, opened 2019-02-21
chore: add administration tools

sylr:
- htop
- iotop
- iftop
- sysstat


jackfrancis:
@sylr FYI I rebased/force-pushed this PR to reconcile it with some changes we're managing to go generated commit requirements in master (sorry for the moving furniture).

jackfrancis:
@weinong @jakepearson @cchildress @juan-lee Anyone have any opinions about introducing the OS tools noted in the PR description to AKS node vms?

cchildress:
These are definitely very helpful debugging tools. I use htop all the time. I find atop and iftop to be quite helpful as well.

If the goal here is to make our nodes easier to debug out of the box I'd say this is a solid step in that direction.

sylr:
@cchildress +1 for iftop, I'll add it

PR #555, opened 2019-02-21
chore: add priorityClassName: system-node-critical to kube-system,…

### LINKED ISSUE #573: Guaranteed Scheduling For Critical Pods
### **Problem Summary:**
Users expect their critical system pods to always be scheduled as needed regardless of any actions they take, like tainting nodes for workload specific scheduling. 
Today because AKS and AKS-engine system pods are controlled by the addon manager and the default pod spec enforced by it does not use any tolerations or priority this leads to the pods being evicted after for example a taint and/or not be scheduled in the tainted nodes anymore.

### **Proposed Solution**
Add PriorityClasses and wildcard tolerations to the critical system pods

jackfrancis:
… master-schedulable components

**Reason for Change**:

For `kube-system` components that are declared to tolerate master node taints, we want to add `priorityClassName: system-node-critical` as well to doubly ensure that they are schedulable, even if the master node taints are customized. Otherwise, master node taint customization is liable to make system-critical components unschedulable.

Included all components that fit the above description for k8s 1.9 and above.

Reference: https://kubernetes.io/docs/tasks/administer-cluster/guaranteed-scheduling-critical-addon-pods/

**Issue Fixed**:

Fixes #573

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
@@ -68,7 +68,7 @@ spec:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
-      priorityClassName: system-cluster-critical
+      priorityClassName: system-node-critical
       tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
--- a/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
@@ -68,8 +68,8 @@ spec:
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
-      tolerations:
       priorityClassName: system-cluster-critical
+      tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
       nodeSelector:
--- a/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
@@ -69,6 +69,7 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       tolerations:
+      priorityClassName: system-cluster-critical
       - key: CriticalAddonsOnly
         operator: Exists
       nodeSelector:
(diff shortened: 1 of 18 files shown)


mboersma:
lgtm

parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
```diff
@@ -69,6 +69,7 @@ spec:
         scheduler.alpha.kubernetes.io/critical-pod: ''
     spec:
       tolerations:
+      priorityClassName: system-cluster-critical
```
jackfrancis: whoops this is a bug!

jackfrancis: fixed

jackfrancis:
omsagent cluster scenario manually tested

jackfrancis:
flannel scenario manually tested

palma21:
Add-on for the reasons:

Adding priorityClass to ensure that kube-system components are schedulable when node taints are customized and added to _any_ node. 

Also before we merge I see

jackfrancis:
@palma21 does the most recent commit address the desired add'l kube-proxy tolerations?

parts/k8s/containeraddons/azure-cni-networkmonitor.yaml
```diff
@@ -25,6 +25,10 @@ spec:
         operator: Equal
         value: "true"
         effect: NoSchedule
+      - operator: "Exists"
+	      effect: NoExecute
+	    - operator: "Exists"
+	      effect: NoSchedule
       nodeSelector:
```
palma21: And this one. Last I promise

parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
```diff
@@ -20,6 +20,10 @@ spec:
         operator: Equal
         value: "true"
         effect: NoSchedule
+      - operator: "Exists"
+        effect: NoExecute
+      - operator: "Exists"
+        effect: NoSchedule
       containers:
```
serbrech: This is to ensure that we never schedule kube-proxy on master, correct?

jackfrancis: It's the opposite: it ensures that kube-proxy will always be able to be scheduled on *any* node, regardless of taint.

PR #556, opened 2019-02-21
chore: ignore artifacts that may be present from prior commits

jackfrancis:
**Reason for Change**:

Re-adding to `.gitignore` previously ignored file references, to unblock back-compat stories.

E.g.:

- checkout an old commit and compile/generate
- checkout current commit and compile/generate
  - this is broken because compile+link detects that there are multiple references, from the previous build generation

See this thread:

https://github.com/Azure/aks-engine/commit/25915938f2c3bccd7fc4dcd4c9185641cbc935cc

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,8 @@ pkg/operations/junit.xml
 pkg/operations/kubernetesupgrade/junit.xml
 pkg/engine/templates_generated.go
 pkg/i18n/translations_generated.go
+pkg/engine/templates.go
+pkg/i18n/translations.go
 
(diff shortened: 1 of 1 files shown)


mboersma:
I understand the problem, but I think this is an isolated, short-term problem resulting from the name change of those two files. I guess adding them back to .gitignore is safe though, so :+1:.

mboersma:
lgtm

PR #558, opened 2019-02-21
test: use "k" script to fetch the right kubectl just-in-time

### LINKED ISSUE #557: update kubectl for 1.14 E2E

mboersma:
**Reason for Change**:

Uses the [`k`](https://github.com/jakepearson/k) script in the [go-dev](https://github.com/deis/docker-go-dev) image to fetch the right version of `kubectl` just-in-time for E2E tests.

**Issue Fixed**:

Fixes #557
Unblocks #435

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This is just one approach to fixing this recurrent problem: other ideas are welcome. It worked fine for me running E2E tests locally (but requires the `k` script to be in your path of course).

### DIFF
--- a/docs/community/developer-guide.md
+++ b/docs/community/developer-guide.md
@@ -194,6 +194,10 @@ environment variables to be set:
 * `SUBSCRIPTION_ID`: Azure subscription UUID
 * `TENANT_ID`: Azure tenant UUID
 
+The end-to-end tests also require the [`k`](https://github.com/jakepearson/k)
+script to be your the search $PATH. This ensures that testing uses a `kubectl`
+client that matches the version of the Kubernetes server.
+
 ### Debugging
 
 For aks-engine code debugging you can use [Delve](https://github.com/derekparker/delve) debugger.
--- a/docs/community/developer-guide.md
+++ b/docs/community/developer-guide.md
@@ -195,7 +195,7 @@ environment variables to be set:
 * `TENANT_ID`: Azure tenant UUID
 
 The end-to-end tests also require the [`k`](https://github.com/jakepearson/k)
-script to be your the search $PATH. This ensures that testing uses a `kubectl`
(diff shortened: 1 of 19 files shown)


jackfrancis:
rock + roll

jackfrancis:
@ritazh does upstream E2E use any of this code? This change would require the execution context of the E2E runner to have a working `k` binary in the path (the `quay.io/deis/go-dev:v1.18.4` docker image that our tooling uses includes this: https://github.com/jakepearson/k

mboersma:
This passed, but the true test is whether #435 will pass after this change. I'll cook up a throwaway branch with both changes and see if Kubernetes 1.14.0-alpha.2 will pass.

jackfrancis:
@mboersma 10-4

also /hold until we get confirmation from @ritazh that this change won't break upstream CI

ritazh:
@jackfrancis upstream e2e does not use this. Is this to fix that issue you saw with the schema change that caused kubectl create to fail when executed from a remote machine?

jackfrancis:
It's to fix older (v1.10?) `kubectl` client commands failing against the 1.14.0-alpha.2 API. Generally, it's to avoid situations like this to use an appropriate client-API implementation going forward.

Thanks for confirming!

mboersma:
What Jack said--we've gotten away with being sloppy and using `kubectl` CLI 1.10.12 for all tests, but that finally broke with Kubernetes 1.14.0-alpha.2.

mboersma:
The 1.14.0-alpha.2 [test run passed](https://msazure.visualstudio.com/One/_build/results?buildId=20328084).

jackfrancis:
ship it

mboersma:
I should document `k` as a new requirement for developers who want to run the e2e tests. Let me add that.

PR #563, opened 2019-02-22
test: enable timeouts for running E2E commands

jackfrancis:
**Reason for Change**:

Enable setting an upper bound time limit on E2E test commands. The specific use-case is ensuring that simple pod-to-pod networking operations are consistently quick and don't fall victim to, for example, occasional > 5 second DNS lookup operations.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/deployment/deployment.go
+++ b/test/e2e/kubernetes/deployment/deployment.go
@@ -17,6 +17,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const commandTimeout = 1 * time.Minute
+
 // List holds a list of deployments returned from kubectl get deploy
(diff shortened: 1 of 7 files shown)


CecileRobertMichon:
what about all the hardcoded 60 second timeouts in cli.go?

jackfrancis:
@CecileRobertMichon that's a different timeout enforcement which is ensuring "fail fast" when something goes wonky with pre-requisite `az` commands.

The purpose of this changeset is to enable timeouts for `kubectl run`-type commands; to be able to detect, for example, regular 5 second name lookups in the pod networking stack.

PR #564, opened 2019-02-22
chore: apiserverCertificate should be apiServerCertificate

jackfrancis:
The apiserver cert param is referred to everywhere as “apiServerCertificate” (note the capital “S”), except in the variables definition, where it is passed into the “params()” func as “apiserverCertificate” (note the lowercase “s”).

Ostensibly the ARM “params” func is case-insensitive, but this is not something we should rely upon.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


serbrech:
lgtm if e2e passes

PR #565, opened 2019-02-22
fix: allow for gzip and b64 encoding of the raw Addon data by decoding the input beforehand

### LINKED ISSUE #488: custom data in addons is not parsed (gunzip+base64 decoded) properly
**Is this a request for help?**:  Help, in a way, yes.

---

---

**What version of aks-engine?**: 0.28.1

---

**Kubernetes version**: any

**What happened**: Supply an addon manifest through aks-engine's `data` apimodel field for addon config in [gzip+base64 OR base64] format. My scenario is to provide a specific manifest for ip-masq-agent, but it probably applies to any addon of this type. The content on disk of the resultant addons config file is the same as the input format (no [gunzip+base64 OR base64] decoding has happened).

**What you expected to happen**: Content to be decoded appropriately and placed as a valid yaml manifest in the appropriate file in `/etc/kubernetes/addons/`

**Anything else we need to know**: Documentation in the section: https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#addons points users to both: "gzip+base64 encode" as well as "base64" encode their addon manifests when putting into the aks-engine api model. Neither is correct at the current time.

tariq1890:
Fixes #488

### DIFF
--- a/pkg/engine/engine.go
+++ b/pkg/engine/engine.go
@@ -646,6 +646,11 @@ func getBase64CustomScript(csFilename string) string {
 	return getBase64CustomScriptFromStr(csStr)
 }
 
+func getStringFromBase64(str string) (string, error) {
+	decodedBytes, err := base64.StdEncoding.DecodeString(str)
+	return string(decodedBytes), err
+}
+
 // getBase64CustomScript will return a base64 of the CSE
 func getBase64CustomScriptFromStr(str string) string {
 	var gzipB bytes.Buffer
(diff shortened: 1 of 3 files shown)


pkg/engine/engine.go
```diff
@@ -719,7 +724,11 @@ func getContainerAddonsString(properties *api.Properties, sourcePath string) str
 		if setting.isEnabled {
 			var input string
 			if setting.rawScript != "" {
-				input = setting.rawScript
+				var err error
+				input, err = getStringFromBase64(setting.rawScript)
```
jackfrancis: How are we dealing with gzip'd input?

tariq1890: We are not. This PR is implemented as per what we've mentioned in the documentation. It's stated that the user must provide the B64 encoded string of the addon YAML

jackfrancis: The documentation states gzip+base64, and in that order:

https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#addons

(also the title of this PR suggests gzip)

tariq1890: Oh yes. There seems to be a discrepancy.

PR #567, opened 2019-02-22
Revert "test: use "k" script to fetch the right kubectl just-in-time"

jackfrancis:
Reverts Azure/aks-engine#558


mboersma:
I think #571 will fix this instead.

PR #574, opened 2019-02-23
fix: Explicitly use overlay2 storage driver while creation of vhd, by defa…

awesomenix:
…ult docker uses aufs

**Reason for Change**:

By default docker uses aufs storage driver, hence while creation of vhd, the container images are always downloaded to wrong partition /var/lib/docker/aufs while we start docker with overlay2 storage driver after creation of vm. Configure docker engine to use overlay2 after installation during vhd creation.

**Issue Fixed**:

**Requirements**:

- [] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -15,6 +15,7 @@ installDeps
 if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
     DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
     installDockerEngine
+    overrideDockerEngineStorageDriver
     installGPUDrivers
 else
     MOBY_VERSION="3.0.4"
--- a/parts/k8s/kubernetesinstalls.sh
+++ b/parts/k8s/kubernetesinstalls.sh
@@ -128,6 +128,27 @@ installDockerEngine() {
     fi
 }
 
+overrideDockerEngineStorageDriver() {
+    echo "stopping docker engine"
+    sudo systemctl stop docker
+    cat << EOF > /tmp/daemon.json 
+{
+    "storage-driver": "overlay2",
+    "live-restore": true,
+    "log-driver": "json-file",
+    "log-opts":  {
+        "max-size": "50m",
+        "max-file": "5"
+    }
+}
+EOF
+    sudo mv /tmp/daemon.json /etc/docker/daemon.json
+    echo "cleaning up aufs storage"
+    sudo rm -rf /var/lib/docker/aufs
+    echo "starting docker engine with overlay2 driver"
+    sudo systemctl start docker
+}
+
 installKataContainersRuntime() {
     # TODO incorporate this into packer CI so that it is pre-baked into the VHD image
     echo "Adding Kata Containers repository key..."


parts/k8s/kubernetesinstalls.sh
```diff
@@ -129,6 +129,27 @@ installDockerEngine() {
...
+    "log-driver": "json-file",
+    "log-opts":  {
+        "max-size": "50m",
+        "max-file": "5"
+    }
+}
+EOF
+    sudo mv /tmp/daemon.json /etc/docker/daemon.json
+    echo "cleaning up aufs storage"
+    sudo rm -rf /var/lib/docker/aufs
```
jackfrancis: I'm curious: in the "docker-engine" VHD CI build scenario, moby (or no other docker) is installed *except* docker-engine. How does `/var/lib/docker/aufs` get there? Is it installed by default by Ubuntu 16.04-LTS (that would seem unlikely).

awesomenix: I think the docker is installed as part of this step installDockerEngine (in packer/install-dependencies.sh), which is used for pulling down the images.

packer/install-dependencies.sh
```diff
@@ -18,6 +18,7 @@ installDeps
 if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
     DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
     installDockerEngine
+    configureDockerEngine
```
CecileRobertMichon: I believe we also need to call this in customscript for clusters not using VHD (eg. gov cloud)

parts/k8s/kubernetesinstalls.sh
```diff
@@ -129,6 +129,27 @@ installDockerEngine() {
     fi
 }
 
+configureDockerEngine() {
```
CecileRobertMichon: I would move this to kubernetesconfigs. All runtime configurations should be in https://github.com/Azure/aks-engine/blob/1bfacd8f278143989e938884b2af560203457a79/parts/k8s/kubernetesconfigs.sh as opposed to kubernetesinstalls since this doesn't involve internet outbound

packer/install-dependencies.sh
awesomenix: But we already configure docker to use overlay2 in case of CSE scenarios, this change is during VHD builds only

CecileRobertMichon: where do we do "we already configure docker to use overlay2 in case of CSE scenarios"?

awesomenix: https://github.com/Azure/aks-engine/blob/124379b05ed2f2183aeb5eec38b2392d97dc860c/parts/k8s/kubernetesagentcustomdata.yml#L119

CecileRobertMichon:
The commit message needs to be changed to use the "fix:" semantics (squash commit will use the first PR commit, not the title)

parts/k8s/kubernetesinstalls.sh
awesomenix: Tried to keep install and config for docker in same file, easier to search, but understand that its better off in kubernetesconfig.sh to keep in sync with code structure.

parts/k8s/kubernetesinstalls.sh
CecileRobertMichon: We use the same function to install docker-engine in non-VHD scenarios (CSE calls installContainerRuntime which calls installDockerEngine) so how is it different for VHDs specifically?

packer/install-dependencies.sh
CecileRobertMichon: custom data applies to both VHD and non-VHD scenarios. Is the issue here that we pre-pull images in the VHD script before the storage driver is set to overlay2?

awesomenix: Yes thats correct.

parts/k8s/kubernetesinstalls.sh
awesomenix: i see that packer/install-dependencies only include the below files


source /home/packer/provision_installs.sh
source /home/packer/provision_source.sh

Doesnt seem to include kubernetesconfigs.sh

```
 {
            "type": "file",
            "source": "packer/cleanup-vhd.sh",
            "destination": "/home/packer/cleanup-vhd.sh"
        },
        {
            "type": "file",
            "source": "parts/k8s/kubernetesinstalls.sh",
            "destination": "/home/packer/provision_installs.sh"
        },
        {
            "type": "file",
            "source": "parts/k8s/kubernetesprovisionsource.sh",
            "destination": "/home/packer/provision_source.sh"
        },
        {
            "type": "file",
            "source": "packer/install-dependencies.sh",
            "destination": "/home/packer/install-dependencies.sh"
        },
```

CecileRobertMichon:
fixed the pipeline and queued a new build

CecileRobertMichon:
VHD build is failing with `2019-02-26T01:22:49.3024544Z �[0;32m    azure-arm: /home/packer/provision_installs.sh: line 9: CONTAINERD_VERSION: unbound variable�[0m`

Looks like a regression from https://github.com/Azure/aks-engine/pull/516 @jackfrancis FYI

Let's not merge this until it's fixed

/hold

jackfrancis:
Good catch @CecileRobertMichon, I think this is the fix: https://github.com/Azure/aks-engine/pull/583

jackfrancis:
FYI rebased/force-pushed this on top of VHD CI-fixing commit in master.

PR #578, opened 2019-02-25
chore: update the compute apiversion to 2018-10-01 to support new vmss features

tariq1890:
We need to use the more recent versions of API compute if we want to be up to date and use the new VMSS features like "VMSS overprovisioning + do not run extensions on overprovisioned instances".


jackfrancis:
Just a slight correction: the new API version delivers "VMSS overprovisioning + do not run extensions on overprovisioned instances".

jackfrancis:
Is there no change to `Gopkg.toml`?

tariq1890:
No there won't be any change, as we already use a new Azure Go SDK library
which has this import version package.

jackfrancis:
Why the change to `Gopkg.lock`?

tariq1890:
The Gopkg locks are changed because we used a new import. The Gopkg lock is a reflection of the imports and the versions of the libraries used in the project.

jackfrancis:
Got it. code lgtm. We'll want to run lots of back-compat tests to validate we can run upgrade against a cluster created with the prior api version spec on the same resources.

tariq1890:
@jackfrancis Sure. Sounds good.

tariq1890:
Upgrade an aks-engine cluster created using `master` worked from this branch :)

serbrech:
I'd say lgtm as well

PR #580, opened 2019-02-26
fix: eviction-hard is not implemented in windows

jsturtevant:
**Reason for Change**:

This removes the configuration for `--eviction-hard` for windows.  Although the values are subtracted from `NodeAllocatable` there is [no enforcement for `--eviction-hard` in windows currently](https://github.com/kubernetes/kubernetes/issues/73417#issuecomment-465806482).

The default linux configuration was being passed to Windows which was causing the error message "eviction manager: no observation found for eviction signal nodefs.inodesFree" show up in error logs.

**Issue Fixed**:

https://github.com/kubernetes/kubernetes/issues/73792

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Could remove just the `inodesFree` from configuration to fix issue above but I thought it was clearer to remove settings that do not work for windows.

I didn't see any unit tests the test overloaded windows values.  Is there preferred way to add those?


PatrickLang:
The E2E failure seemed like a flake. I rescheduled it
/lgtm

PR #583, opened 2019-02-26
fix: CONTAINERD_VERSION unbound variable in VHD build script

jackfrancis:
**Reason for Change**:

The `$CONTAINERD_DOWNLOAD_URL` was present as a global variable in the base k8s CSE install script, and depended upon `$CONTAINERD_VERSION` to construct itself. This removes `$CONTAINERD_DOWNLOAD_URL` from the global variable namespace to allow the CSE script to once again be generally sourced by another script (without having to pass a `$CONTAINERD_VERSION` env var).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/kubernetesinstalls.sh
```diff
@@ -220,7 +220,7 @@ installContainerd() {
     if [[ "$CURRENT_VERSION" == "${CONTAINERD_VERSION}" ]]; then
         echo "containerd is already installed, skipping install"
     else
-        CONTAINERD_TGZ_TMP=$(echo ${CONTAINERD_DOWNLOAD_URL} | cut -d "/" -f 5)
+        CONTAINERD_TGZ_TMP="cri-containerd-${CONTAINERD_VERSION}.linux-amd64.tar.gz"
```
jackfrancis: This (1) removes the need for a 2nd concatenation of the `$CONTAINERD_DOWNLOAD_URL` var, and (2) is arguably more maintainable

parts/k8s/kubernetesinstalls.sh
```diff
@@ -187,6 +186,7 @@ downloadAzureCNI() {
 }
 
 downloadContainerd() {
+    CONTAINERD_DOWNLOAD_URL="${CONTAINERD_DOWNLOAD_URL_BASE}cri-containerd-${CONTAINERD_VERSION}.linux-amd64.tar.gz"
```
CecileRobertMichon: we're already setting `CONTAINERD_DOWNLOAD_URL` in the VHD script itself so this will override it

CecileRobertMichon:
If that helps I have a passing vhd build with https://github.com/Azure/aks-engine/pull/457/files

packer/install-dependencies.sh
```diff
@@ -40,7 +40,7 @@ for CNI_PLUGIN_VERSION in $CNI_PLUGIN_VERSIONS; do
     downloadCNI
 done
 
-for CONTAINERD_VERSION in $CONTAINERD_VERSIONS; do
+for CONTAINERD_VERSION in ${CONTAINERD_VERSIONS}; do
```
CecileRobertMichon: can we move `CONTAINERD_DOWNLOAD_URL_BASE` down here now that it's not used until `downloadContainerd` is called? It had to be at the top of the file before because `CONTAINERD_DOWNLOAD_URL` was set at the top of the sourced file

parts/k8s/kubernetesinstalls.sh
jackfrancis: no longer setting `CONTAINERD_DOWNLOAD_URL` in the VHD script

parts/k8s/kubernetesinstalls.sh
```diff
@@ -278,7 +278,7 @@ installKubeletAndKubectl() {
 pullContainerImage() {
     CLI_TOOL=$1
     DOCKER_IMAGE_URL=$2
-    if [ ! -z "${PRIVATE_AZURE_REGISTRY_SERVER}" ]; then
+    if [[ ! -z "${PRIVATE_AZURE_REGISTRY_SERVER:-}" ]]; then
```
jackfrancis: double `[[` and `]]` just because it's more idiomatic w/ the rest of the bash surface area

packer/install-dependencies.sh
jackfrancis: done

CecileRobertMichon: I'm being picky but the var assignment doesn't need to be inside the for loop since it's static

jackfrancis: done

PR #589, opened 2019-02-26
ci: add storage account cleanup to VHD pipeline

CecileRobertMichon:
**Reason for Change**:

Add cleanup to VHD pipeline to avoid running out of quota.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder.yaml
+++ b/.pipelines/vhd-builder.yaml
@@ -44,7 +44,7 @@ phases:
       -e VHD_NAME=${VHD_NAME} \
       ${DEIS_GO_DEV_IMAGE} make az-copy
     displayName: Copying resource to Classic Storage Account
-    condition: ! ${DRY_RUN}
+    condition: not(${DRY_RUN})
   - script: |
       SA_NAME="$(cat packer-output | grep "creating new storage account" | cut -d " " -f 5)" && \
       docker run --rm \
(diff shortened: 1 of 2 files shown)


.pipelines/vhd-builder.yaml
```diff
@@ -44,6 +44,19 @@ phases:
       -e VHD_NAME=${VHD_NAME} \
       ${DEIS_GO_DEV_IMAGE} make az-copy
     displayName: Copying resource to Classic Storage Account
+    condition: not(${DRY_RUN})
```
CecileRobertMichon: this ensures we only copy the disk to a classic storage account if `DRY_RUN` is set to `False` (`True` by default). That way, we won't clutter the classic SA for PR builds.

.pipelines/vhd-builder.yaml
```diff
@@ -44,6 +44,19 @@ phases:
       -e VHD_NAME=${VHD_NAME} \
       ${DEIS_GO_DEV_IMAGE} make az-copy
     displayName: Copying resource to Classic Storage Account
+    condition: not(variables['DRY_RUN'])
```
CecileRobertMichon: this ensures we only copy the disk to a classic storage account if `DRY_RUN` is set to `False` (`True` by default). That way, we won't clutter the classic SA for PR builds.

PR #590, opened 2019-02-26
test: add VMSS client interfaces and mocks to make upgrade testable

serbrech:
**Reason for Change**:

This is extracting the vmss interfaces and mocks from the `upgrade --force PR` to reduce its footprint.
One test is carried over, which I just marked as skipped here as it lacks the implementation to pass

### DIFF
--- a/pkg/armhelpers/compute.go
+++ b/pkg/armhelpers/compute.go
@@ -36,13 +36,15 @@ func (az *AzureClient) DeleteVirtualMachine(ctx context.Context, resourceGroup, 
 }
 
 // ListVirtualMachineScaleSets returns (the first page of) the vmss resources in the specified resource group.
-func (az *AzureClient) ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (compute.VirtualMachineScaleSetListResultPage, error) {
-	return az.virtualMachineScaleSetsClient.List(ctx, resourceGroup)
+func (az *AzureClient) ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (VirtualMachineScaleSetListResultPage, error) {
+	page, err := az.virtualMachineScaleSetsClient.List(ctx, resourceGroup)
+	return &page, err
 }
 
 // ListVirtualMachineScaleSetVMs returns the list of VMs per VMSS
-func (az *AzureClient) ListVirtualMachineScaleSetVMs(ctx context.Context, resourceGroup, virtualMachineScaleSet string) (compute.VirtualMachineScaleSetVMListResultPage, error) {
-	return az.virtualMachineScaleSetVMsClient.List(ctx, resourceGroup, virtualMachineScaleSet, "", "", "")
+func (az *AzureClient) ListVirtualMachineScaleSetVMs(ctx context.Context, resourceGroup, virtualMachineScaleSet string) (VirtualMachineScaleSetVMListResultPage, error) {
+	page, err := az.virtualMachineScaleSetVMsClient.List(ctx, resourceGroup, virtualMachineScaleSet, "", "", "")
+	return &page, err
 }
 
 // DeleteVirtualMachineScaleSetVM deletes a VM in a VMSS
--- a/pkg/armhelpers/interfaces.go
+++ b/pkg/armhelpers/interfaces.go
@@ -26,6 +26,24 @@ type VirtualMachineListResultPage interface {
 	Values() []compute.VirtualMachine
 }
 
+// VirtualMachineScaleSetListResultPage is an interface for compute.VirtualMachineScaleSetListResultPage to aid in mocking
+type VirtualMachineScaleSetListResultPage interface {
+	Next() error
+	NextWithContext(ctx context.Context) (err error)
+	NotDone() bool
+	Response() compute.VirtualMachineScaleSetListResult
+	Values() []compute.VirtualMachineScaleSet
+}
+
+// VirtualMachineScaleSetVMListResultPage is an interface for compute.VirtualMachineScaleSetListResultPage to aid in mocking
+type VirtualMachineScaleSetVMListResultPage interface {
+	Next() error
+	NextWithContext(ctx context.Context) (err error)
+	NotDone() bool
+	Response() compute.VirtualMachineScaleSetVMListResult
+	Values() []compute.VirtualMachineScaleSetVM
+}
+
 // DeploymentOperationsListResultPage is an interface for resources.DeploymentOperationsListResultPage to aid in mocking
 type DeploymentOperationsListResultPage interface {
 	Next() error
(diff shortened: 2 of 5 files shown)


pkg/armhelpers/compute.go
```diff
@@ -36,13 +36,15 @@ func (az *AzureClient) DeleteVirtualMachine(ctx context.Context, resourceGroup,
 }
 
 // ListVirtualMachineScaleSets returns (the first page of) the vmss resources in the specified resource group.
-func (az *AzureClient) ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (compute.VirtualMachineScaleSetListResultPage, error) {
-	return az.virtualMachineScaleSetsClient.List(ctx, resourceGroup)
+func (az *AzureClient) ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (VirtualMachineScaleSetListResultPage, error) {
```
jackfrancis: FYI this will require a change to AKS implementation, as it consumes this client+method.

pkg/armhelpers/compute.go
```diff
@@ -36,13 +36,15 @@ func (az *AzureClient) DeleteVirtualMachine(ctx context.Context, resourceGroup,
...
+	page, err := az.virtualMachineScaleSetsClient.List(ctx, resourceGroup)
+	return &page, err
 }
 
 // ListVirtualMachineScaleSetVMs returns the list of VMs per VMSS
-func (az *AzureClient) ListVirtualMachineScaleSetVMs(ctx context.Context, resourceGroup, virtualMachineScaleSet string) (compute.VirtualMachineScaleSetVMListResultPage, error) {
-	return az.virtualMachineScaleSetVMsClient.List(ctx, resourceGroup, virtualMachineScaleSet, "", "", "")
+func (az *AzureClient) ListVirtualMachineScaleSetVMs(ctx context.Context, resourceGroup, virtualMachineScaleSet string) (VirtualMachineScaleSetVMListResultPage, error) {
```
jackfrancis: ditto

pkg/armhelpers/interfaces.go
```diff
@@ -74,10 +92,10 @@ type AKSEngineClient interface {
 	DeleteVirtualMachine(ctx context.Context, resourceGroup, name string) error
 
 	// ListVirtualMachineScaleSets lists the vmss resources in the resource group
-	ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (compute.VirtualMachineScaleSetListResultPage, error)
+	ListVirtualMachineScaleSets(ctx context.Context, resourceGroup string) (VirtualMachineScaleSetListResultPage, error)
```
jackfrancis: ditto

pkg/armhelpers/compute.go
serbrech: It shouldn't cause a problem as the interface is 1-1 with the compute type, and all other methods currently depend on these interfaces. If we're lucky, the import will already be present.
Otherwise, good reminder to always depend on our interface types and never on the compute types...

pkg/armhelpers/interfaces.go
```diff
@@ -74,10 +92,10 @@ type AKSEngineClient interface {
 	DeleteVirtualMachine(ctx context.Context, resourceGroup, name string) error
 
 	// ListVirtualMachineScaleSets lists the vmss resources in the resource group
```
CecileRobertMichon: let's standardize all the references to vmss/VMSS/Vmss to `VMSS` in the comments/code/tests

pkg/armhelpers/mockclients.go
```diff
@@ -412,33 +548,10 @@ func (mc *MockAKSEngineClient) ListVirtualMachines(ctx context.Context, resource
...
 }
 
-//GetVirtualMachine mock
-func (mc *MockAKSEngineClient) GetVirtualMachine(ctx context.Context, resourceGroup, name string) (compute.VirtualMachine, error) {
-	if mc.FailGetVirtualMachine {
-		return compute.VirtualMachine{}, errors.New("GetVirtualMachine failed")
-	}
-
+//MakeFakeVirtualMachine returns a fake compute.VirtualMachine
+func (mc *MockAKSEngineClient) MakeFakeVirtualMachine() compute.VirtualMachine {
```
CecileRobertMichon: nit: since this is already a mock function, using "fake" in the name is redundant. Also slight preference for mock over fake in comment wording.

CecileRobertMichon:
s/Vmss/VMSS in the title/commit message for release notes

CecileRobertMichon:
lgtm besides some minor comments on wording/formatting

jackfrancis:
back-compat tests passed

pkg/armhelpers/mockclients.go
serbrech: as I already commented before, the mockClient stubs functions like `CreateVirtualMachine` that are real function on the service. This is a helper builder function that return a default instance of a VirtualMachine. the `Fake` part of the name is to distinguish it from the functions exposed by the actual service that we mock

pkg/armhelpers/interfaces.go
serbrech: There are a lot of places across the code-base using different casing, so let's do this in a different PR

pkg/armhelpers/mockclients.go
jackfrancis: So you're saying existing `MockAKSEngineClient` methods (e.g., `ListVirtualMachineScaleSets`) represent mock equivalents of identically names methods in the real client libraries?

If that's the case, I'd argue we don't actually want to be adding these "fake" methods onto pre-existing mock libraries that fit that description, and better to create static functions that access the mock objects in order to return an appropriate fake, single-use object for testing.

I don't want to block progress, but something to think about as the mock + fake semantics are arguably a bit confusing.

serbrech:
> s/Vmss/VMSS in the title/commit message for release notes
squashed the 2 commits and adjusted the comment

pkg/armhelpers/interfaces.go
jackfrancis: I don't see why we wouldn't do that in this PR.

CecileRobertMichon: to clarify, I meant the references added by this PR, not all the references in the codebase :)

serbrech: I meant I will send a separate PR that fixes all of them across the code base.
but if you insist on doing this one here, it's coming

CecileRobertMichon: fixing them across the codebase would be 💯if you want to do that but not adding new inconsistencies in this PR is the low hanging fruit and was the aim of my comment.

serbrech: done

pkg/armhelpers/mockclients.go
serbrech: yes, I extracted this for reusability as it was inlined in the code and duplicated. I first wanted it standalone, but it relied on other parameters from the mockClient and so, I decided to keep it there for simplicity. it is used by the client to generate the faked results, and it can be used from the tests to build dummy objects.

PR #599, opened 2019-02-27
feat: Azure #402 - Disable windows auto update

### LINKED ISSUE #402: Disable windows auto update
ref: https://kubernetes.slack.com/archives/C0SJ4AFB7/p1548872012173700

Windows VMs are currently set to auto update. Come patch Tuesday an entire FD/UD/Zone can go down as machines updates themselves. Windows autoupdates should be turned off (or use external tools to manage windows servers).

@PatrickLang @craiglpeters

brunhil:
This pull request should fix #402.

**Reason for Change**:
Add functionality to disable automatic windows update

**Issue Fixed**:
Fixes #402 - Disable windows auto update

**Requirements**:
Add ability to disable automatic Windows updates. An entry for _windowsConfiguration_ will now be added under _osProfile_. This will default to true if the user does not provide a value. 

- [X] includes documentation
- [X] adds unit tests

**Notes**: 
This is one of a few approaches I thought of when adding this functionality. Interested to hear if this is acceptable or not. I went with implementation of always adding the entry for  _windowsConfiguration_  under _osProfile_. I thought that explicitly setting this value would be better for consumers.

I wasn't sure if the changes should be put under _api_ or _vlabs_.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -197,16 +197,17 @@ type CustomNodesDNS struct {
 
 // WindowsProfile represents the windows parameters passed to the cluster
 type WindowsProfile struct {
-	AdminUsername         string            `json:"adminUsername"`
-	AdminPassword         string            `json:"adminPassword" conform:"redact"`
-	ImageVersion          string            `json:"imageVersion"`
-	WindowsImageSourceURL string            `json:"windowsImageSourceURL"`
-	WindowsPublisher      string            `json:"windowsPublisher"`
-	WindowsOffer          string            `json:"windowsOffer"`
-	WindowsSku            string            `json:"windowsSku"`
-	WindowsDockerVersion  string            `json:"windowsDockerVersion"`
-	Secrets               []KeyVaultSecrets `json:"secrets,omitempty"`
-	SSHEnabled            bool              `json:"sshEnabled,omitempty"`
+	AdminUsername          string            `json:"adminUsername"`
+	AdminPassword          string            `json:"adminPassword" conform:"redact"`
+	ImageVersion           string            `json:"imageVersion"`
+	WindowsImageSourceURL  string            `json:"windowsImageSourceURL"`
+	WindowsPublisher       string            `json:"windowsPublisher"`
+	WindowsOffer           string            `json:"windowsOffer"`
+	WindowsSku             string            `json:"windowsSku"`
+	WindowsDockerVersion   string            `json:"windowsDockerVersion"`
+	Secrets                []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled             bool              `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates string            `json:"enableAutomaticUpdates,omitempty"`
 }
 
 // ProvisioningState represents the current state of container service resource.
@@ -1158,6 +1159,14 @@ func (w *WindowsProfile) GetWindowsSku() string {
 		return w.WindowsSku
 	}
 	return KubernetesDefaultWindowsSku
+}
+
+// GetEnableWindowsUpdate gets the flag for enable windows update or returns the default value
+func (w *WindowsProfile) GetEnableWindowsUpdate() string {
+	if w.EnableAutomaticUpdates != "" {
+		return w.EnableAutomaticUpdates
+	}
+	return "true"
 }
 
 // HasSecrets returns true if the customer specified secrets to install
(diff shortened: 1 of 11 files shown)


pkg/api/types.go
```diff
@@ -197,16 +197,17 @@ type CustomNodesDNS struct {
...
+	AdminPassword          string            `json:"adminPassword" conform:"redact"`
+	ImageVersion           string            `json:"imageVersion"`
+	WindowsImageSourceURL  string            `json:"windowsImageSourceURL"`
+	WindowsPublisher       string            `json:"windowsPublisher"`
+	WindowsOffer           string            `json:"windowsOffer"`
+	WindowsSku             string            `json:"windowsSku"`
+	WindowsDockerVersion   string            `json:"windowsDockerVersion"`
+	Secrets                []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled             bool              `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates string            `json:"enableAutomaticUpdates,omitempty"`
```
jackfrancis: Let's make this a `*bool` instead.

jackfrancis:
Thanks @brunhil! This mostly looks good. I'd like the property itself to be a `*bool` and a default of false. We can test this and eventually default to true, but the normal pattern for introducing a new feature is to default it to false (i.e., if the feature property isn't explicitly set in the api model we assume a "feature is disabled" cluster configuration)

pkg/api/vlabs/types.go
```diff
@@ -156,16 +156,17 @@ type CustomNodesDNS struct {
...
+	AdminPassword          string            `json:"adminPassword,omitempty"`
+	ImageVersion           string            `json:"imageVersion,omitempty"`
+	WindowsImageSourceURL  string            `json:"WindowsImageSourceUrl"`
+	WindowsPublisher       string            `json:"WindowsPublisher"`
+	WindowsOffer           string            `json:"WindowsOffer"`
+	WindowsSku             string            `json:"WindowsSku"`
+	WindowsDockerVersion   string            `json:"windowsDockerVersion"`
+	Secrets                []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled             bool              `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates string            `json:"enableAutomaticUpdates,omitempty"`
```
CecileRobertMichon: You'll want to add this new field to:
https://github.com/Azure/aks-engine/blob/master/pkg/api/convertertoapi.go#L523 and
https://github.com/Azure/aks-engine/blob/master/pkg/api/converterfromapi.go#L598

CecileRobertMichon: my bad I just saw you already did, resolving my comment

brunhil:
> This mostly looks good. I'd like the property itself to be a `*bool` and a default of false

@jackfrancis, I just want to make sure I'm understanding correctly. Are we talking about defaulting the value of `EnableAutomaticUpdates` or the output value in `osProfile\windowsConfiguration\enableAutomaticUpdates`?

CecileRobertMichon:
The way I understand @brunhil 's description of the issue, the default right now (not specifying `enableAutomaticUpdates` is true) so this optional feature is to override the default and disable updates. Is that correct?

If that's the case, we should keep defaulting to `true` and make opting-into disabled updates optional (which is what the PR is currently doing).

+1 on making the apimodel field of type *bool.

jackfrancis:
@brunhil because those properties are functionally equivalent (one is at the front end, in the cluster configuration api model, and one is at the back end, in the ARM spec), the answer is that the value of one will match the other. So they should have the same default.

Also, what @CecileRobertMichon says is true: the default value for the new `EnableAutomaticUpdates` api model property should match the current effective value of `enableAutomaticUpdates`. So if that effective ARM-applied default is `true`, then yes, the default for the `EnableAutomaticUpdates` api model property should be set to true.

brunhil:
Correct, the default value is true when you don't provide a value for an ARM template. I'll keep the default the true and get the comments addressed.

pkg/api/types.go
```diff
@@ -1160,6 +1161,14 @@ func (w *WindowsProfile) GetWindowsSku() string {
 	return KubernetesDefaultWindowsSku
 }
 
+// GetEnableWindowsUpdate gets the flag for enable windows update or returns the default value
+func (w *WindowsProfile) GetEnableWindowsUpdate() bool {
+	if w.EnableAutomaticUpdates != nil {
+		return *w.EnableAutomaticUpdates
+	}
+	return true
```
jackfrancis: Let's put a new const under `pkg/api/const.go`, e.g., `DefaultEnableAutomaticUpdates`. Set the value of that const to `true`. And then have this convenience method return that const. (instead of using the `true` bool literal here)

PR #601, opened 2019-02-27
ci: DRY_RUN variable is a string not a bool

CecileRobertMichon:
**Reason for Change**:

Azure pipelines treats the variable DRY_RUN as a string (did not find a way to specify it as a bool), so using string comparison. Previously `not('False')` was evaluating as `False`.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon:
Need to validate 

/hold

CecileRobertMichon:
This is ready. The VHD pipeline failed because our SA delete requests are being throttled (I was doing some heavy manual clean-up in the background at the same time) but skipping/not skipping the classic SA copy works as expected.

/hold cancel

PR #603, opened 2019-02-27
[WIP] feat: diet dev toolbox image

mboersma:
**Reason for Change**:

The current `deis/go-dev` docker image is old and huge. This starts afresh with an Alpine-linux based image to meet aks-engine's dev needs.

```console
$ docker images
REPOSITORY             TAG                 IMAGE ID            CREATED             SIZE
mboersma/aks-toolbox   latest              23373fa7cd03        About an hour ago   702MB
deis/go-dev            latest              e8aed44f9cea        5 days ago          3.94GB
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
If we proceed this way, we should refine the image so it works for the AKS projects as well and host it in an Azure repo and in MCR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2
 defaults: &defaults
   working_directory: /go/src/github.com/Azure/aks-engine
   docker:
-    - image: quay.io/deis/go-dev:v1.18.5
+    - image: mboersma/aks-toolbox
   environment:
     GOPATH: /go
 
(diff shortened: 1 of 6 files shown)


mboersma:
I *think* an alpine-based image isn't going to work for this case. I see Azure DevOps mounts an "externals" volume and runs `node` from there as a default program, and then the container exits. There's not a good error message, but I'm guessing that's a `glibc` or other library not being found.

mboersma:
#606 got us halfway there, and using Alpine may be problematic. I'm closing this; we can revisit it later.

PR #604, opened 2019-02-27
chore: modify eviction to memory.available.750Mi

khenidak:
Over the next period I will be PRing some changes to limits and allocation to aks-engine covering

1. (this PR) - Eviction Hard Thresholds 
100Mi (original) is too little. Here is why
   1. kubelet (or rather cadvisor) calculation of free memory is `free` + `cached` by looking at `cat /sys/fs/cgroup/memory/memory.stat`. The problem with this the server can be under sever memory pressure but kubelet will not see that because it sees cached memory (memory that the kernel is unable release back for `free` pool. 

Here is a simple test 
1. Run two SSH to any aks-cluster nodes. 
2. on session #1 run `journalctl -fu kubelet` 
3. on session #2 run `apt update && apt install stress && stress --vm-bytes $(awk '/MemAvailable/{printf "%d\n", $2 * 0.99;}' < /proc/meminfo)k --vm-keep -m 1
` basically allocate as much memory as you can. 

What you will see is free memory is going `~ 100Mi` there will be cached memory. What you will also notice that kubelet logic is **not** kicking in. In order to see what kubelet sees run https://kubernetes.io/docs/tasks/administer-cluster/out-of-resource/memory-available.sh . By the time memory pressure reaches `100Mi` free it will be too late to evict anything. as a matter of fact kubelet will be in hanging state.

The following PRs will cover `reservation` and `oom_killer` handling

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -331,7 +331,7 @@ const (
 	// DefaultKubernetesNodeStatusUpdateFrequency is 10s, see --node-status-update-frequency at https://kubernetes.io/docs/admin/kubelet/
 	DefaultKubernetesNodeStatusUpdateFrequency = "10s"
 	// DefaultKubernetesHardEvictionThreshold is memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5%, see --eviction-hard at https://kubernetes.io/docs/admin/kubelet/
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
Let's make sure the squash commit message follows semantic rules to include it in the release notes

CecileRobertMichon:
title won't be enough here, since it's a single commit GitHub will use the commit message.

holding so we can admin merge and override the message

/hold

PR #607, opened 2019-02-28
fix: Fix the CNI lock file names

jsturtevant:
Fix CNI lock file names and kill the processes before
getting rid of the config files

**Reason for Change**:

On restart of Kubelet Azure CNI would leak a lock file. The kubelet start script had incorrect lock file names which is causing the lock files to be left behind when kubelet starts.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
cc: @ashvindeodhar @PatrickLang


PatrickLang:
/lgtm
/hold

@jsturtevant can you remove the hold after @ashvindeodhar reviews? I'm going to be offline most of the next 2 days

jsturtevant:
/hold cancel

/assign @tariq1890 

@tariq1890 could you review?

PR #615, opened 2019-02-28
chore: add support for k8s version 1.11.8

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#changelog-since-v1117

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests
- [x] Windows assets zip file uploaded to production blob store

**Notes**:


packer/install-dependencies.sh
```diff
@@ -95,7 +95,7 @@ for TILLER_VERSION in ${TILLER_VERSIONS}; do
     pullContainerImage "docker" "gcr.io/kubernetes-helm/tiller:v${TILLER_VERSION}"
 done
 
-CLUSTER_AUTOSCALER_VERSIONS="1.13.1 1.12.2 1.3.4 1.3.3 1.2.2 1.1.2"
+CLUSTER_AUTOSCALER_VERSIONS="1.13.1 1.12.2 1.3.7 1.3.3 1.2.2 1.1.2"
```
jackfrancis: Let's feel free to *add* component versions to VHD but retain old versions as well.

mboersma: There are no references to cluster-autoscaler 1.3.4 any longer, so I dropped it from the VHD list.

mboersma: Actually I guess 1.3.3 is also unused now.

mboersma: I removed the removal--we need to audit AKS as well for usage before aging things out of the VHD image.

jackfrancis:
So we actually do want the cluster-autoscaler `v1.3.7` in the VHD, but we don't need the k8s version itself as that modification already landed.

However, as we've already kicked off the VHD, what is the urgency of bumping cluster-autoscaler to 1.3.7 from 1.3.4?

mboersma:
> what is the urgency of bumping cluster-autoscaler to 1.3.7 from 1.3.4?

Here are the changes: https://github.com/kubernetes/autoscaler/compare/cluster-autoscaler-1.3.4...cluster-autoscaler-1.3.7.  This in particular looks desirable: https://github.com/kubernetes/autoscaler/commit/fae2e71d9f05326cc3ae530926f5675a44c5b9b8

pkg/api/k8s_versions.go
```diff
@@ -128,7 +128,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"aci-connector":                    "virtual-kubelet:latest",
 		ContainerMonitoringAddonName:       "oms:ciprod01092019",
 		AzureCNINetworkMonitoringAddonName: "networkmonitor:v0.0.5",
-		"cluster-autoscaler":               "cluster-autoscaler:v1.3.4",
+		"cluster-autoscaler":               "cluster-autoscaler:v1.3.7",
```
jackfrancis: Based on your findings let's definitely keep this change.

CecileRobertMichon: why not add 1.3.7 to the VHD script (even if we don't build it right away)?

mboersma: Because of cranial flatulence? I've added it to the VHD script now. 😄

jackfrancis:
lgtm pending E2E

Issue #624, opened 2019-03-01
Update AKS Engine "default" Kubernetes version to 1.11


jackfrancis:
Given the stance of this CVE we should assume that 1.11 is more "blessed" in terms of getting back-ported fixes going forward:

https://groups.google.com/forum/#!msg/kubernetes-announce/vmUUNkYfG9g/B9rHFrqLCAAJ

CecileRobertMichon:
fixed in #625

PR #625, opened 2019-03-01
feat: default to Kubernetes 1.11

mboersma:
**Reason for Change**:

**Issue Fixed**:

#624 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/const.go
+++ b/pkg/api/common/const.go
@@ -59,7 +59,7 @@ const (
 
 const (
 	// KubernetesDefaultRelease is the default Kubernetes release
-	KubernetesDefaultRelease string = "1.10"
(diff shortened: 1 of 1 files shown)


pkg/api/common/const.go
```diff
@@ -59,7 +59,7 @@ const (
 
 const (
 	// KubernetesDefaultRelease is the default Kubernetes release
-	KubernetesDefaultRelease string = "1.10"
+	KubernetesDefaultRelease string = "1.11"
```
CecileRobertMichon: this is almost too easy :p

tariq1890: Lol, this is a not-so-simple change for me :P. Got caught off-guard

PR #638, opened 2019-03-04
feat: update Azure-NPM to support new network policy specifications

saiyan86:
**Reason for Change**:

This PR supports new [network policy definition](https://kubernetes.io/docs/concepts/services-networking/network-policies/#behavior-of-to-and-from-selectors) since Kubernetes version 1.11. 
This PR is part of Azure/azure-container-networking#307

This enables network policy to select some pods in some namespaces through "AND" logic. 
A quick example:
```
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          user: alice
      podSelector:
        matchLabels:
          role: client
```

This network policy allows ingress traffic from pods labeled with role: client from namespaces labeled with user: alice.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
```diff
@@ -76,7 +76,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
         - name: azure-npm
-          image: containernetworking/azure-npm:v1.0.13
+          image: containernetworking/azure-npm:v1.0.18-rc2
```
CecileRobertMichon: looks like this container image is missing from the VHD base image, could you please add it?

See the way `containernetworking/networkmonitor` is pre-pulled as an example https://github.com/Azure/aks-engine/blob/master/packer/install-dependencies.sh#L176

saiyan86: Thanks! I already addressed this in https://github.com/Azure/aks-engine/blob/3e7c7b75542b1400ea647a1eb66bea93c03135c0/packer/install-dependencies.sh#L128

jackfrancis:
@saiyan86 do we have E2E test coverage to validate these new scenarios?

Also, let's not merge this PR until the container image reference points to a "final" release version. We can use this PR to test the current RC to validate that is is release-worthy.

saiyan86:
@jackfrancis There's no E2E test yet. I am currently working on it. I manually tested this image though.

serbrech:
Hi @saiyan86 I see that the containers takes mounts the iptables lock file in the daemonset.
can you confirm that the iptables commands that you are using use the `--wait` parameter to respect the lock? 
 
This issue has never been answered or closed : https://github.com/Azure/azure-container-networking/issues/252

saiyan86:
@serbrech Thanks for the reminder! I will address this.

saiyan86:
@jackfrancis I will add e2e tests in this PR. Changing this to WIP.

packer/install-dependencies.sh
```diff
@@ -125,6 +125,11 @@ for AZURE_CNI_NETWORKMONITOR_VERSION in ${AZURE_CNI_NETWORKMONITOR_VERSIONS}; do
     pullContainerImage "docker" "containernetworking/networkmonitor:v${AZURE_CNI_NETWORKMONITOR_VERSION}"
 done
 
+AZURE_NPM_VERSIONS="1.0.18-rc2"
```
CecileRobertMichon: these lines slightly changed in https://github.com/Azure/aks-engine/commit/5d1523de26ef19f719d00aeb6f98406f9afb5c5c#diff-62ed5765f6aaf1d428746b3d7b597100L125

Can you please rebase and add the new echo statement so this is included in release notes?

saiyan86: done!

saiyan86:
@jackfrancis @CecileRobertMichon This PR is done. Could you please take a look? Thanks!

packer/install-dependencies.sh
```diff
@@ -198,6 +198,13 @@ for AZURE_CNI_NETWORKMONITOR_VERSION in ${AZURE_CNI_NETWORKMONITOR_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
 done
 
+AZURE_NPM_VERSIONS="1.0.18-rc3"
```
CecileRobertMichon: are we still waiting for a final release image?

parts/k8s/addons/kubernetesmasteraddons-azure-npm-daemonset.yaml
```diff
@@ -76,7 +76,7 @@ spec:
         beta.kubernetes.io/os: linux
       containers:
         - name: azure-npm
-          image: containernetworking/azure-npm:v1.0.13
+          image: containernetworking/azure-npm:v1.0.18-rc3
```
jackfrancis: as @CecileRobertMichon mentioned, we don't want to replace a final release with a newer RC release

saiyan86: OK. I will push a final release today and update this.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -937,124 +937,194 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				for _, frontendDevPod := range frontendDevPods {
+					pass, err := frontendDevPod.CheckLinuxOutboundConnection(5*time.Second, cfg.Timeout)
 					Expect(err).NotTo(HaveOccurred())
 					Expect(pass).To(BeTrue())
 				}
 
-				By("Ensuring we have outbound internet access from the nginx server pods")
-				serverPods, err := serverDeploy.Pods()
+				By("Ensuring we have outbound internet access from the backend pods")
+				backendPods, err := backendDeployment.Pods()
```
tariq1890: There seems to be duplication here. Let's condense them to a method that takes a list of `Pods` as a parameter.

saiyan86: Because these pods have different purposes, this way we have more readability/granularity.

packer/install-dependencies.sh
saiyan86: updated to v1.0.18!

test/e2e/kubernetes/kubernetes_test.go
tariq1890: I respectfully disagree. If we really need that, we could specify a one line comment and and also add a godoc comment to annotate the newly created method.

saiyan86: Do we need a method? The method will only be called once.
If we really want to condense them together, I think this should be enough.
```
pods := []pod.Pod{}
pods = append(pods, frontendProdPods...)
pods = append(pods, frontendDevPods...)
pods = append(pods, backendPods...)
pods = append(pods, nwpolicyPods...)

for _, pod := range pods {
	pass, err := pod.CheckLinuxOutboundConnection(5*time.Second, cfg.Timeout)
	Expect(pass).To(BeTrue())
}
```

CecileRobertMichon:
FYI you'll need to rebase after https://github.com/Azure/aks-engine/pull/763 is merged to fix the VHD pipeline.

test/e2e/kubernetes/kubernetes_test.go
tariq1890: Since we wrap each podList within a `By()` construct, I'd advocate a method.

serbrech:
/hold for iptables --wait in npm implementation
@saiyan86 What's the status on this fix?

test/e2e/kubernetes/kubernetes_test.go
saiyan86: How about this in pod.go?

```
func (l *List) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool, error) {
...

}

func (l *List) ValidateCurlConnection(uri string, sleep, duration time.Duration) (bool, error) {
...
}

func (l *List) CheckWindowsOutboundConnection(url string, sleep, duration time.Duration) (bool, error) {
...
}


```

saiyan86:
@serbrech Azure/azure-container-networking#315 addressed your comments, please take a look. Thanks!

saiyan86:
@ninzavivek @tamilmani1989 In case you guys want to take a look 😄

serbrech:
> @serbrech [Azure/azure-container-networking#315](https://github.com/Azure/azure-container-networking/pull/315) addressed your comments, please take a look. Thanks!

lgtm, although I suppose this needs to bump the npm image version once it's published?

saiyan86:
> > @serbrech [Azure/azure-container-networking#315](https://github.com/Azure/azure-container-networking/pull/315) addressed your comments, please take a look. Thanks!
> 
> lgtm, although I suppose this needs to bump the npm image version once it's published?

This gonna be included in v1.0.18 image. 
Azure/aks-engine#638 bumps the image version to v1.0.18.

saiyan86:
> /lgtm

Thanks a lot Jack!

PR #650, opened 2019-03-05
fix: always assign target version on upgrade

serbrech:
**Reason for Change**:  

This fixes a bug that did not assign the target version on the containerService object when `--force` was applied to the upgrade command.

**Issue Fixed**:

issue wasn't raised

**Requirements**:

- [x] adds unit tests

**Notes**:
The fix is implemented with the simplest change possible (move the assignment out of the if block) and does not refactor the upgrade command logic to make it less prone to such errors.

related to #525

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -157,7 +157,7 @@ func (uc *upgradeCmd) loadCluster(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error reading ARM file")
 	}
 	defer armTemplateHandle.Close()
-	err = uc.validateCurrentLocalState(armTemplateHandle)
+	err = uc.initializeFromLocalState(armTemplateHandle)
 	if err != nil {
 		return errors.Wrap(err, "error validating the api model")
 	}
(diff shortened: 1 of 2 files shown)


cmd/upgrade.go
```diff
@@ -199,6 +197,7 @@ func (uc *upgradeCmd) validateCurrentLocalState(armTemplateHandle io.Reader) err
 			return errors.Wrap(err, "Invalid upgrade target version. Consider using --force if you really want to proceed")
 		}
 	}
+	uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion = uc.upgradeVersion
```
CecileRobertMichon: Doing this in the validation function `validateCurrentLocalState` is still ambiguous, can we do it when the function returns?

CecileRobertMichon: otherwise maybe rename the function to reflect that it does more than validation

serbrech: I tried, but realized it wasn't achieving much, because lots of other areas are setting state in there.
it also made testing much harder, I had a test harness for the validate function, but there isn't any for loadCluster I think, which will need more setup.
I'll change the name of the method, getting a test harness around the loadCluster function is out of scope for this fix.

serbrech: Maybe `intializeFromLocalState` is more accurate ?
It sets the namesuffix, it sets the agentPools name, it sets the targetVersion.
Validation is secondary when you start setting state all over the place :)

PR #663, opened 2019-03-06
test: pass in skip_test env variable down to the test container

serbrech:
**Reason for Change**:
This allows to pass `SKIP_TEST=true` down to `make test-kubernetes` which effectively just builds a test cluster.


jackfrancis:
lgtm

PR #672, opened 2019-03-07
feat: add selective agentpool upgrade support - VMSS

kkmsft:
Currently the upgrade operaton upgrades the entire cluster, even if the 'AgentPoolsToUpgrade' is specified in the upgrade operation. This change adds the relevant code to skip agentpools which are not specified in the AgentPoolsToUpgrade, thus providing agentpool level upgrade support. This change only supports VMSS.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/utils/util.go
+++ b/pkg/armhelpers/utils/util.go
@@ -160,26 +160,3 @@ func GetK8sVMName(p *api.Properties, agentPoolProfile *api.AgentPoolProfile, age
 
 	return "", errors.Errorf("Failed to reconstruct VM Name")
 }
-
-// IsVMSSUpgradeable - checks if the current vm(in case of VMA) or vm instance (in case of VMSS)
-// is specified in the upgradeMap.
-func IsVMSSUpgradeable(upgradeMap map[string]bool, vmssName string) bool {
-	if len(upgradeMap) > 0 {
-		// Per agentpool upgrade is supported as of now only for linux agentpool names.
-		// When the windows and linux agentpool names will be unified to same format
-		// this will apply across.
-		regexp := regexp.MustCompile(vmssNamingFormat)
-		if capture := regexp.FindStringSubmatch(vmssName); capture != nil {
-			agentPoolName := capture[vmssAgentPoolNameIndex]
-			// Skip the VMSS pool if not requested by caller.
-			if upgradeMap[agentPoolName] {
-				return true
-			} else {
-				return false
-			}
-		}
-	}
-	// If the map is not given, or could not match up any naming convention,
-	// then just return back true to let the upgrade proceed as before.
-	return true
-}
--- a/pkg/armhelpers/utils/util.go
+++ b/pkg/armhelpers/utils/util.go
@@ -160,3 +160,26 @@ func GetK8sVMName(p *api.Properties, agentPoolProfile *api.AgentPoolProfile, age
 
 	return "", errors.Errorf("Failed to reconstruct VM Name")
 }
+
+// IsVMSSUpgradeable - checks if the current vm(in case of VMA) or vm instance (in case of VMSS)
+// is specified in the upgradeMap.
+func IsVMSSUpgradeable(upgradeMap map[string]bool, vmssName string) bool {
+	if len(upgradeMap) > 0 {
+		// Per agentpool upgrade is supported as of now only for linux agentpool names.
+		// When the windows and linux agentpool names will be unified to same format
+		// this will apply across.
+		regexp := regexp.MustCompile(vmssNamingFormat)
+		if capture := regexp.FindStringSubmatch(vmssName); capture != nil {
+			agentPoolName := capture[vmssAgentPoolNameIndex]
+			// Skip the VMSS pool if not requested by caller.
+			if upgradeMap[agentPoolName] {
+				return true
+			} else {
+				return false
+			}
+		}
+	}
+	// If the map is not given, or could not match up any naming convention,
+	// then just return back true to let the upgrade proceed as before.
+	return true
+}
--- a/pkg/operations/kubernetesupgrade/upgradecluster.go
+++ b/pkg/operations/kubernetesupgrade/upgradecluster.go
@@ -18,6 +18,9 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+// IsVMSSToBeUpgradedCb - Call back for checking whether the given vmss is to be upgraded or not.
+type IsVMSSToBeUpgradedCb func(vmss string, cs *api.ContainerService) bool
+
 // ClusterTopology contains resources of the cluster the upgrade operation
 // is targeting
 type ClusterTopology struct {
@@ -34,6 +37,8 @@ type ClusterTopology struct {
 
 	MasterVMs         *[]compute.VirtualMachine
 	UpgradedMasterVMs *[]compute.VirtualMachine
+
+	IsVMSSToBeUpgraded IsVMSSToBeUpgradedCb
 }
 
 // AgentPoolScaleSet contains necessary data required to upgrade a VMSS
@@ -127,7 +132,7 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 			return err
 		}
 		for _, vmScaleSet := range vmScaleSetPage.Values() {
-			if !utils.IsVMSSUpgradeable(uc.AgentPoolsToUpgrade, *vmScaleSet.Name) {
+			if uc.IsVMSSToBeUpgraded != nil && !uc.IsVMSSToBeUpgraded(*vmScaleSet.Name, uc.DataModel) {
 				continue
 			}
 			for vmScaleSetVMsPage, err := uc.Client.ListVirtualMachineScaleSetVMs(ctx, resourceGroup, *vmScaleSet.Name); vmScaleSetVMsPage.NotDone(); err = vmScaleSetVMsPage.NextWithContext(ctx) {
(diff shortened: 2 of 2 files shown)


pkg/armhelpers/utils/util.go
```diff
@@ -26,9 +26,9 @@ const (
 	// here there are 2 capture groups
 	//  the first is the agent pool name, which can contain -s
 	//  the second group is the Cluster ID for the cluster
-	vmssNamingFormat       = "^[0-9a-zA-Z]+-(.+)-([0-9a-fA-F]{8})-vmss$"
-	vmssAgentPoolNameIndex = 1
-	vmssClusterIDIndex     = 2
+	VmssNamingFormat       = "^[0-9a-zA-Z]+-(.+)-([0-9a-fA-F]{8})-vmss$"
```
CecileRobertMichon: let's keep the naming here to follow Go consistent casing guidelines https://github.com/golang/go/wiki/CodeReviewComments#initialisms

CecileRobertMichon: same for all 3 vars

CecileRobertMichon: or `VMSSNamingFormat` if the variable is exported

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -122,11 +123,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 		kubeClient = k
 	}
 
+	clusterID := uc.DataModel.Properties.GetClusterID()
+	agentPoolsToUpgrade := make(map[string]bool)
+	for agentPoolName, val := range uc.AgentPoolsToUpgrade {
+		agentPoolsToUpgrade[agentPoolName] = val
```
CecileRobertMichon: why copy the map into a new one here?

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -122,11 +123,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
+	}
+
 	for vmScaleSetPage, err := uc.Client.ListVirtualMachineScaleSets(ctx, resourceGroup); vmScaleSetPage.NotDone(); err = vmScaleSetPage.NextWithContext(ctx) {
 		if err != nil {
 			return err
 		}
 		for _, vmScaleSet := range vmScaleSetPage.Values() {
+			vmssRegex := regexp.MustCompile(utils.VmssNamingFormat)
+			if capture := vmssRegex.FindStringSubmatch(*vmScaleSet.Name); capture != nil {
+				if !strings.EqualFold(clusterID, capture[utils.VmssClusterIDIndex]) {
```
CecileRobertMichon: I think this might not work for Windows because windows nodes only have a substring of the Cluster ID in their name

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -122,11 +123,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
+					continue
+				}
+				agentPoolName := capture[utils.VmssAgentPoolNameIndex]
+				// Skip the vmss pool if not requested by caller.
```
CecileRobertMichon: s/vmss/VMSS

pkg/armhelpers/utils/util.go
CecileRobertMichon: these should probably be constants... (not added by your PR but in case you want some extra credit :P )

pkg/armhelpers/utils/util.go
```diff
@@ -26,9 +26,9 @@ const (
 	// here there are 2 capture groups
 	//  the first is the agent pool name, which can contain -s
 	//  the second group is the Cluster ID for the cluster
-	vmssNamingFormat       = "^[0-9a-zA-Z]+-(.+)-([0-9a-fA-F]{8})-vmss$"
-	vmssAgentPoolNameIndex = 1
-	vmssClusterIDIndex     = 2
+	VMSSNamingFormat       = "^[0-9a-zA-Z]+-(.+)-([0-9a-fA-F]{8})-vmss$"
+	VMSSAgentPoolNameIndex = 1
+	VMSSClusterIDIndex     = 2
```
CecileRobertMichon: these should probably be constants... (not added by your PR but in case you want some extra credit :P )

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -122,11 +123,28 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
+	}
+
 	for vmScaleSetPage, err := uc.Client.ListVirtualMachineScaleSets(ctx, resourceGroup); vmScaleSetPage.NotDone(); err = vmScaleSetPage.NextWithContext(ctx) {
 		if err != nil {
 			return err
 		}
 		for _, vmScaleSet := range vmScaleSetPage.Values() {
+			vmssRegex := regexp.MustCompile(utils.VMSSNamingFormat)
+			if capture := vmssRegex.FindStringSubmatch(*vmScaleSet.Name); capture != nil {
+				if !strings.EqualFold(clusterID, capture[utils.VMSSClusterIDIndex]) {
```
CecileRobertMichon: (re-adding my comment) I think this might not work for Windows because windows nodes only have a substring of the Cluster ID in their name

CecileRobertMichon:
The commit message is misleading, it sounds like you're adding upgrade support for agentpools... Can we change it to something like "add selective agentpool upgrade support for VMSS"

pkg/operations/kubernetesupgrade/upgradecluster.go
kkmsft: sure, missed that..I some how started living in the new world where naming is the same.. will add the check for windows.. and resend...

kkmsft:
yep, I agree.. I had per agent pool upgrade which did not look okay, so removed it. but "selective" word looks good. will resend shortly..

xizhamsft:
LGTM

pkg/armhelpers/utils/util.go
```diff
@@ -160,3 +160,26 @@ func GetK8sVMName(p *api.Properties, agentPoolProfile *api.AgentPoolProfile, age
 
 	return "", errors.Errorf("Failed to reconstruct VM Name")
 }
+
+// IsVMSSUpgradeable - checks if the current vm(in case of VMA) or vm instance (in case of VMSS)
+// is specified in the upgradeMap.
+func IsVMSSUpgradeable(upgradeMap map[string]bool, vmssName string) bool {
```
tariq1890: I don't think this belongs in the `utils` package.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -127,6 +127,9 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 			return err
 		}
 		for _, vmScaleSet := range vmScaleSetPage.Values() {
+			if !utils.IsVMSSUpgradeable(uc.AgentPoolsToUpgrade, *vmScaleSet.Name) {
```
tariq1890: can we have this method in the operations method itself?

pkg/armhelpers/utils/util.go
kkmsft: what would be a good place for it ?

pkg/operations/kubernetesupgrade/upgradecluster.go
kkmsft: I want to keep the name parsing logic out of the upgrade code. Is there a way we can do this with the common name parsing logic in the utils and only upgrade related part here  ?

pkg/armhelpers/utils/util.go
tariq1890: can we have it in the `kubernetesupgrade` package?

pkg/operations/kubernetesupgrade/upgradecluster.go
tariq1890: Got it. In that case, can we have the method that extracts the agentPoolName (in `armhelpers` package)and then the utility method (perhaps in helpers package. It could even be a private local method.) which checks if the string exists in the map or not.

Thoughts on this?

tariq1890: Reasoning for this is to have small and focused methods.

kkmsft: Thanks Tariq, will give that try and get back in next iteration.

tariq1890: Thanks for your patience with this PR :)

pkg/armhelpers/utils/util.go
CecileRobertMichon: I would rename this function to something other than upgradable - upgradable sounds like it's checking if it _can_ be upgraded as opposed to what it's actually doing: checking if it _should_ be upgraded. The name also conflicts with pre-existing [upgradable](https://github.com/Azure/aks-engine/blob/master/pkg/operations/kubernetesupgrade/upgradecluster.go#L215) function which actually does check if a node can be upgraded.

pkg/armhelpers/utils/util.go
```diff
@@ -160,3 +160,26 @@ func GetK8sVMName(p *api.Properties, agentPoolProfile *api.AgentPoolProfile, age
...
+	if len(upgradeMap) > 0 {
+		// Per agentpool upgrade is supported as of now only for linux agentpool names.
+		// When the windows and linux agentpool names will be unified to same format
+		// this will apply across.
+		regexp := regexp.MustCompile(vmssNamingFormat)
```
CecileRobertMichon: let's add unit tests for this

pkg/operations/kubernetesupgrade/upgradecluster.go
serbrech: I would really prefer to avoid regex because it makes the pattern cryptic, especially that is easy to just split on a character in this case.

PR #675, opened 2019-03-07
fix: always configure the container runtime for GPU in an N series context

jackfrancis:
**Reason for Change**:

We currently configure docker as GPU-enabled only if the particular nvidia addon that we deliver is enabled. This is problematic in the event that someone wants to disable *out* addon and configure their own. In that event, we should *always* configure the container runtime for GPU in an N series context.

Also cleaned up an unused taints variable.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesbase.t
+++ b/parts/k8s/kubernetesbase.t
@@ -16,11 +16,6 @@
       {{range $index, $agent := .AgentPoolProfiles}}
           "{{.Name}}Index": {{$index}},
           {{template "k8s/kubernetesagentvars.t" .}}
-          {{if IsNSeriesSKU .}}
-            {{if IsNVIDIADevicePluginEnabled}}
-            "registerWithGpuTaints": "nvidia.com/gpu=true:NoSchedule",
-            {{end}}
-          {{end}}
           {{if .IsStorageAccount}}
             {{if .HasDisks}}
               "{{.Name}}DataAccountName": "[concat(variables('storageAccountBaseName'), 'data{{$index}}')]",
--- a/pkg/engine/armvariables.go
+++ b/pkg/engine/armvariables.go
@@ -10,7 +10,6 @@ import (
 	"strconv"
 
 	"github.com/Azure/aks-engine/pkg/api"
-	"github.com/Azure/aks-engine/pkg/api/common"
 	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/go-autorest/autorest/to"
 )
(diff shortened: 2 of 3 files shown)


CecileRobertMichon:
I would call this a `fix` since it looks like it fixes a particular use case

lgtm

jackfrancis:
out of band N series tests look good

parts/k8s/kubernetesbase.t
```diff
@@ -16,11 +16,6 @@
       {{range $index, $agent := .AgentPoolProfiles}}
           "{{.Name}}Index": {{$index}},
           {{template "k8s/kubernetesagentvars.t" .}}
-          {{if IsNSeriesSKU .}}
```
jackfrancis: This is cruft. I can't see any usage of this `registerWithGpuTaints` ARM variable anywhere.

CecileRobertMichon:
can we change the commit message to `always configure the container runtime for GPU in an N series context` or something like that? "rationalize nvidia addons surface area" is pretty general and doesn't mean much to me...

parts/k8s/kubernetesbase.t
```diff
@@ -16,11 +16,6 @@
...
-            {{if IsNVIDIADevicePluginEnabled}}
-            "registerWithGpuTaints": "nvidia.com/gpu=true:NoSchedule",
```
CecileRobertMichon: needs to be removed in equivalent https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesbase.t#L21

CecileRobertMichon: https://github.com/Azure/aks-engine/blob/master/pkg/engine/armvariables.go#L31 is the right link

jackfrancis: thanks! done

pkg/engine/armvariables.go
```diff
@@ -27,12 +26,6 @@ func GetKubernetesVariables(cs *api.ContainerService) map[string]interface{} {
 			k8sVars[k] = v
 		}
 
-		if common.IsNvidiaEnabledSKU(profile.VMSize) {
-			if cs.Properties.IsNVIDIADevicePluginEnabled() {
-				k8sVars["registerWithGpuTaints"] = "nvidia.com/gpu=true:NoSchedule"
```
CecileRobertMichon: this might trigger a UT failure heads up

jackfrancis: passed locally for me

PR #693, opened 2019-03-08
fix: commands return errors rather than exit

### LINKED ISSUE #518: Suggested command printed twice
In the current v0.30.1 aks-engine, mistyping a command does invoke Cobra's "[suggested command](https://github.com/spf13/cobra#suggestions-when-unknown-command-happens)" feature, but somehow we're doing it wrong. The suggestion is printed twice:

```
$ aks-engine delpoy
Error: unknown command "delpoy" for "aks-engine"

Did you mean this?
	deploy

Run 'aks-engine --help' for usage.
FATA[0000] unknown command "delpoy" for "aks-engine"

Did you mean this?
	deploy 
```

### LINKED ISSUE #536: Upgrade Command use log.Fatalf to handle errors
No

---
Issue

---
current

---

The upgrade command `run` returns an `error` object, but handle errors in inconsistent ways. Sometimes it returns the error object, and some other times it just calls`log.Fatalf` and exits. This makes testing harder and using the command programmatically more difficult than it should be (i.e. must handle different error patterns).

Non exhaustive samples : 

log.Fatalf :
https://github.com/Azure/aks-engine/blob/93be4e627626c20b37558deae3778ae6e4b4cfaf/cmd/upgrade.go#L229-L232

return err :
https://github.com/Azure/aks-engine/blob/93be4e627626c20b37558deae3778ae6e4b4cfaf/cmd/upgrade.go#L270-L273

mboersma:
**Reason for Change**:

Many places in the `cmd` module were calling `log.Fatalf` which makes that code untestable since it calls `os.Exit`. (This also meant that errors in a couple commands would have a return code of 0.)

This refactors commands so errors are returned all the way to the caller in `main.go` where they can be handled consistently. This also add some basic test coverage at that level, although there is more that can be done in that respect.

**Issue Fixed**:

Fixes #536
Fixes #518

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cmd/version_test.go
+++ b/cmd/version_test.go
@@ -21,7 +21,7 @@ var _ = Describe("the version command", func() {
 
 		command.SetArgs([]string{})
 		err := command.Execute()
-		Expect(err).To(BeNil())
+		Expect(err).NotTo(HaveOccurred())
 	})
 
 	It("should print a json version of AKS Engine", func() {
@@ -46,7 +46,7 @@ var _ = Describe("the version command", func() {
 	It("should error when asked for a yaml version of AKS Engine", func() {
 		output, err := getVersion("yaml")
 
-		Expect(err).NotTo(BeNil())
+		Expect(err).To(HaveOccurred())
 		Expect(err.Error()).To(Equal("output format \"yaml\" is not supported"))
 		Expect(output).To(BeEmpty())
 	})
--- a/cmd/version_test.go
+++ b/cmd/version_test.go
@@ -5,6 +5,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/Azure/aks-engine/pkg/helpers"
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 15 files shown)


cmd/version_test.go
```diff
@@ -13,22 +13,27 @@ import (
...
-		Expect(output.Flags().Lookup("output")).NotTo(BeNil())
+		command := newVersionCmd()
+		Expect(command.Use).Should(Equal(versionName))
+		Expect(command.Short).Should(Equal(versionShortDescription))
+		Expect(command.Long).Should(Equal(versionLongDescription))
+		Expect(command.Flags().Lookup("output")).NotTo(BeNil())
+
+		command.SetArgs([]string{})
+		err := command.Execute()
+		Expect(err).To(BeNil())
```
serbrech: there is a helper for errors `.ToNot(HaveOccurred())` but that's nitpick.

main.go
```diff
@@ -13,6 +15,6 @@ func main() {
 	log.SetFormatter(&log.TextFormatter{ForceColors: true})
 	log.SetOutput(colorable.NewColorableStdout())
 	if err := cmd.NewRootCmd().Execute(); err != nil {
-		log.Fatalln(err)
+		os.Exit(1)
```
serbrech: Does this mean we don't print the error messages anymore?

mboersma: No, cobra was printing them for us in the default implementation of `Execute()`, and then we were printing them again:

```console
$ ./bin/aks-engine get-versions -o yaml
Error: output format "yaml" is not supported
Usage:
  aks-engine get-versions [flags]

Flags:
  -h, --help             help for get-versions
  -o, --output string    Output format. Allowed values: human, json (default "human")
      --version string   Kubernetes version (optional)
      --windows          Kubernetes cluster with Windows nodes (optional)

Global Flags:
      --debug   enable verbose debug logs

FATA[0000] output format "yaml" is not supported
```

This change stops printing that last line.

PR #702, opened 2019-03-08
test: don't hardcode base 64 encoded script values in template tests

CecileRobertMichon:
**Reason for Change**:

Remove hardcoded scripts in unit tests. These cause dev friction as the base64 encoded string needs to be updated each time one of these scripts is changed. The goal of that specific unit test is not to test the content of the scripts themselves but instead to test template generation.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tariq1890:
The key idea is to have those as raw values in the text. Should those methods change, the test cases will have to be updated  as well. This forces the developer to think/look more closely at the changes involved. 

I would vote for keeping the tests as-is.

serbrech:
That's inline with what I was expressing in the stand-up this morning.
lgtm

jackfrancis:
>  Should those methods change, the test cases will have to be updated as well.

We should unit test the methods themselves for those behaviors, not the ARM resource object composition implementations.

jackfrancis:
lgtm pending unit tests

PR #703, opened 2019-03-08
fix: implement cosmosEtcd no_template

tariq1890:
No_template implementation of CosmosDB Etcd


jackfrancis:
lgtm pending unit tests

PR #705, opened 2019-03-08
fix: add err != nil checks for json Unmarshal in scale and upgrade ops

tariq1890:
Add appropriate err != nil checks to avoid panics when a json string isn't unmarshalled succesfully.


jackfrancis:
FYI rebased/force-pushed to fix unit tests

tariq1890:
Sure. Thanks

tariq1890:
@jackfrancis can this be merged?

PR #714, opened 2019-03-10
feat: windows agentpool support for AKS

xizhamsft:
**Reason for Change**:

AKS windows agent pool support needs a few changes
1. Windows VM name prefix should not include agent pool index since AKS agent pool can be dynamically added and deleted. The new naming prefix get rid of the clusterId and agent pool index. It is now "k8s"+agentpoolname. It is up to AKS code to ensure the agent pool name can't be more than 6 chars.
2. Included a new flag NewWindowsNaming in agent pool profile data model to indicate new naming. Right now, this flag is NOT exposed to AKS-engine users. Only AKS can set this flag.
2. The generation of kubelet config for windows agent pool assumes kubernetes config being null for windows agent pool definition. This might not always be the case for AKS.
3. The two kubelet flags removed for linux agent pool when secure kubelet is not enabled are NOT removed for windows agentpool.  We should remove them too to be consistent with Linux agent pool.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -128,17 +128,25 @@ func (cs *ContainerService) setKubeletConfig() {
 		if profile.KubernetesConfig == nil {
 			profile.KubernetesConfig = &KubernetesConfig{}
 			profile.KubernetesConfig.KubeletConfig = make(map[string]string)
-			if profile.OSType == "Windows" {
-				for key, val := range staticWindowsKubeletConfig {
-					profile.KubernetesConfig.KubeletConfig[key] = val
-				}
+		}
+
+		if profile.OSType == "Windows" {
+			for key, val := range staticWindowsKubeletConfig {
+				profile.KubernetesConfig.KubeletConfig[key] = val
 			}
 		}
+
 		setMissingKubeletValues(profile.KubernetesConfig, o.KubernetesConfig.KubeletConfig)
 
 		if profile.OSType == "Windows" {
 			// Remove Linux-specific values
 			delete(profile.KubernetesConfig.KubeletConfig, "--pod-manifest-path")
+
+			if !to.Bool(o.KubernetesConfig.EnableSecureKubelet) {
+				for _, key := range []string{"--anonymous-auth", "--client-ca-file"} {
+					delete(profile.KubernetesConfig.KubeletConfig, key)
+				}
+			}
 		}
 
 		// For N Series (GPU) VMs
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -30,7 +30,18 @@ func (cs *ContainerService) setKubeletConfig() {
 	// Start with copy of Linux config
 	staticWindowsKubeletConfig := make(map[string]string)
 	for key, val := range staticLinuxKubeletConfig {
-		staticWindowsKubeletConfig[key] = val
+		switch key {
+		case "--pod-manifest-path": // Don't add Linux-specific config
+			break
+		case "--anonymous-auth", "--client-ca-file":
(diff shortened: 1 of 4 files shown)


xizhamsft:
@CecileRobertMichon, @jackfrancis, could you take a look?

pkg/api/defaults-kubelet.go
```diff
@@ -128,17 +128,25 @@ func (cs *ContainerService) setKubeletConfig() {
...
 			}
 		}
+
 		setMissingKubeletValues(profile.KubernetesConfig, o.KubernetesConfig.KubeletConfig)
 
 		if profile.OSType == "Windows" {
 			// Remove Linux-specific values
 			delete(profile.KubernetesConfig.KubeletConfig, "--pod-manifest-path")
+
+			if !to.Bool(o.KubernetesConfig.EnableSecureKubelet) {
```
CecileRobertMichon: let's add this to existing unit tests

jackfrancis: Since @xizhamsft found and fixed a bug in our implementation, I can offer unit tests :)

@xizhamsft could you check over my logical tweaks and confirm that things are as they should be? (Also validate that the unit tests express functional intent)

Sorry, this code area is a bit hairy!

xizhamsft: @jackfrancis , could you push your change again? I did notice you made some changes too and did a force push. Now your changes are lost

xizhamsft: @jackfrancis , LGTM

jackfrancis:
@acs-bot It *is* his PR I assure you I'm just camping! ;)

jackfrancis:
@xizhamsft is `NewWindowsVMNaming` meant to suggest that a profile is using the "v2 Name" instead of the "legacy name"?

Would `NameV2` be a more concise way to express this? @CecileRobertMichon do you have any thoughts?

xizhamsft:
@jackfrancis , agree. "New" windows VM Naming is not very specific, like the "New" iPad, :-)! Any other good suggestions? NameV2 is fine to me.

CecileRobertMichon:
+1 on versioning the name with v2 or something like that. "New" won't be accurate 6 months from now if we decide to change the name again.

jackfrancis:
Another possibility would be adding a `namingVersion` string that would help with the per-version naming logic that we need to deal with. That way the "name" property remains, which is the most semantic and comprehensible label for a "name".

pkg/api/types.go
```diff
@@ -808,7 +809,11 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile) string {
 	vmPrefix := ""
 	if index != -1 {
 		if a.IsWindows() {
-			vmPrefix = nameSuffix[:4] + p.K8sOrchestratorName() + fmt.Sprintf("%02d", index)
+			if a.NewWindowsVMNaming != nil && *a.NewWindowsVMNaming {
```
jackfrancis: Using my suggestion this would change to:

`if a.NameVersion == "v2" {`

@xizhamsft @CecileRobertMichon  thoughts?

CecileRobertMichon: I like that, it's less binary and gives us more room for future changes

xizhamsft: Do we need to differentiate between windows name version vs linux name version? Chances are linux name version never need to be changed. So this should be more like "windowsNameVersion"?

jackfrancis: Yes, it should be `WindowsNameVersion` (`windowsNameVersion` in JSON)

jackfrancis:
Just a reminder @xizhamsft: manually re-pull from this branch as it has commits of mine that you may not have locally. :)

jackfrancis:
lgtm

PR #715, opened 2019-03-10
style: enable vet shadow and remove the shadowed declaration go anti-pattern

tariq1890:
Add stricter go vetting to aks-engine.


mboersma:
lgtm

PR #732, opened 2019-03-11
fix: use the correct go sdk import for system assigned role assignments

tariq1890:
As per the `kubernetesmastervars.t` we need to use 2018-01-01-preview for system authorization. The no_template code needs to use the right import version to reflect that.


jackfrancis:
lgtm pending unit tests

PR #487, opened 2019-02-12
fix: do not create ELB for private clusters

mkosieradzki:
**Reason for Change**:

Prevents creation of ELB for private clusters.

**Issue Fixed**:

Fixes #474 

**Requirements**:

- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #747, opened 2019-03-12
fix: removed unnecessary elb artifacts when PrivateCluster is enabled

mkosieradzki:
**Reason for Change**:

Removes remaining elb artifacts in private cluster. Related to: #487 .

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Hi @mkosieradzki, is this to enable a "private cluster" + "standard load balancer" cluster configuration scenario?

mkosieradzki:
@jackfrancis the scenario you have mentioned has been enabled in #487 and it already works for me. I have just noticed that there are some unncessary leftovers in this scenario and added some cleanup.

PR #748, opened 2019-03-12
style: refactorings suggested by the gocritic linter

mboersma:
**Reason for Change**:

Increases the legibility of some code.

The [`gocritic`](https://github.com/go-critic/go-critic) linter suggests a lot of refactorings I don't think we would want to pursue, so I didn't enable it in our golangci-lint configuration.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/agentPoolOnlyApi/v20180331/validate.go
+++ b/pkg/api/agentPoolOnlyApi/v20180331/validate.go
@@ -262,26 +262,20 @@ func validateAgentPoolVNET(a []*AgentPoolProfile) error {
 
 			if subscription == "" {
 				subscription = subnetSubscription
-			} else {
-				if subscription != subnetSubscription {
-					return ErrorSubscriptionNotMatch
-				}
+			} else if subscription != subnetSubscription {
+				return ErrorSubscriptionNotMatch
 			}
 
 			if resourceGroup == "" {
 				resourceGroup = subnetResourceGroup
-			} else {
-				if resourceGroup != subnetResourceGroup {
-					return ErrorResourceGroupNotMatch
-				}
+			} else if resourceGroup != subnetResourceGroup {
+				return ErrorResourceGroupNotMatch
 			}
 
 			if vnet == "" {
 				vnet = subnetVnet
-			} else {
-				if vnet != subnetVnet {
-					return ErrorVnetNotMatch
-				}
+			} else if vnet != subnetVnet {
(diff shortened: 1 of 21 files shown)


test/e2e/engine/template.go
```diff
@@ -206,17 +206,19 @@ func (e *Engine) GetWindowsTestImages() (*WindowsTestImages, error) {
 		return nil, errors.New("Can't guess a Windows version without Windows nodes in the cluster")
 	}
 
-	if strings.Contains(e.ExpandedDefinition.Properties.WindowsProfile.GetWindowsSku(), "1809") || strings.Contains(e.ExpandedDefinition.Properties.WindowsProfile.GetWindowsSku(), "2019") {
+	windowsSku := e.ExpandedDefinition.Properties.WindowsProfile.GetWindowsSku()
+	switch {
+	case strings.Contains(windowsSku, "1809"):
```
jackfrancis: I don't think this is equivalent. Are we assuming the 1809 case will fall through to the 2019 case?

I think a preferable change would be to use a single func that can evaluate a string against a set of string values for equality.

mboersma: I added the `fallthrough` I forgot. I can pursue your suggested refactoring if that's preferable.

tariq1890: +1 to jack's comment.

tariq1890:
Thanks for this @mboersma !

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -262,26 +262,20 @@ func validateAgentPoolVNET(a []*AgentPoolProfile) error {
 
 			if subscription == "" {
 				subscription = subnetSubscription
-			} else {
```
CecileRobertMichon: gotta love those else if statements

test/e2e/engine/template.go
CecileRobertMichon: can't we do  `case strings.Contains(windowsSku, "1809"), strings.Contains(windowsSku, "2019"):` if the two cases are the same result?

mboersma: We can, and I think that reads better than using `fallthrough` anyway.

mboersma: @jackfrancis @tariq1890 are you ok with the current changes?

PR #751, opened 2019-03-12
test: more apiloader unit tests

jackfrancis:
**Reason for Change**:

add failure condition unit test for LoadContainerServiceFromFile and tests for DeserializeContainerService

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/apiloader.go
+++ b/pkg/api/apiloader.go
@@ -71,7 +71,7 @@ func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
 	}
 }
 
-// DeserializeContainerService loads an AKS Cluster API Model, validates it, and returns the unversioned representation
+// DeserializeContainerService loads an AKS Engine Cluster API Model, validates it, and returns the unversioned representation
 func (a *Apiloader) DeserializeContainerService(contents []byte, validate, isUpdate bool, existingContainerService *ContainerService) (*ContainerService, string, error) {
 	m := &TypeMeta{}
 	if err := json.Unmarshal(contents, &m); err != nil {
@@ -79,8 +79,15 @@ func (a *Apiloader) DeserializeContainerService(contents []byte, validate, isUpd
 	}
 
 	version := m.APIVersion
-	service, err := a.LoadContainerService(contents, version, validate, isUpdate, existingContainerService)
-	return service, version, err
+	var cs *ContainerService
+	var err error
+	switch version {
+	case "2017-08-31", "2018-03-31":
+		cs, _, err = a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate, isUpdate, "", existingContainerService)
+	default:
+		cs, err = a.LoadContainerService(contents, version, validate, isUpdate, existingContainerService)
+	}
+	return cs, version, err
 }
 
 // LoadContainerService loads an AKS Cluster API Model, validates it, and returns the unversioned representation
--- a/pkg/api/apiloader.go
+++ b/pkg/api/apiloader.go
@@ -21,7 +21,6 @@ import (
 	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/aks-engine/pkg/i18n"
 	"github.com/pkg/errors"
-	log "github.com/sirupsen/logrus"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/api/apiloader.go
```diff
@@ -81,13 +80,6 @@ func (a *Apiloader) DeserializeContainerService(contents []byte, validate, isUpd
 
 	version := m.APIVersion
 	service, err := a.LoadContainerService(contents, version, validate, isUpdate, existingContainerService)
-	if service == nil || err != nil {
```
jackfrancis: I don't see any AKS codepath for the `DeserializeContainerService` method. The only invocations are:

- AKS Engine deploy command
- called from `LoadContainerServiceFromFile` (which is never invoked in an AKS context)

jackfrancis: ignore this, we use `DeserializeContainerService` to test AKS as well as AKS Engine model deserialization

Refactored the way that we actually implement this overloading (the switch/case implementation below makes it so we don't call `a.LoadContainerService` unnecessarily in the AKS case

pkg/api/apiloader_test.go
```diff
@@ -105,6 +106,12 @@ func TestLoadContainerServiceFromFile(t *testing.T) {
 	if containerService.Properties.OrchestratorProfile.OrchestratorVersion != "1.8.12" {
 		t.Errorf("Failed to set orcherstator version when it is set in the json, expected 1.8.12 but got %s", containerService.Properties.OrchestratorProfile.OrchestratorVersion)
 	}
+
+	// Test error scenario
+	containerService, _, err = apiloader.LoadContainerServiceFromFile("../this-file-doesnt-exist.json", true, false, nil)
+	if err == nil {
```
tariq1890:
```suggestion
	if err != nil {
```

pkg/api/apiloader_test.go
```diff
@@ -105,6 +106,12 @@ func TestLoadContainerServiceFromFile(t *testing.T) {
...
+		t.Error(err.Error())
```
tariq1890:
```suggestion
		t.Errorf("unexpected error while trying to Load the Apimodel from file: %s", err.Error())
```

pkg/api/apiloader_test.go
```diff
@@ -398,6 +405,45 @@ func TestLoadContainerServiceWithEmptyLocationCustomCloud(t *testing.T) {
...
 
+func TestDeserializeContainerService(t *testing.T) {
+	apiloader := &Apiloader{
+		Translator: nil,
+	}
+
+	// Test AKS Engine api model
+	cs, version, err := apiloader.DeserializeContainerService([]byte(exampleAPIModel), false, false, nil)
+	if err != nil {
+		t.Fatalf("unexpected error deserializing the example apimodel: %s", err)
```
tariq1890: Any specific reason for using `t.Fatalf`?

pkg/api/apiloader_test.go
jackfrancis: No, this test is validating the error scenario.

pkg/api/apiloader_test.go
jackfrancis: copy/pasta, will change

pkg/api/apiloader.go
```diff
@@ -72,23 +71,23 @@ func LoadDefaultContainerServiceProperties() (TypeMeta, *vlabs.Properties) {
...
-		if isAgentPoolOnlyClusterJSON(contents) {
-			log.Info("No masterProfile: interpreting API model as agent pool only")
-			service, _, err = a.LoadContainerServiceForAgentPoolOnlyCluster(contents, version, validate, isUpdate, "", existingContainerService)
-			return service, version, err
-		}
+	var cs *ContainerService
+	var err error
+	switch version {
+	case "2017-08-31", "2018-03-31":
```
CecileRobertMichon: @jackfrancis we should flag this in release notes as a potentially AKS impacting refactor (the commit is test which implies it's only unit tests so it could be easily overlooked).

CecileRobertMichon: `LoadContainerServiceForAgentPoolOnlyCluster` also supports apvlabs.APIVersion (`vlabs`)

jackfrancis: AKS doesn't call this function. As far as I can tell the purpose of this function from an AKS perspective is to test AKS-like api models under `pkg/engine/testdata`. AKS never passes a byte array as input (i.e., raw JSON) to aks-engine for processing.

CecileRobertMichon: Sounds good, thanks for clarifying

PR #753, opened 2019-03-12
Use mcr for hyperkube

cpuguy83:
This implementation is a work-around until we have all k8s images on
mcr.

- Remove the `KubernetesImageBase` setting from the generated default
spec
- `KubernetesImageBase` gets set later, if the user spec did not have
`KubernetesImageBase`, then we'll use gcr for everything except for
hyperkube.

### DIFF
--- a/pkg/engine/params_k8s.go
+++ b/pkg/engine/params_k8s.go
@@ -16,6 +16,13 @@ import (
 	"github.com/Azure/aks-engine/pkg/api/common"
 )
 
+const (
+	// TODO: migrate all k8s images over to mcr and move the azure image base to
+	// the generated default spec
+	defaultKubernetesImageBase = "k8s.gcr.io/"
+	azureKubernetesImageBase   = "mcr.microsoft.com/k8s/core/"
+)
+
 func assignKubernetesParameters(properties *api.Properties, parametersMap paramsMap,
 	cloudSpecConfig api.AzureEnvironmentSpecConfig, generatorCode string) {
 	addValue(parametersMap, "generatorCode", generatorCode)
@@ -27,7 +34,15 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		k8sVersion := orchestratorProfile.OrchestratorVersion
 		k8sComponents := api.K8sComponentsByVersionMap[k8sVersion]
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
+
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
+
+		// This is a temporary work-around as we migrate images over to mcr.
+		hyperkubeImageBase := kubernetesImageBase
+		if kubernetesImageBase == "" {
+			kubernetesImageBase = defaultKubernetesImageBase
+			hyperkubeImageBase = azureKubernetesImageBase
+		}
 
 		if kubernetesConfig != nil {
 			if to.Bool(kubernetesConfig.UseCloudControllerManager) {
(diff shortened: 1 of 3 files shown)


pkg/engine/params_k8s.go
```diff
@@ -27,8 +34,16 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		k8sVersion := orchestratorProfile.OrchestratorVersion
 		k8sComponents := api.K8sComponentsByVersionMap[k8sVersion]
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
+
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
 
+		// This is a temporary work-around as we migrate images over to mcr.
+		hyperkubeImageBase := kubernetesImageBase
+		if kubernetesImageBase == "" {
```
CecileRobertMichon: Just to make sure I understand, this will be empty for public cloud and non-empty for sovereign clouds + custom image base passed in?

cpuguy83: That's the general idea, yes, but...
Where are sovereign cloud defaults defined?

CecileRobertMichon: Yes, see https://github.com/Azure/aks-engine/pull/753/files#diff-f41c9478d3eabb30a6e28202cbfe7e18R233 for an example

cpuguy83: Ok so yes.
The idea here is a user spec could very well define k8s.gcr.io and I don't want to overwrite that value with our own.
But if the user config does not specify a value we can go ahead and insert in our own value for the images that are currently available on mcr.

cpuguy83: There may be a better place to define this logic, though.

cpuguy83: I'm not sure I'm seeing a better place taking another glance at it.

CecileRobertMichon:
let's change the squash commit and PR title to `chore: Pull Hyperkube image from MCR` (I wouldn't call this a `fix` since it's not fixing any bugs).

pkg/engine/params_k8s.go
CecileRobertMichon: do we also want to use MCR in any sovereign clouds? Or is it meant to be public cloud only?

CecileRobertMichon: I'm not sure relying on `kubernetesImageBase` to be empty is a very reliable way to do this

CecileRobertMichon: In any case, we definitely need unit tests for this

pkg/engine/params_k8s.go
```diff
@@ -16,6 +16,13 @@ import (
 	"github.com/Azure/aks-engine/pkg/api/common"
 )
 
+const (
+	// TODO: migrate all k8s images over to mcr and move the azure image base to
```
CecileRobertMichon:
```suggestion
	// TODO: migrate all k8s images over to MCR and move the Azure image base to
```

pkg/engine/params_k8s.go
```diff
@@ -27,8 +34,16 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		k8sVersion := orchestratorProfile.OrchestratorVersion
 		k8sComponents := api.K8sComponentsByVersionMap[k8sVersion]
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
+
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
 
+		// This is a temporary work-around as we migrate images over to mcr.
```
CecileRobertMichon:
```suggestion
		// This is a temporary work-around as we migrate images over to MCR.
```

PR #755, opened 2019-03-13
feat: support deploy command on Azure Stack

### LINKED ISSUE #754: feat: Support deploy command on Azure Stack

---

FEATURE REQUEST
---

---

**Kubernetes version**:

**What happened**:

can not use aks-engine deploy command on Azure Stack

feat: Support deploy command on Azure Stack

honcao:
**Reason for Change**:

Fix #754

**Issue Fixed**:

#754

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


cmd/root.go
```diff
@@ -222,6 +239,39 @@ func (authArgs *authArgs) getClient() (armhelpers.AKSEngineClient, error) {
 	return client, nil
 }
 
+func (authArgs *authArgs) getazsClient() (armhelpers.AKSEngineClient, error) {
```
jackfrancis: Let's rename this func to `getAzureStackClient` so it can't be accidentally mistaken for `getazClient` by humans. :)

jackfrancis:
Should there be updates to `Gopkg.toml` to reflect the addition of new azure-sdk-for-go versions?

cmd/root.go
honcao: sure :)

honcao:
Gopkg.toml already has the constraint and dep ensure will put azure stack API version into vender when the code have the reference.

[[constraint]]
  name = "github.com/Azure/azure-sdk-for-go"
  version = "24.0.0"

jackfrancis:
Could you comment on the `armhelpers` function interface changes? That will require AKS code to be changed to adapt to the new interfaces.

honcao:
Yes, I already made the changes in AKS code to comply the new interface.
It follows the same pattern with other *ResultPage type. like VirtualMachineListResultPage.

pkg/armhelpers/disk.go
```diff
@@ -25,6 +23,7 @@ func (az *AzureClient) DeleteManagedDisk(ctx context.Context, resourceGroupName
 }
 
 // ListManagedDisksByResourceGroup lists managed disks in a resource group.
-func (az *AzureClient) ListManagedDisksByResourceGroup(ctx context.Context, resourceGroupName string) (result compute.DiskListPage, err error) {
```
jackfrancis: This is an example of a breaking change that will slow down delivery of this PR to master. If possible, implement what is needed while keeping the `ListManagedDisksByResourceGroup` method as-is.

pkg/armhelpers/mockclients.go
```diff
@@ -744,8 +744,8 @@ func (mc *MockAKSEngineClient) DeleteManagedDisk(ctx context.Context, resourceGr
 }
 
 // ListManagedDisksByResourceGroup is a wrapper around disksClient.ListManagedDisksByResourceGroup
-func (mc *MockAKSEngineClient) ListManagedDisksByResourceGroup(ctx context.Context, resourceGroupName string) (result compute.DiskListPage, err error) {
```
jackfrancis: ditto about breaking change

pkg/armhelpers/providers.go
```diff
@@ -6,11 +6,11 @@ package armhelpers
...
 	"context"
 
-	"github.com/Azure/azure-sdk-for-go/services/resources/mgmt/2018-05-01/resources"
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
 // ListProviders returns all the providers for a given AzureClient
-func (az *AzureClient) ListProviders(ctx context.Context) (resources.ProviderListResultPage, error) {
-	return az.providersClient.List(ctx, to.Int32Ptr(100), "")
+func (az *AzureClient) ListProviders(ctx context.Context) (ProviderListResultPage, error) {
```
jackfrancis: ditto about breaking change

pkg/armhelpers/disk.go
honcao: @jackfrancis 
This change can not be avoid.
DiskListPage belongs to package github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute, 
so I would need covert the return object from 2017-03-30/compute to 2018-10-01/compute

But the DiskListPage struct did not have public fields, I could not do the deep copy.

the current implementation is I add a wrapper to on top of DiskListPage and expose all the method it supports (for example Next())

// DiskListPage contains a page of Disk values.
type DiskListPage struct {
                fn func(context.Context, DiskList) (DiskList, error)
                dl DiskList
}

cmd/scale.go
```diff
@@ -123,20 +123,8 @@ func (sc *scaleCmd) load(cmd *cobra.Command) error {
 	sc.logger = log.New().WithField("source", "scaling command line")
 	var err error
 
-	if err = sc.authArgs.validateAuthArgs(); err != nil {
```
honcao: @jackfrancis AzureStack flow would need load the environment information, so move the validation order.

cmd/scale.go
```diff
@@ -155,6 +143,24 @@ func (sc *scaleCmd) load(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	if sc.containerService.Properties.IsAzureStackCloud() {
+		writeCustomCloudProfile(sc.containerService)
+		api.AzureCloudSpecEnvMap[api.AzureStackCloud] = *sc.containerService.Properties.CustomCloudProfile.AzureEnvironmentSpecConfig
```
honcao: @jackfrancis  Set the CloudSpec for azure stack. I could create another wrapper struct and expose set and get function. Please let me know what is your thoughts.

cmd/scale.go
```diff
@@ -155,6 +143,24 @@ func (sc *scaleCmd) load(cmd *cobra.Command) error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	if sc.containerService.Properties.IsAzureStackCloud() {
+		writeCustomCloudProfile(sc.containerService)
```
honcao: @jackfrancis , this function will write azure.Environment into temp file and set the environment variable to consume it.

parts/k8s/kubernetesagentresourcesvmas.t
```diff
@@ -46,8 +46,10 @@
           {{end}}
         ]
 {{if not IsAzureCNI}}
+  {{if not IsAzureStackCloud}}
         ,
         "enableIPForwarding": true
```
honcao: @jackfrancis  AzureStack enable the IP forwarding by default and can not change the setting. So I add the logic to omit the filed for Azure Stack.

pkg/engine/params_k8s.go
```diff
@@ -28,6 +28,11 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		k8sComponents := api.K8sComponentsByVersionMap[k8sVersion]
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
+		hyperKubeImageBase := kubernetesConfig.KubernetesImageBase
+
+		if properties.IsAzureStackCloud() {
```
honcao: @jackfrancis , the hyperkube image and other dependencies image are hosted in different repository for Azure Stack.  This change will use the information in kubernetesSpecConfig for the dependencies images and use kubernetesImageBase under kubernetesConfig for hyperkube images.

CecileRobertMichon:
This PR decreased unit test coverage on the project by more than 2.5%, @honcao can you please add unit tests for the newly added functions in a follow up?

tariq1890:
:(

honcao:
@CecileRobertMichon @tariq1890 ,  I will investigate the decrease of test coverage.

PR #749, opened 2019-03-12
chore: update cluster-autoscaler versions

jackfrancis:
**Reason for Change**:

Consume new versions of cluster-autoscaler

**Issue Fixed**:

Fixes #741 

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #756, opened 2019-03-13
chore: enable Kubernetes v1.14.0-beta.2

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#changelog-since-v1140-beta1

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -107,7 +107,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.3":         true,
 	"1.13.4":         true,
 	"1.14.0-alpha.1": false,
-	"1.14.0-alpha.2": true,
+	"1.14.0-alpha.2": false,
 	"1.14.0-beta.1":  true,
 	"1.14.0-beta.2":  true,
 }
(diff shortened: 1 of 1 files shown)


pkg/api/common/versions.go
```diff
@@ -109,6 +109,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.14.0-alpha.1": false,
 	"1.14.0-alpha.2": true,
```
CecileRobertMichon: remove alpha.2?

mboersma: I thought in the past we left all alpha + beta versions enabled, then set them all to `false` when there was an official release to replace them. But I see we disabled alpha.1 so I assume we're applying our support policy to prereleases as well? Works for me.

CecileRobertMichon: yeah I'm not sure, I saw that alpha.1 was deprecated which is why I asked. I don't think we're used to having more than 1 pre-release. Either way is fine I think, it doesn't matter much.

CecileRobertMichon:
I reverted #749 to fix unit tests

CecileRobertMichon:
VHD pipeline fix is in #763

CecileRobertMichon:
both 1.14 tests passed in a previous test https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=156 so this can probably merge without waiting for E2E.

PR #769, opened 2019-03-14
fix: add changes to allow for scaling of VMSS nodes in no_template refactor

tariq1890:


cmd/scale.go
```diff
@@ -369,12 +376,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
...
 
 	transformer := transform.Transformer{Translator: translator.Translator}
-	// Our templates generate a range of nodes based on a count and offset, it is possible for there to be holes in the template
-	// So we need to set the count in the template to get enough nodes for the range, if there are holes that number will be larger than the desired count
-	countForTemplate := sc.newDesiredAgentCount
-	if highestUsedIndex != 0 {
-		countForTemplate += highestUsedIndex + 1 - currentNodeCount
-	}
+
 	addValue(parametersJSON, sc.agentPool.Name+"Count", countForTemplate)
```
jackfrancis: The reason for continuing to add the new "count" value in the params is to persist the "new" desired cluster state (being generous here with the word state)?

Basically, if the new template generation implementation no longer consumes the count via parameters, can we just remove this statement?

tariq1890: It's only for VMSS. VMAS still depends on parameters/variables. Which is why the scale for VMSS was failing in the first place.

tariq1890: The variable dependency in VMSS was removed in no_template.

Issue #664, opened 2019-03-06
move logs under /var/log from VHD creation to somewhere else


weinong:
**Is this a request for help?**:

---
Yes
**Is this an ISSUE or FEATURE REQUEST?** (choose one):

---
Issue
**What version of aks-engine?**:

---

**Kubernetes version**:

**What happened**:
When we look at the logs such as /var/log/kern.log on VM created from AKS VHD, we see logs from when the VHD was created. For example, I only booted up the VM once, but I saw it twice in kern.log.

**What you expected to happen**:

**How to reproduce it** (as minimally and precisely as possible):

**Anything else we need to know**:

PR #777, opened 2019-03-15
fix: move /var/log/ during VHD build to a vhd specific directory

CecileRobertMichon:
**Reason for Change**:

When we look at the logs such as /var/log/kern.log on VM created from AKS VHD, we see logs from when the VHD was created. For example, I only booted up the VM once, but I saw it twice in kern.log.

**Issue Fixed**:

#664 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/cleanup-vhd.sh
+++ b/packer/cleanup-vhd.sh
@@ -1,7 +1,11 @@
 #!/bin/bash -eux
 
-## Cleanup packer SSH key and machine ID generated for this boot
+# Cleanup packer SSH key and machine ID generated for this boot
 rm -f /root/.ssh/authorized_keys
 rm -f /home/packer/.ssh/authorized_keys
 rm -f /etc/machine-id
 touch /etc/machine-id
+
+# Move logs under /var/log from VHD creation
+sudo mkdir /var/log.vhd
+sudo mv /var/log /var/log.vhd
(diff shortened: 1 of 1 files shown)


packer/cleanup-vhd.sh
```diff
@@ -1,7 +1,11 @@
...
 rm -f /root/.ssh/authorized_keys
 rm -f /home/packer/.ssh/authorized_keys
 rm -f /etc/machine-id
-touch /etc/machine-id
\ No newline at end of file
+touch /etc/machine-id
+
+# Move logs under /var/log from VHD creation
+sudo mkdir /var/log.vhd
+sudo mv /var/log /var/log.vhd
```
mboersma: This puts the VHD creation logs under `/var/log.vhd/var/log`. Is that the intention, or should we just rename it?

```
sudo mv /var/log /var/log.vhd
sudo mkdir /var/log
```

CecileRobertMichon: does it? My intention was to put it under `/var/log.vhd`

mboersma: It does--I think the two commands I posted are what you want.

CecileRobertMichon: fixed

mboersma:
lgtm

PR #771, opened 2019-03-14
fix: actually apt-mark hold walinuxagent during all CSE runs

jackfrancis:
**Reason for Change**:

By not marking walinuxagent for "hold" (i.e., do not modify via apt) during CSE, even VHD scenarios, we risk walinuxagent being updated via CSE, which will never finish, because walinuxagent updates require a non-active CSE operation.

Additionally, we were releasing the hold too early, as apt system.daily operations are able to update waagent, and despite the invocation being sent to the background, the fact that it is a child process of the parent CSE runtime process in practice means we see these symptoms.

**Issue Fixed**:

Fixes #768 Fixes #758 

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #778, opened 2019-03-15
fix: mark walinux for hold in cloud-init

jackfrancis:
**Reason for Change**:

Follow-up of #771

Testing against the changes in #771 revealed that cloud-init itself is invoking `apt` to install some packages:

```
2019-03-15 19:26:12,725 - helpers.py[DEBUG]: Running update-sources using lock (<FileLock using file '/var/lib/cloud/instances/3557FA54-5097-654B-9116-21C59991226E/sem/update_sources'>)
2019-03-15 19:26:12,726 - util.py[DEBUG]: Running command ['eatmydata', 'apt-get', '--option=Dpkg::Options::=--force-confold', '--option=Dpkg::options::=--force-unsafe-io', '--assume-yes', '--quiet', 'update'] with allowed return codes [0] (shell=False, capture=False)
2019-03-15 19:26:25,480 - util.py[DEBUG]: apt-update [eatmydata apt-get --option=Dpkg::Options::=--force-confold --option=Dpkg::options::=--force-unsafe-io --assume-yes --quiet update] took 12.754 seconds
2019-03-15 19:26:25,480 - helpers.py[DEBUG]: update-sources already ran (freq=once-per-instance)
2019-03-15 19:26:25,480 - util.py[DEBUG]: Running command ['eatmydata', 'apt-get', '--option=Dpkg::Options::=--force-confold', '--option=Dpkg::options::=--force-unsafe-io', '--assume-yes', '--quiet', 'install', 'jq', 'traceroute'] with allowed return codes [0] (shell=False, capture=False)
2019-03-15 19:26:34,734 - util.py[DEBUG]: apt-install [eatmydata apt-get --option=Dpkg::Options::=--force-confold --option=Dpkg::options::=--force-unsafe-io --assume-yes --quiet install jq traceroute] took 9.254 seconds
2019-03-15 19:26:34,735 - handlers.py[DEBUG]: finish: modules-final/config-package-update-upgrade-install: SUCCESS: config-package-update-upgrade-install ran successfully
```

This means in practice that when we attempt to `apt-mark hold walinuxagent` in CSE it may fail due to cloud-init already having reserved the lock.

Additionally, we are explicitly pinning the version of walinuxagent to further protect against runtime changes to the walinuxagent. This last change is a defensive manouver. Based on production data following the introduction of `2.2.37` we want to introduce new version changes purposefully.

We can consider reverting to the prior behavior once we recover create stability, if the static version pinning produces undesired effects for long-lived clusters who wish to update their `walinuxagent` versions "out of band".

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -306,5 +306,5 @@ coreos:
 runcmd:
 - set -x
 - . /opt/azure/containers/provision_source.sh
-- {{GetKubernetesAgentPreprovisionYaml .}}
+- aptmarkWALinuxAgent hold{{GetKubernetesAgentPreprovisionYaml .}}
 {{end}}
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -107,6 +107,14 @@ write_files:
     RemainAfterExit=yes
     ExecStart=/usr/local/bin/health-monitor.sh container-runtime
 
+- path: /etc/apt/preferences
+  permissions: "0644"
+  owner: root
+  content: |
+    Package: walinuxagent
+    Pin: version 2.2.32.2
+    Pin-Priority: 550
+
 {{if .KubernetesConfig.RequiresDocker}}
     {{if not .IsCoreOS}}
 - path: /etc/systemd/system/docker.service.d/clear_mount_propagation_flags.conf
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -179,6 +179,7 @@ if $REBOOTREQUIRED; then
   fi
 else
   if [[ $OS == $UBUNTU_OS_NAME ]]; then
-      runAptDaily &
+      /usr/lib/apt/apt.systemd.daily &
+      aptmarkWALinuxAgent unhold &
   fi
 fi
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -57,10 +57,6 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
-if [[ $OS == $UBUNTU_OS_NAME ]]; then
-    holdWALinuxAgent "hold"
-fi
-
 if [[ ! -z "${MASTER_NODE}" ]] && [[ -z "${COSMOS_URI}" ]]; then
     	installEtcd
 fi
@@ -179,7 +175,7 @@ if $REBOOTREQUIRED; then
   echo 'reboot required, rebooting node in 1 minute'
   /bin/bash -c "shutdown -r 1 &"
   if [[ $OS == $UBUNTU_OS_NAME ]]; then
-      holdWALinuxAgent "unhold"
+      aptmarkWALinuxAgent "unhold"
   fi
 else
   if [[ $OS == $UBUNTU_OS_NAME ]]; then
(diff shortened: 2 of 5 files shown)


parts/k8s/kubernetescustomscript.sh
```diff
@@ -57,10 +57,6 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
-if [[ $OS == $UBUNTU_OS_NAME ]]; then
```
jackfrancis: I think we need to get rid of the "apt-mark hold" invocation in CSE, as the hold command will wait for the release of apt locks, which may take a long time if cloud-init has already reserved them.

parts/k8s/kubernetesmastercustomdata.yml
```diff
@@ -504,5 +504,6 @@ coreos:
 {{else}}
 runcmd:
 - set -x
-- {{GetKubernetesMasterPreprovisionYaml}}
+- . /opt/azure/containers/provision_source.sh
+- aptmarkWALinuxAgent hold{{GetKubernetesMasterPreprovisionYaml}}
```
CecileRobertMichon: shouldn't hold be "hold"?

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -306,5 +306,5 @@ coreos:
 runcmd:
 - set -x
 - . /opt/azure/containers/provision_source.sh
-- {{GetKubernetesAgentPreprovisionYaml .}}
-{{end}}
+- aptmarkWALinuxAgent hold{{GetKubernetesAgentPreprovisionYaml .}}
```
CecileRobertMichon: same here

parts/k8s/kubernetesmastercustomdata.yml
jackfrancis: I'm skittish about quotes in runcmd + cloud-init. Strictly speaking the `"` are unnecessary:

```
azureuser@k8s-master-21272034-0:~$ cat test.sh 
#!/bin/bash

testfunc() {
    if [[ "$1" == "hold" ]]; then
        echo "yes"
    else
        echo "no"
    fi
}

testfunc "hold"
testfunc hold
testfunc nothold
azureuser@k8s-master-21272034-0:~$ ./test.sh 
yes
yes
no
```

jackfrancis: Should we standardize to not using `"hold"` anywhere and prefer `hold`?

CecileRobertMichon: Yeah let's standardize to `hold` and `unhold`

parts/k8s/kubernetescustomscript.sh
```diff
@@ -179,7 +175,7 @@ if $REBOOTREQUIRED; then
   echo 'reboot required, rebooting node in 1 minute'
   /bin/bash -c "shutdown -r 1 &"
   if [[ $OS == $UBUNTU_OS_NAME ]]; then
-      holdWALinuxAgent "unhold"
+      aptmarkWALinuxAgent unhold
```
CecileRobertMichon: should this be a background process to return control to the user without waiting for it to finish?

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -107,6 +107,14 @@ write_files:
     RemainAfterExit=yes
     ExecStart=/usr/local/bin/health-monitor.sh container-runtime
 
+- path: /etc/apt/preferences
+  permissions: "0644"
+  owner: root
+  content: |
+    Package: walinuxagent
+    Pin: version 2.2.32.2
```
CecileRobertMichon: is that the version that is currently in VHD? or the version we are currently being upgraded to ?

jackfrancis: The version in the VHD

PR #785, opened 2019-03-18
 fix: rehabilitate constant generator script 

mboersma:
**Reason for Change**:

The script that generates `pkg/helpers/azureconst.go` would not compile or run, so this fixes it and updates it to Python code standards.

It also adds the new `southafricanorth` and `southafricawest` regions that were picked up the by generator script, and adjusts tests to match that and the added VM types.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tariq1890:
What are we trying to do here?

tariq1890:
This takes away critical support for the Azure ML/AI team. We need the `Standard PB6` SKUs to be present.

I have run the existing python script and it has worked.

mboersma:
> I have run the existing python script and it has worked.

I don't see how...

```
$ make generate-azure-constants
python pkg/helpers/Get-AzureConstants.py
Traceback (most recent call last):
  File "pkg/helpers/Get-AzureConstants.py", line 160, in <module>
    main()
  File "pkg/helpers/Get-AzureConstants.py", line 149, in main
    dcosMasterMap = getDcosMasterMap(allSizes)
NameError: global name 'allSizes' is not defined
make: *** [generate-azure-constants] Error 1
```

It appears we didn't quite finish refactoring it last fall. I fixed this, and then could not restrain myself from making the script actually look like Python.

> We need the Standard PB6 SKUs to be present.

I thought this might be a problem, so thanks for reviewing! Those aren't listed when run in my sub, but I'll restore them to the code and tests.

mboersma:
I restored the Standard_PBxx types and verified that `make generate-azure-constants` regenerates azureconst.go reliably with no diffs.

jackfrancis:
/lgtm

thanks for taking this on independently @mboersma!

PR #796, opened 2019-03-19
fix: Bug fixes for the HostedMasterProfile Scenario 

tariq1890:
i) Fixed a bug where unnecessary arm variables were being declared for HostedMasterProfiles
ii) Fixed a bug in virtualNetwork Subnet Address prefixes for HostedMasterProfiles
iii) Fixed a minor commandScript discrepancy in agent CSE VM Extensions


pkg/engine/vmextensions.go
```diff
@@ -132,7 +132,7 @@ func createAgentVMASCustomScriptExtension(cs *api.ContainerService, profile *api
...
+		commandExec := fmt.Sprintf("[concat('retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; %s for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),' GPU_NODE=%s SGX_NODE=%s /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1%s\"')]", outBoundCmd, nVidiaEnabled, sgxEnabled, runInBackground)
```
jackfrancis: does the removed space actually have any effect here?

tariq1890: It doesn't, but it brings us closer to creating an ARM template that matches those before the no_template refactor.

PR #797, opened 2019-03-19
chore: pull hyperkube and pause images from MCR

### LINKED ISSUE #122: Use azure k8s builds
We are now hosting kubernetes on MCR/Azure storage and should use these instead of the ones from gcr.

The hypercube images are available at `mcr.microsoft.com/k8s/core/hyperkube-amd64`

The list of tags currently available is:

|tags|
| -- | 
| v1.8.15-azure-c2bd642 |
| v1.8.14-azure-9d72faf |
| v1.9.11-azure-1bfeeb6 |
| v1.11.2-azure-bb9ffb1 |
| v1.10.9-azure-e6ab4ee |
| v1.11.3-azure-a452946 |
| v1.12.1-azure-4ed3216 |
| v1.9.10-azure-0985707 |
| v1.10.8-azure-7eab6a4 |
| v1.10.7-azure-0c38c36 |

Release tarballs are also available for Linux and Windows at `https://azurek8scishared.blob.core.windows.net/kubernetes-release/<version tag>`.
The tarballs don't yet exactly match upstream tarballs exactly, but the layout to get the binaries is the same.

cpuguy83:
**Reason for Change**:

This allows us to provide custom image repos/paths/tags for kubernetes components AND adds overrides to get images from MCR (for available images)

**Issue Fixed**:

Fixes #122 

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -376,7 +376,7 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
       HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
-      HYPERKUBE_URL="mcr.microsoft.com/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      HYPERKUBE_URL="upstream.azurecr.io/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
       CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
       pullContainerImage "docker" ${CONTAINER_IMAGE}
       echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -376,7 +376,7 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
       HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
-      HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      HYPERKUBE_URL="mcr.microsoft.com/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
       CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
       pullContainerImage "docker" ${CONTAINER_IMAGE}
       echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -83,7 +83,7 @@ var (
 		ContainerdDownloadURLBase:        "https://storage.googleapis.com/cri-containerd-release/",
 		ImageRepoOverrides: map[string]ImageRepoOverride{
 			"k8s.gcr.io/hyperkube-amd64": {
-				Registry: "mcr.microsoft.com",
+				Registry: "upstream.azurecr.io",
 				Repo:     "k8s/core/hyperkube-amd64",
 				// TODO: Setup tag overrides so we can map upstream versions to Azure k8s versions
 				// For now just using the upstream version works fine.
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -36,19 +36,20 @@ type DCOSSpecConfig struct {
 
 //KubernetesSpecConfig is the kubernetes container images used.
 type KubernetesSpecConfig struct {
-	KubernetesImageBase              string `json:"kubernetesImageBase,omitempty"`
-	TillerImageBase                  string `json:"tillerImageBase,omitempty"`
-	ACIConnectorImageBase            string `json:"aciConnectorImageBase,omitempty"`
-	NVIDIAImageBase                  string `json:"nvidiaImageBase,omitempty"`
-	AzureCNIImageBase                string `json:"azureCNIImageBase,omitempty"`
-	CalicoImageBase                  string `json:"CalicoImageBase,omitempty"`
-	EtcdDownloadURLBase              string `json:"etcdDownloadURLBase,omitempty"`
-	KubeBinariesSASURLBase           string `json:"kubeBinariesSASURLBase,omitempty"`
-	WindowsTelemetryGUID             string `json:"windowsTelemetryGUID,omitempty"`
-	CNIPluginsDownloadURL            string `json:"cniPluginsDownloadURL,omitempty"`
-	VnetCNILinuxPluginsDownloadURL   string `json:"vnetCNILinuxPluginsDownloadURL,omitempty"`
-	VnetCNIWindowsPluginsDownloadURL string `json:"vnetCNIWindowsPluginsDownloadURL,omitempty"`
-	ContainerdDownloadURLBase        string `json:"containerdDownloadURLBase,omitempty"`
+	KubernetesImageBase              string                       `json:"kubernetesImageBase,omitempty"`
+	TillerImageBase                  string                       `json:"tillerImageBase,omitempty"`
+	ACIConnectorImageBase            string                       `json:"aciConnectorImageBase,omitempty"`
+	NVIDIAImageBase                  string                       `json:"nvidiaImageBase,omitempty"`
+	AzureCNIImageBase                string                       `json:"azureCNIImageBase,omitempty"`
+	CalicoImageBase                  string                       `json:"CalicoImageBase,omitempty"`
+	EtcdDownloadURLBase              string                       `json:"etcdDownloadURLBase,omitempty"`
+	KubeBinariesSASURLBase           string                       `json:"kubeBinariesSASURLBase,omitempty"`
+	WindowsTelemetryGUID             string                       `json:"windowsTelemetryGUID,omitempty"`
+	CNIPluginsDownloadURL            string                       `json:"cniPluginsDownloadURL,omitempty"`
+	VnetCNILinuxPluginsDownloadURL   string                       `json:"vnetCNILinuxPluginsDownloadURL,omitempty"`
+	VnetCNIWindowsPluginsDownloadURL string                       `json:"vnetCNIWindowsPluginsDownloadURL,omitempty"`
+	ContainerdDownloadURLBase        string                       `json:"containerdDownloadURLBase,omitempty"`
+	ImageRepoOverrides               map[string]ImageRepoOverride `json:"imageRepoOverrides,omitempty"`
 }
 
 //AzureEndpointConfig describes an Azure endpoint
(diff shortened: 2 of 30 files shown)


CecileRobertMichon:
Does this PR replace #753 ?

cpuguy83:
It does.

cpuguy83:
Still a stray gcr pause image in the kubelet config.

packer/install-dependencies.sh
```diff
@@ -268,7 +268,7 @@ echo "  - busybox" >> ${RELEASE_NOTES_FILEPATH}
 K8S_VERSIONS="1.9.10 1.9.11 1.10.12 1.10.13 1.11.7 1.11.8 1.12.5 1.12.6 1.13.3 1.13.4"
 
 for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
-    HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+    HYPERKUBE_URL="mcr.microsoft.com/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
```
CecileRobertMichon: you'll want to add this new repo to `cleanUpContainerImages` in `kubernetesinstalls.sh`

cpuguy83:
I'll also note that I did not expose this new field on vlabs so a user can't (I'm assuming) actually set any overrides yet with this PR and it's only used internally to use mcr.

packer/install-dependencies.sh
cpuguy83: I *think* what's there is good, it's grepping on `hyperkube` and `KUBERNETS_VERSION`, but maybe I'm missing something.

pkg/api/azenvtypes.go
```diff
@@ -78,6 +79,22 @@ var (
...
+				Registry: "mcr.microsoft.com",
+				Repo:     "k8s/core/hyperkube-amd64",
+				// TODO: Setup tag overrides so we can map upstream versions to Azure k8s versions
+				// For now just using the upstream version works fine.
+			},
+			"pause": {
+				Registry: "mcr.microsoft.com",
+				Repo:     "k8s/core/pause",
+				Tags: map[string]string{
+					"3.1": "1.1.0",
```
cpuguy83: We probably want to support globbing for these keys, but maybe as a follow-up.

cpuguy83:
Unit tests for all the things are in.

pkg/api/defaults.go
```diff
@@ -105,7 +105,20 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpdate bool) {
 
 		if o.KubernetesConfig.KubernetesImageBase == "" {
 			o.KubernetesConfig.KubernetesImageBase = cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
+
+			// Only copy overrides when the user did not specify an image base for back-compat
+			//
+			// TODO: inject overrides when the user specified overrides but we have overrides
+			// that the user does not?
+			// TODO: Should we just always do this even if the user specified an image base?
```
cpuguy83: Looks like maybe on upgrade overrides are not applied because the derived model from the original deploy has "KubernetesImageBase" set?

packer/install-dependencies.sh
CecileRobertMichon: You're right, no change needed.

CecileRobertMichon:
Unit test job failed on `dep check`:

`# Gopkg.lock is out of sync:
github.com/opencontainers/go-digest: in Gopkg.lock's input-imports, but neither imported nor required`

pkg/api/defaults.go
cpuguy83: Just had a thought, maybe overrides are keyed by image reference instead of component name.
e.g. instead of `{"hyperkube": {// ovrerride }`, do `{"k8s.gcr.io/hyperkube-amd64": { // override }}`

This would require overrides to be more explicit and per cloud, but I think that's generally what we want?

CecileRobertMichon:
The VHD pipeline is failing because 1.14.0 isn't available.

CecileRobertMichon:
linter is failing the unit tests job

cpuguy83:
> --- FAIL: TestSetCustomCloudProfileDefaults (0.00s)

This is failing when the test does a `reflect.DeepEqual` with the custom config against some built in config. The issue is the built-in config has overrides and the custom config does not... it's not clear to me if I've missed something in code or if the test is now just not valid as is anymore.

CecileRobertMichon:
@cpuguy83 can you try rebasing? @honcao recently merged a change that changes `TestSetCustomCloudProfileDefaults`.

cpuguy83:
Same problem :(

cpuguy83:
Trying to rebase this... It's unfortunate there's more hacky code that has gone in here that's made it incredibly difficult to do, at least without writing even more hacky code.

I'm wondering if there's some way we can coalesce the KubernetsSpecConfig (from Azure env) with KubernetsConfig before we are trying to set components up.
I'm really a bit over my head with all of this, no idea what these types are being used for or why there is this split.

pkg/engine/params_k8s.go
```diff
@@ -10,14 +10,26 @@ import (
...
 
-	"github.com/Azure/go-autorest/autorest/to"
-
 	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/aks-engine/pkg/api/common"
+	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/pkg/errors"
 )
 
+var compToParamImage = map[string]string{
```
jadarsie: Tiny thing. I have the impression that `componentToParamImage` may be slightly easier to read. 
In my head, `comp` did not immediately trigger `component` and I think that `component` was favored across the code base.

jadarsie:
lgtm, @honcao could you please also double check?

jchauncey:
whats the status of this? Im waiting on it so i can complete the rest of the image move work.

packer/install-dependencies.sh
```diff
@@ -375,7 +375,7 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
       HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
-      HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      HYPERKUBE_URL="mcr.microsoft.com/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
```
jchauncey: We should decide if this should be the MCR registry or the new ACR registry. Technically its not supposed to live in MCR.

pkg/api/azenvtypes.go
```diff
@@ -80,6 +81,22 @@ var (
 		VnetCNILinuxPluginsDownloadURL:   "https://acs-mirror.azureedge.net/cni/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
 		VnetCNIWindowsPluginsDownloadURL: "https://acs-mirror.azureedge.net/cni/azure-vnet-cni-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
 		ContainerdDownloadURLBase:        "https://storage.googleapis.com/cri-containerd-release/",
+		ImageRepoOverrides: map[string]ImageRepoOverride{
+			"k8s.gcr.io/hyperkube-amd64": {
+				Registry: "mcr.microsoft.com",
```
jchauncey: See above comment about if this should be mcr or not.

pkg/api/azenvtypes.go
```diff
@@ -80,6 +81,22 @@ var (
...
+				Repo:     "k8s/core/hyperkube-amd64",
+				// TODO: Setup tag overrides so we can map upstream versions to Azure k8s versions
+				// For now just using the upstream version works fine.
+			},
+			"k8s.gcr.io/pause-amd64": {
+				Registry: "mcr.microsoft.com",
```
jchauncey: So Im leaning towards keeping this in mcr unless someone has a strong objection to that?

pkg/engine/params_k8s.go
jchauncey: I agree with this as well

pkg/api/azenvtypes.go
cpuguy83: Where should it go to? Are all the versions available there?

jchauncey: I have the following versions:

```
k8s/core/hyperkube-amd64:v1.12.6
k8s/core/hyperkube-amd64:v1.12.7
k8s/core/hyperkube-amd64:v1.13.5
k8s/core/hyperkube-amd64:v1.14.2
k8s/core/hyperkube-amd64:v1.15.0-beta.0
k8s/core/hyperkube-amd64:v1.16.0-alpha.0
```

jchauncey: I tried building earlier versions (1.11 and 1.10) but something has changed with how those are built so my CI job cant build those versions.

jchauncey: The full url would be `upstream.azurecr.io/k8s/core/hyperkube-amd64`

cpuguy83: Ok, I've updated. e2e tests will let us know if we are missing any required versions.

cpuguy83:
How do I get the real errors from these e2e tests?

CecileRobertMichon:
@cpuguy83 looking at https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=1146&view=logs&jobId=7277e1b3-ee82-57b0-6a9a-af833125b779 it looks like the VM extensions are failing with exit code 30 which means kubelet fails to start (see  https://github.com/Azure/aks-engine/blob/master/docs/howto/troubleshooting.md#vmextensionprovisioningerror-or-vmextensionprovisioningtimeout for help on troubleshooting these).

Since it's failing 100% of the time, it should be pretty easy to repro if you deploy a cluster from your branch. Kubelet logs are probably a good place to start to figure out what's going on...

CecileRobertMichon:
the VHD build is failing to pull upstream.azurecr.io/k8s/core/hyperkube-amd64:v1.14.1

```
 timeout 1200 docker pull upstream.azurecr.io/k8s/core/hyperkube-amd64:v1.14.1[0m
2019-05-21T21:40:59.2281221Z [0;32m    azure-arm: Error response from daemon: manifest for upstream.azurecr.io/k8s/core/hyperkube-amd64:v1.14.1 not found: manifest unknown: manifest unknown[0m
```

jchauncey:
We didnt have a 1.14.1 image just a 1.14.2. Im building 1.14.1 now. Retry in a bit

packer/install-dependencies.sh
```diff
@@ -376,7 +376,7 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
       HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
-      HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      HYPERKUBE_URL="upstream.azurecr.io/k8s/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
```
qike-ms: Just want to double check if we have all the versions we support today in MCR.

feiskyer: see the comment https://github.com/Azure/aks-engine/pull/797#discussion_r286144437, not all the versions are available. We may need to block this until all the AKS supported versions are pushed to MCR.

jchauncey:
Should we close this PR in favor of the ones @jackfrancis  has opened?

honcao:
@rjaini could you please review this change and verify that it works with Azure Stack.

jchauncey:
I would wait until we have word on whether we are going to abandon this PR in favor of the ones Jack has opened which provide a cleaner abstraction.

Issue #679, opened 2019-03-07
kube-proxy stuck in ContainerCreating


jackfrancis:
We're seeing this regularly across our E2E tests:

```
kube-system   kube-proxy-29jv2                                1/1     Running             0          32m     10.239.255.241   k8s-master-35441517-2   <none>           <none>
kube-system   kube-proxy-5zf7s                                1/1     Running             0          3m27s   10.239.2.35      k8s-agent2-35441517-1   <none>           <none>
kube-system   kube-proxy-996r5                                1/1     Running             0          11m     10.239.1.4       k8s-agent1-35441517-1   <none>           <none>
kube-system   kube-proxy-bx6gc                                0/1     ContainerCreating   0          46m     10.239.255.239   k8s-master-35441517-0   <none>           <none>
kube-system   kube-proxy-xsmrn                                1/1     Running             0          40m     10.239.255.240   k8s-master-35441517-1   <none>           <none>
```

See a single kube-proxy not able to be fully reconciled on a single node.

Pod logs aren't interesting:

```
2019/03/07 18:40:59 $ k logs kube-proxy-bx6gc -c kube-proxy -n kube-system
2019/03/07 18:40:59 #### $ k logs kube-proxy-bx6gc -c kube-proxy -n kube-system completed in 376.260953ms
2019/03/07 18:40:59 
Error from server (BadRequest): container "kube-proxy" in pod "kube-proxy-bx6gc" is waiting to start: ContainerCreating
```

jackfrancis:
@feiskyer @andyzhangx @ritazh have you seen the above symptoms in any upstream-originating k8s e2e tests?

feiskyer:
@jackfrancis Could you run kubectl describe for the pod? There may be something wrong when creating the pod sandbox (usually because of network plugins issues)

jackfrancis:
@feiskyer will do, see:

https://github.com/Azure/aks-engine/pull/691

Do pod sandbox creation errors due to network plugin issues reconcile themselves? What would be the expected remediation steps if we see that?

feiskyer:
It depends the errors. sometimes delete pod and let the controller recreate a new one may help.

jackfrancis:
@feiskyer here's a repro w/ kubectl describe pod:

```
2019/03/18 19:09:57 #### $ k describe pod kube-proxy-fhnc9 -n kube-system completed in 394.489276ms
2019/03/18 19:09:57 
Name:               kube-proxy-fhnc9
[...]

PR #799, opened 2019-03-20
[WIP] test: try to detect orphan pods stuck in ContainerCreating

jackfrancis:
**Reason for Change**:

Our E2E test runs regularly encounter "orphaned" pods after upgrade, especially kube-proxy. This issue outlines the symptoms:

https://github.com/Azure/aks-engine/issues/679

Essentially, after an upgrade, the "old" pod is left lying around, with a reference to a now non-existent daemonset. These E2E changes accommodate force cleanup so we can get clean test signal while we do root cause analysis as to why these orphaned pods are not properly cleaned up by cordon/drain.

See this thread for folks struggling to delete orphaned pods:

https://github.com/kubernetes/kubernetes/issues/65569

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -800,7 +800,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(pass).To(BeTrue())
 
 			By("Cleaning up after ourselves")
-			err = p.Delete(deleteResourceRetries)
+			err = p.Delete(deleteResourceRetries, false)
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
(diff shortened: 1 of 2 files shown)


jackfrancis:
@feiskyer let me know if you have any ideas why an old `kube-proxy` would be left around as an orphaned pod after upgrade. Ostensibly, the cordon/drain would have cleaned it up before the new node came online.

feiskyer:
If you are using the version >=1.12, then probably a bug of daemonset scheduler. the fix is here: https://github.com/kubernetes/kubernetes/pull/73401

jackfrancis:
@feiskyer will keep an eye out for which k8s versions this repros on.

Is that fix not yet in released versions of Kubernetes?

feiskyer:
1.12.6 and 1.13.4

jackfrancis:
Thanks... Hmm... the repro from earlier today manifested on an upgrade *from* 1.11.8 *to* 1.12.6. So perhaps that isn't the bug we're seeing.

PR #811, opened 2019-03-20
test: enforce --image-pull-policy=IfNotPresent for windows deployment…

jackfrancis:
…s in e2e tests

**Reason for Change**:

Anecdotal evidence suggests invoking kubectl run w/ `--image-pull-policy=IfNotPresent` is required to ensure not pulling the same images over and over again.

In the Windows "IIS scale out" test this saves about 5 mins, it seems.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
The default `imagePullPolicy` is supposed to be `IfNotPresent` already according to [k8s docs](https://kubernetes.io/docs/concepts/containers/images/#updating-images). But regardless, if this actually helps it should be a safe change.

mboersma:
lgtm

PR #812, opened 2019-03-20
fix: remove shell output from VHD release notes

CecileRobertMichon:
**Reason for Change**:

The output of the DevOps pipelines isn't always as serial as it should be, remove all shell commands that accidentally show up in the release notes artifacts.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer.mk
+++ b/packer.mk
@@ -20,4 +20,4 @@ generate-sas: az-login
 	az storage container generate-sas --name vhds --permissions lr --connection-string "${CLASSIC_SA_CONNECTION_STRING}" --start ${START_DATE} --expiry ${EXPIRY_DATE} | tr -d '"' | tee -a vhd-sas && cat vhd-sas
 
 make vhd-notes:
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
/hold

until I can verify desired outcome in vhd build

CecileRobertMichon:
verified

/hold cancel

jackfrancis:
I think we need to include a beginning of line regex, like `'^+`, right?

CecileRobertMichon:
> I think we need to include a beginning of line regex, like '^+, right?

Good point, right now we don't have any other "+" characters anywhere else but we might add some later.

It's not at the beginning of the line exactly since the lines start with four spaces but '    ^+.*' works

jackfrancis:
I think something like this is more resilient:

`'^[[:space:]]\{4\}\+'`

If we always know the first four chars are a space followed by a literal `x`

CecileRobertMichon:
I miscounted there are actually 5 spaces, but I applied your suggestion with 5 instead of 4.

CecileRobertMichon:
/hold

the last commit doesn't work

CecileRobertMichon:
/hold  cancel

verified the pipeline output is what we want

PR #816, opened 2019-03-21
fix: add more kubernetesConfig nil checks to account for more resilience

tariq1890:
More nil checks! One can never be too safe!

### DIFF
--- a/pkg/engine/masterarmresources.go
+++ b/pkg/engine/masterarmresources.go
@@ -39,7 +39,12 @@ func createKubernetesMasterResourcesVMAS(cs *api.ContainerService) []interface{}
 		masterResources = append(masterResources, createRouteTable())
 	}
 
-	isPrivateCluster := to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateCluster.Enabled)
+	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
+
+	var isPrivateCluster bool
+	if kubernetesConfig != nil && kubernetesConfig.PrivateCluster != nil {
+		isPrivateCluster = to.Bool(kubernetesConfig.PrivateCluster.Enabled)
+	}
 
 	if !isPrivateCluster {
 		publicIPAddress := CreatePublicIPAddress()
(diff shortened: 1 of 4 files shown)


pkg/engine/masterarmresources.go
```diff
@@ -52,14 +57,18 @@ func createKubernetesMasterResourcesVMAS(cs *api.ContainerService) []interface{}
...
+		}
 
 		if provisionJumpbox {
 			jumpboxVM := createJumpboxVirtualMachine(cs)
 			masterResources = append(masterResources, jumpboxVM)
 			jumpboxIsManagedDisks :=
-				cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateJumpboxProvision() &&
-					cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateCluster.JumpboxProfile.StorageProfile == api.ManagedDisks
+				kubernetesConfig.PrivateJumpboxProvision() &&
+					kubernetesConfig.PrivateCluster.JumpboxProfile.StorageProfile == api.ManagedDisks
```
jackfrancis: do we know that `JumpboxProfile` will never be nil here?

pkg/engine/virtualmachines.go
```diff
@@ -220,12 +226,15 @@ func createJumpboxVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
 		DataDisks: &[]compute.DataDisk{},
 	}
 
-	jumpBoxIsManagedDisks := cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateJumpboxProvision() && cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateCluster.JumpboxProfile.StorageProfile == api.ManagedDisks
+	var jumpBoxIsManagedDisks bool
+	if kubernetesConfig != nil && kubernetesConfig.PrivateCluster != nil {
+		jumpBoxIsManagedDisks = kubernetesConfig.PrivateJumpboxProvision() && kubernetesConfig.PrivateCluster.JumpboxProfile.StorageProfile == api.ManagedDisks
```
jackfrancis: ditto here

pkg/engine/masterarmresources.go
tariq1890: The `PrivateJumpboxProvision()` function checks for `JumpoboxProfile`. So we are covered that way.

pkg/engine/virtualmachines.go
tariq1890: same answer as above.

pkg/engine/virtualmachines.go
```diff
@@ -220,12 +226,15 @@ func createJumpboxVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
...
+	}
 
 	if jumpBoxIsManagedDisks {
 		storageProfile.OsDisk = &compute.OSDisk{
 			CreateOption: compute.DiskCreateOptionTypesFromImage,
-			DiskSizeGB:   to.Int32Ptr(int32(cs.Properties.OrchestratorProfile.KubernetesConfig.PrivateCluster.JumpboxProfile.OSDiskSizeGB)),
+			DiskSizeGB:   to.Int32Ptr(int32(kubernetesConfig.PrivateCluster.JumpboxProfile.OSDiskSizeGB)),
```
jackfrancis: ditto

jackfrancis:
/lgtm

pending unit tests

PR #822, opened 2019-03-21
fix: add extra nil guard checks for LinuxProfiles

tariq1890:
Is there such a thing called "too many nil checks"?


jackfrancis:
> Is there such a thing called "too many nil checks"?

Yes, some would call that golang.

jackfrancis:
/lgtm

pending E2E

PR #824, opened 2019-03-21
fix: add nil checks before accessing the service principal profile

tariq1890:
Holding onto dear life with nil checks


jackfrancis:
/lgtm

pending unit tests

PR #832, opened 2019-03-22
feat: AKS-Engine Install Script for Linux 

palma21:
**Reason for Change**:

Providing a simple install script to automatically pull and install into the Path the desired AKS Engine version. Defaults to latest
Also upgrades existing versions.

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests


scripts/get_akse.sh
```diff
@@ -0,0 +1,223 @@
+#!/usr/bin/env bash
+
+# Copyright Jorge Palma.
```
mboersma: We're supposed to have the Microsoft standard copyright header on each code file, which I think would include scripts as well, although we haven't been doing that. I'm not a lawyer, but I think asserting a different copyright might be problematic.

palma21: Yup, copy pasta foo. I technically think it should be alright but def not gonna dive into that. Changed.

mboersma:
lgtm

PR #857, opened 2019-03-22
fix: increase timeout tolerance for nvidia drivers installation

jackfrancis:
**Reason for Change**:

Increase the timeout to 10 mins for nvidia drivers installation.

We have suggestions that nvidia drivers installation is being interrupted before it has time to finish:

```
+ timeout 120 sh /usr/local/nvidia/nvidia-drivers-396.26 --silent --accept-license --no-drm --dkms --utility-prefix=/usr/local/nvidia --opengl-prefix=/usr/local/nvidia
Verifying archive integrity... OK
Uncompressing NVIDIA Accelerated Graphics Driver for Linux-x86_64 396.26.................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................

WARNING: The nvidia-drm module will not be installed. As a result, DRM-KMS
         will not function with this installation of the NVIDIA driver.


WARNING: nvidia-installer was forced to guess the X library path '/usr/lib'
         and X module path '/usr/lib/xorg/modules'; these paths were not
         queryable from the system.  If X fails to find the NVIDIA X driver
         module, please install the `pkg-config` utility and the X.Org
         SDK/development package for your distribution and reinstall the
         driver.

Received signal SIGTERM; aborting.
/usr/local/nvidia/nvidia-drivers-396.26: 1: /usr/local/nvidia/nvidia-drivers-396.26: cannot create /dev/tty: No such device or address
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -408,7 +408,7 @@ configAddons() {
 }
 
 configGPUDrivers() {
-    retrycmd_if_failure 30 1 120 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
+    retrycmd_if_failure 30 1 1200 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
     echo "${GPU_DEST}/lib64" > /etc/ld.so.conf.d/nvidia.conf
     retrycmd_if_failure 120 5 25 ldconfig || exit $ERR_GPU_DRIVERS_START_FAIL
     umount -l /usr/lib/x86_64-linux-gnu
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -408,7 +408,7 @@ configAddons() {
 }
 
 configGPUDrivers() {
-    retrycmd_if_failure 3 1 1200 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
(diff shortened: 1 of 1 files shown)


mboersma:
lgtm

parts/k8s/kubernetesconfigs.sh
```diff
@@ -408,7 +408,7 @@ configAddons() {
 }
 
 configGPUDrivers() {
-    retrycmd_if_failure 30 1 120 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
+    retrycmd_if_failure 30 1 1200 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
```
CecileRobertMichon: this is too aggressive and will cause CSE timeouts (+ 90 minutes) on the ARM deployment if it actually fails every time.

jackfrancis: Good point, what do you think about my latest commit that reduces the retry count to 3?

parts/k8s/kubernetesconfigs.sh
```diff
@@ -408,7 +408,7 @@ configAddons() {
 }
 
 configGPUDrivers() {
-    retrycmd_if_failure 30 1 120 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
+    retrycmd_if_failure 3 1 1200 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit $ERR_GPU_DRIVERS_START_FAIL
```
CecileRobertMichon: that's still 1 hour but should be okay for ARM timeout. Can the install really take up to 20 minutes?

jackfrancis: See the output above proving that 2 mins is too short. Would increasing timeout to a lower (but higher than 2) minute interval be better?

CecileRobertMichon: how about 10 minutes?

CecileRobertMichon:
We should monitor GPU CSE success rate after this change since it's cutting down the number of retries (in favor of a longer timeout). If the success rate go down that might suggest it's preferable to go back to more retries (with shorted timeout).

PR #862, opened 2019-03-22
fix: fix sshd logging to auth.log for OMS

CecileRobertMichon:
**Reason for Change**:

https://icm.ad.msft.net/imp/v3/incidents/details/108362084/home
OMS agent relies on auth.log for getting ssh logs for security monitoring purposes. https://docs.microsoft.com/en-us/azure/log-analytics/log-analytics-data-sources-syslog#configure-syslog-on-linux-agent
https://github.com/Azure/acs-engine/pull/1956 disabled `ForwardToSyslog` which broke this scenario. See for additional context: https://github.com/Azure/acs-engine/pull/1956#issuecomment-382822102

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesconfigs.sh
+++ b/parts/k8s/kubernetesconfigs.sh
@@ -288,7 +288,7 @@ ensureJournal(){
     echo "Storage=persistent" >> /etc/systemd/journald.conf
     echo "SystemMaxUse=1G" >> /etc/systemd/journald.conf
     echo "RuntimeMaxUse=1G" >> /etc/systemd/journald.conf
-    echo "ForwardToSyslog=no" >> /etc/systemd/journald.conf
+    echo "ForwardToSyslog=yes" >> /etc/systemd/journald.conf
     systemctlEnableAndStart systemd-journald || exit $ERR_SYSTEMCTL_START_FAIL
 }
 


CecileRobertMichon:
@JunSun17 @mpalumbo7 FYI

@jackfrancis @feiskyer please speak up if you know of any reasons that we shouldn't make this change.

jackfrancis:
The argument against making this change is the additional I/O (and arguably CPU/mem, though probably not significant) overhead required to stream journald to syslog. @feiskyer @khenidak @cpuguy83 @cchildress any thoughts?

JunSun17:
Thanks for fixing this! So does this only impact auth.log or all other logs? Just not sure how much other impact it might bring beside bring back auth.log.

CecileRobertMichon:
Regarding extra I/O concern, I think it's a tradeoff. This is the only way to restore OMS security monitoring AFAIK. I don't think the extra logs would be significant I/O overhead but they would be the price to pay for auth logging. FWIW, `ForwardToSyslog` defaults to `yes` on vanilla Azure Ubuntu VMs so we would be going back to the default behavior.

PR #863, opened 2019-03-22
style: remove unused constants

mboersma:
**Reason for Change**:

Removes unused constants after the template refactoring was merged.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I thought we had the "unused" linter enabled so that `make test-style` would find this in CI. I will investigate.


CecileRobertMichon:
lgtm

PR #870, opened 2019-03-25
perf: improve CSE script

### LINKED ISSUE #869: Disable unattended-upgrades at first boot
When using a VMSS, if you scale down and up a lot the tuning of the provision script is important.

The script has a lot of failsafe implemented, mostly to deal with `apt` not being concurrent and the fact that this script is ran at the same time a unattended-upgrades (which is ran a each boot).

I think to avoid deadlocks, possibility of failures and decrease the time of the provision script, `unattended-upgrades` should be disabled and re-enabled at the end of the provision script.

I'll send a patch.

sylr:
- Remove apt locks with unattended-upgrade
- Upgrade OS at first boot
- Improve CSE execution duration

Fixes #869

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -52,6 +52,26 @@ write_files:
   content: !!binary |
     {{WrapAsVariable "systemConf"}}
 
+- path: /etc/apt/apt.conf.d/10periodic
+  permissions: "0644"
+  owner: root
+  content: |
+    APT::Periodic::Update-Package-Lists "1";
+    APT::Periodic::Download-Upgradeable-Packages "1";
+    APT::Periodic::AutocleanInterval "7";
+    APT::Periodic::Unattended-Upgrade "1";
+
+- path: /etc/apt/apt.conf.d/99periodic
+  permissions: "0644"
+  owner: root
+  content: |
+    // Disable unattended upgrades at first boot
+    // This file must be removed by the provision script
+    APT::Periodic::Update-Package-Lists "0";
+    APT::Periodic::Download-Upgradeable-Packages "0";
+    APT::Periodic::AutocleanInterval "0";
+    APT::Periodic::Unattended-Upgrade "0";
+
 - path: /usr/local/bin/health-monitor.sh
   permissions: "0544"
   encoding: gzip
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -161,6 +161,9 @@ if $FULL_INSTALL_REQUIRED; then
     fi
 fi
 
+# Re-enable unattended-upgrades
+rm -f /etc/apt/apt.conf.d/99periodic
+
 echo "Custom script finished successfully"
 
 echo `date`,`hostname`, endcustomscript>>/opt/m
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -43,12 +43,6 @@ if [[ $OS == $COREOS_OS_NAME ]]; then
     KUBECTL=/opt/kubectl
 fi
 
-if [ -f /var/run/reboot-required ]; then
-    REBOOTREQUIRED=true
-else
-    REBOOTREQUIRED=false
-fi
-
 if [ -f /var/log.vhd/azure/golden-image-install.complete ]; then
     echo "detected golden image pre-install"
     FULL_INSTALL_REQUIRED=false
@@ -56,6 +50,8 @@ if [ -f /var/log.vhd/azure/golden-image-install.complete ]; then
 else
     FULL_INSTALL_REQUIRED=true
 fi
+
+upgradeOs
 
 if [[ ! -z "${MASTER_NODE}" ]] && [[ -z "${COSMOS_URI}" ]]; then
     	installEtcd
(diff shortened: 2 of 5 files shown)


sylr:
```admin@agentzone2-vmss-000000:~$ cat /opt/m
Mon Mar 25 14:44:11 UTC 2019,agentzone2-vmss-000000, startcustomscript
Mon Mar 25 14:45:19 UTC 2019,agentzone2-vmss-000000, endcustomscript
```

CecileRobertMichon:
This fixes #869 correct?

sylr:
@CecileRobertMichon yes

parts/k8s/kubernetesagentcustomdata.yml
```diff
@@ -52,6 +52,26 @@ write_files:
   content: !!binary |
     {{WrapAsVariable "systemConf"}}
 
+- path: /etc/apt/apt.conf.d/10periodic
```
CecileRobertMichon: Should the same files be added to master custom data?

CecileRobertMichon:
https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=223&view=logs E2E passed but artifact upload failed. We can consider pr-e2e green ✅

parts/k8s/kubernetesagentcustomdata.yml
sylr: Indeed

mtparet:
Nice! Does it apply also for workers nodes ?
Each time we upgraded, VMs needed reboot immediately after showing up..

CecileRobertMichon:
@sylr the E2E tests should be fixed now, can you please rebase?

sylr:
Done.

CecileRobertMichon:
@sylr you have a merge conflict

CecileRobertMichon:
@jackfrancis for review

parts/k8s/kubernetescustomscript.sh
```diff
@@ -171,12 +170,18 @@ if ! $FULL_INSTALL_REQUIRED; then
...
+else
+    REBOOTREQUIRED=false
+fi
+
 if $REBOOTREQUIRED; then
   echo 'reboot required, rebooting node in 1 minute'
-  /bin/bash -c "shutdown -r 1 &"
   if [[ $OS == $UBUNTU_OS_NAME ]]; then
-      aptmarkWALinuxAgent unhold &
+      aptmarkWALinuxAgent unhold
```
jackfrancis: The reason we were sending this to the background is to accelerate a CSE return code (and thus speed up deployments). Because unholding the walinux mark requires the apt locks to be released, it could take several minutes to reconcile. We don't want to block CSE response on that.

parts/k8s/kubernetescustomscript.sh
```diff
@@ -171,12 +170,18 @@ if ! $FULL_INSTALL_REQUIRED; then
...
   fi
+  nohup /bin/bash -c "sleep 3 && reboot" &
```
jackfrancis: Does an `apt-mark hold` persist reboots? If it doesn't, we don't need to worry about the fact that we're potentially rebooting before that completes.

parts/k8s/kubernetesinstalls.sh
```diff
@@ -16,6 +16,11 @@ removeMoby() {
     sudo apt-get purge -y moby-engine moby-cli
 }
 
+upgradeOs() {
+    apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
+    apt_get_dist_upgrade
```
jackfrancis: We would want to add `|| exit $ERR_APT_UPDATE_TIMEOUT` to the dist-upgrade as well.

parts/k8s/kubernetescustomscript.sh
sylr: There are no more apt locks since I disabled unattended-upgrades and because I decreased to time before the reboot from 1m to 3s I want to make sure this is correctly executed before the reboot occurs.

jackfrancis: Got it. That makes sense. Except... we've seen instances where cloud-init initiates a background apt operation. That could result in apt locks persisting, whcih might delay apt-mark unhold for a while. Let's put some mileage on this change and compare deployment times before/after. I think the change is ultimately an improvement, just wanna be cautious before introducing additional serial/blocking operations before exiting CSE.

sylr: In any case we got to unhold before the reboot and like it's the last apt thing before reboot it does not make sens to background it.

parts/k8s/kubernetesinstalls.sh
sylr: Do we really want to fail the entire script if this fails ?

sylr:
```
[Fail] ‌Azure Container Cluster using the Kubernetes Orchestrator ‌after the cluster has been up for awhile ‌[It] dns-liveness pod should not have any restarts
```

I believe this fails because of the reboot of the node after the dist-upgrade which results on having the pod being restarted.

parts/k8s/kubernetesinstalls.sh
jackfrancis: Yes, because we want everything in CSE to be 100% deterministic + successful, and we want to know if certain steps don't meet that criteria. If this is a "run in the background w/ best effort" kind of this we should move it out of CSE and into some other area of enforcement.

sylr: All right.

CecileRobertMichon:
lgtm

jackfrancis:
thx again @sylr!

PR #872, opened 2019-03-25
feat: add support for Kubernetes 1.12.7

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#v1127

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/versions.go
```diff
@@ -98,8 +98,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.12.1":         false,
 	"1.12.2":         false,
 	"1.12.4":         false,
-	"1.12.5":         true,
+	"1.12.5":         false,
 	"1.12.6":         true,
+	"1.12.7":         true,
```
CecileRobertMichon: I think you forgot to add it to the vhd script

jackfrancis: Done @mboersma FYI

PR #873, opened 2019-03-25
feat: add support for Kubernetes 1.11.9

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#v1119

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
`TestK8sVars` needs some attention—I'll do that next.

pkg/engine/armvariables_test.go
```diff
@@ -119,13 +119,13 @@ func TestK8sVars(t *testing.T) {
 		"mountetcdScript":                           GetKubernetesB64Mountetcd(),
 		"nsgID":                                     "[resourceId('Microsoft.Network/networkSecurityGroups',variables('nsgName'))]",
 		"nsgName":                                   "[concat(variables('masterVMNamePrefix'), 'nsg')]",
-		"orchestratorNameVersionTag":                "Kubernetes:1.11.8",
+		"orchestratorNameVersionTag":                "Kubernetes:1.11.9",
```
CecileRobertMichon: are we going to have to update these tests every time there's a new default k8s version patch? Is there any way we can make them expect the latest patch instead of hardcoding to a specific patch?

mboersma: I'm sure there is a way. I'll enter a ticket to refactor these tests so we're not chasing 1.11.x patches.

PR #876, opened 2019-03-25
feat: add support for Kubernetes 1.14.0

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
I'll rebase this after #873 merges and then run tests.

mboersma:
The 1.14 [known issues](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#known-issues) mentions this:
> There is a known issue coredns/coredns#2629 in CoreDNS 1.3.1, wherein if the Kubernetes API shuts down while CoreDNS is connected, CoreDNS will crash. The issue is fixed in CoreDNS 1.4.0 in coredns/coredns#2529.

But the Docker image `k8s.gcr.io/coredns:1.4.0` is not available, only `k8s.gcr.io/coredns.1.3.1` as is mentioned elsewhere throughout the release notes. So we'll use that.

mboersma:
There was [one failure](https://github.com/Azure/aks-engine/pull/876/checks?check_run_id=84689382) on the first run. I'm rebasing from master now.

PR #877, opened 2019-03-25
chore: Use moby for GPU instances

### LINKED ISSUE #534: Enable nvidia drivers w/ moby
nvidia drivers currently have a hard dependency on docker-engine. Let's get this removed so they can run on moby-backed nodes.

cpuguy83:
This was stuck on docker 1.13 due to packaging issues with nvidia-docker
(which has a hard dep on docker specific package versions).

This change manually installs nvidia-docker2, but uses apt to install
the rest of the nvidia-* packages which don't have this hard dep.

nvidia-docker2 is really just a docker CLI clone (not sure if there are
changes to it) and a custom `/etc/docker/daemon.json` which sets the
default runtime to nvidia-container-runtime.

**Reason for Change**:

**Issue Fixed**:

Fixes #534

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -52,13 +52,14 @@ if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
     installDockerEngine
     overrideDockerEngineStorageDriver
     echo "  - docker-engine v${DOCKER_ENGINE_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-    installGPUDrivers
-    echo "  - nvidia-docker2 nvidia-container-runtime" >> ${RELEASE_NOTES_FILEPATH}
 else
     MOBY_VERSION="3.0.4"
     installMoby
     echo "  - moby v${MOBY_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
 fi
+
+installGPUDrivers
+echo "  - nvidia-docker2 nvidia-container-runtime" >> ${RELEASE_NOTES_FILEPATH}
 
 installClearContainersRuntime
 
--- a/parts/k8s/kubernetesinstalls.sh
+++ b/parts/k8s/kubernetesinstalls.sh
@@ -16,6 +16,10 @@ removeMoby() {
     sudo apt-get purge -y moby-engine moby-cli
 }
 
+removeDockerEngine() {
+    apt_get_remove 20 30 120 docker-engine
+}
+
 installEtcd() {
     CURRENT_VERSION=$(etcd --version | grep "etcd Version" | cut -d ":" -f 2 | tr -d '[:space:]')
     if [[ "$CURRENT_VERSION" == "${ETCD_VERSION}" ]]; then
@@ -48,7 +52,29 @@ installGPUDrivers() {
     retrycmd_if_failure_no_stats 120 5 25 cat /tmp/nvidia-docker.list > /etc/apt/sources.list.d/nvidia-docker.list
     apt_get_update
     retrycmd_if_failure 30 5 3600 apt-get install -y linux-headers-$(uname -r) gcc make dkms || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-docker2=${NVIDIA_DOCKER_VERSION}+docker${DOCKER_VERSION} nvidia-container-runtime=${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker${DOCKER_VERSION} || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+
+    # only install the runtime since nvidia-docker2 has a hard dep on docker CE packages.
+    # we will manually install nvidia-docker2
+    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker18.09.2-1" || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+
+    tmpDir=$(mktmp -d)
+    (
+      set -e -o pipefail
+      cd "${tmpDir}"
+
+      retrycmd_if_failure 30 5 3600 apt-get download nvidia-docker2="${NVIDIA_DOCKER_VERSION}+docker18.09.2-1"
+
+      dpkg-deb -R "nvidia-docker2*.deb" "${tmpDir}/pkg"
+      cp -r "${tmpDir}/pkg/usr/*" /usr/
+    )
+    status=$?
+
+    rm -rf "${tmpDir}"
+    if [ ! ${status} -eq 0 ]; then
+      exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+    fi
+
+
     retrycmd_if_failure 120 5 25 pkill -SIGHUP dockerd || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
     retrycmd_if_failure 30 5 60 curl -fLS https://us.download.nvidia.com/tesla/$GPU_DV/NVIDIA-Linux-x86_64-${GPU_DV}.run -o ${GPU_DEST}/nvidia-drivers-${GPU_DV} || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
     mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
@@ -104,6 +130,7 @@ installMoby() {
     if [[ "$CURRENT_VERSION" == "${MOBY_VERSION}" ]]; then
         echo "dockerd $MOBY_VERSION is already installed, skipping Moby download"
     else
+        removeDockerEngine || true
         removeMoby
         retrycmd_if_failure_no_stats 120 5 25 curl https://packages.microsoft.com/config/ubuntu/${UBUNTU_RELEASE}/prod.list > /tmp/microsoft-prod.list || exit $ERR_MOBY_APT_LIST_TIMEOUT
         retrycmd_if_failure 10 5 10 cp /tmp/microsoft-prod.list /etc/apt/sources.list.d/ || exit $ERR_MOBY_APT_LIST_TIMEOUT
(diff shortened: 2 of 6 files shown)


packer/install-dependencies.sh
```diff
@@ -47,17 +47,16 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
...
+MOBY_VERSION="3.0.4"
+installMoby
+echo "  - moby v${MOBY_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
+
 if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
-    DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
-    installDockerEngine
     overrideDockerEngineStorageDriver
     echo "  - docker-engine v${DOCKER_ENGINE_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
     installGPUDrivers
```
CecileRobertMichon: Why are we only installing GPU drivers with the docker-engine feature flag if this PR is to "use moby for GPU instances"?

parts/k8s/kubernetesinstalls.sh
```diff
@@ -86,11 +113,7 @@ installSGXDrivers() {
 
 installContainerRuntime() {
     if [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
-        if [[ "$DOCKER_ENGINE_REPO" != "" ]]; then
```
CecileRobertMichon: I think we still want to allow a user to specify docker-engine

parts/k8s/kubernetesinstalls.sh
```diff
@@ -16,6 +16,10 @@ removeMoby() {
     sudo apt-get purge -y moby-engine moby-cli
 }
 
+removeDockerEngine() {
+	apt_get_remove 20 30 120 docker-engine
```
CecileRobertMichon: Should `removeMoby` and `removeDockerEngine` be using the same method for uninstall?

indent looks off

packer/install-dependencies.sh
```diff
@@ -47,17 +47,16 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
...
+
+MOBY_VERSION="3.0.4"
+installMoby
+echo "  - moby v${MOBY_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
+
 if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
-    DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
-    installDockerEngine
     overrideDockerEngineStorageDriver
     echo "  - docker-engine v${DOCKER_ENGINE_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
```
CecileRobertMichon: if we're not installing dockerEngine anymore this log line isn't true

parts/k8s/kubernetesinstalls.sh
CecileRobertMichon: https://github.com/Azure/aks-engine/blob/master/pkg/engine/params_k8s.go#L240 is what you want to change

CecileRobertMichon:
docs: https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#L537 needs to be changed

CecileRobertMichon:
validation: https://github.com/Azure/aks-engine/blob/master/pkg/api/vlabs/validate.go#L780 needs to be updated

CecileRobertMichon:
tests: https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults_test.go#L1180 needs to be updated

jackfrancis:
This is great, thanks @cpuguy83!

Here's a proposal: let's keep the large majority of stuff unchanged, and simply evolve the nvidia installation script so that it is installable whether docker-engine or moby are present?

We can then test our way into (1) converting N series cluster configurations into being moby-backed by default and (2) deprecate the docker-engine VHD.

Basicaly, if all we need to do is the manual apt stuff in order to get nvidia drivers working w/ out docker-engine being present, let's restrict this PR to just that. All the other stuff we can do incrementally in the future.

@CecileRobertMichon thoughts?

parts/k8s/kubernetesinstalls.sh
```diff
@@ -48,7 +52,30 @@ installGPUDrivers() {
     retrycmd_if_failure_no_stats 120 5 25 cat /tmp/nvidia-docker.list > /etc/apt/sources.list.d/nvidia-docker.list
     apt_get_update
     retrycmd_if_failure 30 5 3600 apt-get install -y linux-headers-$(uname -r) gcc make dkms || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-docker2=${NVIDIA_DOCKER_VERSION}+docker${DOCKER_VERSION} nvidia-container-runtime=${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker${DOCKER_VERSION} || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+
+    # only install the runtime since nvidia-docker2 has a hard dep on docker CE packages.
+    # we will manually install nvidia-docker2
+    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker18.09.2-1" || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
```
tariq1890:
```suggestion
    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker18.09.2-1" || exit ${ERR_GPU_DRIVERS_INSTALL_TIMEOUT}
```

parts/k8s/kubernetesinstalls.sh
```diff
@@ -48,7 +52,30 @@ installGPUDrivers() {
...
+
+      dpkg-deb -R "nvidia-docker2*.deb" "${tmpDir}/pkg"
+      rm -rf "${tmpDir}/pkg/DEBIAN"
+      cp -r "${tmpDir}/pkg/*" /
+    )
+    status=$?
+
+    rm -rf "${tmpDir}"
+    if [ ! ${status} -eq 0 ]; then
+      exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
```
tariq1890:
```suggestion
      exit ${ERR_GPU_DRIVERS_INSTALL_TIMEOUT}
```

parts/k8s/kubernetesinstalls.sh
```diff
@@ -48,7 +52,30 @@ installGPUDrivers() {
...
+      retrycmd_if_failure 30 5 3600 apt-get download nvidia-docker2="${NVIDIA_DOCKER_VERSION}+docker18.09.2-1"
+
+      dpkg-deb -R "nvidia-docker2*.deb" "${tmpDir}/pkg"
+      rm -rf "${tmpDir}/pkg/DEBIAN"
+      cp -r "${tmpDir}/pkg/*" /
+    )
+    status=$?
+
+    rm -rf "${tmpDir}"
+    if [ ! ${status} -eq 0 ]; then
```
tariq1890:
```suggestion
    if [[ ! ${status} -eq 0 ]]; then
```

tariq1890:
@jackfrancis @cpuguy83  Can we change the line 137 in `kubernetesinstalls.sh` so that moby-cli picks up `${MOBY-VERSION}` instead of a hardcoded version?

This is assuming we have v3.0.4 published for moby-cli

CecileRobertMichon:
the squash commit message needs to be updated to use `chore:` instead of `[chore] `

parts/k8s/kubernetesinstalls.sh
cpuguy83: No need for the double bracket here.

cpuguy83:
@tariq1890 The CLI has a different release cycle from the engine, I don't think we want to mix versions here.

packer/install-dependencies.sh
cpuguy83: Seems like this was a bad assumption before.
Is there some env var that we can look at for GPU instances?

CecileRobertMichon: This the script that we run in the pipeline that builds our VHDs, it doesn't run at provisioning. Packer runs the script to install a bunch of software on a VM. There is no notion of GPU/not GPU instance at that stage. The env var that you are looking for that tells if a VM is a GPU node at provisioning is in kubernetescustomscript.sh

parts/k8s/kubernetesinstalls.sh
cpuguy83: Possibly. The reason I did not is `purge` was removing `/etc/docker/daemon.json`, I'm not sure why it wasn't replaced later on... but in any case it caused the deployment to fail, timing out on `[ -f /etc/docker/daemon.json ]` to return ok.

packer/install-dependencies.sh
cpuguy83: So we definitely don't want to install GPU drivers unless it's a GPU enabled node as this changes how containers are executed.
I guess we could modify the installation to only install the updated `/etc/docker/daemon.json` (which is what modifies the container runtime to use nvidia-docker's runtime) at at provision time.

What do you think? Any other ideas?

CecileRobertMichon: There are two different things happening:

1) kubernetesinstalls: downloads all necessary software, this is called by both the VHD packer script (install-dependencies.sh) and CSE for VMs not using the VHD image (base Ubuntu image)
2) kubernetesconfigs: runtime configurations + uninstall software that isn't needed for a particular VM instance

We only run `ensureGPUDrivers`  https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesconfigs.sh#L420 on GPU VMs at provisioning. 

We should pull all the GPU drivers in the VHD and only enable the nvidia-runtime if the node is a GPU instance at provisioning. Also, we should uninstall the GPU drivers if the node is NOT a GPU instance. Currently, we don't do this because we have two different VHD flavors: Moby for non-GPU and docker-engine for GPU nodes. In the future, both GPU and non-GPU nodes will build from the same Moby VHD so we need to add extra uninstalling of the nvidia-drivers in CSE when the node doesn't need them.

Keep in mind that all of this also needs to work for clusters that opt out of the VHD and build on top of the Vanilla Ubuntu image (this is a requirement for sovereign clouds that don't support our base image). So the full GPU flow needs to be functional installing when everything during CSE.

jackfrancis:
closing in favor of #897, which will incorporate the needful business logic into a PR that actually retires docker-engine scenarios entirely

PR #880, opened 2019-03-25
fix: use --dns-prefix for output directory when running deploy command

### LINKED ISSUE #597: Incorrect output directory while using aks-engine deploy with --dns-prefix specified
NO
---

ISSUE
---

0.31.1
---

**Kubernetes version**:
1.13.x

**What happened**:
While using aks-engine deploy with --dns-prefix specified, files are generated to _output instead of  _output/DNSPrefix

Files are generated to  _output/DNSPrefix while using aks-engine deploy with --dns-prefix specified

 Deploy cluster with command aks-engine deploy --dns-prefix xxx, check the generated output folder.

I checked the code in deploy.go, dc.outputDirectory is set before  dc.containerService.Properties.MasterProfile.DNSPrefix gets correct value.
Line 192
`dc.outputDirectory = path.Join("_output", dc.containerService.Properties.MasterProfile.DNSPrefix)`
Line 261
`dc.containerService.Properties.MasterProfile.DNSPrefix = dc.dnsPrefix`

alexeldeib:
**Reason for Change**:

**Issue Fixed**:

Fixes #597 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


alexeldeib:
@CecileRobertMichon I wanted to insert a test case to check that outputDirectory is correct after autofillApiModel when providing the flag, to do so it looks like I would need to stub out `apiloader.LoadContainerServiceFromFile` to return one of the static configs. Does that sound okay to you, or should I leave this change as-is? 

Sorry for picking on you, I saw you tagged the original issue 😄

CecileRobertMichon:
@alexeldeib you could use `../pkg/engine/testdata` for static test files the way https://github.com/Azure/aks-engine/blob/master/cmd/deploy_test.go#L508 does it. Would that solve it? Not sure I understand which part you need `apiloader.LoadContainerServiceFromFile` for, could you give me an example?

alexeldeib:
The value was set twice, once during `loadApiModel` and once during `autofill`. I just wanted a simple test to do run load + autofill and check the value at the end to catch regressions. In the existing tests I see `getApiModel` helps stub out the load functionality, but that short circuits the scenario which would have caused the issue. I wanted to make loadApiModel a bit more testable, I was thinking to stub `apiloader.LoadContainerServiceFromFile` since this is the portion which does the actual file read.

The corresponding test case would be fairly trivial, something like: 
```
func TestOutputDirectoryWithDNSPrefix(t *testing.T) {
	apiloader := &api.Apiloader{
		Translator: nil,
	}

	apimodel := getAPIModel(ExampleAPIModel, false, "", "")

	cs, ver, err := apiloader.DeserializeContainerService([]byte(apimodel), false, false, nil)
	if err != nil {
		t.Fatalf("unexpected error deserializing the example apimodel: %s", err)
	}
	deployCmd := &deployCmd{
		apimodelPath:     "./this/is/unused.json",
		outputDirectory:  "",
		dnsPrefix:        "dnsPrefix1",
		forceOverwrite:   true,
		location:         "westus",
		containerService: cs,
		apiVersion:       ver,
		client:           &armhelpers.MockAKSEngineClient{},
		authProvider: &mockAuthProvider{
			authArgs: &authArgs{},
		},
	}

	err = d.loadAPIModel(r, []string{})
	if err != nil {
		t.Fatalf("Failed to call LoadAPIModel: %s", err)
	}
	
	err = autofillApimodel(deployCmd)

	if err != nil {
		t.Fatalf("unexpected error autofilling the example apimodel: %s", err)
	}

	if deployCmd.outputDirectory != path.Join("_output", deployCmd.dnsPrefix) {
		t.Fatalf("Calculated output directory should be _output/{{dnsPrefix}}, actual value %s", deployCmd.outputDirectory)
	}
}
```

Don't want to bike shed a fairly minor change -- thought I might be missing a simple fix

alexeldeib:
Oh -- this is definitely already being done and I misunderstood, nevermind 😊 

```
func TestDeployCmdMergeAPIModel(t *testing.T) {
	d := &deployCmd{}
	d.apimodelPath = "../pkg/engine/testdata/simple/kubernetes.json"
	err := d.mergeAPIModel()
	if err != nil {
		t.Fatalf("unexpected error calling mergeAPIModel with no --set flag defined: %s", err.Error())
	}
...
}
```
```
        d.apimodelPath = "../pkg/engine/testdata/simple/kubernetes.json"
        ...
	err = d.loadAPIModel(r, []string{})
	if err != nil {
		t.Fatalf("Failed to call LoadAPIModel: %s", err)
	}
```

alexeldeib:
@CecileRobertMichon Took me a while but finally added a test for that case 😊 Was trying to call `autofill` twice and shooting myself in the foot. 

Let me know if you'd like any additional changes here!

alexeldeib:
Going to try to re-run that since I snuck in one more commit -- not sure if i'm able...

CecileRobertMichon:
@alexeldeib the new testdata file is causing a unit test error:
`    engine_test.go:53: Loading file testdata/emptyDnsPrefix/kubernetes.json got error: missing Properties.MasterProfile.DNSPrefix`

You can use `make test` to run unit tests locally.

alexeldeib:
Sorry, I thought I did but I guess I didn't pull in the correct changeset.

I see in [pkg/api/vlabs/types.go](https://github.com/Azure/aks-engine/blob/master/pkg/api/types.go), DNSPrefix field is defined:
```
DNSPrefix                string            `json:"dnsPrefix" validate:"required"`
```
but in `pkg/api/types.go` and several of the other version specific packages:
```
DNSPrefix                string            `json:"dnsPrefix"`
```
Looks like the unit test failures are validation failures from the `` `validate:"required"` ``, which doesn't seem to be either correct or necessary? Is there a reason this is enforced on the vlabs version?

alexeldeib:
Moreover, it seems like the current unit tests prevent the correct usage of leaving `dnsPrefix` empty in apiModel and populating it via CLI? If the expected behavior is clear, I can make it consistent across the tests.

CecileRobertMichon:
The `dnsPrefix` field being required for vlabs is by design.

The `deploy` command allows autofilling some fields via cli, so it does not validate the apimodel file the first time, only after autofill. However, the the engine tests expect every file in `testdata` to be a valid apimodel. 

Is there any way you can test your feature without adding a testdata file that doesn't pass validation (ie. with no "dnsPrefix")? Could you maybe use what https://github.com/Azure/aks-engine/blob/master/cmd/deploy_test.go#L225 is doing, ie. use `ExampleAPIModel` which already doesn't have a dnsPrefix and then call `autofillApimodel ` and check the resulting `dc.outputDirectory`?

CecileRobertMichon:
FYI @alexeldeib if you make your Azure org membership public (https://github.com/orgs/Azure/people?utf8=%E2%9C%93&query=alexeldeib) you'll be able to run pipelines yourself.

alexeldeib:
Thanks! I'm still validating that change locally, but I made my membership public so I'll trigger the CI once it looks good.

alexeldeib:
@CecileRobertMichon Changed my membership but no luck yet^ 🤕 I believe this passes tests now.

mboersma:
Test failures don't appear to be related.

PR #858, opened 2019-03-22
[MERGE ME WITH NEXT VHD UPDATE] fix: change the filepath of the VHD logs in CSE

CecileRobertMichon:
**Reason for Change**:

#777 moved the VHD logs. Update CSE to expect the logs in the right place.

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:
Merge this when the next VHD is merged.

PR #884, opened 2019-03-26
chore: enable 2019.03.25 VHD image

### LINKED ISSUE #761: ubuntu-18.04 distro + AzureCNI results in OS Provisioning Timeouts (OSPTO)
As discovered in #223, cloud-init 18.04 changed the default for ` apply_network_config` default to true.
https://git.launchpad.net/cloud-init/tree/cloudinit/sources/DataSourceAzure.py#n213

This results in no network path from VM ( potentially due to race with older IP address set within the VM from from Metadata Service). Because AzureCNI uses more IPs than kubenet, it is much more likely to reproduce this behavior, which causes most 18.04 + AzureCNI deployments to result in OSPTO.

There is a separate bug in cloud-init where the userdata supplied when a VM is provisioned cannot override the configuration of the Azure datasource in custom data. 

Workaround (not tested):
In the VM used to build the VHD, create a file in /etc/cloud/cloud.cfg.d called 80_azure_net_config.cfg with this content:
 
```
datasource:
    Azure:
        apply_network_config: false
```

jackfrancis:
**Reason for Change**:

Enable new VHD image. See release notes for details.

**Issue Fixed**:

Fixes #761

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetescustomscript.sh
+++ b/parts/k8s/kubernetescustomscript.sh
@@ -49,7 +49,7 @@ else
     REBOOTREQUIRED=false
 fi
 
-if [ -f /var/log/azure/golden-image-install.complete ]; then
+if [ -f /var/log.vhd/azure/golden-image-install.complete ]; then
     echo "detected golden image pre-install"
     FULL_INSTALL_REQUIRED=false
     rm -rf /home/packer
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -135,7 +135,7 @@ var (
 		ImageOffer:     "aks",
 		ImageSku:       "aks-ubuntu-1604-201903",
 		ImagePublisher: "microsoft-aks",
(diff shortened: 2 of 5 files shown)


CecileRobertMichon:
This also fixes #761

parts/k8s/kubernetescustomscript.sh
```diff
@@ -49,7 +49,7 @@ else
     REBOOTREQUIRED=false
 fi
 
-if [ -f /var/log/azure/golden-image-install.complete ]; then
+if [ -f /var/log.vhd/azure/golden-image-install.complete ]; then
```
CecileRobertMichon: See #858 for more details

mboersma:
lgtm

CecileRobertMichon:
/hold

until image is live

CecileRobertMichon:
/hold cancel

image is live

PR #897, opened 2019-03-27
feat: enable nvidia drivers on moby-backed VMs, deprecate docker-engine entirely

jackfrancis:
**Reason for Change**:

Standardizes nvidia driver installation onto the moby-backed "aks" distro. The scope of this PR:

- officially retire the  N-series-specific VHD (via the "aks-docker-engine" distro)
- officially remove any cluster configuration scenario that will deliver docker-engine
  - prior to this PR, docker-engine was added "implicitly" via a `containerRuntime` value of `docker` on an agent pool backed by N series VMs

This PR as-is only works for N series agent pool VMs on the "ubuntu" and "ubuntu-18.04" distros (i.e., the non-VHD-backed cluster configurations). To work w/ "aks" and "aks-18.04" (i.e., the VHD-backed cluster configurations) we'll first have to build a VHD using the changes to VHD build CI included in this PR, and then update the distro references to the new VHD that has been built w/ the moby installation changes in this PR.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -48,16 +48,14 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
...
-    echo "  - docker-engine v${DOCKER_ENGINE_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-    installGPUDrivers
-    echo "  - nvidia-docker2 nvidia-container-runtime" >> ${RELEASE_NOTES_FILEPATH}
+    # deprecated
+    exit 1
 else
     MOBY_VERSION="3.0.4"
     installMoby
     echo "  - moby v${MOBY_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
+    installGPUDrivers
```
CecileRobertMichon: Since we are now installing this on the "regular" VHD flavor, we should remove it at provisioning for VMs that aren't GPU enabled.

parts/k8s/kubernetesinstalls.sh
```diff
@@ -48,7 +48,29 @@ installGPUDrivers() {
     retrycmd_if_failure_no_stats 120 5 25 cat /tmp/nvidia-docker.list > /etc/apt/sources.list.d/nvidia-docker.list
     apt_get_update
     retrycmd_if_failure 30 5 3600 apt-get install -y linux-headers-$(uname -r) gcc make dkms || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
-    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-docker2=${NVIDIA_DOCKER_VERSION}+docker${DOCKER_VERSION} nvidia-container-runtime=${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker${DOCKER_VERSION} || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
+
+    # only install the runtime since nvidia-docker2 has a hard dep on docker CE packages.
+    # we will manually install nvidia-docker2
+    retrycmd_if_failure 30 5 3600 apt-get -o Dpkg::Options::="--force-confold" install -y nvidia-container-runtime="${NVIDIA_CONTAINER_RUNTIME_VERSION}+docker18.09.2-1" || exit $ERR_GPU_DRIVERS_INSTALL_TIMEOUT
```
CecileRobertMichon: let's make sure this works for non-VHD scenarios (Ubuntu distro + GPU enabled).

parts/k8s/kubernetesinstalls.sh
```diff
@@ -48,7 +48,29 @@ installGPUDrivers() {
...
+
+    tmpDir=$(mktemp -d)
+    (
+      set -e -o pipefail
+      cd "${tmpDir}"
+
+      retrycmd_if_failure 30 5 3600 apt-get download nvidia-docker2="${NVIDIA_DOCKER_VERSION}+docker18.09.2-1"
```
CecileRobertMichon: we should exit with an error code if this fails

pkg/engine/params_k8s.go
```diff
@@ -244,13 +244,6 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 			addValue(parametersMap, "kubernetesEndpoint", properties.HostedMasterProfile.FQDN)
 		}
 
-		// GPU nodes need docker-engine as the container runtime
-		if properties.HasNSeriesSKU() {
```
CecileRobertMichon: should we also remove from https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesparams.t#L322?

pkg/api/defaults.go
```diff
@@ -473,10 +469,6 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 				// (1) At present, "aks-docker-engine" and "ubuntu" are the only working distro base for running GPU workloads on N Series SKUs
 				// (2) Previous versions of aks-engine had working implementations using the "aks" distro value,
 				//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
-			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) && profile.IsNSeriesSKU() {
-				if profile.Distro == AKS {
-					profile.Distro = AKSDockerEngine
```
CecileRobertMichon: how will this work for upgrade/scale? Shouldn't we override the distro to AKS? Otherwise it will keep AKSDockerEngine which doesn't exist anymore

jackfrancis:
@cpuguy83 I've stolen your hard work and am experimenting on getting nvidia drivers working on moby-backed vms. The installation process seems to succeed, but I'm getting this when attempting to schedule workloads onto the nodes:

```
  Warning  FailedCreatePodSandBox  21s (x12 over 117s)  kubelet, k8s-agent1-19153332-vmss000000  Failed create pod sandbox: rpc error: code = Unknown desc = failed to start sandbox container for pod "blobfuse-flexvol-installer-rqckx": Error response from daemon: OCI runtime create failed: container_linux.go:344: starting container process caused "process_linux.go:297: getting the final child's pid from pipe caused \"EOF\"": unknown
```

Does the above ring a bell?

cpuguy83:
@jackfrancis Does this have memory constraints on it?

jackfrancis:
Yes. But the constraints are unchanged compared to docker-engine-configured clusters.

parts/k8s/kubernetesinstalls.sh
jackfrancis: yep, this entire thing is being tested against ubuntu distro

parts/k8s/kubernetesinstalls.sh
cpuguy83: `set -e` does that.

jackfrancis: done

pkg/engine/params_k8s.go
jackfrancis: done

pkg/api/defaults.go
jackfrancis: done

parts/k8s/kubernetesconfigs.sh
```diff
@@ -420,5 +420,4 @@ configGPUDrivers() {
 ensureGPUDrivers() {
     configGPUDrivers
     systemctlEnableAndStart nvidia-modprobe || exit $ERR_GPU_DRIVERS_START_FAIL
-    retrycmd_if_failure 30 1 120 systemctl restart kubelet || exit $ERR_GPU_DRIVERS_START_FAIL
 }
```
cpuguy83: See https://github.com/Azure/aks-engine/pull/877/files#diff-b57d179f607a757733708864ea1bc0cbR418 to enable the nvidia runtime.

cpuguy83: The nvidia-docker2 package does have a file that it would normally copy to /etc/docker/daemon.json, but this would overwrite any custom settings (which I know I've seen elsewhere in the codebase)

pkg/api/defaults.go
```diff
@@ -460,22 +460,17 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 					if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
 						profile.Distro = Ubuntu
 					} else {
-						if profile.IsNSeriesSKU() {
```
CecileRobertMichon: validate.go and docs also need to be changed

pkg/api/defaults_test.go
```diff
@@ -1208,139 +1208,6 @@ func TestSetVMSSDefaultsAndZones(t *testing.T) {
 
 }
 
-func TestAKSDockerEngineDistro(t *testing.T) {
```
CecileRobertMichon: let's update the existing tests to test that GPU nodes get the "AKS" or "AKS-18.04" distro

pkg/api/defaults.go
```diff
@@ -460,22 +460,17 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
-				// (2) Previous versions of aks-engine had working implementations using the "aks" distro value,
-				//     so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
-			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) && profile.IsNSeriesSKU() {
-				if profile.Distro == AKS {
-					profile.Distro = AKSDockerEngine
+				// Previous versions of aks-engine required the docker-engine distro for N series vms,
+				// so we need to hard override it in order to produce a working cluster in upgrade/scale contexts
+			} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) {
+				if profile.Distro == AKSDockerEngine {
+					profile.Distro = AKS
```
CecileRobertMichon: this should test for 18.04 as well

parts/k8s/kubernetesconfigs.sh
jackfrancis: @cpuguy83 Is this what you're referring to?

https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetesagentcustomdata.yml#L146

cpuguy83: There was somewhere that sets overlay2 as a storage driver as I recall.
But yeah that file you have referenced there is what you want.

jackfrancis: Extra overlay2 stuff for docker-engine has been removed.

I see what you mean above. Basically, we're currently enabling the nvidia runtime via a cloud-init-provided `/etc/docker/daemon.json`, but you're saying that the drivers installation will override that? I actually have a cluster built against this PR as-is (without your manual changes in https://github.com/Azure/aks-engine/pull/877/files#diff-b57d179f607a757733708864ea1bc0cbR418) and it looks O.K.:

```
azureuser@k8s-agent1-12179023-vmss000000:~$ cat /etc/docker/daemon.json
{
  "live-restore": true,
  "log-driver": "json-file",
  "log-opts":  {
     "max-size": "50m",
     "max-file": "5"
  }
  ,"default-runtime": "nvidia",
  "runtimes": {
     "nvidia": {
         "path": "/usr/bin/nvidia-container-runtime",
         "runtimeArgs": []
    }
  }
}
```

(Also, the E2E tests w/ cuda-vector job checks out.)

jackfrancis:
// TODO validate that we're actually remove docker-engine support altogether in a sane way

CecileRobertMichon:
Can we make the commit message more explicit about switching GPU to Moby and removing docker-engine support?

packer/install-dependencies.sh
```diff
@@ -48,12 +48,8 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
 EOF
 
 if [[ ${FEATURE_FLAGS} == *"docker-engine"* ]]; then
-    DOCKER_ENGINE_REPO="https://apt.dockerproject.org/repo"
-    installDockerEngine
-    overrideDockerEngineStorageDriver
-    echo "  - docker-engine v${DOCKER_ENGINE_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-    installGPUDrivers
```
CecileRobertMichon: Why are we not pre-pulling GPU drivers in VHD anymore?

CecileRobertMichon:
`Standardizes nvidia driver installation onto the moby-backed "aks" distro` and `Moving away from an N-series-specific VHD (via the "aks-docker-engine" distro) means we have to install the nvidia drivers at runtime` are contradictory. The first sentence and the PR title imply that we are installing the drivers in the aks distro image while the second sentence and the code say that we aren't installing the drivers in the VHD but at runtime.

pkg/api/defaults.go
CecileRobertMichon: clusterdefinitions.md still needs to be updated (two places reference the docker-engine distro and the restrictions on GPU) as well as gpu.md

packer/install-dependencies.sh
jackfrancis: I'm rethinking this, we'll download (as appropriate) prereq's during VHD CI so we don't have to during VHD bootstrapping,.

pkg/api/defaults.go
jackfrancis: Done

pkg/api/defaults.go
jackfrancis: Not sure what you mean

packer/install-dependencies.sh
```diff
@@ -47,18 +47,10 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
...
-    echo "  - nvidia-docker2 nvidia-container-runtime" >> ${RELEASE_NOTES_FILEPATH}
-else
-    MOBY_VERSION="3.0.4"
-    installMoby
-    echo "  - moby v${MOBY_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-fi
+MOBY_VERSION="3.0.4"
+installMoby
+installGPUDrivers
```
CecileRobertMichon: let's re-add the release notes log line for the nvidia driver install

parts/k8s/kubernetesconfigs.sh
```diff
@@ -420,5 +437,4 @@ configGPUDrivers() {
 ensureGPUDrivers() {
     configGPUDrivers
     systemctlEnableAndStart nvidia-modprobe || exit $ERR_GPU_DRIVERS_START_FAIL
-    retrycmd_if_failure 30 1 120 systemctl restart kubelet || exit $ERR_GPU_DRIVERS_START_FAIL
```
CecileRobertMichon: why remove this?

parts/k8s/kubernetescustomscript.sh
```diff
@@ -75,6 +75,8 @@ if [[ "${GPU_NODE}" = true ]]; then
         installGPUDrivers
     fi
     ensureGPUDrivers
+else
+    cleanUpGPUDrivers
```
CecileRobertMichon: 👍

parts/k8s/kubernetesconfigs.sh
jackfrancis: the k8s runtime hasn't been commenced yet, I think this is a no op (or actually a bug)

jackfrancis: there is a comparable restart later in the CSE flow

parts/k8s/kubernetesconfigs.sh
```diff
@@ -408,6 +408,23 @@ configAddons() {
 }
 
 configGPUDrivers() {
+    # only install the runtime since nvidia-docker2 has a hard dep on docker CE packages.
```
CecileRobertMichon: does any of this rely on a working outbound connection?

pkg/api/addons.go
```diff
@@ -97,9 +97,9 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 			{
 				Name:           DefaultBlobfuseFlexVolumeAddonName,
 				CPURequests:    "50m",
-				MemoryRequests: "10Mi",
+				MemoryRequests: "100Mi",
 				CPULimits:      "50m",
-				MemoryLimits:   "10Mi",
+				MemoryLimits:   "100Mi",
```
CecileRobertMichon: does this belong in this PR? maybe we should change the other addons in a separate PR

pkg/api/defaults.go
CecileRobertMichon: you can disregard my comment actually, I was thinking we needed to preserve 18.04 if the distro was AKS1804 but this is only for clusters using the AKSDockerEngine distro so my comment doesn't apply.

parts/k8s/kubernetesconfigs.sh
jackfrancis: the apt stuff does

pkg/api/addons.go
jackfrancis: Yes, as-is they were not schedulable on moby + N series clusters.

parts/k8s/kubernetesconfigs.sh
CecileRobertMichon: any way to move that stuff in installGPUDrivers? I'm guessing if you didn't already it's probably because we can't but if we leave it here we're basically closing the door on no outbound clusters with GPU for Azure Stack.

jackfrancis: Let's mark it as a follow-up item, as it'll be be tricky to do that. Basically what we'll have to do is pre-download everything and then manually use dpkg to install local packages.

CecileRobertMichon:
bumping my comment from last week

Can we make the commit message more explicit about switching GPU to Moby and removing docker-engine support?

pkg/api/defaults_test.go
jackfrancis: I think the point of this PR is: there are no longer any VM SKU -- Distro relationships. (Happy to add more tests generally, but I don't think we want to add that one.)

pkg/api/vlabs/validate.go
```diff
@@ -768,17 +768,6 @@ func (a *AgentPoolProfile) validateCustomNodeLabels(orchestratorType string) err
 	return nil
 }
 
-func (a *AgentPoolProfile) validateKubernetesDistro() error {
-	switch a.Distro {
-	case AKS, AKS1804, Ubuntu1804:
-		if a.IsNSeriesSKU() {
-			return errors.Errorf("The %s VM SKU must use the %s or %s Distro as they require the docker-engine container runtime with Ubuntu 16.04-LTS", a.VMSize, AKSDockerEngine, Ubuntu)
```
CecileRobertMichon: since we're removing the "GPU needs to use ubuntu 16.04" validation let's run a test to verify that GPU + 18.04 works

pkg/api/defaults_test.go
CecileRobertMichon: I think what I meant was "test that no configurations get the AKSDockerEngine distro" since that is deprecated but still exists. Ie. we should never assign the AKSDockerEngine distro, especially for a scale or update scenario.

parts/k8s/kubernetesconfigs.sh
CecileRobertMichon: sure, let's open an issue and add it to the backlog

pkg/api/vlabs/validate.go
jackfrancis: verified

pkg/api/vlabs/validate.go
```diff
@@ -364,6 +364,23 @@ func (a *Properties) validateMasterProfile() error {
 	if m.SinglePlacementGroup != nil && m.AvailabilityProfile == AvailabilitySet {
 		return errors.New("singlePlacementGroup is only supported with VirtualMachineScaleSets")
 	}
+
+	// Check that distro has a valid value.
+	valid := false
```
CecileRobertMichon: Nice add. Can you put this in a function and reuse it for the master profile and agent pool validations?

jackfrancis: done

Issue #809, opened 2019-03-20
Windows + VMSS external svc routing broken


jackfrancis:
Repro:

Create a deployment + service (e.g., IIS). Like this:

```
$ k get pods --all-namespaces
NAMESPACE     NAME                                                  READY   STATUS    RESTARTS   AGE
default       dns-liveness                                          1/1     Running   0          18m
default       iis-kubernetes-westus2-87576-68128-745594868b-l4fn7   1/1     Running   0          12m
default       php-apache-long-running-7549b5558f-qzsfm              1/1     Running   0          17m
kube-system   azure-cni-networkmonitor-jrr4g                        1/1     Running   0          20m
kube-system   azure-ip-masq-agent-5qb4s                             1/1     Running   0          20m
kube-system   blobfuse-flexvol-installer-7tlsc                      1/1     Running   0          20m
kube-system   coredns-6fcd4ffdc-rlghv                               1/1     Running   0          20m
kube-system   keyvault-flexvolume-6fvks                             1/1     Running   0          20m
kube-system   kube-addon-manager-k8s-master-20982975-0              1/1     Running   0          19m
kube-system   kube-apiserver-k8s-master-20982975-0                  1/1     Running   0          20m
kube-system   kube-controller-manager-k8s-master-20982975-0         1/1     Running   0          20m
kube-system   kube-proxy-2mcns                                      1/1     Running   0          20m
kube-system   kube-scheduler-k8s-master-20982975-0                  1/1     Running   0          19m
kube-system   kubernetes-dashboard-54b795b69b-6d4hs                 1/1     Running   0          20m
kube-system   metrics-server-6bf85bb69b-5jqhs                       1/1     Running   0          20m
kube-system   tiller-deploy-5d9ffb885b-2s2b9                        1/1     Running   0          20m
$ k get svc --all-namespaces
NAMESPACE     NAME                                 TYPE           CLUSTER-IP     EXTERNAL-IP      PORT(S)         AGE
default       iis-kubernetes-westus2-87576-68128   LoadBalancer   10.0.180.62    52.247.222.249   80:30497/TCP    7m24s
default       kubernetes                           ClusterIP      10.0.0.1       <none>           443/TCP         20m
default       php-apache-long-running              ClusterIP      10.0.209.58    <none>           80/TCP          15m
kube-system   kube-dns                             ClusterIP      10.0.0.10      <none>           53/UDP,53/TCP   20m
[...]

PR #900, opened 2019-03-27
fix: AKS VMSS windows agent pool upgrade

xizhamsft:
**Reason for Change**:
Upgrade windows nodes's kubernetes version requires keeping the custom data of the windows profile because the kubernetes version is fixed in the custom data.
Also added retry with a 5 minutes timeout for copying node properties because windows node takes a little bit longer to show up.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/utils/util.go
+++ b/pkg/armhelpers/utils/util.go
@@ -32,12 +32,16 @@ const (
 
 	k8sWindowsOldVMNamingFormat = "^([a-fA-F0-9]{5})([0-9a-zA-Z]{3})([9])([a-zA-Z0-9]{3,5})$"
 	k8sWindowsVMNamingFormat    = "^([a-fA-F0-9]{4})([0-9a-zA-Z]{3})([0-9]{3,8})$"
+
+	windowsVmssNamingFormatV2       = "aks([0-9a-zA-Z]+)$"
+	windowsVmssAgentPoolNameIndexV2 = 1
 )
 
 var vmnameLinuxRegexp *regexp.Regexp
 var vmssnameRegexp *regexp.Regexp
 var vmnameWindowsRegexp *regexp.Regexp
 var oldvmnameWindowsRegexp *regexp.Regexp
+var windowsVmssRegexV2 *regexp.Regexp
 
 func init() {
 	vmnameLinuxRegexp = regexp.MustCompile(k8sLinuxVMNamingFormat)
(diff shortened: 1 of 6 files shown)


xizhamsft:
@kkmsft , please take a look

pkg/armhelpers/utils/util.go
```diff
@@ -91,8 +96,14 @@ func K8sLinuxVMNameParts(vmName string) (poolIdentifier, nameSuffix string, agen
 }
 
 // VmssNameParts returns parts of Linux VM name e.g: k8s-agentpool1-11290731-0
+// For windows, the VMSS name is 'aks' concatenated with agentpool name
 func VmssNameParts(vmssName string) (poolIdentifier, nameSuffix string, err error) {
-	vmssNameParts := vmssnameRegexp.FindStringSubmatch(vmssName)
+	vmssNameParts := windowsVmssRegexV2.FindStringSubmatch(vmssName)
+	if len(vmssNameParts) == 2 {
```
jackfrancis: Do we have Windows/Linux context in the code areas that call this func? Determining Windows vs Linux this way seems edge case-prone. It might be better to pass in to this func whether or not to use the Windows or Linux regex, or to have separate funcs for each.

jackfrancis:
manually validated back-compat upgrade scenario w/ VMAS hybrid cluster

note: Windows + VMSS still has some issues, e.g.: https://github.com/Azure/aks-engine/issues/809

pkg/armhelpers/utils/util.go
xizhamsft: added a new isWindows flag for vmss scale sets to upgrade.

PatrickLang:
/LGTM
If upgrades aren't handled in the Azure pipeline, then a manual test is needed

pkg/armhelpers/utils/util.go
```diff
@@ -100,6 +105,17 @@ func VmssNameParts(vmssName string) (poolIdentifier, nameSuffix string, err erro
 	return vmssNameParts[vmssAgentPoolNameIndex], vmssNameParts[vmssClusterIDIndex], nil
 }
 
+// WindowsVmssNameParts returns the agent pool name from the VMSS name.
+// For windows, the VMSS name is 'aks' concatenated with agentpool name
+func WindowsVmssNameParts(vmssName string) (poolIdentifier string, err error) {
```
CecileRobertMichon: can this be unit tested?

xizhamsft: added a simple unit test for WindowsVmssNameParts.

xizhamsft:
> /LGTM
> If upgrades aren't handled in the Azure pipeline, then a manual test is needed
Yes. Manually tested in AKS.

PR #909, opened 2019-03-28
fix: remove packer user

### LINKED ISSUE #899: Remove packer user at provisioning with AKS distro
packer user is created as part of VHD Packer build process. We delete `/home/packer` during deployments but we don't remove the packer user. We should remove it for consistency.

On a VM built with AKS distro:

`packer:x:1000:1000:Ubuntu:/home/packer:/bin/bash`

alexeldeib:
**Reason for Change**:

**Issue Fixed**:

Fixes #899 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/vhd-image-builder.json
+++ b/packer/vhd-image-builder.json
@@ -69,7 +69,7 @@
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
                 "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
-                "sudo deluser --remove-home packer"
+                "sudo rm -rf /home/packer"
             ]
         }
     ]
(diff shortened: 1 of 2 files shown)


jackfrancis:
lgtm pending successful VHD Pipeline job

mboersma:
D'oh, I guess it's not that simple:

```
==> azure-arm: Cleanup requested, deleting resource group ...[0m
[1;32m==> azure-arm: 
[0;32m    azure-arm: /usr/sbin/deluser: `/usr/sbin/userdel packer' returned error code 8. Exiting.[0m
[0;32m    azure-arm: userdel: user packer is currently used by process 2153[0m
[0;32m    azure-arm: Warning: group `packer' has no more members.[0m
```

alexeldeib:
If it seems too good to be true...

I'll try to find a good place to move this during the setup scripts.

CecileRobertMichon:
@alexeldeib https://github.com/Azure/aks-engine/blob/master/parts/k8s/kubernetescustomscript.sh#L55 should work

alexeldeib:
@CecileRobertMichon is there a different command to requeue vhd check? It seems to have disappeared

CecileRobertMichon:
Nope, that's because the VHD pipeline is smart and only runs when it's needed :) You're not changing the VHD codepaths anymore so the VHD build isn't a required validation.

alexeldeib:
D'oh 💯

CecileRobertMichon:
second attempt e2e passed, pipeline failed on artifact publish step

/lgtm

PR #915, opened 2019-03-29
docs: Cleaned up examples

larryms:
**Reason for Change**:

docs: cleans up example docs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
`--output tsv` is much better. :+1: Thanks for this and the typo fix @larryms!

Could you `git commit --amend` this so the commit message starts with `docs: `? We use that format to generate the changelog for each release.

PR #916, opened 2019-03-29
chore: use V1beta1 Policy client, not the deprecated default

mboersma:
**Reason for Change**:

The default `Policy()` is [deprecated](https://github.com/Azure/aks-engine/blob/master/vendor/k8s.io/client-go/kubernetes/clientset.go#L320) (and is gone in current versions of kubernetes/client-go). AKS Engine should be using `PolicyV1beta1()`.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/azurestack/kubeclient.go
+++ b/pkg/armhelpers/azurestack/kubeclient.go
@@ -113,7 +113,7 @@ func (c *KubernetesClientSetClient) EvictPod(pod *v1.Pod, policyGroupVersion str
 			Namespace: pod.Namespace,
 		},
 	}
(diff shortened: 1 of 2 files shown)


jackfrancis:
Is `PolicyV1beta1` a valid interface call as far back as k8s 1.9? E.g.:

```
$ az aks get-versions -l westus2 | grep 1.9
      "orchestratorVersion": "1.9.10",
          "orchestratorVersion": "1.9.11"
      "orchestratorVersion": "1.9.11",
```

Issue #878, opened 2019-03-25
armvariables_tests.go shouldn't have a hard-coded k8s version


mboersma:
In #873 it appears the unit tests broke by the addition of a new Kubernetes version. This should not be the case: tests should use the dynamic version functions to ensure they move forward when our Kubernetes support does.

PR #918, opened 2019-03-29
feat: default to k8s 1.14 for Windows and 1.12 for Linux

### LINKED ISSUE #798: set default k8s version to 1.14 for windows clusters
Since 1.14 is the Kubernetes version that Windows GA support is leaning on, we should set the default cluster version accordingly.

Currently the default version is 1.9.

mboersma:
**Reason for Change**:

With the release of Kubernetes 1.14, Windows containers support became GA. Additionally, Kubernetes 1.11 is no longer supported.

Also removes Kubernetes 1.11 for Windows from the test matrix.

**Issue Fixed**:

Fixes #798 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/const.go
```diff
@@ -59,9 +59,9 @@ const (
 
 const (
 	// KubernetesDefaultRelease is the default Kubernetes release
-	KubernetesDefaultRelease string = "1.11"
```
CecileRobertMichon: there might be unit tests that require updating with this change

mboersma: Yes, I pushed the update commit. Also see #878.

jackfrancis:
@mboersma let's expand the scope of this PR to include removing Windows 1.11 tests from PR E2E. AKS is enabling Windows clusters in public preview @ 1.12 and above only, and as described above GA is pinned to 1.14.

PR #937, opened 2019-04-03
revert: improve CSE script

jackfrancis:
Reverts Azure/aks-engine#870

Following this change, we're seeing regressions in CSE performance. Let's consider discrete changes here for re-introduction (especially the "disable unattended upgrades during first boot", as that may increase the stability of purposeful apt operations).

### DIFF
--- a/parts/k8s/kubernetesagentcustomdata.yml
+++ b/parts/k8s/kubernetesagentcustomdata.yml
@@ -52,26 +52,6 @@ write_files:
   content: !!binary |
     {{WrapAsVariable "systemConf"}}
 
-- path: /etc/apt/apt.conf.d/10periodic
-  permissions: "0644"
-  owner: root
-  content: |
-    APT::Periodic::Update-Package-Lists "1";
-    APT::Periodic::Download-Upgradeable-Packages "1";
-    APT::Periodic::AutocleanInterval "7";
-    APT::Periodic::Unattended-Upgrade "1";
-
-- path: /etc/apt/apt.conf.d/99periodic
-  permissions: "0644"
-  owner: root
-  content: |
-    // Disable unattended upgrades at first boot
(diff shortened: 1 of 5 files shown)


CecileRobertMichon:
@sylr FYI

/lgtm

CecileRobertMichon:
let's make the squash commit message `revert: improve CSE script` 

/hold

mboersma:
lgtm

sylr:
What do you mean by regression ? Isn't the problem that, with this patch, nodes are actually rebooted when they need to, which should be all the time as long as the image does not ship with the latest kernel.

jackfrancis:
Hi @sylr, if the functional intention of this change is to regularly reboot nodes during provisioning, we want that to be configurable, as it otherwise slows down cluster provisioning time. We'll follow up on that as a distinct PR.

sylr:
The main goal of the PR was to avoid apt locks by disabling the `unattended-upgrades` and triggers updates manually by calling `apt-get dist-upgrade`.

Rebooting the node if needed was already implemented but was basically a noop because the check happened before `unattended-upgrades` finished its work :

https://github.com/Azure/aks-engine/blob/patch-release-v0.31.1/parts/k8s/kubernetescustomscript.sh#L39-L43

I just moved it after all system upgrades are finished.

Not doing this is really dangerous security wise because if you provision new nodes with an old image (the one available when you created the cluster) you can run nodes with unpatched kernels and that's not worth the few seconds gained at first boot.

jackfrancis:
I think you are correct. The generic point is that this is a change and should be guarded behind a configuration flag that is disabled by default: folks who want this arguably improved implementation can get it.

sylr:
I understand this is a change but I disagree it should be disabled by default.

PR #938, opened 2019-04-03
fix: increase cordon/drain timeout to 20 mins during upgrade

### LINKED ISSUE #930: Upgrade Fails due to node drain timeout
Is this a request for help?: NO

Is this an ISSUE or FEATURE REQUEST? (choose one): ISSUE

What version of aks-engine?: v0.33.2

Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm) Kubernetes (upgrade from 1.13.3 to 1.13.5)

What happened: When doing an upgrade, there is now a drain process due to . However, the timeout for the drain is only 1 minute, and this is nowhere near long enough with persistent Azure disks, which take several minutes to drain from a node because they are dang slow to reschedule on other nodes.

The upgrade fails every time due to this timeout 

What you expected to happen:

Drain should not timeout or set an option to increase it.

How to reproduce it (as minimally and precisely as possible):

Create a cluster with a bunch of PVs. Upgrade it.

Anything else we need to know:

jackfrancis:
**Reason for Change**:

There are cluster configuration scenarios (e.g., vms that have lots of persistent volumes) for which a 1 minute cordon/drain timeout isn't sufficiently long. Let's unblock these quickly by increasing the timeout to 20 mins. A // TODO for later is to make this a user-configurable timeout.

**Issue Fixed**:

Fixes #930 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -21,7 +21,7 @@ import (
 const (
 	interval           = time.Second * 1
 	retry              = time.Second * 5
-	cordonDrainTimeout = time.Minute * 10
+	cordonDrainTimeout = time.Minute * 1
 )
 
 // Compiler to verify QueueMessageProcessor implements OperationsProcessor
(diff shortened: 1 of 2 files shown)


mboersma:
lgtm

CecileRobertMichon:
lgtm

I would call this a `fix: `

CecileRobertMichon:
Do we have a test signal for an upgrade with this change?

jackfrancis:
@CecileRobertMichon yes

PR #946, opened 2019-04-03
fix: fix disable rbac + enable aggregated API upgrade bug

### LINKED ISSUE #905: Aks-engine upgrade command fails on cluster deployed with ACSE v.0.18.8 and enabledRbac set to false
Yes.
---

**Is this an ISSUE or FEATURE REQUEST?**
Issue.
---

acs-engine 0.18.8 & aks-engine 0.33.1
---

The template used is here: 
```
{
    "apiVersion": "vlabs",
    "properties": {
        "orchestratorProfile": {
            "orchestratorVersion": "1.10.4",
            "kubernetesConfig": {
                "enableRbac": false
            }
        },
        "agentPoolProfiles": [
            {
                "storageProfile": "ManagedDisks",
                "name": "mtfoo0327",
                "count": 2,
                "osType": "Linux",
                "vmSize": "Standard_A2m_v2",
                "availabilityProfile": "VirtualMachineScaleSets"
            }
        ],
        "servicePrincipalProfile": {
            "clientId": "",
            "secret": ""
        },
        "linuxProfile": {
            "adminUsername": "foo",
            "ssh": {
                "publicKeys": [
                    {
                        "keyData": ""
                    }
                ]
            }
        },
        "masterProfile": {
            "storageProfile": "ManagedDisks",
            "count": 1,
            "dnsPrefix": "k8s-master-foo0327",
            "vmSize": "Standard_F2s"
        }
    }
}

```
**Kubernetes version**:
1.10.4 -> 1.11.8

**What happened**:
I have clusters that were provisioned with acs-engine version 0.18.8 with the flag `enableRbac` set to `false`. When upgrading the acs-engine to aks-engine 0.33.1 and then attempting the aks-engine upgrade command I receive the following error: 
```
{"code":"DeploymentFailed","message":"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details.","details":[{"code":"Conflict","message":"{\r\n \"status\": \"Failed\",\r\n \"error\": {\r\n \"code\": \"ResourceDeploymentFailure\",\r\n \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n \"details\": [\r\n {\r\n \"code\": \"VMExtensionProvisioningError\",\r\n \"message\": \"VM has reported a failure when processing extension 'cse-master-0'. Error message: \\\"Enable failed: failed to execute command: command terminated with exit status=30\\n[stdout]\\n\\n[stderr]\\nConnection to k8s.gcr.io 443 port [tcp/https] succeeded!\\nConnection to gcr.io 443 port [tcp/https] succeeded!\\nConnection to docker.io 443 port [tcp/https] succeeded!\\n\\\".\"\r\n }\r\n ]\r\n }\r\n}"}]}
```
This happens specifically when setting `"enableRbac": false` as a cluster with Rbac set to true upgrades successfully. Inside of the VM we can see that the apiserver has crashed and querying the docker container logs for the apiserver shows the error: 
``` 
Error: unable to load client CA file: unable to load client CA file: open /etc/kubernetes/certs/proxy-ca.crt: no such file or directory
```
Going through the apimodel.json that is generated, I have looked at every file path for every parameter and searched for the corresponding file on the VM. As well as confirming the docker error above, I found that there are four files in total that are listed in the generated apimodel.json that are not located on the VM after an attempted upgrade:

- /var/log/audit.log
- /etc/kubernetes/certs/proxy.crt
- /etc/kubernetes/certs/proxy.key
- /etc/kubernetes/certs/proxy-ca.crt

I'm assuming that the audit.log isn't mission critical but I wanted to be thorough. 

I can confirm that the proxy certs are, indeed, on the VM before the upgrade takes place, which would infer that they aren't being copied over correctly during the upgrade when enableRbac is set to false. 

I expect the upgrade to complete successfully.

- Using the provided template, run the `acs-engine generate` command using acs-engine version 0.18.8
- Run `az group create` to create a clean resource group.
- Run `az group deployment create` using the the output folder created in step one.   
- Using aks-engine version 0.33.1, Run `aks-engine upgrade` command using the output folder from above as the deployment-dir and 1.11.8 as the target version. 

I have included the cloud-init-output.log and the cluster-provision.log as attachments. 

[cloud-init-output.log](https://github.com/Azure/aks-engine/files/3017129/cloud-init-output.log)
[cluster-provision.log](https://github.com/Azure/aks-engine/files/3017130/cluster-provision.log)

We will continue looking at this issue on our side and will post any further information to this space. 

Thank you for your time.

jackfrancis:
**Reason for Change**:

For a period of time, there was a bug in default assignment w/ respect to `enableAggregatedAPI` and `enableRBAC`. Specifically: `enableAggregatedAPI` configuration depends upon `enableRBAC` (the former can only be `true` if the latter is also `true`).

Proper default enforcement was implemented here:

https://github.com/Azure/acs-engine/pull/3813

Prior to that it was possible to create a cluster w/ a pathological config (primarily because CSE was permissive to `enableAggregatedAPI` being set to true without RBAC, see https://github.com/Azure/acs-engine/pull/3751 for when that started to change).

The purpose of this PR is to detect pathological configurations in upgrade scenarios, and to fix them via defaults enforcements flow.

**Issue Fixed**:

Fixes #905 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -217,6 +217,10 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpdate bool) {
 				// TODO make EnableAggregatedAPIs a pointer to bool so that a user can opt out of it
 				a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs = true
 			}
+		} else if isUpdate && a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
+			// Upgrade scenario:
+			// We need to force set EnableAggregatedAPIs to false if RBAC was previously disabled
+			a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs = false
 		}
 
 		if a.OrchestratorProfile.KubernetesConfig.EnableSecureKubelet == nil {
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -1244,7 +1244,7 @@ func TestAzureCNIVersionString(t *testing.T) {
 	}
 }
 
-func TestDefaultDisableRbac(t *testing.T) {
+func TestEnableAggregatedAPIs(t *testing.T) {
 	mockCS := getMockBaseContainerService("1.10.3")
 	properties := mockCS.Properties
 	properties.OrchestratorProfile.OrchestratorType = Kubernetes
@@ -1253,6 +1253,17 @@ func TestDefaultDisableRbac(t *testing.T) {
 
 	if properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
 		t.Fatalf("got unexpected EnableAggregatedAPIs config value for EnableRbac=false: %t",
+			properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs)
+	}
+
+	mockCS = getMockBaseContainerService("1.10.3")
+	properties = mockCS.Properties
+	properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	properties.OrchestratorProfile.KubernetesConfig.EnableRbac = to.BoolPtr(true)
+	mockCS.setOrchestratorDefaults(true)
+
+	if !properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
+		t.Fatalf("got unexpected EnableAggregatedAPIs config value for EnableRbac=true: %t",
 			properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs)
 	}
 }
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -1266,7 +1266,7 @@ func TestAlwaysSetEnableAggregatedAPIsToFalseIfRBACDisabled(t *testing.T) {
 	mockCS.setOrchestratorDefaults(true)
 
 	if properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
-		t.Fatalf("got unexpected EnableAggregatedAPIs config value for EnableRbac=false: %t",
+		t.Fatalf("expected EnableAggregatedAPIs to be manually set to false in update scenario, instead got %t",
 			properties.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs)
 	}
 }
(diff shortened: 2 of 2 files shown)


pkg/api/defaults.go
```diff
@@ -217,6 +217,10 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpdate bool) {
 				// TODO make EnableAggregatedAPIs a pointer to bool so that a user can opt out of it
 				a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs = true
 			}
+		} else if isUpdate && a.OrchestratorProfile.KubernetesConfig.EnableAggregatedAPIs {
+			// Upgrade scenario:
+			// We need to force set EnableAggregatedAPIs to false if RBAC was previously disabled
```
CecileRobertMichon: "if RBAC was previously disabled" --> where is that logic? 

This seems to always set EnableAggregatedAPIs to `false` if it was true during update which doesn't seem to be the behavior we want.

pkg/api/defaults_test.go
```diff
@@ -1257,6 +1257,20 @@ func TestDefaultDisableRbac(t *testing.T) {
 	}
 }
 
+func TestAlwaysSetEnableAggregatedAPIsToFalseIfRBACDisabled(t *testing.T) {
+	mockCS := getMockBaseContainerService("1.10.3")
+	properties := mockCS.Properties
+	properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	properties.OrchestratorProfile.KubernetesConfig.EnableRbac = to.BoolPtr(false)
```
CecileRobertMichon: what about if RBAC is enabled?

pkg/api/defaults.go
jackfrancis: This is in the "else" block of a `if a.OrchestratorProfile.KubernetesConfig.IsRBACEnabled() {` statement. So the brute force override only happens if RBAC is disabled.

pkg/api/defaults_test.go
jackfrancis: Added coverage in a follow-up commit

pkg/api/defaults.go
CecileRobertMichon: aahh didn't read the diff correctly

PR #947, opened 2019-04-04
feat: retrieve endpoints from metadata endpoint for Azure Stack

honcao:
**Reason for Change**:

feat: retrieve endpoints from metadata endpoint for Azure Stack

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -216,6 +216,11 @@ func (dc *deployCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
 	}
 
 	if dc.containerService.Properties.IsAzureStackCloud() {
+		dc.containerService.SetCustomCloudProfileEnvironment()
+		err = dc.containerService.SetCustomCloudProfileEnvironment()
+		if err != nil {
+			return fmt.Errorf("Failed to set environment - %s", err)
+		}
 		writeCustomCloudProfile(dc.containerService)
 	}
 
--- a/parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
+++ b/parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
@@ -14,6 +14,9 @@ spec:
       imagePullPolicy: IfNotPresent
       command: ["/hyperkube", "controller-manager"]
       args: [<args>]
+      env:
+      - name: AZURE_ENVIRONMENT_FILEPATH
+        value: "/etc/kubernetes/azurestackcloud.json"
       volumeMounts:
         - name: etc-kubernetes
           mountPath: /etc/kubernetes
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -423,8 +423,9 @@ const (
 	// AzurePublicCloud is a const string reference identifier for public cloud
 	AzurePublicCloud = "AzurePublicCloud"
 	// AzureChinaCloud is a const string reference identifier for china cloud
-	AzureChinaCloud  = "AzureChinaCloud"
-	azureGermanCloud = "AzureGermanCloud"
+	AzureChinaCloud = "AzureChinaCloud"
+	// AzureGermanCloud is a const string reference identifier for german cloud
+	AzureGermanCloud = "AzureGermanCloud"
 	// AzureUSGovernmentCloud is a const string reference identifier for us government cloud
 	AzureUSGovernmentCloud = "AzureUSGovernmentCloud"
 	// AzureStackCloud is a const string reference identifier for Azure Stack cloud
@@ -432,16 +433,27 @@ const (
 )
 
 const (
-	//AzureADIdentitySystem Identity System
+	// AzureADIdentitySystem is a const string reference identifier for Azure AD identity System
 	AzureADIdentitySystem = "azure_ad"
-	//ADFS Identity System
+	// ADFSIdentitySystem is a const string reference identifier for ADFS identity System
 	ADFSIdentitySystem = "adfs"
 )
 
 const (
-	//ClientSecretAuthMethod Authentication method
+	// AzureStackDependenciesLocationPublic indicates to get dependencies from in AzurePublic cloud
+	AzureStackDependenciesLocationPublic = "public"
+	// AzureStackDependenciesLocationChina indicates to get dependencies from AzureChina cloud
+	AzureStackDependenciesLocationChina = "china"
+	// AzureStackDependenciesLocationGerman indicates to get dependencies from AzureGerman cloud
+	AzureStackDependenciesLocationGerman = "german"
+	// AzureStackDependenciesLocationUSGovernment indicates to get dependencies from AzureUSGovernment cloud
+	AzureStackDependenciesLocationUSGovernment = "usgovernment"
+)
+
+const (
+	// ClientSecretAuthMethod indicates to use client seret for authentication
 	ClientSecretAuthMethod = "client_secret"
-	//ClientCertificateAuthMethod Authentication method
+	// ClientCertificateAuthMethod indicates to use client certificate for authentication
 	ClientCertificateAuthMethod = "client_certificate"
 )
 
--- a/pkg/api/defaults-custom-cloud-profile.go
+++ b/pkg/api/defaults-custom-cloud-profile.go
@@ -3,34 +3,113 @@
 
 package api
 
-import "github.com/Azure/aks-engine/pkg/helpers"
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
 
-func (p *Properties) setCustomCloudProfileDefaults() {
+	"github.com/Azure/aks-engine/pkg/helpers"
+	"github.com/Azure/go-autorest/autorest/azure"
+)
 
+func (cs *ContainerService) setCustomCloudProfileDefaults() error {
+	p := cs.Properties
 	if p.IsAzureStackCloud() {
+		p.CustomCloudProfile.AuthenticationMethod = helpers.EnsureString(p.CustomCloudProfile.AuthenticationMethod, ClientSecretAuthMethod)
+		p.CustomCloudProfile.IdentitySystem = helpers.EnsureString(p.CustomCloudProfile.IdentitySystem, AzureADIdentitySystem)
+		p.CustomCloudProfile.DependenciesLocation = DependenciesLocation(helpers.EnsureString(string(p.CustomCloudProfile.DependenciesLocation), AzureStackDependenciesLocationPublic))
+		p.SetAzureStackCloudSpec()
+		err := cs.SetCustomCloudProfileEnvironment()
+		if err != nil {
+			return fmt.Errorf("Failed to set environment - %s", err)
+		}
+	}
+	return nil
+}
 
-		//azureStackCloudSpec is the default configurations for azure stack with public Azure.
-		azureStackCloudSpec := AzureEnvironmentSpecConfig{
-			CloudName: AzureStackCloud,
-			//DockerSpecConfig specify the docker engine download repo
-			DockerSpecConfig: DefaultDockerSpecConfig,
-			//KubernetesSpecConfig is the default kubernetes container image url.
-			KubernetesSpecConfig: DefaultKubernetesSpecConfig,
-			DCOSSpecConfig:       DefaultDCOSSpecConfig,
-			EndpointConfig: AzureEndpointConfig{
-				ResourceManagerVMDNSSuffix: "",
-			},
-			OSImageConfig: map[Distro]AzureOSImageConfig{
-				Ubuntu: DefaultUbuntuImageConfig,
-				RHEL:   DefaultRHELOSImageConfig,
-				CoreOS: DefaultCoreOSImageConfig,
-				AKS:    DefaultAKSOSImageConfig,
-			},
+// SetCustomCloudProfileEnvironment retrieves the endpoints from Azure Stack metadata endpoint and sets the values for azure.Environment
+func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
+	p := cs.Properties
+	if p.IsAzureStackCloud() {
+		if p.CustomCloudProfile.Environment == nil {
+			p.CustomCloudProfile.Environment = &azure.Environment{}
 		}
 
-		//Set default value for ResourceManagerVMDNSSuffix
-		azureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
+		env := p.CustomCloudProfile.Environment
+		if env.Name == "" || env.ResourceManagerEndpoint == "" || env.ServiceManagementEndpoint == "" || env.ActiveDirectoryEndpoint == "" || env.GraphEndpoint == "" || env.ResourceManagerVMDNSSuffix == "" {
+			env.Name = AzureStackCloud
+			if !strings.HasPrefix(p.CustomCloudProfile.PortalURL, fmt.Sprintf("https://portal.%s.", cs.Location)) {
+				return fmt.Errorf("portalURL needs to start with https://portal.%s. ", cs.Location)
+			}
+			azsFQDNSuffix := strings.Replace(p.CustomCloudProfile.PortalURL, fmt.Sprintf("https://portal.%s.", cs.Location), "", -1)
+			azsFQDNSuffix = strings.TrimSuffix(azsFQDNSuffix, "/")
+			env.ResourceManagerEndpoint = fmt.Sprintf("https://management.%s.%s/", cs.Location, azsFQDNSuffix)
+			metadataURL := fmt.Sprintf("%s/metadata/endpoints?api-version=1.0", strings.TrimSuffix(env.ResourceManagerEndpoint, "/"))
 
+			// Retrieve the metadata
+			httpClient := &http.Client{
+				Timeout: 30 * time.Second,
+			}
+			endpointsresp, err := httpClient.Get(metadataURL)
+			if err != nil || endpointsresp.StatusCode != 200 {
+				return fmt.Errorf("%s . apimodel invalid: failed to retrieve Azure Stack endpoints from %s", err, metadataURL)
+			}
+
+			body, err := ioutil.ReadAll(endpointsresp.Body)
+			if err != nil {
+				return fmt.Errorf("%s . apimodel invalid: failed to read the response from %s", err, metadataURL)
+			}
+
+			endpoints := AzureStackMetadataEndpoints{}
+			err = json.Unmarshal(body, &endpoints)
+			if err != nil {
+				return fmt.Errorf("%s . apimodel invalid: failed to parse the response from %s", err, metadataURL)
+			}
+
+			if endpoints.GraphEndpoint == "" || endpoints.Authentication == nil || endpoints.Authentication.LoginEndpoint == "" || len(endpoints.Authentication.Audiences) == 0 || endpoints.Authentication.Audiences[0] == "" {
+				return fmt.Errorf("%s . apimodel invalid: invalid response from %s", err, metadataURL)
+			}
+
+			env.GraphEndpoint = endpoints.GraphEndpoint
+			env.ServiceManagementEndpoint = endpoints.Authentication.Audiences[0]
+			env.GalleryEndpoint = endpoints.GalleryEndpoint
+			env.ActiveDirectoryEndpoint = endpoints.Authentication.LoginEndpoint
+			if p.CustomCloudProfile.IdentitySystem == ADFSIdentitySystem {
+				env.ActiveDirectoryEndpoint = strings.TrimSuffix(env.ActiveDirectoryEndpoint, "/")
+				env.ActiveDirectoryEndpoint = strings.TrimSuffix(env.ActiveDirectoryEndpoint, "adfs")
+			}
+
+			env.ManagementPortalURL = endpoints.PortalEndpoint
+			env.ResourceManagerVMDNSSuffix = fmt.Sprintf("cloudapp.%s", azsFQDNSuffix)
+			env.StorageEndpointSuffix = fmt.Sprintf("%s.%s", cs.Location, azsFQDNSuffix)
+			env.KeyVaultDNSSuffix = fmt.Sprintf("vault.%s.%s", cs.Location, azsFQDNSuffix)
+		}
+	}
+	return nil
+}
+
+// SetAzureStackCloudSpec sets the cloud spec for Azure Stack .
+func (p *Properties) SetAzureStackCloudSpec() {
+	if p.IsAzureStackCloud() {
+		var azureStackCloudSpec AzureEnvironmentSpecConfig
+		switch p.CustomCloudProfile.DependenciesLocation {
+		case AzureStackDependenciesLocationPublic:
+			azureStackCloudSpec = AzureCloudSpecEnvMap[AzurePublicCloud]
+		case AzureStackDependenciesLocationChina:
+			azureStackCloudSpec = AzureCloudSpecEnvMap[AzureChinaCloud]
+		case AzureStackDependenciesLocationGerman:
+			azureStackCloudSpec = AzureCloudSpecEnvMap[AzureGermanCloud]
+		case AzureStackDependenciesLocationUSGovernment:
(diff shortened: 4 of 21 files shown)


honcao:
FYI @deaborch @jadarsie @rjaini @vikasnav @Bhuvaneswari-Santharam

pkg/api/types.go
```diff
@@ -708,12 +711,27 @@ type V20180331ARMManagedContainerService struct {
...
+	Audiences     []string `json:"audiences,omitempty"`
+}
+
 // CustomCloudProfile represents the custom cloud profile
 type CustomCloudProfile struct {
 	Environment                *azure.Environment          `json:"environment,omitempty"`
 	AzureEnvironmentSpecConfig *AzureEnvironmentSpecConfig `json:"azureEnvironmentSpecConfig,omitempty"`
 	IdentitySystem             string                      `json:"identitySystem,omitempty"`
 	AuthenticationMethod       string                      `json:"authenticationMethod,omitempty"`
+	ConnectionMode             string                      `json:"connectionMode,omitempty"`
```
CecileRobertMichon: Do we expect to have more than two ConnectionMode valid values in the future? If not, would it make sense to make this a boolean `Connected` with a default of `true`?

pkg/api/const.go
```diff
@@ -434,10 +435,23 @@ const (
...
+	//Connected environment
+	Connected = "connected"
+	//AzureChina environment
+	AzureChina = "azure_china"
+	//AzureGerman environment
+	AzureGerman = "azure_german"
+	//AzureUSGovernment environment
+	AzureUSGovernment = "azure_us_government"
+	//DisConnect environment
+	DisConnect = "disconnected"
```
CecileRobertMichon:
```suggestion
	Disconnected = "disconnected"
```

CecileRobertMichon: what is `disconnected `? I don't see it being used anywhere?

pkg/api/vlabs/types.go
```diff
@@ -519,6 +519,7 @@ type CustomCloudProfile struct {
 	AzureEnvironmentSpecConfig *AzureEnvironmentSpecConfig `json:"azureEnvironmentSpecConfig,omitempty"`
 	IdentitySystem             string                      `json:"identitySystem,omitempty"`
 	AuthenticationMethod       string                      `json:"authenticationMethod,omitempty"`
+	ConnectionMode             string                      `json:"connectionMode,omitempty"`
```
jadarsie: My understanding is that field `ConnectionMode` groups together a bunch of base URLs.
If that's the case, the current field name does not reflect that fact immediately.

I may prefer slightly better something like `DependenciesSources` or `DependenciesLocation` with valid values being "default", "china", and "none".

jadarsie:
PR title has a typo `retrive` => `retrieve`

pkg/api/const.go
```diff
@@ -434,10 +435,23 @@ const (
 const (
 	//AzureADIdentitySystem Identity System
```
jadarsie: Picky. MS documentation refers to `AAD` and `ADFS` as `Identity Providers`, not `Identity Systems`

https://docs.microsoft.com/en-us/azure/azure-stack/azure-stack-identity-overview

parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
```diff
@@ -14,6 +14,9 @@ spec:
       imagePullPolicy: IfNotPresent
       command: ["/hyperkube", "controller-manager"]
       args: [<args>]
+      env:
+      - name: AZURE_ENVIRONMENT_FILEPATH
+        value: "/etc/kubernetes/azurestackcloud.json"
```
jadarsie: Variable `AZURE_ENVIRONMENT_FILEPATH` is set in the custom data script.

Is it possible to use the variable value (`$AZURE_ENVIRONMENT_FILEPATH`) instead typing the same string twice?

I fear we can introduce regressions in the future if someone changes one string and not the other.

pkg/api/types.go
```diff
@@ -708,12 +711,27 @@ type V20180331ARMManagedContainerService struct {
...
+type AzureStackMetadataEndpoints struct {
+	GalleryEndpoint string                            `json:"galleryEndpoint,omitempty"`
+	GraphEndpoint   string                            `json:"graphEndpoint,omitempty"`
+	PortalEndpoint  string                            `json:"portalEndpoint,omitempty"`
+	Authentication  *AzureStackMetadataAuthentication `json:"authentication,omitempty"`
+}
+
+// AzureStackMetadataAuthentication is the type for Azure Stack metadata authentication endpoints
+type AzureStackMetadataAuthentication struct {
+	LoginEndpoint string   `json:"loginEndpoint,omitempty"`
```
jadarsie: I think the code base has a naming convention for properties that represent URLs or Base URLs.
In this case it would be `LoginEndpointURL` or `LoginEndpointURLBase` depending on the value.

@CecileRobertMichon, may be able to confirm/deny?

parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
jadarsie: Never mind, I looks like this is not possible

pkg/api/types.go
```diff
@@ -1546,6 +1564,120 @@ func (p *Properties) GetCustomCloudIdentitySystem() string {
...
+		if env.ServiceManagementEndpoint == "" || env.ActiveDirectoryEndpoint == "" || env.GraphEndpoint == "" || env.ResourceManagerVMDNSSuffix == "" {
+			metadataURL := fmt.Sprintf("%s/metadata/endpoints?api-version=1.0", strings.TrimSuffix(env.ResourceManagerEndpoint, "/"))
+
+			// Retrieve the metadata
+			httpClient := &http.Client{
+				Timeout: 15 * time.Second,
+			}
+			endpointsresp, err := httpClient.Get(metadataURL)
+			if err != nil || endpointsresp.StatusCode != 200 {
+				log.Fatalf("%s . apimodel invalid: failed to retrieve Azure Stack endpoints from %s", err, metadataURL)
```
jadarsie: The error message may be misleading. A non-200 HTTP response is not always caused by a bad URL.

pkg/api/vlabs/types.go
CecileRobertMichon: +1 

I didn't understand was `ConnectionMode` was at first when going over the code.

pkg/api/types.go
jadarsie: The same comment applies for the next 3 validations. The root cause may not be an `invalid apimodel`.

parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
jackfrancis: It seems like we should consider an Azure Stack-specific controller manager spec, or at least only expose this environment variable if we're in an Azure Stack context.

(And we can figure out a way to source the filename from a common place as well, but let's figure out how to expose Azure Stack-specific stuff in an Azure Stack-exclusive context first.)

pkg/api/vlabs/types.go
```diff
@@ -519,6 +519,7 @@ type CustomCloudProfile struct {
 	AzureEnvironmentSpecConfig *AzureEnvironmentSpecConfig `json:"azureEnvironmentSpecConfig,omitempty"`
```
jadarsie: This comment is out of place. It is based on you email. Given this input:

```
"customCloudProfile": {
  "identitySystem":"adfs",
  "authenticationMethod":"client_certificate",
  "connectionMode":"connected",
  "environment": {
    "name": "azureStackCloud",
    "resourceManagerEndpoint": "https://management.local.azurestack.external/"
  }
}
```
- Is it possible for `resourceManagerEndpoint` to deviate from this pattern: `https://management.{{location}}.azurestack.external/`?
- If not, then AKSe can compute it to further simplify the input cluster definition

parts/k8s/manifests/kubernetesmaster-kube-controller-manager-custom.yaml
jadarsie: @jackfrancis, if I understand correctly you comment, I think Hongbin has already addressed the concern. He has already introduced an AzS specific controller manager:

`kubernetesmaster-kube-controller-manager- **custom** .yaml`

pkg/api/vlabs/types.go
honcao: azurestack.external is input parameter during deploying azure stack.

pkg/api/types.go
honcao: it aligins with other existing error message.

pkg/api/const.go
honcao: I will ask PM for this and address it in the following changes.
In the getazuretsackstampinfo cmdlet in Azure Stack, they are AzureAD and ADFS.
the convension in aks engine for other parameters is using _ as seperator and lower case

jadarsie: I did not mean to say that we should use " " as separator or change the case. I am just pointing out that it may be more accurate to use the term `provider` over `system`. Just a small thing.

pkg/api/types.go
jadarsie: I like consistency as well. I'll rather be extra careful and avoid confusing users or CSS as they can end up barking at the wrong tree.

pkg/api/vlabs/types.go
jadarsie: k, I see

pkg/api/const.go
honcao: for none ( or disconnected) we do not need download any dependencies. but still need an clould spec for set up the parameters for ARM template, so set it to AzurePublicCloud

pkg/api/const.go
honcao: I see, in getazuretsackstampinfo cmdlet, it is IdentitySystem. let me check with PM.

pkg/api/types.go
honcao: agree, I would like to have PM reviewed all the message we have in the furture and make the change together.

CecileRobertMichon:
Please format all code comments according to golang convention (https://blog.golang.org/godoc-documenting-go-code)
`// A comment should start with a space and be a full sentence.`

pkg/api/const.go
```diff
@@ -434,10 +435,23 @@ const (
...
+	//DependenciesLocationDefault dependencies are in AzurePublic cloud
+	DependenciesLocationDefault = "default"
+	//DependenciesLocationChina dependencies are in AzureChina cloud
+	DependenciesLocationChina = "china"
+	//DependenciesLocationGerman dependencies are in AzureGerman cloud
+	DependenciesLocationGerman = "german"
+	//DependenciesLocationUSGovernment dependencies are us_government cloud
+	DependenciesLocationUSGovernment = "us_government"
+	//DependenciesLocationNone dependencies are in the VHD
+	DependenciesLocationNone = "none"
```
CecileRobertMichon: VHD is not its own cloud location. There is an overlap between cloud and pre-pulling dependencies on as VM image. It could be public cloud + VHD, China cloud + VHD, etc.

Also, I do not see `DependenciesLocationNone` being used anywhere. We shouldn't add a constant unless it is used in code.

pkg/api/const.go
```diff
@@ -434,10 +435,23 @@ const (
...
 	//AzureADIdentitySystem Identity System
 	AzureADIdentitySystem = "azure_ad"
-	//ADFS Identity System
+	//ADFSIdentitySystem System
 	ADFSIdentitySystem = "adfs"
 )
 
+const (
+	//DependenciesLocationDefault dependencies are in AzurePublic cloud
+	DependenciesLocationDefault = "default"
```
CecileRobertMichon: Are these constants Azure Stack specific? If so, would it make sense to clarify that by either naming or putting them in an Azure Stack specific field? Something like `AzureStackLocation`.

pkg/api/const.go
```diff
@@ -423,21 +423,33 @@ const (
...
 
+const (
+	//AzureStackDependenciesLocationDefault indicates to get dependencies from in AzurePublic cloud
+	AzureStackDependenciesLocationDefault = "default"
+	//AzureStackDependenciesLocationChina indicates to get dependencies from AzureChina cloud
+	AzureStackDependenciesLocationChina = "china"
+	//AzureStackDependenciesLocationGerman indicates to get dependencies from AzureGerman cloud
+	AzureStackDependenciesLocationGerman = "german"
+	//AzureStackDependenciesLocationUSGovernment indicates to get dependencies from AzureUSGovernment cloud
+	AzureStackDependenciesLocationUSGovernment = "us_government"
```
CecileRobertMichon: let's avoid using an underscore here

pkg/api/const.go
```diff
@@ -423,21 +423,33 @@ const (
...
+	//AzureADIdentitySystem is a const string reference identifier for Azure AD identity System
 	AzureADIdentitySystem = "azure_ad"
-	//ADFS Identity System
+	//ADFSIdentitySystem is a const string reference identifier for ADFS identity System
 	ADFSIdentitySystem = "adfs"
 )
 
+const (
+	//AzureStackDependenciesLocationDefault indicates to get dependencies from in AzurePublic cloud
+	AzureStackDependenciesLocationDefault = "default"
```
CecileRobertMichon: should this be "public" instead?

honcao: fixed.

pkg/api/const.go
honcao: fixed.

pkg/api/types.go
CecileRobertMichon: No strong opinion here, I'll leave it up to you. If the expected is a full URL then `URL` suffix makes sense. `URLBase` should only be if we expect a suffix and this is only the base domain (eg. `gcr.io`).

pkg/api/types.go
```diff
@@ -1546,6 +1564,119 @@ func (p *Properties) GetCustomCloudIdentitySystem() string {
...
+			azsFQDNSuffix = strings.TrimSuffix(azsFQDNSuffix, "/")
+			env.ResourceManagerVMDNSSuffix = fmt.Sprintf("cloudapp.%s", azsFQDNSuffix)
+			env.StorageEndpointSuffix = fmt.Sprintf("%s.%s", cs.Location, azsFQDNSuffix)
+			env.KeyVaultDNSSuffix = fmt.Sprintf("vault.%s.%s", cs.Location, azsFQDNSuffix)
+		}
+	}
+}
+
+// SetCustomAzureCloudSpec sets the cloud spec for Azure Stack .
+func (p *Properties) SetCustomAzureCloudSpec() {
```
CecileRobertMichon: this doesn't seem like it belongs in types.go... Is it setting defaults? If so it should be in defaults.go

pkg/api/types.go
```diff
@@ -1546,6 +1564,119 @@ func (p *Properties) GetCustomCloudIdentitySystem() string {
 	return AzureADIdentitySystem
 }
 
+// SetCustomCloudProfileEnvironment retrives the endpoints from Azure Stack metadata endpoint and set the values for azure.Environment
+func (cs *ContainerService) SetCustomCloudProfileEnvironment() {
```
CecileRobertMichon: same comment here about moving to defaults.go

pkg/api/types.go
```diff
@@ -1546,6 +1564,119 @@ func (p *Properties) GetCustomCloudIdentitySystem() string {
 	return AzureADIdentitySystem
 }
 
+// SetCustomCloudProfileEnvironment retrives the endpoints from Azure Stack metadata endpoint and set the values for azure.Environment
```
CecileRobertMichon: Are the below two functions unit tested?

cmd/deploy.go
```diff
@@ -216,6 +216,7 @@ func (dc *deployCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
 	}
 
 	if dc.containerService.Properties.IsAzureStackCloud() {
+		dc.containerService.SetCustomCloudProfileEnvironment()
```
CecileRobertMichon: what about using the `generate` + `az deploy` way of deploying (https://github.com/Azure/aks-engine/blob/master/docs/tutorials/deploy.md#aks-engine-the-long-way) 

would it make sense in general to move this function outside of commands and into template generation since deploy, generate, upgrade and scale all call it?

pkg/api/types.go
CecileRobertMichon: or even better in defaults-custom-cloud-profile.go since it's Azure Stack specific

pkg/api/vlabs/types.go
```diff
@@ -519,6 +519,7 @@ type CustomCloudProfile struct {
...
 	IdentitySystem             string                      `json:"identitySystem,omitempty"`
 	AuthenticationMethod       string                      `json:"authenticationMethod,omitempty"`
+	DependenciesLocation       string                      `json:"dependenciesLocation,omitempty"`
```
CecileRobertMichon: Let's make DependenciesLocation a type since we only accept specific values (see Distro and OSType as examples)

pkg/api/types.go
```diff
@@ -1546,6 +1564,119 @@ func (p *Properties) GetCustomCloudIdentitySystem() string {
...
+				Timeout: 15 * time.Second,
+			}
+			endpointsresp, err := httpClient.Get(metadataURL)
+			if err != nil || endpointsresp.StatusCode != 200 {
+				log.Fatalf("%s . apimodel invalid: failed to retrieve Azure Stack endpoints from %s", err, metadataURL)
+			}
+
+			body, err := ioutil.ReadAll(endpointsresp.Body)
+			if err != nil {
+				log.Fatalf("%s . apimodel invalid: failed to read the response from %s", err, metadataURL)
```
CecileRobertMichon: Instead of logging errors and exiting here, make the function return type an error and return the error. The only place that should exit and log the error is the original caller, main.go. 

Same comment for all the other `log.Fatalf` calls in the code.

pkg/api/types.go
honcao: TestSetCustomCloudProfileEnvironmentDefaults or TestSetCustomCloudProfileDefaults

cmd/deploy.go
honcao: For generate and deploy, it is implicitly invoked by SetPropertiesDefaults.

SetCustomCloudProfileEnvironment() is to set the default value and invoked by all 4 commands.
this is the method with this pattern. I am open for either way. 
@jackfrancis what do you think?

CecileRobertMichon: I don't see `SetCustomCloudProfileEnvironment` being called in `deploy.go`. Are you saying that it's already invoked in `SetPropertiesDefault`?

honcao: yes, currently it is correct SetCustomCloudProfileEnvironment is invoked in SetPropertiesDefault.

for deploy command, SetCustomCloudProfileEnvironment need to be invoked before get getClient,  
I could not find easy way to move SetPropertiesDefault before getClient, I explicitly invoke  SetCustomCloudProfileEnvironment before getClient

so SetCustomCloudProfileEnvironment will be invoked twice for deploy command.
the second call SetCustomCloudProfileEnvironment will not make any changes since all required values are set.

So we have two options here

1. remove SetCustomCloudProfileEnvironment from SetPropertiesDefault and call SetCustomCloudProfileEnvironment for generate and deploy command.

2. (the current code) call  SetCustomCloudProfileEnvironment inside SetPropertiesDefault and only call SetCustomCloudProfileEnvironment for deploy command.

Both way are ok for me.
@CecileRobertMichon please let me know your preference.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -3,34 +3,110 @@
...
+func (cs *ContainerService) setCustomCloudProfileDefaults() {
+	p := cs.Properties
 	if p.IsAzureStackCloud() {
+		p.CustomCloudProfile.AuthenticationMethod = helpers.EnsureString(p.CustomCloudProfile.AuthenticationMethod, ClientSecretAuthMethod)
+		p.CustomCloudProfile.IdentitySystem = helpers.EnsureString(p.CustomCloudProfile.IdentitySystem, AzureADIdentitySystem)
+		p.CustomCloudProfile.DependenciesLocation = DependenciesLocation(helpers.EnsureString(string(p.CustomCloudProfile.DependenciesLocation), AzureStackDependenciesLocationPublic))
+		p.SetCustomAzureCloudSpec()
+		err := cs.SetCustomCloudProfileEnvironment()
+		if err != nil {
+			log.Fatalf("Failed to set environment - %s", err)
```
CecileRobertMichon: Instead of logging errors and exiting here, make the function return type an error and return the error. The only place that should exit and log the error is the original caller, main.go.

Same comment for all the other log.Fatalf calls in the code.

cmd/deploy.go
CecileRobertMichon: Ok makes sense, thanks for the explanation. I think the way you're doing it now is fine in that case, unless there's a way to call getClient after `SetPropertiesDefault` (but looks like you tried that and it wasn't straightforward.)

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -3,34 +3,110 @@
...
+			env.ResourceManagerVMDNSSuffix = fmt.Sprintf("cloudapp.%s", azsFQDNSuffix)
+			env.StorageEndpointSuffix = fmt.Sprintf("%s.%s", cs.Location, azsFQDNSuffix)
+			env.KeyVaultDNSSuffix = fmt.Sprintf("vault.%s.%s", cs.Location, azsFQDNSuffix)
+		}
+	}
+	return nil
+}
+
+// SetCustomAzureCloudSpec sets the cloud spec for Azure Stack .
+func (p *Properties) SetCustomAzureCloudSpec() {
```
mboersma: Do we think this function will grow to include other types of customizations? If not (or not right away), maybe it would be clearer to include AzureStack in the name, like `SetAzureStackCloudSpec`.

honcao: I will make the change.

pkg/api/defaults-custom-cloud-profile.go
mboersma: I agree: this should be a `log.Errorf` instead, and the error should be returned. If we have code that calls `os.Exit`—which `Fatalf` does—it basically can't be unit-tested. (See also #536 for an example of how we've tried to get rid of `Fatalf`.)

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -3,34 +3,113 @@
...
-			EndpointConfig: AzureEndpointConfig{
-				ResourceManagerVMDNSSuffix: "",
-			},
-			OSImageConfig: map[Distro]AzureOSImageConfig{
-				Ubuntu: DefaultUbuntuImageConfig,
-				RHEL:   DefaultRHELOSImageConfig,
-				CoreOS: DefaultCoreOSImageConfig,
-				AKS:    DefaultAKSOSImageConfig,
-			},
+// SetCustomCloudProfileEnvironment retrives the endpoints from Azure Stack metadata endpoint and set the values for azure.Environment
```
mboersma: Typo: "retrives" -> "retrieves". Also "set" should be "sets," as in `...metadata endpoint and **sets** the values..."

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -3,34 +3,113 @@
...
+func (cs *ContainerService) setCustomCloudProfileDefaults() {
+	p := cs.Properties
 	if p.IsAzureStackCloud() {
+		p.CustomCloudProfile.AuthenticationMethod = helpers.EnsureString(p.CustomCloudProfile.AuthenticationMethod, ClientSecretAuthMethod)
+		p.CustomCloudProfile.IdentitySystem = helpers.EnsureString(p.CustomCloudProfile.IdentitySystem, AzureADIdentitySystem)
+		p.CustomCloudProfile.DependenciesLocation = DependenciesLocation(helpers.EnsureString(string(p.CustomCloudProfile.DependenciesLocation), AzureStackDependenciesLocationPublic))
+		p.SetAzureStackCloudSpec()
+		err := cs.SetCustomCloudProfileEnvironment()
+		if err != nil {
+			log.Printf("Failed to set environment - %s", err)
```
CecileRobertMichon: this should return an error instead of logging it

Issue #935, opened 2019-04-03
Upgrade fails due to drain node error for VMSS instance with hostnames ending with letters


mo-saeed:
**Is this a request for help?**: NO

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

**What version of aks-engine?**: v0.33.2

**Kubernetes version**:
1.13.4

**What happened**:

I got an error trying to upgrade kubernetes version from 1.13.4 to 1.13.5:

`INFO[0971] Draining node k8s-minion-42312186-vmss00000B
ERRO[0971] Error draining VM in VMSS: nodes "k8s-minion-42312186-vmss00000B" not found
Error: upgrading cluster: nodes "k8s-minion-42312186-vmss00000B" not found`

```kubectl get node
NAME                             STATUS   ROLES    AGE  
k8s-master-42312186-0            Ready    master   11h  
k8s-minion-42312186-vmss00000b   Ready    agent    33m  
k8s-minion-42312186-vmss00000c   Ready    agent    29m 
```
It seems that there's an issue with the azure instance names when it ends with Letters "e.g. a or b". in the scalet-set, instance names are ending with UPPERCASE Letters but kubernetes nodes names end with LOWERCASE letters which makes draining nodes not possible.

**What you expected to happen**:
Drain node should be successful regardless instance names in VMSS

**How to reproduce it** (as minimally and precisely as possible):
Upgrade kubernetes cluster with VMSS including instances with names ending with letters not numbers

CecileRobertMichon:
This is an interesting edge case. We usually don't have uppercase letters in VM names (the prefix is always lowercase and the suffix is numbers only). Just curious, did you scale up and down a lot to get to letter indices for your vmss instances?

I think we should be able to fix this easily by adding a `strings.ToLower()` on the VM name before proceeding with the drain and any other kube client operations in the upgrade flow.

mo-saeed:
@CecileRobertMichon Yes due to the load change on this cluster, we had to scale down/up many times and that led to having nodes ending with letters. After index "10" node names will always end with letter.

I hope that will be fixed soon.

Thanks a lot !

mo-saeed:
FYI, this is how computer name is defined in azure scale set:
https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-computer-name

mo-saeed:
[...]

PR #951, opened 2019-04-05
fix: handle uppercase letters in VM names in upgrade

CecileRobertMichon:
**Reason for Change**:

**Issue Fixed**:

#935 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"strings"
 	"time"
 
 	"github.com/Azure/aks-engine/pkg/api"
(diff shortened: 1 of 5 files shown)


CecileRobertMichon:
/hold

running manual upgrade test

pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -54,13 +55,15 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 		kubeAPIServerURL = kan.UpgradeContainerService.Properties.MasterProfile.FQDN
 	}
 
+	nodeName := strings.ToLower(*vmName)
```
jackfrancis: Just curious: are we validating that `*vmName` isn't nil anywhere in this flow, e.g., before it's passed into this method?

CecileRobertMichon: It _should_ never be nil since it comes directly from the list of VMs returned by the client but I've added an extra nil check just in case.

CecileRobertMichon:
upgrade test passed

/hold cancel

CecileRobertMichon:
ready for final lgtm + merge

Issue #953, opened 2019-04-05
Shell scripts should be linted


mboersma:
A significant part of AKS Engine's provisioning logic is contained in shell scripts, and there are several others in the /scripts directory. `shellcheck` is cruel but useful—let's use it.

PR #955, opened 2019-04-07
refactor: avoid shell globbing

davidsantiago-bib:
Double quote to prevent globbing.


mboersma:
Thanks for the cleanup @davidsantiago-bib!

See also #953: ideally `make test-style` should be running `shellcheck` on all our scripts. This helps.

PR #959, opened 2019-04-08
chore: update go toolchain to 1.12.2

mboersma:
**Reason for Change**:

See https://github.com/deis/docker-go-dev/releases/tag/v1.21.0
and https://github.com/golang/go/issues?q=milestone%3AGo1.12.2

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I already set this go-dev image as the default in our Jenkins instance. 🤞

### DIFF
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -17,7 +17,7 @@ pr:
 resources:
   containers:
   - container: dev1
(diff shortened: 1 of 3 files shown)


CecileRobertMichon:
We also need to update the image in the VHD pipeline

PR #961, opened 2019-04-08
chore: CIS hardening: root pw and apt-get dist-upgrade

jackfrancis:
**Reason for Change**:

Adds 2 CIS-related changes to provisioning process:

- run `apt-get dist-upgrade` at VHD (and CSE provisioning time, if appropriate) to freshen the apt dependency graph for component versions
- set a root password for single user mode

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -17,12 +17,6 @@ fi
 
 hardenNode
 installDeps
-
-ETCD_VERSION="3.2.25"
-ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
-installEtcd
-echo "  - etcd v${ETCD_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-
 cat << EOF >> ${RELEASE_NOTES_FILEPATH}
   - apt-transport-https
   - blobfuse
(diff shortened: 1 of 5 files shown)


packer/install-dependencies.sh
```diff
@@ -15,12 +15,14 @@ if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
...
+hardenNode
+installDeps
+
 ETCD_VERSION="3.2.25"
 ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
 installEtcd
 echo "  - etcd v${ETCD_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
 
-installDeps
 cat << EOF >> ${RELEASE_NOTES_FILEPATH}
```
CecileRobertMichon: the log lines should move up with the function call

parts/k8s/kubernetescustomscript.sh
```diff
@@ -187,3 +188,4 @@ else
       aptmarkWALinuxAgent unhold &
   fi
 fi
+set_log_perms
```
CecileRobertMichon: we move the /var/log directory in the Packer clean up step, we should do this after that is done. https://github.com/Azure/aks-engine/blob/master/packer/cleanup-vhd.sh#L11

PR #975, opened 2019-04-09
[CIS conformance] fix: disable ReadOnlyPort since it's only for heapster

andyzhangx:
**Reason for Change**:

`read-only-port`(default: `10255`) is only for heapster, while our aks-engine and aks has now switched to use metric server which uses `10250` by default.
This PR disabled `read-only-port` by set it as 0, here are the k8s ports definitons:
https://github.com/kubernetes/kubernetes/blob/b8b7ab39ec457b11caffbe0d31bd527cc8aa6f09/pkg/master/ports/ports.go#L38-L43

https://github.com/Azure/acs-engine/pull/2091

**Issue Fixed**:

461

Original CIS conformance test result:
```
[FAIL] 2.1.4 Ensure that the --read-only-port argument is set to 0 (Scored)
== Remediations ==
2.1.4 If using a Kubelet config file, edit the file to set readOnlyPort to 0 .
/etc/systemd/system/kubelet.service.d/10-kubeadm.conf on each worker node and
set the below parameter in KUBELET_SYSTEM_PODS_ARGS variable.
--read-only-port=0
systemctl daemon-reload
```

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
/assign @jackfrancis 
cc @feiskyer @leifei87

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -120,7 +120,9 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") {
 		// for enabling metrics-server v0.3.0+
 		defaultKubeletConfig["--authentication-token-webhook"] = "true"
-		defaultKubeletConfig["--read-only-port"] = "0" // we only have metrics-server v0.3 support in 1.16.0 and above
+		if !cs.Properties.IsHostedMasterProfile() { // Skip for AKS until it supports metrics-server v0.3
+			defaultKubeletConfig["--read-only-port"] = "0" // we only have metrics-server v0.3 support in 1.16.0 and above
+		}
 	}
 
 	// If no user-configurable kubelet config values exists, use the defaults
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -117,9 +117,10 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		defaultKubeletConfig["--tls-cipher-suites"] = TLSStrongCipherSuitesKubelet
 	}
 
-	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0-beta.1") {
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") {
 		// for enabling metrics-server v0.3.0+
 		defaultKubeletConfig["--authentication-token-webhook"] = "true"
+		defaultKubeletConfig["--read-only-port"] = "0" // we only have metrics-server v0.3 support in 1.16.0 and above
 	}
 
 	// If no user-configurable kubelet config values exists, use the defaults
--- a/docs/topics/clusterdefinitions.md
+++ b/docs/topics/clusterdefinitions.md
@@ -223,21 +223,29 @@ Below is a list of kubelet options that aks-engine will configure by default:
 
 | kubelet option                      | default value                                                                                                                                                 |
 | ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| "--cadvisor-port"    | "0"                                                                                                                                                         |
 | "--cloud-config"                    | "/etc/kubernetes/azure.json"                                                                                                                                  |
 | "--cloud-provider"                  | "azure"                                                                                                                                                       |
 | "--cluster-domain"                  | "cluster.local"                                                                                                                                               |
+| "--event-qps"    | "0"                                                                                                                                                         |
 | "--pod-infra-container-image"       | "pause-amd64:_version_"                                                                                                                                       |
+| "--network-plugin"                           | "cni"                                            |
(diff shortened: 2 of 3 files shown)


jackfrancis:
We should make this change only if the AKS-Engine-provided heapster addon isn't used. Also, for AKS scenarios, we will need to be able to know if the node is being built inside a cluster w/ heapster disabled.

pkg/api/defaults-kubelet.go
```diff
@@ -78,6 +78,7 @@ func (cs *ContainerService) setKubeletConfig() {
 		"--pod-max-pids":                    strconv.Itoa(DefaultKubeletPodMaxPIDs),
 		"--image-pull-progress-deadline":    "30m",
 		"--enforce-node-allocatable":        "pods",
+		"--read-only-port":                  "0",
```
CecileRobertMichon: Please add the flag to unit tests (after applying Jack's comment about conditionally adding this flag)

CecileRobertMichon:
The commit message should be `fix(CIS conformance): ...` in order to be picked up by release notes.

https://www.conventionalcommits.org/en/v1.0.0-beta.3/#commit-message-for-a-fix-using-an-optional-issue-number

andyzhangx:
> We should make this change only if the AKS-Engine-provided heapster addon isn't used. Also, for AKS scenarios, we will need to be able to know if the node is being built inside a cluster w/ heapster disabled.

@jackfrancis you are correct, let me mark this PR as WIP, looks like more work need to do.
In AKS, heapster is still used on v1.9.10
/hold

andyzhangx:
our AKS metric server(v0.2.1) is still using `KubeletReadOnlyPort`(10255), we cannot switch to `KubeletPort(10250)` now, after metric server v0.3.0, it uses `10250` port.

https://github.com/kubernetes-incubator/metrics-server/blob/020262240108f783e5d0e46891627fa9c8385776/cmd/metrics-server/app/start.go#L107

https://github.com/kubernetes/kubernetes/blob/99e61466ab694b3652db2c063b9996a5d324a57a/pkg/master/ports/ports.go#L43

Current AKS metric server(v0.2.1) logs:
```
I0416 14:18:10.371136       1 heapster.go:71] /metrics-server --source=kubernetes.summary_api:''
I0416 14:18:10.371350       1 heapster.go:72] Metrics Server version v0.2.1
I0416 14:18:10.371816       1 configs.go:61] Using Kubernetes client with master "https://10.0.0.1:443" and version
I0416 14:18:10.371894       1 configs.go:62] Using kubelet port 10255
I0416 14:18:10.373185       1 heapster.go:128] Starting with Metric Sink
I0416 14:18:10.562120       1 serving.go:308] Generated self-signed cert (apiserver.local.config/certificates/apiserver.crt, apiserver.local.config/certificates/apiserver.key)
I0416 14:18:10.929694       1 heapster.go:101] Starting Heapster API server...
[restful] 2019/04/16 14:18:10 log.go:33: [restful/swagger] listing is available at https:///swaggerapi
[restful] 2019/04/16 14:18:10 log.go:33: [restful/swagger] https:///swaggerui/ is mapped to folder /swagger-ui/
I0416 14:18:10.931327       1 serve.go:85] Serving securely on 0.0.0.0:443
```

jackfrancis:
@andyzhangx I'm re-engaging this PR now that we have 1.16.0 w/ metrics-server 0.3.4. I think this is ready to go.

(also, I noticed a bunch of kubelet default values aren't documented properly, cleaned that up while I'm here)

andyzhangx:
/hold
@jackfrancis we also need to make sure `ReadOnlyPort` is still enabled for AKS since metrics server on aks is still using v0.2.x

jackfrancis:
AKS isn’t using 1.16 yet so this change doesn’t affect AKS scenarios.

andyzhangx:
> AKS isn’t using 1.16 yet so this change doesn’t affect AKS scenarios.
but finally AKS will use 1.16, need to make sure if it's on an AKS cluster, `ReadOnlyPort` is still enabled on 1.16

pkg/api/defaults-kubelet.go
```diff
@@ -117,9 +117,12 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		defaultKubeletConfig["--tls-cipher-suites"] = TLSStrongCipherSuitesKubelet
 	}
 
-	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0-beta.1") {
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") {
 		// for enabling metrics-server v0.3.0+
 		defaultKubeletConfig["--authentication-token-webhook"] = "true"
+		if !cs.Properties.IsHostedMasterProfile() { // Skip for AKS until it supports metrics-server v0.3
```
jackfrancis: @andyzhangx This should protect AKS from getting this change too early.

I think this is ready for merge consideration

andyzhangx: thanks
/hold cancel

andyzhangx:
With this fix, all agent node (on aks-engine) CIS failures are fixed, it's a milestone!

PR #976, opened 2019-04-09
fix: potential go routine leak in upgrader if copy node properties always fails

xizhamsft:
**Reason for Change**:

As we added retry in copying node properties during VMSS agent pool upgrade, it is possible the go routine will never end if the copying keeps failure. This will not fail the upgrade because the code will continue after the time out. But we should fix the go routine leak.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -544,27 +544,33 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
 
 				ku.logger.Infof("Copying custom annotations, labels, taints from old node %s to new node %s...", vmToUpgrade.Name, newNodeName)
 				ch := make(chan struct{}, 1)
-				go func() {
-					for {
-						err = ku.copyCustomPropertiesToNewNode(client, strings.ToLower(vmToUpgrade.Name), strings.ToLower(newNodeName))
-						if err != nil {
-							ku.logger.Warningf("Failed to copy custom annotations, labels, taints from old node %s to new node %s: %v", vmToUpgrade.Name, newNodeName, err)
-							time.Sleep(time.Second * 5)
-						} else {
-							ch <- struct{}{}
-						}
-					}
-				}()
 
-				for {
-					select {
-					case <-ch:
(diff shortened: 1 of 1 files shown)


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -544,26 +544,37 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
...
-					case <-time.After(nodePropertiesCopyTimeout):
-						ku.logger.Errorf("Copying custom annotations, labels, taints from old node %s to new node %s can't complete within %v", vmToUpgrade.Name, newNodeName, nodePropertiesCopyTimeout)
-					}
-					break
+				}(ctx)
+
+				select {
+				case <-ch:
+					ku.logger.Infof("Finished trying to copy custom annotations, labels, taints from old node %s to new node %s.", vmToUpgrade.Name, newNodeName)
+				case <-time.After(nodePropertiesCopyTimeout):
```
jackfrancis: I know this usage was carried over from the existing implementation, but I think this `time.After` calculation is unnecessary.

I think a more idiomatic implementation would be having two channels: one for success condition and one for error condition.

See here for a simple example:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/pod/pod.go#L289

xizhamsft: @jackfrancis , per your request, removed the time.After for the channel.

CecileRobertMichon:
Did we run an upgrade test on this change?

CecileRobertMichon:
@xizhamsft do we still want this change? if so it needs to be rebased.

jackfrancis:
@CecileRobertMichon @xizhamsft I actually have a prototype in this PR that should be non-leaking:

https://github.com/Azure/aks-engine/pull/1345/files#diff-02355cd08d21ade04ad914b30dd6726cR28

(Note: that PR is not in the AKS codepath, it is AKS Engine-only)

I've been successfully testing it out O.K. If we get that to master I propose we use the above as a pattern, after which we should audit all similar goroutines in the code and update them accordingly. (And we could close this PR in favor of a near future one)

Thoughts @xizhamsft ?

PR #977, opened 2019-04-09
chore(CIS): add streaming-connection-idle-timeout

andyzhangx:
**Reason for Change**:

Add "--streaming-connection-idle-timeout=true" flag, according to CIS conformance, kubectl exec will drop after 5 minutes unattended:
```
[FAIL] 2.1.5 Ensure that the --streaming-connection-idle-timeout argument is not set to 0 (Scored)
== Remediations ==
2.1.5 If using a Kubelet config file, edit the file to set streamingConnectionIdleTimeout to a
value other than 0.
/etc/systemd/system/kubelet.service on each worker node and
set the below parameter in KUBELET_SYSTEM_PODS_ARGS variable.
--streaming-connection-idle-timeout=5m
systemctl daemon-reload
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/assign @jackfrancis
cc @leifei87 @feiskyer


CecileRobertMichon:
The commit message should be fix(CIS conformance): ... in order to be picked up by release notes.

https://www.conventionalcommits.org/en/v1.0.0-beta.3/#commit-message-for-a-fix-using-an-optional-issue-number

andyzhangx:
> The commit message should be fix(CIS conformance): ... in order to be picked up by release notes.

My commit msg is like following, I think it's ok, it already starts with `fix:` @CecileRobertMichon 
```
fix: add streaming-connection-idle-timeout
```

CecileRobertMichon:
let's standardize on https://github.com/Azure/aks-engine/pull/1013#issuecomment-481880375

andyzhangx:
@CecileRobertMichon I have changed the commit msg as following, PTAL, thanks.
```
chore(CIS): add streaming-connection-idle-timeout
```

PR #998, opened 2019-04-10
refactor: move isPrivateCluster to types.go

CecileRobertMichon:
**Reason for Change**:

Noticed that some isPrivateCluster assignments were missing nil checks and got carried away... This makes sure the function is in one central place and is unit tested.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -873,6 +873,65 @@ func TestRequireRouteTable(t *testing.T) {
 	}
 }
 
+func TestIsPrivateCluster(t *testing.T) {
+	cases := []struct {
+		p        Properties
+		expected bool
+	}{
+		{
+			p: Properties{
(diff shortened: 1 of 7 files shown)


pkg/api/types_test.go
```diff
@@ -873,6 +873,65 @@ func TestRequireRouteTable(t *testing.T) {
...
 
+func TestIsPrivateCluster(t *testing.T) {
+	cases := []struct {
+		p        Properties
+		expected bool
+	}{
+		{
+			p: Properties{
+				OrchestratorProfile: &OrchestratorProfile{
+					OrchestratorType: DCOS,
```
jackfrancis: I'd kinda rather we're testing the `Kubernetes` default case (no `PrivateCluster` object in the api model at all).

mboersma:
lgtm

PR #1005, opened 2019-04-10
test: add --feature-gates tests

jackfrancis:
**Reason for Change**:

Adds unit tests for controller-manager and kubet `--feature-gates` functionality. Also gets rid of an unnecessary no-op assignment (tests added primarily to prove the functional outcome of that removal).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/defaults-kubelet.go
```diff
@@ -92,7 +92,6 @@ func (cs *ContainerService) setKubeletConfig() {
 
 	// If no user-configurable kubelet config values exists, use the defaults
 	setMissingKubeletValues(o.KubernetesConfig, defaultKubeletConfig)
-	addDefaultFeatureGates(o.KubernetesConfig.KubeletConfig, o.OrchestratorVersion, "", "")
```
jackfrancis: The primary motivation for this PR was observing this weird thing in the default kubelet assignments flow.

pkg/api/defaults-controller-manager_test.go
```diff
@@ -84,6 +84,25 @@ func TestControllerManagerConfigDefaultFeatureGates(t *testing.T) {
...
+		t.Fatalf("got unexpected '--feature-gates' Controller Manager config value for \"--feature-gates\": \"LocalStorageCapacityIsolation=true,ServiceNodeExclusion=true\": %s",
+			cm["--feature-gates"])
+	}
+
+	// test user-overrides
+	cs = CreateMockContainerService("testcluster", "1.14.1", 3, 2, false)
+	cm = cs.Properties.OrchestratorProfile.KubernetesConfig.ControllerManagerConfig
+	cm["--feature-gates"] = "TaintBasedEvictions=true"
+	cs.setControllerManagerConfig()
+	if cm["--feature-gates"] != "LocalStorageCapacityIsolation=true,ServiceNodeExclusion=true,TaintBasedEvictions=true" {
```
jackfrancis: We can do this literal string comparison because the conversion from `map[string]string` to comma-delimited string is deterministically ordered alphabetically.

CecileRobertMichon:
yay more tests 🎉💯

pkg/api/defaults-kubelet.go
CecileRobertMichon: so is the thinking that this was unnecessary? any chance this was meant to be a guard for backcompat upgrade or anything like that?

jackfrancis: As I read the functional implementation of `addDefaultFeatureGates`, it does nothing. That func just adds to the pre-existing `--feature-gates` string. So passing in `""` does... nothing.

Also, the fact that the next statement actually does add stuff is additionally suspicious.

I was a little confused too, which is why I wrote the tests to prove that we weren't doing anything. Feel free to look through the tests and note any user scenarios that should be added. We can then do a before/after to prove that the tests both with and without the above line (to prove no-op).

CecileRobertMichon: 👍

PR #1007, opened 2019-04-10
chore: run apt-get dist-upgrade during VHD or for base OS image distro

### LINKED ISSUE #974: CIS: run `apt-get dist-upgrade` during VHD creation

jackfrancis:
**Reason for Change**:

Run `apt-get dist-upgrade` after `apt-get update` on "full install" scenarios (e.g., VHD or base image)

**Issue Fixed**:

Fixes #974

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -180,23 +180,23 @@ apt_get_install() {
     wait_for_apt_locks
 }
 apt_get_dist_upgrade() {
-	    retries=10
-	    apt_dist_upgrade_output=/tmp/apt-get-dist-upgrade.out
-	    for i in $(seq 1 $retries); do
-	        wait_for_apt_locks
-	        dpkg --configure -a
-	        apt-get -f -y install
-	        apt-get dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -E "^([WE]:.*)|([eE]rr.*)$"
-            [ $? -ne 0  ] && cat $apt_dist_upgrade_output && break || \
-            cat $apt_update_output
-            if [ $i -eq $retries ]; then
-                return 1
-            else sleep 5
-            fi
-	    done
-	    echo Executed apt-get dist-upgrade $i times
-	    wait_for_apt_locks
-	}
+  retries=10
+  apt_dist_upgrade_output=/tmp/apt-get-dist-upgrade.out
+  for i in $(seq 1 $retries); do
+    wait_for_apt_locks
+    dpkg --configure -a
+    apt-get -f -y install
+    apt-get dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -E "^([WE]:.*)|([eE]rr.*)$"
+    [ $? -ne 0  ] && cat $apt_dist_upgrade_output && break || \
+    cat $apt_update_output
+    if [ $i -eq $retries ]; then
+      return 1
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -155,7 +156,7 @@ apt_get_update() {
         cat $apt_update_output
         if [ $i -eq $retries ]; then
             return 1
-        else sleep 30
+        else sleep 5
```
jackfrancis: Standardizing to a 5 second sleep between apt update/upgrade retry operations

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -178,6 +179,24 @@ apt_get_install() {
...
+            [ $? -ne 0  ] && cat $apt_dist_upgrade_output && break || \
+            cat $apt_update_output
+            if [ $i -eq $retries ]; then
+                return 1
+            else sleep 5
+            fi
+	    done
+	    echo Executed apt-get dist-upgrade $i times
+	    wait_for_apt_locks
+	}
```
CecileRobertMichon: weird indent here

jackfrancis: agreed, fixed

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -31,6 +31,7 @@ installDeps() {
     retrycmd_if_failure_no_stats 120 5 25 curl -fsSL https://packages.microsoft.com/config/ubuntu/${UBUNTU_RELEASE}/packages-microsoft-prod.deb > /tmp/packages-microsoft-prod.deb || exit $ERR_MS_PROD_DEB_DOWNLOAD_TIMEOUT
     retrycmd_if_failure 60 5 10 dpkg -i /tmp/packages-microsoft-prod.deb || exit $ERR_MS_PROD_DEB_PKG_ADD_FAIL
     apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
+    apt_get_dist_upgrade || exit $ERR_APT_DIST_UPGRADE_TIMEOUT
```
CecileRobertMichon: Agreed we don't want to run this in the VHD case systematically since that would break the no outbound scenario. However, `apt-get dist-upgrade` seems to be most relevant when building from a VHD that was built a few days ago as the OS image might be outdated by then. For clusters deployed using the "full install" or ubuntu distro flows, `dist-upgrade` isn't as relevant since the VM will have just been built using the latest image available on Marketplace.

jackfrancis: "latest image available" is always a moving target, though, so this fills a non-trivial gap during VHD build time: if, say, the image version we're building upon is n days old this will add an additional n days of apt hygiene to the dependency graph.

CecileRobertMichon: Makes sense. Also I hadn't realized this is actually running as part of the VHD build too since it's in installDeps. The PR title is misleading, it implies it's only for the full install flow.

CecileRobertMichon:
lgtm

I would change the title to `chore: run apt-get dist-upgrade before apt-get install` or something that suggests this isn't just for the "full install" scenario which implies the FULL_INSTALL block in code that refers to cluster provisioning without VHD.

PR #1013, opened 2019-04-10
chore(CIS): assign root pw

### LINKED ISSUE #964: CIS 1.4.3 Ensure authentication required for single user mode
```
#!/bin/bash
#
# set root password to 52 random characters (only if root accts is 'locked')

# 1.4.3 Ensure authentication required for single user mode

# assign a passwd to root account (only if it is locked):
grep '^root:[!*]:' /etc/shadow

if [ $? -eq '0' ] ; then
  SALT=`openssl rand -base64 5`
  SECRET=`openssl rand -base64 37`
  CMD="import crypt, getpass, pwd; print crypt.crypt('$SECRET', '\$6\$$SALT\$')"
  HASH=`python <<< "$CMD"`

  echo 'root:'$HASH | /usr/sbin/chpasswd -e 2>/dev/null;
fi
```

jackfrancis:
**Reason for Change**:

Assigns a root password if none exists. Doing this via the cis.sh interface, which will be invoked by VHD and base OS image creation flows.

Also, as a point of hygiene, moving the `installDeps` above `installEtcd`, as the former should be done prior to the latter.

Includes E2E test coverage, which required that we shore up ssh-agent forwarding. For now this test is only enabled on ubuntu distro cluster configurations (as the enforcement isn't present on VHDs yet).

**Issue Fixed**:

Fixes #964

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/e2e-step-template.yaml
+++ b/.pipelines/e2e-step-template.yaml
@@ -20,7 +20,7 @@ steps:
       export CLIENT_ID=$(SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES)
       export CLIENT_SECRET=$(SERVICE_PRINCIPAL_CLIENT_SECRET_E2E_KUBERNETES)
       export CLIENT_OBJECTID=$(SERVICE_PRINCIPAL_OBJECT_ID_E2E_KUBERNETES)
-      bash -c 'eval `ssh-agent` && make test-kubernetes'
+      make test-kubernetes
     displayName: ginkgo k8s e2e tests
     workingDirectory: $(modulePath)
   - task: PublishPipelineArtifact@0
--- a/.pipelines/e2e-step-template.yaml
+++ b/.pipelines/e2e-step-template.yaml
@@ -20,7 +20,7 @@ steps:
       export CLIENT_ID=$(SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES)
       export CLIENT_SECRET=$(SERVICE_PRINCIPAL_CLIENT_SECRET_E2E_KUBERNETES)
       export CLIENT_OBJECTID=$(SERVICE_PRINCIPAL_OBJECT_ID_E2E_KUBERNETES)
-      make test-kubernetes
+      bash -c 'eval `ssh-agent` && make test-kubernetes'
     displayName: ginkgo k8s e2e tests
     workingDirectory: $(modulePath)
   - task: PublishPipelineArtifact@0
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -12,7 +12,6 @@ cat /proc/version | tee -a ${RELEASE_NOTES_FILEPATH}
 echo ""
 echo "Components downloaded in this VHD build (some of the below components might get deleted during cluster provisioning if they are not needed):" >> ${RELEASE_NOTES_FILEPATH}
 
-applyCIS
 installDeps
 cat << EOF >> ${RELEASE_NOTES_FILEPATH}
   - apt-transport-https
@@ -378,3 +377,5 @@ echo "START_OF_NOTES"
 cat ${RELEASE_NOTES_FILEPATH}
 echo "END_OF_NOTES"
 set -x
+
+applyCIS
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -12,15 +12,7 @@ cat /proc/version | tee -a ${RELEASE_NOTES_FILEPATH}
 echo ""
 echo "Components downloaded in this VHD build (some of the below components might get deleted during cluster provisioning if they are not needed):" >> ${RELEASE_NOTES_FILEPATH}
 
-if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  overrideNetworkConfig
-fi
-
-ETCD_VERSION="3.2.25"
-ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
-installEtcd
-echo "  - etcd v${ETCD_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-
+applyCIS
 installDeps
 cat << EOF >> ${RELEASE_NOTES_FILEPATH}
   - apt-transport-https
@@ -47,6 +39,15 @@ cat << EOF >> ${RELEASE_NOTES_FILEPATH}
   - xz-utils
   - zip
 EOF
+
+if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
+  overrideNetworkConfig
+fi
+
+ETCD_VERSION="3.2.25"
+ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
+installEtcd
+echo "  - etcd v${ETCD_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
 
 MOBY_VERSION="3.0.4"
 installMoby
--- a/parts/k8s/cloud-init/artifacts/cis.sh
+++ b/parts/k8s/cloud-init/artifacts/cis.sh
@@ -1,3 +1,16 @@
 #!/bin/bash
 
-# Placeholder for future CIS-related shell utilities
+assignRootPW() {
+  grep '^root:[!*]:' /etc/shadow
+  if [ $? -eq '0' ] ; then
+    SALT=`openssl rand -base64 5`
+    SECRET=`openssl rand -base64 37`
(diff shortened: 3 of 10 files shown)


CecileRobertMichon:
You can define a scope in a commit message like so:

`chore(CIS):  assign root pw`

https://www.conventionalcommits.org/en/v1.0.0-beta.3/#commit-message-with-scope

What do you think about on making that scope the standard for all the CIS PRs to come?

CecileRobertMichon:
also we should agree on chore vs fix so they all appear in the same section of release notes. @andyzhangx has been using fix: in his PRs.

CecileRobertMichon:
your squash commit message here will be your single commit, not the PR title FYI

packer/install-dependencies.sh
```diff
@@ -12,15 +12,7 @@ cat /proc/version | tee -a ${RELEASE_NOTES_FILEPATH}
...
-if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  overrideNetworkConfig
-fi
-
-ETCD_VERSION="3.2.25"
-ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
-installEtcd
-echo "  - etcd v${ETCD_VERSION}" >> ${RELEASE_NOTES_FILEPATH}
-
+applyCIS
```
CecileRobertMichon: some of other CIS stuff needs to happen after we setup the other things in this script so should we call this at the end of the Packer script instead of before everything else?

jackfrancis: Yeah, that works.

jackfrancis:
@CecileRobertMichon @andyzhangx I don't think these changes are fixes, so I vote for:

`chore(CIS): <description of change>`

CecileRobertMichon:
+1, they aren't fixing any bugs.

jackfrancis:
FYI I added an E2E test for this which will fail in this PR (because it is using a VHD that doesn't have this change)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -112,33 +112,20 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 		})
 
-		It("should display the installed docker runtime on the master node", func() {
+		It("should display the installed docker runtime on all nodes", func() {
```
jackfrancis: changed this test description to reflect reality

pkg/api/types.go
```diff
@@ -1015,6 +1015,21 @@ func (p *Properties) HasZonesForAllAgentPools() bool {
...
+// IsUbuntuDistroForAllAgentPools returns true if all of the agent pools plus masters are running the base Ubuntu image
+func (p *Properties) IsUbuntuDistroForAllAgentPools() bool {
+	if len(p.AgentPoolProfiles) > 0 {
+		for _, ap := range p.AgentPoolProfiles {
+			if ap.Distro != Ubuntu {
+				return false
+			}
+		}
+	}
+	if p.MasterProfile.Distro == Ubuntu {
```
CecileRobertMichon: this could be `return p.MasterProfile.Distro == Ubuntu`

pkg/api/types.go
```diff
@@ -1015,6 +1015,21 @@ func (p *Properties) HasZonesForAllAgentPools() bool {
 	return false
 }
 
+// IsUbuntuDistroForAllAgentPools returns true if all of the agent pools plus masters are running the base Ubuntu image
+func (p *Properties) IsUbuntuDistroForAllAgentPools() bool {
```
CecileRobertMichon: Function name doesn't accurately reflect what it's actually doing since it implies the check is for agent pools only, should be `HasNonUbuntuDistroProfiles` (then reverse the true/false returned), `IsUbuntuDistroForAllNodes`  or something like that.

pkg/api/types.go
```diff
@@ -1015,6 +1015,21 @@ func (p *Properties) HasZonesForAllAgentPools() bool {
...
+	if len(p.AgentPoolProfiles) > 0 {
+		for _, ap := range p.AgentPoolProfiles {
+			if ap.Distro != Ubuntu {
```
CecileRobertMichon: `&& ap.Distro != Ubuntu1804` for 18.04 testing

pkg/api/types.go
CecileRobertMichon: `return p.MasterProfile.Distro == Ubuntu || p.MasterProfile.Distro == Ubuntu1804` actually

pkg/api/types.go
jackfrancis: done

pkg/api/types.go
jackfrancis: done

pkg/api/types.go
jackfrancis: done

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -88,6 +88,22 @@ var _ = BeforeSuite(func() {
 
 var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", func() {
 	Describe("regardless of agent pool type", func() {
+		It("should set up ssh key forwarding to the master node", func() {
```
CecileRobertMichon: this is failing e2e

.pipelines/e2e-step-template.yaml
```diff
@@ -20,7 +20,7 @@ steps:
       export CLIENT_ID=$(SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES)
       export CLIENT_SECRET=$(SERVICE_PRINCIPAL_CLIENT_SECRET_E2E_KUBERNETES)
       export CLIENT_OBJECTID=$(SERVICE_PRINCIPAL_OBJECT_ID_E2E_KUBERNETES)
-      make test-kubernetes
+      bash -c 'eval `ssh-agent` && make test-kubernetes'
```
CecileRobertMichon: do we need to add this to Jenkins tests as well? What about running tests locally? Should it just be part of the make target?

jackfrancis: So this actually works. (which is an improvement) Next step is to see if we can do one of your alternate, preferable proposals.

pkg/api/types_test.go
```diff
@@ -743,6 +743,140 @@ func TestHasAvailabilityZones(t *testing.T) {
...
+					},
+				},
+			},
+			expected: false,
+		},
+	}
+
+	for _, c := range cases {
+		if c.p.IsUbuntuDistroForAllNodes() != c.expected {
+			t.Fatalf("expected IsUbuntuDistroForAllNodes() to return %t but instead returned %t", c.expected, c.p.IsUbuntuDistroForAllNodes())
```
CecileRobertMichon: can you add a check for the `isUbuntu` functions in here?

parts/k8s/cloud-init/artifacts/cis.sh
```diff
@@ -1,3 +1,16 @@
 #!/bin/bash
 
-# Placeholder for future CIS-related shell utilities
\ No newline at end of file
+assignRootPW() {
+  grep '^root:[!*]:' /etc/shadow
+  if [ $? -eq '0' ] ; then
```
mboersma: This could be simplified to:
`if grep '^root:[!*]:' /etc/shadow; then`

parts/k8s/cloud-init/artifacts/cis.sh
```diff
@@ -1,3 +1,16 @@
...
+    SALT=`openssl rand -base64 5`
+    SECRET=`openssl rand -base64 37`
+    CMD="import crypt, getpass, pwd; print crypt.crypt('$SECRET', '\$6\$$SALT\$')"
+    HASH=`python <<< "$CMD"`
```
mboersma: `python -c "$CMD"` is a more standard way to run a script in a string.

pkg/api/types_test.go
jackfrancis: done

parts/k8s/cloud-init/artifacts/cis.sh
jackfrancis: punt on this change

parts/k8s/cloud-init/artifacts/cis.sh
jackfrancis: This one we can do, proved that it's equivalent:

```
FrancisBookMS:aks-engine jackfrancis$ HASH=`python <<< "print('hello world')"`
FrancisBookMS:aks-engine jackfrancis$ echo $HASH
hello world
FrancisBookMS:aks-engine jackfrancis$ HASH=`python -c "print('hello world')"`
FrancisBookMS:aks-engine jackfrancis$ echo $HASH
hello world
```

PR #1027, opened 2019-04-11
refactor: shell script cleanup

mboersma:
**Reason for Change**:

Fixes three linting errors pointed out by `shellcheck`:
- [use `-n` instead of `! -z`](https://github.com/koalaman/shellcheck/wiki/SC2236)
- [remove literal carriage returns](https://github.com/koalaman/shellcheck/wiki/SC1017)
- [sudo doesn't affect redirects](https://github.com/koalaman/shellcheck/wiki/SC2024)

**Issue Fixed**:

Refs #953

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
It seems especially important to enable `shellcheck` on the suddenly active CSE script area. I'd like to get there, but I'll try to keep the PRs bite-size en route.


parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
```diff
@@ -2,9 +2,9 @@
 set -x
 source /opt/azure/containers/provision_source.sh
 
-sudo echo "  dns-search <searchDomainName>" >> /etc/network/interfaces.d/50-cloud-init.cfg
```
mboersma: Note that [SC2024](https://github.com/koalaman/shellcheck/wiki/SC2024) implies this was not working as expected. Maybe it wasn't an issue because the script is already run in a `root` context?

mboersma:
I'm going to close (and the two subsequent PRs I had queued up) this until after the [CIS backlog](https://github.com/Azure/aks-engine/projects/7) is completed.

PR #1028, opened 2019-04-11
fix: Windows + VMSS external svc routing broken

adelina-t:
The VMSSs for Linux and Windows agentpools follow a different
naming patterns. However, when populating azure.json, aks-engine
does not take that into account and we end up with the wrong name
for the primareyScaleSetName variable. Thus, the LB created for
service will have a empty backend pool, as there isn't a
VMSS with the desired name.

Properly populating the ARM variable for the primaryScaleSet will
fix this issue.

Fixes: #809

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon:
@adelina-t the unit tests failed on the lint step.

adelina-t:
Will check. Thank you!

jackfrancis:
These test failures are real, looking into them

pkg/api/types.go
```diff
@@ -898,12 +898,22 @@ func (p *Properties) GetNSGName() string {
 
 // GetPrimaryAvailabilitySetName returns the name of the primary availability set of the cluster
 func (p *Properties) GetPrimaryAvailabilitySetName() string {
```
jackfrancis: @xizhamsft this method is called to populate AKS metadata for the cluster record. It doesn't make sense that we'd have a static value for `PrimaryAvailabilitySetName` when there is no AvailabilitySet in the cluster. Does changing this method to return an empty string have any breaking effects on the AKS VMSS scenario?

pkg/engine/armvariables.go
```diff
@@ -340,9 +335,9 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 		} else {
 			masterVars["primaryAvailabilitySetName"] = ""
 		}
-		masterVars["primaryScaleSetName"] = ""
 		masterVars["vmType"] = "standard"
 	}
+	masterVars["primaryScaleSetName"] = cs.Properties.GetPrimaryScaleSetName()
```
jackfrancis: We can just set this variable using the getter now that the getter returns an empty string if the primary agent pool is not vmss.

pkg/engine/armvariables_test.go
```diff
@@ -124,7 +124,7 @@ func TestK8sVars(t *testing.T) {
 		"nsgName":                        "[concat(variables('masterVMNamePrefix'), 'nsg')]",
 		"orchestratorNameVersionTag":     "Kubernetes:" + testK8sVersion,
 		"primaryAvailabilitySetName":     "",
-		"primaryScaleSetName":            "[concat(parameters('orchestratorName'), '-agentpool1-',parameters('nameSuffix'), '-vmss')]",
+		"primaryScaleSetName":            cs.Properties.GetPrimaryScaleSetName(),
```
jackfrancis: This is the actual fix.

jackfrancis: In other words, we're validating that the value of the ARM variable `primaryScaleSetName` derives from the canonical `primaryScaleSetName()` method, not from an ARM expression string.

pkg/engine/template_generator.go
```diff
@@ -968,12 +968,6 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"GetMasterEtcdClientPort": func() int {
 			return DefaultMasterEtcdClientPort
 		},
-		"GetPrimaryAvailabilitySetName": func() string {
```
jackfrancis: These template layer functions are never called.

Issue #1032, opened 2019-04-11
Enable errcheck linter


mboersma:
[`errcheck`](https://github.com/kisielk/errcheck) is enabled by default in golangci-lint, but we have it disabled for historical reasons. In Go, ignoring errors returned from a function is often a bug. This should be enabled as part of `make test-style`.

PR #1033, opened 2019-04-11
refactor: add error checking to tests

mboersma:
**Reason for Change**:

Checks all errors that were previously ignored in test code.

**Issue Fixed**:

Refs #1032 
Refs #992, but purposely does not overlap with those changes.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
This is the first baby step en route to enabling `errcheck` by default. These seemed likely to be the least controversial changes, since they are only in test code and do not produce any new failures.


test/e2e/runner.go
```diff
@@ -93,22 +93,32 @@ func main() {
 		if provision || cfg.ForceDeploy {
 			log.Printf("Soak cluster %s does not exist or has expired\n", rg)
 			log.Printf("Deleting Resource Group:%s\n", rg)
-			acct.DeleteGroup(rg, true)
+			if err = acct.DeleteGroup(rg, true); err != nil {
+				log.Fatalf("Unexpected error deleting Resource Group: %s", err)
```
CecileRobertMichon: will this fail if the group doesn't exist?

test/e2e/runner.go
```diff
@@ -93,22 +93,32 @@ func main() {
...
+			}
 			log.Printf("Deleting Storage files:%s\n", rg)
-			sa.DeleteFiles(cfg.SoakClusterName)
+			if err = sa.DeleteFiles(cfg.SoakClusterName); err != nil {
+				log.Fatalf("Unexpected error deleting Storage files: %s", err)
```
CecileRobertMichon: Same question here

test/e2e/runner.go
mboersma: It will—it just calls `az group delete` under the covers, so if the resource group doesn't exist, the command returns 1 even with the `--no-wait` flag.

Maybe it's safer here to discard the error then:
`_ = acct.DeleteGroup(rg, true)`

jackfrancis:
Yeah, can we just not change any of the E2E behaviors at this time?

mboersma:
> there may be some cases in E2E where we weren't exiting on errors on purpose 

Agreed, although in those cases we should be explicit about ignoring the error by assigning it to `_`.

PR #1036, opened 2019-04-11
refactor: remove engine pkg custom script unnecessary functions

CecileRobertMichon:
**Reason for Change**:

Removes lines of code. Custom data no longer uses `GetKubernetesB64` functions anymore so they do not need to be exported now that they are only used by the engine package.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
lgtm

PR #1031, opened 2019-04-11
chore(CIS): Ensure permissions on all log files are configured

CecileRobertMichon:
**Reason for Change**:

CIS 4.2.4 Ensure permissions on all logfiles are configured 

**Issue Fixed**:

Fixes #1012 

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1039, opened 2019-04-12
chore(CIS): CIS network configuration enforcement

### LINKED ISSUE #988: CIS Network configuration enforcement
in `/etc/sysctl.d/60-CIS.conf`:

```
# 3.1.2 Ensure packet redirect sending is disabled (Scored)

net.ipv4.conf.all.send_redirects = 0
net.ipv4.conf.default.send_redirects = 0

# 3.2.1 Ensure source routed packets are not accepted (Scored) 

net.ipv4.conf.all.accept_source_route = 0
net.ipv4.conf.default.accept_source_route = 0

# 3.2.2 Ensure ICMP redirects are not accepted (Scored)

net.ipv4.conf.all.accept_redirects = 0
net.ipv4.conf.default.accept_redirects = 0

# 3.2.3 Ensure secure ICMP redirects are not accepted (Scored)

net.ipv4.conf.all.secure_redirects = 0
net.ipv4.conf.default.secure_redirects = 0

# 3.2.4 Ensure suspicious packets are logged (Scored)

net.ipv4.conf.all.log_martians = 1
net.ipv4.conf.default.log_martians = 1

# 3.3.1 Ensure IPv6 router advertisements are not accepted (Not Scored)

net.ipv6.conf.all.accept_ra = 0
net.ipv6.conf.default.accept_ra = 0

# 3.3.2 Ensure IPv6 redirects are not accepted (Not Scored)

net.ipv6.conf.all.accept_redirects = 0
net.ipv6.conf.default.accept_redirects = 0
```

jackfrancis:
**Reason for Change**:

Adds network configuration changes to accommodate CIS benchmarks.

Proposes a new "VHD file delivery" pattern that is partly re-usable. Primarily: we will maintain the file in one place, i.e., `parts/k8s/cloud-init/artifacts/`; however, the delivery implementation is distinct between "full install" flow vs VHD-backed flow.

Includes some new E2E stuffs to accomodate delivering a common shell script to run for validation across all nodes.

**Issue Fixed**:

Fixes #988 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -179,6 +179,7 @@ ps auxfww > /opt/azure/provision-ps.log &
 
 if $FULL_INSTALL_REQUIRED; then
   applyCIS || exit $ERR_CIS_HARDENING_ERROR
+  sysctl_reload 20 5 10 || exit $ERR_SYSCTL_RELOAD
```
jackfrancis: @CecileRobertMichon This is how we enable these changes for non-VHD scenario

CecileRobertMichon: Cool thanks. How about running that in the VHD build? Or will VMs starting from that VHD automatically pick up the configs on the disk?

CecileRobertMichon:
Commit message should be `chore(CIS): ...`

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: We don't need it in the VHD build process unless we want to run validation tests, because the configuration will be applied at next boot.

jackfrancis: Nevermind, we don't have a VHD implementation for these cloud-init-delivered changes, hmm...

CecileRobertMichon: yeah that was what I meant

CecileRobertMichon:
squash commit message is picked up from single commit in this case @jackfrancis.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -29,6 +29,12 @@ cis_script=/opt/azure/containers/provision_cis.sh
 wait_for_file 3600 1 $cis_script || exit $ERR_FILE_WATCH_TIMEOUT
 source $cis_script
 
+if [[ $OS == $UBUNTU_OS_NAME ]]; then
```
jackfrancis: @CecileRobertMichon moved this up here to run on every ubuntu CSE invocation (VHD or full install)

packer/install-dependencies.sh
```diff
@@ -4,10 +4,14 @@ source /home/packer/provision_source.sh
 source /home/packer/cis.sh
 
 RELEASE_NOTES_FILEPATH=/var/log/azure/golden-image-install.complete
+SYSCTL_CONFIG_SRC=/home/packer/sysctl-d-60-CIS.conf
+SYSCTL_CONFIG_DEST=/etc/sysctl.d/60-CIS.conf
 
 echo "Starting build on " `date` > ${RELEASE_NOTES_FILEPATH}
 echo "Using kernel:" >> ${RELEASE_NOTES_FILEPATH}
 cat /proc/version | tee -a ${RELEASE_NOTES_FILEPATH}
+cp $SYSCTL_CONFIG $SYSCTL_CONFIG_DEST && chmod 0644 $SYSCTL_CONFIG_DEST
```
jackfrancis: O.K. on second thought enabling this in VHD is pretty trivial and maintainable, we still have one source of truth with the exception of file mode enforcement.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -63,7 +63,10 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
-if [[ $OS != $COREOS_OS_NAME ]] && [[ $FULL_INSTALL_REQUIRED ]]; then
+if [[ $OS == $UBUNTU_OS_NAME ]] && [[ $FULL_INSTALL_REQUIRED ]]; then
```
jackfrancis: Now that this should be there via VHD we only do this in the full install scenario. Also changed `$OS != $COREOS_OS_NAME` to `$OS == $UBUNTU_OS_NAME` as that's a more expressive/accurate condition given the current level of OS support (i.e., we only support CoreOS and Ubuntu, so it's sort of unnecessarily obtuse to do the not condition IMO)

packer/install-dependencies.sh
CecileRobertMichon: not sure what that brings if we're already adding the file in cloud init. We should keep this pattern consistent, right now all other .conf files are written as part of cloud-init.

jackfrancis: It's to satisfy customers who want these configurations to not be applied "at runtime".

CecileRobertMichon: Makes sense. We should do this for all the other CIS files too then (etc/issue, etc.)

jackfrancis: Yeah, let's use this PR as a prototype, get the patterns how we want them

packer/install-dependencies.sh
```diff
@@ -4,10 +4,14 @@ source /home/packer/provision_source.sh
 source /home/packer/cis.sh
 
 RELEASE_NOTES_FILEPATH=/var/log/azure/golden-image-install.complete
+SYSCTL_CONFIG_SRC=/home/packer/sysctl-d-60-CIS.conf
+SYSCTL_CONFIG_DEST=/etc/sysctl.d/60-CIS.conf
 
 echo "Starting build on " `date` > ${RELEASE_NOTES_FILEPATH}
 echo "Using kernel:" >> ${RELEASE_NOTES_FILEPATH}
 cat /proc/version | tee -a ${RELEASE_NOTES_FILEPATH}
+cp $SYSCTL_CONFIG_SRC $SYSCTL_CONFIG_DEST && chmod 0644 $SYSCTL_CONFIG_DEST
```
jackfrancis: @CecileRobertMichon can you validate that this `cp` command is doing what I think it's doing? (i.e., paving a file onto the OS fs in order to ensure the file is baked into the VHD)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -108,7 +110,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			if eng.ExpandedDefinition.Properties.MasterProfile.IsUbuntu() {
 				kubeConfig, err := GetConfig()
 				Expect(err).NotTo(HaveOccurred())
-				master := fmt.Sprintf("azureuser@%s", kubeConfig.GetServerName())
+				master := fmt.Sprintf("%s@%s", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, kubeConfig.GetServerName())
```
jackfrancis: cleaned up a bunch of `"azureuser"` string literals in favor of deriving this from the api model

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -168,6 +170,37 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 		})
 
+		It("should validate Ubuntu host OS network configuration on all nodes", func() {
+			if eng.ExpandedDefinition.Properties.IsUbuntuDistroForAllNodes() {
+				kubeConfig, err := GetConfig()
```
jackfrancis: We can clean up some of this boilerplate in the future, but this is (IMO) O.K. as a pattern we can copy/paste develop for the remaining CIS changes

packer/vhd-image-builder.json
```diff
@@ -69,6 +69,11 @@
             "source": "packer/install-dependencies.sh",
             "destination": "/home/packer/install-dependencies.sh"
         },
+        {
+            "type": "file",
+            "source": "parts/k8s/cloud-init/artifacts/sysctl-d-60-CIS.conf",
+            "destination": "/home/packer/sysctl-d-60-CIS.conf"
```
CecileRobertMichon: no need to use cp, you can use this destination interface to pave the file where you want it to go, so here:

```suggestion
            "destination": "/etc/sysctl.d/60-CIS.conf"
```

packer/install-dependencies.sh
CecileRobertMichon: see my comment in vhd-image-builder.json

packer/vhd-image-builder.json
CecileRobertMichon: https://www.packer.io/docs/provisioners/file.html

CecileRobertMichon: "The file Packer provisioner uploads files to machines built by Packer. The recommended usage of the file provisioner is to use it to upload files, and then use shell provisioner to move them to the proper place, set permissions, etc."

jackfrancis: so chould we make the destination change or no? sounds like we keep the `chmod` statement in the install script

packer/install-dependencies.sh
CecileRobertMichon: once https://github.com/Azure/aks-engine/pull/1031 merges you can add the permissions there

packer/vhd-image-builder.json
CecileRobertMichon: I think you can upload it to the right place right away since /etc/ already exists. Worth trying at least. And then set permissions in the function added in #1031.

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -126,6 +126,15 @@ write_files:
   content: !!binary |
     {{CloudInitData "etcIssue"}}
 
+{{if not .IsVHDDistro}}
```
jackfrancis: @CecileRobertMichon thoughts on this pattern?

packer/install-dependencies.sh
```diff
@@ -4,10 +4,12 @@ source /home/packer/provision_source.sh
 source /home/packer/cis.sh
 
 RELEASE_NOTES_FILEPATH=/var/log/azure/golden-image-install.complete
+SYSCTL_CONFIG_SRC=/home/packer/sysctl-d-60-CIS.conf
```
CecileRobertMichon: is this still needed?

parts/k8s/cloud-init/nodecustomdata.yml
CecileRobertMichon: just one thought, do we want this for Coreos? If not `.IsUbuntu` might be a better fit

jackfrancis: The idea is to not use cloud-init for VHD-backed flows, because we've already done this work via packer.

packer/vhd-image-builder.json
jackfrancis: Done, VHD CI test should validate if that worked, as the `chmod 644 /etc/sysctl.d/60-CIS.conf` statement in `applyCIS` fail if the file doesn't exist.

jackfrancis:
Validated on an "ubuntu" distro cluster:

```
  should validate Ubuntu host OS network configuration on all nodes
  /Users/jackfrancis/work/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:173

$ k config view -o json

$ k get nodes -o json

$ scp -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -o StrictHostKeyChecking=no scripts/net-config-validate.sh azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com:/tmp/net-config-validate.sh
2019/04/12 14:52:06 Warning: Permanently added 'kubernetes-westus2-75990.westus2.cloudapp.azure.com,13.66.173.23' (ECDSA) to the list of known hosts.


$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -o ConnectTimeout=30 -o StrictHostKeyChecking=no azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com -p 22 scp -o StrictHostKeyChecking=no /tmp/net-config-validate.sh k8s-agentpool1-14444666-vmss000000:/tmp/net-config-validate.sh

$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -p 22 -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR k8s-agentpool1-14444666-vmss000000 "/tmp/net-config-validate.sh"
2019/04/12 14:52:08 0
0
0
0
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/send_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/send_redirects
+ cat /proc/sys/net/ipv4/conf/all/accept_source_route
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/accept_source_route
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/accept_redirects
+ grep 0
0
0
0
0
1
1
0
0
0
0
+ cat /proc/sys/net/ipv4/conf/default/accept_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/secure_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/secure_redirects
+ grep 0
+ grep 1
+ cat /proc/sys/net/ipv4/conf/all/log_martians
+ cat /proc/sys/net/ipv4/conf/default/log_martians
+ grep 1
+ cat /proc/sys/net/ipv6/conf/all/accept_ra
+ grep 0
+ cat /proc/sys/net/ipv6/conf/default/accept_ra
+ grep 0
+ cat /proc/sys/net/ipv6/conf/all/accept_redirects
+ grep 0
+ cat /proc/sys/net/ipv6/conf/default/accept_redirects
+ grep 0


$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -o ConnectTimeout=30 -o StrictHostKeyChecking=no azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com -p 22 scp -o StrictHostKeyChecking=no /tmp/net-config-validate.sh k8s-agentpool1-14444666-vmss000001:/tmp/net-config-validate.sh

$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -p 22 -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR k8s-agentpool1-14444666-vmss000001 "/tmp/net-config-validate.sh"
2019/04/12 14:52:10 + grep 0
+ cat /proc/sys/net/ipv4/conf/all/send_redirects
0
+ cat /proc/sys/net/ipv4/conf/default/send_redirects
+ grep 0
0
+ cat /proc/sys/net/ipv4/conf/all/accept_source_route
+ grep 0
0
+ cat /proc/sys/net/ipv4/conf/default/accept_source_route
+ grep 0
0
+ cat /proc/sys/net/ipv4/conf/all/accept_redirects
+ grep 0
0
0
0
0
+ cat /proc/sys/net/ipv4/conf/default/accept_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/secure_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/secure_redirects
+ grep 0
1
+ cat /proc/sys/net/ipv4/conf/all/log_martians
+ grep 1
+ cat /proc/sys/net/ipv4/conf/default/log_martians
+ grep 1
1
+ cat /proc/sys/net/ipv6/conf/all/accept_ra
+ grep 0
0
+ cat /proc/sys/net/ipv6/conf/default/accept_ra
+ grep 0
0
+ cat /proc/sys/net/ipv6/conf/all/accept_redirects
0
+ grep 0
+ cat /proc/sys/net/ipv6/conf/default/accept_redirects
0
+ grep 0


$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -o ConnectTimeout=30 -o StrictHostKeyChecking=no azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com -p 22 scp -o StrictHostKeyChecking=no /tmp/net-config-validate.sh k8s-master-14444666-0:/tmp/net-config-validate.sh

$ ssh -A -i /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-westus2-75990-ssh -p 22 -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR azureuser@kubernetes-westus2-75990.westus2.cloudapp.azure.com ssh -o ConnectTimeout=10 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=ERROR k8s-master-14444666-0 "/tmp/net-config-validate.sh"
2019/04/12 14:52:12 0
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/send_redirects
+ cat /proc/sys/net/ipv4/conf/default/send_redirects
+ grep 0
0
+ cat /proc/sys/net/ipv4/conf/all/accept_source_route
+ grep 0
0
+ cat /proc/sys/net/ipv4/conf/default/accept_source_route
+ grep 0
0
0
0
0
0
1
1
0
+ cat /proc/sys/net/ipv4/conf/all/accept_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/accept_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/secure_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/default/secure_redirects
+ grep 0
+ cat /proc/sys/net/ipv4/conf/all/log_martians
+ grep 1
+ cat /proc/sys/net/ipv4/conf/default/log_martians
+ grep 1
+ cat /proc/sys/net/ipv6/conf/all/accept_ra
+ grep 0
+ cat /proc/sys/net/ipv6/conf/default/accept_ra
+ grep 0
0
+ cat /proc/sys/net/ipv6/conf/all/accept_redirects
+ grep 0
0
+ cat /proc/sys/net/ipv6/conf/default/accept_redirects
+ grep 0
0

•
```

parts/k8s/cloud-init/nodecustomdata.yml
CecileRobertMichon: yes but my question is should we also add this file to coreos VMs?

jackfrancis: CoreOS is out of scope for CIS at this time

CecileRobertMichon: the way you're paving the file right now it IS being added to CoreOS VMs. What I'm saying is that maybe we should exclude CoreOS as well to not change that scenario (and use `.IsUbuntu` instead).

jackfrancis: I see what you're saying.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -65,6 +65,15 @@ write_files:
   content: !!binary |
     {{CloudInitData "systemConf"}}
 
+{{if .MasterProfile.IsUbuntuNonVHD}}
```
jackfrancis: @CecileRobertMichon this new method more explicitly provides what we're after

pkg/api/types.go
```diff
@@ -1175,6 +1180,11 @@ func (m *MasterProfile) IsUbuntu() bool {
 	return m.IsUbuntu1604() || m.IsUbuntu1804()
 }
 
+// IsUbuntuNonVHD returns true if the distro uses a base Ubuntu image
+func (m *MasterProfile) IsUbuntuNonVHD() bool {
+	return m.IsUbuntu() && !m.IsVHDDistro()
```
CecileRobertMichon: this is equivalent to the already existing `IsUbuntu` function, I don't think we need a new one.

jackfrancis:
```
	switch m.Distro {
	case AKS, Ubuntu, ACC1604:
		return true
```

CecileRobertMichon:
```
// IsUbuntu returns true if the master profile distro is any ubuntu distro
func (m *MasterProfile) IsUbuntu() bool {
	return m.IsUbuntu1604() || m.IsUbuntu1804()
}
```

jackfrancis: We are pasting the same thing. :)

`IsUbuntu()` returns true if the distro is `AKS` or `AKS1804`. That doesn't fulfill the requirements of "ubuntu non-VHD".

CecileRobertMichon: Oh wow I totally misread that, my bad

CecileRobertMichon: I thought `IsUbuntu1604` was literally the `Ubuntu1604` distro... that's a bit confusing.

parts/k8s/cloud-init/artifacts/cse_main.sh
CecileRobertMichon: I included this in #1043 since I'm already fixing that if statement

jackfrancis: thanks, I fixed the `[[ $FULL_INSTALL_REQUIRED ]];` thing in this PR as well

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -178,7 +181,9 @@ mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
 ps auxfww > /opt/azure/provision-ps.log &
 
 if $FULL_INSTALL_REQUIRED; then
-  applyCIS || exit $ERR_CIS_HARDENING_ERROR
+  if [[ $OS != $COREOS_OS_NAME ]]; then
```
CecileRobertMichon: let's change to `if [[ $OS == $UBUNTU_OS_NAME ]] && [ "$FULL_INSTALL_REQUIRED" = "true" ]; then` to match the above.

pkg/api/types.go
CecileRobertMichon: I still think we want this to be `return m.Distro == Ubuntu || m.Distro == Ubuntu1804` to exclude ACC1604 and ONLY include the vanilla Ubuntu image explicitly.

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: That messes up our else, I can change the inner block to if ubuntu, tho

pkg/api/types.go
jackfrancis: Why do we want to exclude `ACC1604`?

CecileRobertMichon: I don't know if we do but it's a special case https://github.com/Azure/aks-engine/blob/master/docs/topics/sgx.md and a custom image that the Azure Confidential Compute team maintains. @dmitsh might be able to tell us if we should apply CIS hardening for those.

CecileRobertMichon: I think it should be out of scope for now and we can optionally opt it in later if it turns out they do want it. We don't have good testing around it though so we risk breaking it with this PR if we don't validate that scenario.

jackfrancis: As long as it's an Ubuntu 16.04-LTS or 18.04-LTS base let's include it, it should work fine. These changes should not effect any k8s app layer behaviors, they should simply provide a better hardened OS underpinning for Ubuntu based on CIS recommendations.

jackfrancis: Making it out of scope essentially means bringing it out of the `IsUbuntu` conveniences. We can't reasonably manually work around the `ACC1604` case any other way.

CecileRobertMichon: I see what you're saying. We also need `IsUbuntuNonVHD` to be equivalent to 
```
$FULL_INSTALL_REQUIRED; then
```
otherwise the file permission step will fail.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -210,4 +211,16 @@ systemctl_restart() {
             sleep $wait_sleep
         fi
     done
+}
+sysctl_reload() {
+    retries=$1; wait_sleep=$2; timeout=$3
+    for i in $(seq 1 $retries); do
+        timeout $timeout sysctl --system
+        [ $? -eq 0  ] && break || \
```
mboersma: There's an extra space after `0`. Also I think this would be simpler as:
```shell
timeout $timeout sysctl --system && break || \
```

pkg/api/types.go
```diff
@@ -1200,6 +1210,11 @@ func (a *AgentPoolProfile) IsCoreOS() bool {
 	return a.OSType == Linux && a.Distro == CoreOS
 }
 
+// IsVHDDistro returns true if the distro uses VHD SKUs
+func (a *AgentPoolProfile) IsVHDDistro() bool {
```
mboersma: Too bad this logic has to be the same as the master version, but I guess there isn't a common interface.

jackfrancis: Indeed, we have this debt among all of the profile methods across master vs agent pool.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
jackfrancis: This is copy/pasta, let's use your suggested fix in a follow-up PR to all funcs in this helpers file

mboersma: Agreed—let me know whenever the CIS dust has settled enough, because I have lots of script refactoring and linting ready to unleash.

PR #1040, opened 2019-04-12
feat: pre-pull Azure Stack's custom Hyperkube

### LINKED ISSUE #990: Pre-pull Azure Stack's dependencies
No
---

FEATURE REQUEST
---

Master
---

**Kubernetes version**:

jadarsie:
**Reason for Change**:
Stores Azure Stack's custom Hyperkube build in the VHD image.

**Issue Fixed**:
Fixes #990

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
- The custom Hyperkube build are downloaded during the VHD provision process. 
- For now it is just single image, will add more entries to the array in a future PR
- I added an extra optional parameter to all functions that deal with Hyperkube to slightly alter their behavior (extractHyperkube, installKubeletAndKubectl, cleanUpContainerImages)


jadarsie:
I just blindly created this PR. Please let me know what validations do you usually run these set of scripts.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -253,7 +253,10 @@ installImg() {
 
 extractHyperkube() {
     CLI_TOOL=$1
-    path="/home/hyperkube-downloads/${KUBERNETES_VERSION}"
+    # CUSTOM_CLOUD_SUFFIX only set when extracting custom Hyperkube builds (e.x.: Azure Stack)
+    CUSTOM_CLOUD_SUFFIX=$2
+    kubernetes_version="${KUBERNETES_VERSION}${CUSTOM_CLOUD_SUFFIX}"
```
CecileRobertMichon: use a different var name here to avoid confusion with `KUBERNETES_VERSION`

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -298,8 +304,15 @@ pullContainerImage() {
 
 cleanUpContainerImages() {
     # TODO remove all unused container images at runtime
+    # CUSTOM_CLOUD_SUFFIX only set when extracting custom Hyperkube builds (e.x.: Azure Stack)
+    CUSTOM_CLOUD_SUFFIX=$1
     docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v ${KUBERNETES_VERSION} | grep 'hyperkube') &
     docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v ${KUBERNETES_VERSION} | grep 'cloud-controller-manager') &
+    if [[ "${CUSTOM_CLOUD_SUFFIX}" == "${AZURE_STACK_SUFFIX}" ]]; then
```
CecileRobertMichon: Can we check the value of `TARGET_ENVIRONMENT` here instead of setting a `CUSTOM_CLOUD_SUFFIX` arg?

parts/k8s/cloud-init/artifacts/cse_install.sh
CecileRobertMichon: on second thought, wouldn't it be cleaner to not touch this function and add the `AZURE_STACK_SUFFIX` to the `KUBERNETES_VERSION` var directly in install_dependencies.sh?

jadarsie: Sorry, I pushed changes before seeing these comments. Yes, no problem.

packer/install-dependencies.sh
```diff
@@ -365,11 +364,11 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
 done
 
 # TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
-KUBERNETES_VERSION_SUFFIX=${AZURE_STACK_SUFFIX}
 AZURE_STACK_K8S_VERSIONS="
 1.11.7
 "
 for AZS_KUBERNETES_VERSION in ${AZURE_STACK_K8S_VERSIONS}; do
+    KUBERNETES_VERSION=${AZS_KUBERNETES_VERSION}${AZURE_STACK_SUFFIX}
```
honcao: please use full name AZURESTACK to avoid the confusion.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -53,6 +53,9 @@ OS=$(cat /etc/*-release | grep ^ID= | tr -d 'ID="' | awk '{print toupper($0)}')
 UBUNTU_OS_NAME="UBUNTU"
 RHEL_OS_NAME="RHEL"
 COREOS_OS_NAME="COREOS"
+AZURE_SUFFIX=""
+AZURE_STACK_SUFFIX="-azs"
+AZURE_STACK_HYPERKUBE_REPOSITORY="mcr.microsoft.com/azurestack/k8s/core/hyperkube-amd64"
```
honcao: for now, Please msazurestackdocker/hyperkube-amd64  it will be for current public preview.
And later on, we can move to MCR

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -33,8 +33,10 @@ if [[ "${TARGET_ENVIRONMENT,,}" == "${AZURE_STACK_ENV}"  ]]; then
     config_script_custom_cloud=/opt/azure/containers/provision_configs_custom_cloud.sh
     wait_for_file 3600 1 $config_script_custom_cloud || exit $ERR_FILE_WATCH_TIMEOUT
     source $config_script_custom_cloud
+    CUSTOM_CLOUD_SUFFIX=${AZURE_STACK_SUFFIX}
 fi
 
+KUBERNETES_VERSION_SUFFIX="${CUSTOM_CLOUD_SUFFIX:-$AZURE_SUFFIX}"
```
CecileRobertMichon: instead of using empty variables, can we do:

```
if [[ "${TARGET_ENVIRONMENT,,}" == "${AZURE_STACK_ENV}"  ]]; then 
    KUBERNETES_VERSION=${KUBERNETES_VERSION}${AZURE_STACK_SUFFIX}
fi
```

?

jadarsie: Maybe, but I am a little hesitant to override `KUBERNETES_VERSION` in this case.

The reason is that I fear someone in the future my break Azure Stack's deployments by adding a feature that assumes the `X.X.X` format for `KUBERNETES_VERSION`. But I understand why you want that inside the `if` statement.

Example of this could be `cleanUpContainerImages`. I looks like it won't correctly if we add the suffix?

jadarsie: Thought about this on my way home. My concern is missing the point as we will only offer validated VHDs. Will update.

packer/install-dependencies.sh
```diff
@@ -363,6 +363,19 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
 done
 
+# TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
```
CecileRobertMichon: you can remove that TODO line, it doesn't apply to Azure Stack versions since it's a subset of versions.

packer/install-dependencies.sh
```diff
@@ -363,6 +363,19 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
...
+AZURE_STACK_K8S_VERSIONS="
+1.11.7
```
CecileRobertMichon: we usually support the latest two patches of each minor version, is supporting 1.11.7 a requirement for you or would 1.11.8 and 1.11.9 enough?

packer/install-dependencies.sh
jadarsie: I am hoping that Azure Stack versions can be validated some day in a similar way.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -33,6 +33,8 @@ if [[ "${TARGET_ENVIRONMENT,,}" == "${AZURE_STACK_ENV}"  ]]; then
     config_script_custom_cloud=/opt/azure/containers/provision_configs_custom_cloud.sh
     wait_for_file 3600 1 $config_script_custom_cloud || exit $ERR_FILE_WATCH_TIMEOUT
     source $config_script_custom_cloud
+    KUBERNETES_VERSION_NO_SUFFIX="${KUBERNETES_VERSION}"
+    KUBERNETES_VERSION=${KUBERNETES_VERSION}${AZURE_STACK_SUFFIX}
```
CecileRobertMichon: this breaks the scenario : Azure public cloud + Ubuntu distro (requiring full install)

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -33,6 +33,8 @@ if [[ "${TARGET_ENVIRONMENT,,}" == "${AZURE_STACK_ENV}"  ]]; then
     config_script_custom_cloud=/opt/azure/containers/provision_configs_custom_cloud.sh
     wait_for_file 3600 1 $config_script_custom_cloud || exit $ERR_FILE_WATCH_TIMEOUT
     source $config_script_custom_cloud
+    KUBERNETES_VERSION_NO_SUFFIX="${KUBERNETES_VERSION}"
```
CecileRobertMichon: instead of changing `KUBERNETES_VERSION`, let's make a new var `AZURE_STACK_KUBERNETES_VERSION` and use that in the `extractHyperkube` function if and only if the target env is Azure Stack. Are there any other places where the Azure Stack prefix is needed besides `extractHyperkube`?

packer/install-dependencies.sh
jadarsie: Our plan is to support two patches per minor version as well. 
It was unclear to me if 1.11.7 was still needed. Checked with Hongbin.

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: extractHyperkube, installKubeletAndKubectl, cleanUpContainerImages

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: ok, I will change the functions

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: let me know please if this my last reply changes the new approach in someway

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -180,6 +182,7 @@ ps auxfww > /opt/azure/provision-ps.log &
 if $FULL_INSTALL_REQUIRED; then
   applyCIS || exit $ERR_CIS_HARDENING_ERROR
 else
+  KUBERNETES_VERSION="${KUBERNETES_VERSION_NO_SUFFIX}"
```
CecileRobertMichon: why is this needed? `${KUBERNETES_VERSION_NO_SUFFIX}` will be unbound if the environment isn't Azure Stack

parts/k8s/cloud-init/artifacts/cse_main.sh
CecileRobertMichon: I'm trying to think of the best way to do this. We need to put all these changes in  `AZURE_STACK_ENV` flows as much as possible so we don't break non-Azure-Stack scenarios.

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: Needed because my understanding of your previous comments was to adjust the consumer code in order to keep the functions unchaged.

The unbound situation is an issue.

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: np, let me know please the approach you guys prefer
- alter functions vs alter code consuming functions?
- global variables vs function parameters?

CecileRobertMichon:
Before this change:

`/dev/sda1        30G   18G   12G  60%`

After this change:

` /dev/sda1        30G   19G   11G  64%`

parts/k8s/cloud-init/artifacts/cse_main.sh
CecileRobertMichon: A simple fix here would be to simply do this re-assignment only for Azure Stack. 

The only concern I have with this approach is that if new functions that rely on `KUBERNETES_VERSION` are introduced in CSE it might break Azure Stack to have the suffix still around. But I guess we can deal with that problem once it comes. Do you have any other ideas on how to make this more elegant?

jackfrancis:
@CecileRobertMichon thank you for tracking that!

I know this is a crazy one-liner, but we could add something like this in either VHD CI and/or E2E:

`[ -s $(df -P | grep '/dev/sda1' | awk '0+$5 >= 75 {print}') ]` where `75` is the percent threshold we want to alert at.

CecileRobertMichon:
@jackfrancis will do in a separate PR: #1064.

parts/k8s/cloud-init/artifacts/cse_main.sh
jadarsie: Given how the scripts are organized, I agree that overriding `KUBERNETES_VERSION` is probably not recommended. I pushed it because you suggested the change and I do not want to break your patterns/design.

However, things get more complicated because you also seem to dislike the idea of changing the functions. I think we won't be able to have it both ways.

> Do you have any other ideas on how to make this more elegant?

Another option is to pass parameters to the functions instead of relying on global variables like `KUBERNETES_VERSION`. That would be a bigger change though.

I can go either way, just let me know.

jackfrancis:
Can't we just add `1.11.8-azs` and `1.11.9-azs` to the existing versions array in the VHD CI script? Then  we just set the value of `KUBERNETES_VERSION` in `pkg/engine/armvariables.go` to have the proper string value if we're in an Azure Stack context.

CecileRobertMichon:
+1 on @jackfrancis's suggestion.

This should also work with the `cleanUpContainerImages` function the way you wrote it right now to clean up the Azure Stack images depending on context.

packer/install-dependencies.sh
```diff
@@ -361,19 +361,27 @@ K8S_VERSIONS="
 1.12.7
 1.12.6
 1.11.9
+1.11.9-azs
 1.11.8
+1.11.8-azs
```
CecileRobertMichon: I think you need to add back your previous changes to `cleanUpContainerImages` to clean up the azure stack image in public cloud environments and vice versa.

packer/install-dependencies.sh
```diff
@@ -361,19 +361,27 @@ K8S_VERSIONS="
...
-    HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
-    extractHyperkube "docker"
-    CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
-    pullContainerImage "docker" ${CONTAINER_IMAGE}
-    echo "  - ${HYPERKUBE_URL}" >> ${RELEASE_NOTES_FILEPATH}
-    echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
+    if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
+      HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      extractHyperkube "docker"
+      echo "  - ${HYPERKUBE_URL}" >> ${RELEASE_NOTES_FILEPATH}
```
CecileRobertMichon: these two lines 
```
extractHyperkube "docker"
echo "  - ${HYPERKUBE_URL}" >> ${RELEASE_NOTES_FILEPATH}
```
Can be extracted from the if/else statement since they are repeated in both.

pkg/engine/armvariables.go
```diff
@@ -74,6 +74,11 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 	isCustomVnet := cs.Properties.AreAgentProfilesCustomVNET()
 	hasAgentPool := len(profiles) > 0
 	hasCosmosEtcd := masterProfile != nil && to.Bool(masterProfile.CosmosEtcd)
+	
+	kubernetesVersion := orchProfile.OrchestratorVersion
+	if cs.Properties.IsAzureStackCloud() {
```
CecileRobertMichon: can we add a unit test for this?

packer/install-dependencies.sh
jadarsie: Quick valudation for the change in `cleanUpContainerImages `

```
KUBERNETES_VERSION="1.11.8"
printf '%s\n%s\n' msazurestackdocker/hyperkube-amd64:v1.11.8-azs k8s.gcr.io/hyperkube-amd64:v1.11.8 | \
grep -v "${KUBERNETES_VERSION}$" | \
grep 'hyperkube'

KUBERNETES_VERSION="1.11.8-azs"
printf '%s\n%s\n' msazurestackdocker/hyperkube-amd64:v1.11.8-azs k8s.gcr.io/hyperkube-amd64:v1.11.8 | \
grep -v "${KUBERNETES_VERSION}$" | \
grep 'hyperkube'
```

pkg/engine/armvariables.go
jadarsie: I fixed an existing test that I think covers the new execution flow.

pkg/engine/armvariables_test.go
```diff
@@ -527,7 +527,7 @@ func TestK8sVars(t *testing.T) {
...
...VIDER_RATELIMIT=',toLower(string(parameters('cloudproviderConfig').cloudProviderRatelimit)),' CLOUDPROVIDER_RATELIMIT_QPS=',parameters('cloudproviderConfig').cloudProviderRatelimitQPS,' CLOUDPROVIDER_RATELIMIT_BUCKET=',parameters('cloudproviderConfig').cloudProviderRatelimitBucket,' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension'),' USER_ASSIGNED_IDENTITY_ID=',variables('userAssignedClientID'),' USE_INSTANCE_METADATA=',variables('useInstanceMetadata'),' LOAD_BALANCER_SKU=',variables('loadBalancerSku'),' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB'),' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount'),' CONTAINER_RUNTIME=',parameters('containerRuntime'),' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase'),' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec'),' KMS_PROVIDER_VAULT_NAME=',variables('clusterKeyVaultName'),' IS_HOSTED_MASTER=false',' PRIVATE_AZURE_REGISTRY_SERVER=',parameters('privateAzureRegistryServer'),' AUTHENTICATION_METHOD=',variables('customCloudAuthenticationMethod'),' IDENTITY_SYSTEM=',variables('customCloudIdentifySystem'))]", testK8sVersion),
```
jackfrancis: Instead of manually adding the "-azs" suffix inline simply change the value of `testK8sVersion` (or make an Azure Stack-specific variable to add to the end of the `fmt.Sprintf()` func.

packer/install-dependencies.sh
```diff
@@ -361,19 +361,25 @@ K8S_VERSIONS="
...
 1.9.11
 1.9.10
 "
 for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
-    HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+    if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
+      HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
+    else
+      HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
+      CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
```
jackfrancis: I think all of this should be outside of the if block, right? The only conditional change is to `HYPERKUBE_URL`, correct?

jadarsie: Only the hypercube image is different, cloud-controller-manager remains the same

jackfrancis: The point is, this implementation doesn't actually pre-pull the image in the Azure Stack case. The if should be:

```
if [[ $KUBERNETES_VERSION == *"azs"* ]]; then
  HYPERKUBE_URL="msazurestackdocker/hyperkube-amd64:v${KUBERNETES_VERSION}"
else
fi
```

Unless I'm missing something.

jadarsie: Function `extractHyperkube` calls `pullContainerImage $CLI_TOOL ${HYPERKUBE_URL}`.

That clears your concern, right?

PR #997, opened 2019-04-10
refactor: return error instead of invoking log.Fatal

honcao:
**Reason for Change**:

Fix issue #963

**Issue Fixed**:

#963

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1041, opened 2019-04-12
refactor: don't call log.Fatal except from main.go

### LINKED ISSUE #963: refactor: do not call log.Fatalf except main.go

---

ISSUE
---

latest master
---

**Kubernetes version**:

**What happened**:

refactor: do not call log.Fatalf except main.go

mboersma:
**Reason for Change**:

Removes usage of [`log.Fatal`](https://golang.org/pkg/log/#Fatal) from non-test code.

`log.Fatal` and friends call `os.Exit(1)`, breaking the return flow of Go errors and making unit testing impossible.

**Issue Fixed**:

Fixes #963
See also #997

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/transform/transform.go
+++ b/pkg/engine/transform/transform.go
@@ -427,13 +427,13 @@ func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry
 func (t *Transformer) NormalizeResourcesForK8sAgentUpgrade(logger *logrus.Entry, templateMap map[string]interface{}, isMasterManagedDisk bool, agentPoolsToPreserve map[string]bool) error {
 	logger.Infoln(fmt.Sprintf("Running NormalizeResourcesForK8sMasterUpgrade...."))
 	if err := t.NormalizeResourcesForK8sMasterUpgrade(logger, templateMap, isMasterManagedDisk, agentPoolsToPreserve); err != nil {
-		log.Fatalln(err)
+		log.Println(err)
 		return err
 	}
 
 	logger.Infoln(fmt.Sprintf("Running NormalizeForK8sVMASScalingUp...."))
 	if err := t.NormalizeForK8sVMASScalingUp(logger, templateMap); err != nil {
-		log.Fatalln(err)
+		log.Println(err)
 		return err
 	}
 
--- a/pkg/engine/transform/transform.go
+++ b/pkg/engine/transform/transform.go
@@ -5,7 +5,6 @@ package transform
 
 import (
 	"fmt"
-	"log"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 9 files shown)


pkg/engine/virtualmachines.go
```diff
@@ -131,13 +131,16 @@ func CreateVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
 	}
 
 	t, err := InitializeTemplateGenerator(Context{})
+	if err != nil {
+		panic(err)
```
mboersma: `panic()` is arguably as problematic as `log.Fatal`, but that was already the error-handling paradigm in this function.

pkg/engine/transform/transform.go
```diff
@@ -427,13 +427,13 @@ func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry
 func (t *Transformer) NormalizeResourcesForK8sAgentUpgrade(logger *logrus.Entry, templateMap map[string]interface{}, isMasterManagedDisk bool, agentPoolsToPreserve map[string]bool) error {
 	logger.Infoln(fmt.Sprintf("Running NormalizeResourcesForK8sMasterUpgrade...."))
 	if err := t.NormalizeResourcesForK8sMasterUpgrade(logger, templateMap, isMasterManagedDisk, agentPoolsToPreserve); err != nil {
-		log.Fatalln(err)
+		log.Println(err)
```
CecileRobertMichon: do we need to print the error before we return it? If we're doing things right the error should get logged later on by the caller.

pkg/engine/transform/transform.go
```diff
@@ -427,13 +427,13 @@ func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry
...
 		return err
 	}
 
 	logger.Infoln(fmt.Sprintf("Running NormalizeForK8sVMASScalingUp...."))
 	if err := t.NormalizeForK8sVMASScalingUp(logger, templateMap); err != nil {
-		log.Fatalln(err)
+		log.Println(err)
```
CecileRobertMichon: same comment

pkg/engine/virtualmachines.go
CecileRobertMichon: Could we change the function to return an error instead? I did a quick search and it doesn't seem like it's called in that many places so it wouldn't be that big of a change. Definitely agree we should try to avoid using panic().

pkg/engine/virtualmachinescalesets.go
```diff
@@ -200,13 +200,16 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 	}
 
 	t, err := InitializeTemplateGenerator(Context{})
+	if err != nil {
+		panic(err)
```
CecileRobertMichon: same thing here

pkg/operations/remote_ssh.go
```diff
@@ -17,7 +17,8 @@ func RemoteRun(user string, addr string, port int, sshKey []byte, cmd string) (s
 	// Create the Signer for this private key.
 	signer, err := ssh.ParsePrivateKey(sshKey)
 	if err != nil {
-		log.Fatalf("unable to parse private key: %v", err)
+		log.Printf("unable to parse private key: %v", err)
```
CecileRobertMichon: same question here about not printing the error before returning it.

pkg/engine/transform/transform.go
mboersma: Perhaps printing the error is redundant. I was trying to keep the changes minimal and obvious by just translating `Fatalln` to `Println`.

I'll take a look at removing these and the others you mentioned.

PR #1042, opened 2019-04-12
chore: remove comments from cilium addons spec

jackfrancis:
**Reason for Change**:

This saves > 25% against the payload of this yaml

Also includes a reference to `docker.io/cilium/operator:v1.4` instead of `docker.io/cilium/operator:latest` as per @skinny 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-cilium-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-cilium-daemonset.yaml
@@ -7,158 +7,28 @@ metadata:
   labels:
     addonmanager.kubernetes.io/mode: "EnsureExists"
 data:
-  # This etcd-config contains the etcd endpoints of your cluster. If you use
-  # TLS please make sure you follow the tutorial in https://cilium.link/etcd-config
   etcd-config: |-
     ---
     endpoints:
       - https://cilium-etcd-client.kube-system.svc:2379
-    #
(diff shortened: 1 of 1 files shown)


jackfrancis:
@skinny FYI

skinny:
Also the operator image tag in the current addon is set to :latest which breaks a bit against the 1.4 addon, but I will make a seperate PR for that after the weekend

jackfrancis:
> The yaml payload size is not the biggest deal right?

Actually it is. As long as we use Azure's cloud-init interface, we are bound to data limits. E.g.:

```
Deployment failed. Correlation ID: d8046089-f0da-4856-ad31-2c9cd052c35e. {
  "error": {
    "code": "InvalidParameter",
    "message": "Custom data in OSProfile must be in Base64 encoding and with a maximum length of 87380 characters.",
    "target": "customData"
  }
}
```

skinny:
Ah well that sort of rules out the chance of keeping this file in tact. While you merge this PR, can you also change the `:latest` version of the operator in this file (`image: docker.io/cilium/operator:latest`) to `image: docker.io/cilium/operator:v1.4` to keep stuff running now 1.5 is around the corner

jackfrancis:
@skinny done

PR #892, opened 2019-03-27
feat: revive CoreOS support

alexeldeib:
**Reason for Change**:

Fixes #584 

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:
I took the absolute dead simplest approach I could to get this working. Right now the masters and agents successfully form a cluster. Seems like there are some issues with rpc-statd and possibly volume mounts. Got some running pods though!

image

PR #1043, opened 2019-04-12
fix: only run installDeps in non-VHD scenarios

CecileRobertMichon:
**Reason for Change**:

#892 introduced a regression by using by checking `if [[ $FULL_INSTALL_REQUIRED ]]` which always evaluates to true since that var is assigned as a string. The correct way to check if it is true is with `[ "$FULL_INSTALL_REQUIRED" == "true" ]`. Also change the `OS` check to only run these lines for `UBUNTU since that's more explicit (and allows us to add more OSes later on without breaking this flow.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -63,7 +63,7 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
-if [[ $OS != $COREOS_OS_NAME ]] && [[ $FULL_INSTALL_REQUIRED ]]; then
+if [[ $OS == $UBUNTU_OS_NAME ]] && [ "$FULL_INSTALL_REQUIRED" == "true" ]; then
     installDeps
 else 
     echo "Golden image; skipping dependencies installation"
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -63,7 +63,7 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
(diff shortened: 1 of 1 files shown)


mboersma:
lgtm

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -63,7 +63,7 @@ else
     FULL_INSTALL_REQUIRED=true
 fi
 
-if [[ $OS != $COREOS_OS_NAME ]] && [[ $FULL_INSTALL_REQUIRED ]]; then
+if [[ $OS == $UBUNTU_OS_NAME ]] && [ "$FULL_INSTALL_REQUIRED" = "true" ]; then
```
jackfrancis: `[ "$FULL_INSTALL_REQUIRED" = "true" ]` is inconsistent with all the other `[[ $FULL_INSTALL_REQUIRED ]]` usages.

CecileRobertMichon: there are no other `[[ $FULL_INSTALL_REQUIRED ]]` usages

CecileRobertMichon: only `if $FULL_INSTALL_REQUIRED; then`

CecileRobertMichon: that's what broke it

CecileRobertMichon:
verified:

```
+ [[ UBUNTU == UBUNTU ]]
+ '[' false = true ']'
+ echo 'Golden image; skipping dependencies installation'
Golden image; skipping dependencies installation
```

PR #1044, opened 2019-04-12
fix: Ensure pods scheduled onto new nodes during upgrade respect the original node's labels/taints

chengliangli0918:
**Reason for Change**:
Ensure pods scheduled onto new nodes during upgrade respect the original node's labels/taints: 
Edge case to fix: on a multiple node AKS/aks-engine cluster, manually set taints to prevent a pod from being scheduled on any agent nodes, so that pod is always in Pending. After upgrading, that pod is scheduled on one of new nodes, although the annotations/labels/taints are copied over from old nodes to new nodes correspondingly

**Issue Fixed**:
1. Ensure pods scheduled onto new nodes during upgrade respect the original node's labels/taints by evicting any pods might be scheduled on to the newly created agent nodes before copying over node properties (annotations/labels/taints) from old node to new node

2. Move the recent retry logic added to VMSS agent nodes by xizhamsft to common place in order that both VMAS/VMSS agent pool share the same code path to copy over node properties from old nodes to new nodes.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -706,6 +725,7 @@ func (ku *Upgrader) copyCustomNodeProperties(client armhelpers.KubernetesClient,
 		}
 	}
 
+	newNode.Spec.Unschedulable = false
```
CecileRobertMichon: should this be done after the node is updated?

chengliangli0918: Based on the tests for the edge case that this PR tries to fix, it seems ok to put "copying node properties and re-enabling scheduling' in the same UpdateNode() call.

palma21: If it's done after the update then the pending pods will be incorrectly scheduled on this node before it receives the ported taints. Unless you see a better way to do it

CecileRobertMichon: Isn't it the opposite? We want to make the node schedulable _after_ it has the correct taints no? otherwise pods might get scheduled incorrectly in the short time before it gets the ported taints.

palma21: Correct, so making the node schedulable after all the taints are applied only. 
I believe we're saying the same thing. So will stand down for Charlie to explain the actual code.

chengliangli0918: Although using single UpdateNode() call to apply node properties and make it schedulable works fine according to tests.  I change the code to use two UpdateNode() calls separately to apply the node properties first and make the node schedulable next.

CecileRobertMichon:
@jackfrancis what's the reason for holding this one? I think maybe you meant to hold #976

jackfrancis:
I noticed that E2E looked good and thought this one needed more review/testing. Just didn't want it to auto-merge too quickly.

jackfrancis:
This lgtm

@chengliangli0918 just a thought: in the future, should we taint an upgraded node at creation time so that no workloads are initially scheduled? And then we do all of the post-creation operations, and then finally we untaint the node? Basically something like this should reduce the edge case surface area of workloads being scheduled "during" an upgrade.

chengliangli0918:
agree. In the future, we should set UnSchedule=true when the node is created, and then set UnSchedule=false if creation and post-creation configuration is done. This will need code change for scenarios, like new cluster creation, cluster scaling up, cluster upgrading for both AKS/AKS-engine. This might need more efforts to achieve.
  
> This lgtm
> @chengliangli0918 just a thought: in the future, should we taint an upgraded node at creation time so that no workloads are initially scheduled? And then we do all of the post-creation operations, and then finally we untaint the node? Basically something like this should reduce the edge case surface area of workloads being scheduled "during" an upgrade.

palma21:
@jackfrancis are we ready to merge this?

jackfrancis:
Running upgrade validation test now

jackfrancis:
FYI rebased/force-pushed so we can do back-compat tests

PR #1045, opened 2019-04-12
test: ensure node is ready before running validations

jackfrancis:
**Reason for Change**:

In some upgrade tests we're seeing the `k get nodes` return list of nodes after scale in. This change should allow us to further filter out non-Ready nodes to ensure we don't do e.g., ssh to node tests against nodes in a state of deletion that don't actually have functional VMs behind them.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
lgtm

CecileRobertMichon:
shouldn't the test fail in the first place if some nodes are not ready?

PR #1024, opened 2019-04-10
chore(CIS): ensure local login warning banner is configured properly

jackfrancis:
**Reason for Change**:

**Issue Fixed**:

Fixes #982

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1057, opened 2019-04-15
chore: backport tests + cloud-init usage for CIS changes

jackfrancis:
**Reason for Change**:

Follow-up from PRs #1024 and #1031 to include validation tests; and to only pave files via cloud-init in non-VHD flows.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cis.sh
+++ b/parts/k8s/cloud-init/artifacts/cis.sh
@@ -24,6 +24,11 @@ assignFilePermissions() {
     lastlog
     waagent.log
     syslog
+    unattended-upgrades/unattended-upgrades.log
+    azure-vnet-ipam.log
+    azure-vnet-telemetry.log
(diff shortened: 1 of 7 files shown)


parts/k8s/cloud-init/artifacts/cis.sh
```diff
@@ -24,6 +24,14 @@ assignFilePermissions() {
     lastlog
     waagent.log
     syslog
+    unattended-upgrades/unattended-upgrades.log
```
jackfrancis: These files may be added based on cluster config, so we need to prep their file perms properly

jackfrancis:
tested these changes (using "ubuntu" distro) against a variety of cluster configs:

- the cluster configs that PRs build/test against
- omsagent
- GPU-enabled

Everything checks out.

PR #1063, opened 2019-04-15
chore(CIS): add apt package validation, ensure postfix is not present

### LINKED ISSUE #985: CIS 2.2.15 Ensure mail transfer agent is configured for local-only mode
in `/etc/postfix/main.cf`:

```
# See /usr/share/postfix/main.cf.dist for a commented, more complete version


# Debian specific:  Specifying a file name will cause the first
# line of that file to be used as the name.  The Debian default
# is /etc/mailname.
#myorigin = /etc/mailname

smtpd_banner = $myhostname ESMTP $mail_name (Ubuntu)
biff = no

# appending .domain is the MUA's job.
append_dot_mydomain = no

# Uncomment the next line to generate "delayed mail" warnings
#delay_warning_time = 4h

readme_directory = no

# TLS parameters
smtpd_tls_cert_file=/etc/ssl/certs/ssl-cert-snakeoil.pem
smtpd_tls_key_file=/etc/ssl/private/ssl-cert-snakeoil.key
smtpd_use_tls=yes
smtpd_tls_session_cache_database = btree:${data_directory}/smtpd_scache
smtp_tls_session_cache_database = btree:${data_directory}/smtp_scache

# See /usr/share/doc/postfix/TLS_README.gz in the postfix-doc package for
# information on enabling SSL in the smtp client.

smtpd_relay_restrictions = permit_mynetworks permit_sasl_authenticated defer_unauth_destination
myhostname = localhost
alias_maps = hash:/etc/aliases
alias_database = hash:/etc/aliases
myorigin = /etc/mailname
mydestination = $myhostname, localhost
relayhost = 
mynetworks = 127.0.0.0/8 [::ffff:127.0.0.0]/104 [::1]/128
mailbox_size_limit = 0
recipient_delimiter = +
#inet_interfaces = all
inet_protocols = all

# 2.2.15 Ensure mail transfer agent is configured for local-only mode

inet_interfaces = localhost
```

jackfrancis:
**Reason for Change**:

CIS work item #985 is to validate postfix configuration; however, we don't ship postfix w/ node vms. This PR validates that, and adds some additional positive package validation to the packages we install via VHD CI or CSE.

**Issue Fixed**:

Fixes #985

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/types.go
```diff
@@ -1037,6 +1037,18 @@ func (p *Properties) IsUbuntuDistroForAllNodes() bool {
 	return p.MasterProfile.Distro == Ubuntu || p.MasterProfile.Distro == Ubuntu1804
 }
 
+// IsUbuntuForAllNodes returns true if all of the agent pools plus masters are running an Ubuntu-based OS
+func (p *Properties) IsUbuntuForAllNodes() bool {
```
CecileRobertMichon: All these new "is..." functions are getting confusing IMO... This one is very similar to `IsUbuntuDistroForAllNodes` even though it means something very different. Can we do anything to either make the names more descriptive/distinct or reuse logic instead of having one function for each? eg. `func (p *Properties) IsDistroForAllNodes(distros []Distro) bool {` with a better name.

CecileRobertMichon: I don't mind doing this refactor in a follow up PR as a proposal and you can tell me if it makes sense.

CecileRobertMichon: Also `IsUbuntu` does:

```
// IsUbuntu returns true if the master profile distro is any ubuntu distro
func (a *AgentPoolProfile) IsUbuntu() bool {
	return a.IsUbuntu1604() || a.IsUbuntu1804()
}
```
Which is different than what we want here I believe.

jackfrancis: Alternatively we can get rid of this method and wrap the entire validation script inside an "is my OS ubuntu" if condition. Would you prefer that?

CecileRobertMichon: Yes definitely.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -236,6 +236,32 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 		})
 
+		It("should validate installed software packages", func() {
```
CecileRobertMichon: This shouldn't run on Windows nodes

jackfrancis: fixed in latest commit

test/e2e/kubernetes/node/node.go
```diff
@@ -61,12 +61,13 @@ type Address struct {
 	Type    string `json:"type"`
 }
 
-// Info contains information like what version the kubelet is running
+// Info contains node information like what version the kubelet is running
 type Info struct {
```
jackfrancis: The actual k8s node data type spec calls this `NodeInfo` (hence the change above). But gofmt complains if you start a method with the same word that is the type you're adding a method against (i.e., it doesn't want you to call something `node.NodeAction` and wants you to call it `node.Action` instead).

So, I kept the `Info` struct name here, but fixed the other definitions above so that they would actually correctly parse the node JSON output.

test/e2e/kubernetes/node/node.go
```diff
@@ -94,6 +95,19 @@ func (n *Node) IsReady() bool {
 	return false
 }
 
+// IsLinux returns if the node is in a Ready state
+func (n *Node) IsLinux() bool {
+	return n.Status.NodeInfo.OperatingSystem == "linux"
+}
+
+// IsUbuntu returns if the node is in a Ready state
```
CecileRobertMichon: comment copy paste

test/e2e/kubernetes/node/node.go
```diff
@@ -94,6 +95,19 @@ func (n *Node) IsReady() bool {
 	return false
 }
 
+// IsLinux returns if the node is in a Ready state
```
CecileRobertMichon: ditto

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -186,6 +186,7 @@ apt_get_dist_upgrade() {
   apt_dist_upgrade_output=/tmp/apt-get-dist-upgrade.out
   for i in $(seq 1 $retries); do
     wait_for_apt_locks
+    export DEBIAN_FRONTEND=noninteractive
```
jackfrancis: Added this because postfix + dist-upgrade was adding a package dependency that yielded an interactive, blocking message in the dist-upgrade operation. Glad we caught this in testing. :)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -236,6 +236,34 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+			cmd := exec.Command("scp", "-i", masterSSHPrivateKeyFilepath, "-o", "StrictHostKeyChecking=no", filepath.Join(ScriptsDir, installedPackagesValidateScript), master+":/tmp/"+installedPackagesValidateScript)
+			util.PrintCommand(cmd)
+			out, err := cmd.CombinedOutput()
+			log.Printf("%s\n", out)
+			Expect(err).NotTo(HaveOccurred())
+			var conn *remote.Connection
+			conn, err = remote.NewConnection(kubeConfig.GetServerName(), "22", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, masterSSHPrivateKeyFilepath)
+			Expect(err).NotTo(HaveOccurred())
+			for _, node := range nodeList.Nodes {
+				if node.IsUbuntu() {
```
CecileRobertMichon: I like this way of checking for Ubuntu way better 👍

parts/k8s/cloud-init/artifacts/cse_helpers.sh
CecileRobertMichon: I thought we weren't installing postfix?

CecileRobertMichon: This PR is only changing E2E besides this change so why is this change needed?

jackfrancis: Sorry, yes!

Actually, it's possible my testing stumbled into a bug w/ dist-upgrade.

Basically, we need to be ensuring all apt operations in scripts have `DEBIAN_FRONTEND=noninteractive` context

PR #1073, opened 2019-04-16
fix: make conntrack more liberal on packets

### LINKED ISSUE #1068: Make conntrack more liberal on packets
See:

https://github.com/kubernetes/kubernetes/pull/74840

Let's follow the recommendation here:

https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/

e.g., `echo 1 > /proc/sys/net/ipv4/netfilter/ip_conntrack_tcp_be_liberal`

mboersma:
**Reason for Change**:

Works around a [TCP reset bug](https://github.com/kubernetes/kubernetes/issues/74839) in Kubernetes by setting the value of kernel parameter `net.netfilter.nf_conntrack_tcp_be_liberal` to `1` (true) before starting `kubelet`. 

Also piggy-backs on the `net-config-validate.sh` script to test this parameter as well as`net.ipv4.tcp_retries2` (which was also set by `kubelet.service`).

See [Debugging an Intermittent Connection Reset](https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/) for a good explanation of the issue.

**Issue Fixed**:

Fixes #1068 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/kubelet.service
+++ b/parts/k8s/cloud-init/artifacts/kubelet.service
@@ -12,6 +12,9 @@ ExecStartPre=/bin/mkdir -p /var/lib/kubelet
 ExecStartPre=/bin/mkdir -p /var/lib/cni
 ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
 ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
+# This is a workaround to this upstream Kubernetes issue:
+#  https://github.com/kubernetes/kubernetes/issues/74839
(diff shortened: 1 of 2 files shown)


jackfrancis:
lgtm

vishnukraj1111:
Just to provide an update on this. The fix mentioned in most of the places echo 1 > /proc/sys/net/netfilter/ip_conntrack_tcp_be_liberal cause the conntrack table to be full and all connectivity to the servers were lost.

mboersma:
Eeek—that's good to know. We reverted this change; I guess we won't reinstate it.

jallirs:
What was the actual fix for this issue? Also could someone correct https://kubernetes.io/blog/2019/03/29/kube-proxy-subtleties-debugging-an-intermittent-connection-reset/ with an "update" to show ip_conntrack_tcp_be_liberal not actually being a good fix?

SSanjeevi:
Any actual fix happened for this?

PR #1076, opened 2019-04-17
feat: grant user assigned identity 'Reader' role for hosted cluster

norshtein:
**Reason for Change**:

Background: there is an upcoming feature in AKS: support authenticating to Azure using MSI. In the implementation, agent nodes will use an auto-generated user assigned identity and master components won't use this identity to authenticate. In this case, it enough to grant the generated user assigned identity only "Reader" role in the created resource group. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/armresources.go
+++ b/pkg/engine/armresources.go
@@ -111,7 +111,7 @@ func createKubernetesAgentVMASResources(cs *api.ContainerService, profile *api.A
 				Name: to.StringPtr(fmt.Sprintf("[variables('%sAvailabilitySet')]",
 					profile.Name)),
 				AvailabilitySetProperties: &compute.AvailabilitySetProperties{},
-				Type: to.StringPtr("Microsoft.Compute/availabilitySets"),
+				Type:                      to.StringPtr("Microsoft.Compute/availabilitySets"),
 			},
 		}
 
--- a/pkg/engine/armresources.go
+++ b/pkg/engine/armresources.go
@@ -22,9 +22,15 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		userAssignedIDEnabled = useManagedIdentity && kubernetesConfig.UserAssignedID != ""
 	}
 
+	isHostedMaster := cs.Properties.IsHostedMasterProfile()
 	if userAssignedIDEnabled {
 		userAssignedID := createUserAssignedIdentities()
-		msiRoleAssignment := createMSIRoleAssignment()
+		var msiRoleAssignment RoleAssignmentARM
+		if isHostedMaster {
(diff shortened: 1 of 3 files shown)


norshtein:
The code is auto-formatted by VSCode, seems the formatted code is not in the format the linter want, so I manually change the format to keep aligned with `master` branch, hope CI can pass this time...

pkg/engine/armresources.go
```diff
@@ -22,9 +22,15 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		userAssignedIDEnabled = useManagedIdentity && kubernetesConfig.UserAssignedID != ""
 	}
 
+	isHostedMaster := cs.Properties.IsHostedMasterProfile()
 	if userAssignedIDEnabled {
 		userAssignedID := createUserAssignedIdentities()
-		msiRoleAssignment := createMSIRoleAssignment()
+		var msiRoleAssignment RoleAssignmentARM
+		if isHostedMaster {
+			msiRoleAssignment = createMSIRoleAssignment(IdentityReaderRole)
```
andyzhangx: @jackfrancis @CecileRobertMichon Shall we change this `ReaderRole` to `ContributorRole` on agent node? Since we are going to support CSI driver with MSI credential. And CSI driver will be installed on agent nodes, e.g. disk provisioner of azure disk CSI driver will be installed as a StatefulsSet on one agent node, `ReaderRole` MSI won't work since it needs write disk permission.

norshtein: If I understand this correctly, CSI should be an option, something like `cs.Properties.****.UseCSI`, 
and its default value is `false`, so I personally prefer to add a judgment somewhere: if CSI is used, use `Contributor` role on agent node; otherwise, use `Reader` role.

andyzhangx: In the future, it's not an option, it's a must, I think we need to enable it now.
CSI is beta now, would step into GA soon.

norshtein: Got that. This PR only affects cluster having hosted master. For a complete cluster using user assigned identity, the identity still has `Contributor` role, so it won't affect CSI's usage. When CSI becomes mandatory in AKS, we can revert this PR and grant the user assigned identity `Contributor` role. Does that make sense?

andyzhangx: @norshtein then why not enabled it now? AKS v1.13.x is already available now, if we enable `Contributor` role, then we could test CSI driver on AKS at the first moment when MSI is enabled on AKS.

norshtein: What's the status of CSI driver on AKS? Are users able to enable(and disable) CSI?

PR #1077, opened 2019-04-17
chore: add go mod support

andyliuliming:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1078, opened 2019-04-17
fix: go mod would not copy the non-go files by default

andyliuliming:
**Reason for Change**:

fix: go mod would not copy the non-go files by default, so do go-bindata here is better. which will help the projects which vendored aks-engine .

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


andyliuliming:
I'm not sure whether it will have impact on the i18n.

maybe the i18n is defered to the projects which vendored aks-engine?

mboersma:
@andyliuliming can you make sure your commit messages follow https://www.conventionalcommits.org/? Otherwise they don't show up in the changelog we generate.

andyliuliming:
> @andyliuliming can you make sure your commit messages follow https://www.conventionalcommits.org/? Otherwise they don't show up in the changelog we generate.

done, thanks.

mboersma:
We have tried before to check the `go-bindata`-generated files into version control, but found it created a `git rebase` nightmare. We reverted that change in #546. I don't think we want to do this again.

The tradeoff is that you have to run `make build` or `make generate` after `go get github.com/Azure/aks-engine`, but that's the compromise we chose.

jackfrancis:
What @mboersma said is correct. We can't accept this change as-is.

andyliuliming:
but if we switch to use the go module.
the go mod vendor will not copy the parts files for us.
so the go generate will fail at the repos which try to vendor aks-engien :(

mboersma:
So #1077 won't work without this change?

Theoretically, the binary data could be broken into small segments to ensure there was minimal contention over changes. If there were a very close mapping between the script or code and its `go-bindata` representation, then real-world merge conflicts on PRs would be meaningful.

But I doubt we would want to make things that complicated—if that approach would even work. In practice, checking in these files required *every* PR in the queue to be rebased every time one was merged.

andyliuliming:
> So #1077 won't work without this change?
> 
> Theoretically, the binary data could be broken into small segments to ensure there was minimal contention over changes. If there were a very close mapping between the script or code and its `go-bindata` representation, then real-world merge conflicts on PRs would be meaningful.
> 
> But I doubt we would want to make things that complicated—if that approach would even work. In practice, checking in these files required _every_ PR in the queue to be rebased every time one was merged.

then one no-compress option will help :)

re-send one PR: https://github.com/Azure/aks-engine/pull/1088

PR #1083, opened 2019-04-17
test: single host OS DNS test script

jackfrancis:
**Reason for Change**:

This simplifies the Linux host OS DNS E2E tests a bit.

1. consolidates all tests into a single shell script
2. runs this validation earlier in the E2E run as later tests will not work if host DNS (for example) is not healthy; in such cases we want to fail on the actual failure, which is host OS DNS
3. adds additional vm DNS resolution coverage: now all node vms will be validated against being able to resolve all nodes (previously we were just checking that the master could resolve node vns)
4. adds retry tolerance

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -106,6 +106,38 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(success).To(BeTrue())
 		})
 
+		It("should validate host OS DNS", func() {
+			kubeConfig, err := GetConfig()
+			Expect(err).NotTo(HaveOccurred())
+			master := fmt.Sprintf("%s@%s", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, kubeConfig.GetServerName())
+			nodeList, err := node.Get()
+			Expect(err).NotTo(HaveOccurred())
+			hostOSDNSValidateScript := "host-os-dns-validate.sh"
+			envString := "NODE_HOSTNAMES='"
+			for _, node := range nodeList.Nodes {
+				envString += fmt.Sprintf("%s ", node.Metadata.Name)
+			}
+			envString += "'"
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -106,6 +106,40 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+			Expect(err).NotTo(HaveOccurred())
+			for _, node := range nodeList.Nodes {
+				if node.IsLinux() {
+					err := conn.CopyToRemote(node.Metadata.Name, "/tmp/"+hostOSDNSValidateScript)
+					Expect(err).NotTo(HaveOccurred())
+					netConfigValidationCommand := fmt.Sprintf("\"%s /tmp/%s\"", envString, hostOSDNSValidateScript)
+					cmd = exec.Command("ssh", "-A", "-i", masterSSHPrivateKeyFilepath, "-p", masterSSHPort, "-o", "ConnectTimeout=10", "-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-o", "LogLevel=ERROR", master, "ssh", "-o", "ConnectTimeout=10", "-o", "StrictHostKeyChecking=no", "-o", "UserKnownHostsFile=/dev/null", "-o", "LogLevel=ERROR", node.Metadata.Name, netConfigValidationCommand)
+					util.PrintCommand(cmd)
+					out, err = cmd.CombinedOutput()
+					log.Printf("%s\n", out)
```
CecileRobertMichon: is it valuable to print the output if there's no error? Our E2E output already tends to be very bloated and hard to read so just want to make sure we don't print unnecessary stuff.

jackfrancis: It's valuable to be able to look at passing tests to compare to failed, flaky tests. (IMO)

CecileRobertMichon:
asking @mboersma for secondary review since this one is a little more complex.

PR #1085, opened 2019-04-17
feat: refactor to allow for isPreview to orchestrator version

GaneshaThirumurthi:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/orchestrators.go
+++ b/pkg/api/orchestrators.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 
 	"github.com/Azure/aks-engine/pkg/api/common"
-	v20170930 "github.com/Azure/aks-engine/pkg/api/v20170930"
 	"github.com/Azure/aks-engine/pkg/api/vlabs"
 	"github.com/blang/semver"
 	"github.com/pkg/errors"
@@ -80,15 +79,15 @@ func GetOrchestratorVersionProfileListVLabs(orchestrator, version string, window
 	return orchList, nil
 }
 
-// GetOrchestratorVersionProfileListV20170930 returns v20170930 OrchestratorVersionProfileList object per (optionally) specified orchestrator and version
-func GetOrchestratorVersionProfileListV20170930(orchestrator, version string) (*v20170930.OrchestratorVersionProfileList, error) {
+// GetOrchestratorVersionProfileList returns unversioned OrchestratorVersionProfileList object per (optionally) specified orchestrator and version
+func GetOrchestratorVersionProfileList(orchestrator, version string) (*OrchestratorVersionProfileList, error) {
 	apiOrchs, err := getOrchestratorVersionProfileList(orchestrator, version, false)
 	if err != nil {
 		return nil, err
 	}
-	orchList := &v20170930.OrchestratorVersionProfileList{}
+	orchList := &OrchestratorVersionProfileList{}
 	for _, orch := range apiOrchs {
-		orchList.Properties.Orchestrators = append(orchList.Properties.Orchestrators, ConvertOrchestratorVersionProfileToV20170930(orch))
+		orchList.Properties.Orchestrators = append(orchList.Properties.Orchestrators, orch)
 	}
 	return orchList, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/api/types.go
```diff
@@ -246,10 +246,27 @@ type OrchestratorVersionProfile struct {
 	OrchestratorProfile
 	// Whether this orchestrator version is deployed by default if orchestrator release is not specified
 	Default bool `json:"default,omitempty"`
+	// Whether version is in preview mode
+	IsPreview bool `json:"isPreview,omitempty"`
+	// Metadata for customers to help them select a healthy kubernetes version
```
CecileRobertMichon: this comment seems a bit vague, what does "help them select a healthy kubernetes version" mean?

what is a "healthy" k8s version? 

Kubernetes should start with an upper case.

pkg/api/orchestrators.go
```diff
@@ -80,15 +79,15 @@ func GetOrchestratorVersionProfileListVLabs(orchestrator, version string, window
 	return orchList, nil
 }
 
-// GetOrchestratorVersionProfileListV20170930 returns v20170930 OrchestratorVersionProfileList object per (optionally) specified orchestrator and version
-func GetOrchestratorVersionProfileListV20170930(orchestrator, version string) (*v20170930.OrchestratorVersionProfileList, error) {
+// GetOrchestratorVersionProfileList returns unversioned OrchestratorVersionProfileList object per (optionally) specified orchestrator and version
+func GetOrchestratorVersionProfileList(orchestrator, version string) (*OrchestratorVersionProfileList, error) {
```
CecileRobertMichon: there is already an existing `getOrchestratorVersionProfileList` function. How does this one compare to it?

pkg/api/orchestrators.go
```diff
@@ -80,15 +79,15 @@ func GetOrchestratorVersionProfileListVLabs(orchestrator, version string, window
...
 	apiOrchs, err := getOrchestratorVersionProfileList(orchestrator, version, false)
 	if err != nil {
 		return nil, err
 	}
-	orchList := &v20170930.OrchestratorVersionProfileList{}
+	orchList := &OrchestratorVersionProfileList{}
```
CecileRobertMichon: not sure we want to do this... @weinong PTAL

CecileRobertMichon:
@GaneshaThirumurthi please update the PR description for context.

pkg/api/types.go
GaneshaThirumurthi: Yes, it is vague. Jake's team has asked me for this parameter to allow them to provide customers with additional information. When you select a kubernetes version, you also happen to selecting the versions for many other components that are bundled. He wants to be able to inform customers of problems with any of those versions via this metadata. That way customers can make informed decisions about the version they select. 

I will upper case it and ask the boulder team if there is a better comment that could be used instead.

pkg/api/orchestrators.go
GaneshaThirumurthi: Good question, the public function is just wrapping the private function. The reason for this is that OrchestratorVersionProfileList has required metadata like ID and Name that AKS needs. If having two functions with the same name is confusing, we could rename the public function to be GetExtendedOrchestratorVersionProfileList. Also, this function could be housed in AKS RP instead of AKS-engine - I am going to double check that it can be moved and then move it over to avoid bloating aks-engine.

jackfrancis:
As I understand what is intended here from a high level, I'll make a suggestion:

Let's include a general `IsPreview` bool scenario that is attached to the hosted master api model object. AKS can then turn that on/off based on introspecting the entire cluster configuration. This would be (arguably) cleaner, more general, and would allow for continual maintenance, as the set of configuration options that AKS wants to represent as being equivalent to "preview" configuration is always changing.

@GaneshaThirumurthi @jakepearson thoughts?

GaneshaThirumurthi:
@jackfrancis I was able to refactor some aks code so that most of this PR is not needed. I am going to close this PR and open a new one with a much smaller, non-critical/ non-time-sensitive scope. Sorry for not closing this faster.

PR #1087, opened 2019-04-17
feat: support VMSS agent nodes with public IP.

### LINKED ISSUE #130: nodes with public IP addresses
To support use-cases like agones:

https://github.com/GoogleCloudPlatform/agones

We need nodes w/ public IP addresses. We should make this a 1st class option.

JunSun17:
**Reason for Change**:

Support VMSS agent nodes with public IP.

**Issue Fixed**:

Fixes #130

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
This feature is controlled by EnableVMSSNodePublicIP flag in agent pool. When activated, it will create a public IP for each VMSS nodes.

A reference doc is here: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-networking#creating-a-scale-set-with-public-ip-per-virtual-machine


CecileRobertMichon:
This fixes https://github.com/Azure/aks-engine/issues/130 correct?

@JunSun17 can you please make sure your commit message and PR title use the semantic commit model so they get picked up in the release notes?

pkg/api/types.go
```diff
@@ -1101,6 +1102,11 @@ func (m *MasterProfile) IsCustomVNET() bool {
 	return len(m.VnetSubnetID) > 0
 }
 
+// HasVMSSNodePublicIP returns true if EnableVMSSNodePublicIP flag is true.
+func (a *AgentPoolProfile) HasVMSSNodePublicIP() bool {
+	return a.EnableVMSSNodePublicIP != nil && *a.EnableVMSSNodePublicIP
```
CecileRobertMichon: no need for this extra function, you can use the built-in `to.Bool()` func wherever this function is used instead:

`to.Bool(a.EnableVMSSNodePublicIP)`

docs/topics/clusterdefinitions.md
```diff
@@ -568,6 +568,7 @@ A cluster can have 0 to 12 agent pool profiles. Agent Pool Profiles are used for
...
 | vmssOverProvisioningEnabled | no                                                                   | Use [Overprovisioning](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-design-overview#overprovisioning) with VMSS. This configuration is only valid on an agent pool with an `"availabilityProfile"` value of `"VirtualMachineScaleSets"`. Defaults to `false`                                                                                                                                                                                                                                                      |
+| enableVMSSNodePublicIP | no                                                                   | Enable creation of public IP on VMSS nodes. This configuration is only valid on an agent pool with an `"availabilityProfile"` value of `"VirtualMachineScaleSets"`. Defaults to `false`                                                                                                                                                                                                                                                      |
```
CecileRobertMichon: If this is to be an officially supported feature for aks-engine we need to also add the new apimodel field to vlabs. I think there's evidence that users want this (#130) so I vote that we do.

JunSun17:
> This fixes #130 correct?
> 
> @JunSun17 can you please make sure your commit message and PR title use the semantic commit model so they get picked up in the release notes?



> This fixes #130 correct?
> 
> @JunSun17 can you please make sure your commit message and PR title use the semantic commit model so they get picked up in the release notes?

It is a feature asked by EA, and the requirement is similar to #130.

pkg/api/vlabs/validate.go
```diff
@@ -407,6 +407,12 @@ func (a *Properties) validateAgentPoolProfiles(isUpdate bool) error {
 			}
 		}
 
+		if to.Bool(agentPoolProfile.EnableVMSSNodePublicIP) {
```
CecileRobertMichon: let's add this case to unit tests as well

JunSun17: Sure, done.

jackfrancis:
lgtm

CecileRobertMichon:
/lgtm

for the label

PR #1088, opened 2019-04-18
fix: commit generated go-bindata files with "--no-compress" option

andyliuliming:
so do go-bindata here will help projects which vendored aks-engine.
and add no-compress option to go-bindata, so we will not need to rebase it everytime.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
> add no-compress option to go-bindata, so we will not need to rebase it everytime

:+1: That's awesome if the fix is something simple like that we have overlooked.

/azp run pr-e2e

jackfrancis:
Just to be clear, the only de-optimization from this change will be the storage overhead of this file in the repo? ( And also to github's backend merge conflict validation service :| )

andyliuliming:
> Just to be clear, the only de-optimization from this change will be the storage overhead of this file in the repo? ( And also to github's backend merge conflict validation service :| )

yes, and the file size is not too big. so I think it's acceptable.

and for the future change, we can also verify if we change the files under "parts" folder, and the real change shows in the go files too :)

jackfrancis:
How often do we generate `pkg/i18n/translations_generated.go`? Should we add `--no-compress` to that as well?

andyliuliming:
> How often do we generate `pkg/i18n/translations_generated.go`? Should we add `--no-compress` to that as well?
do not know how often that file would be generated :)
this file is genearted by pkg/i18n/resourceloader.go
and '--no-compress' is added there too.

andyliuliming:
Hi @jackfrancis @mboersma , could you please help merge this PR if no other comments? so I can do the according change in ACSRP repo. thanks.

jackfrancis:
@mboersma to validate docs and E2E changes I just made

.pipelines/pr-e2e.yaml
```diff
@@ -55,6 +55,9 @@ jobs:
   - script: make validate-copyright-headers test-style
     displayName: Run linting rules
     workingDirectory: $(modulePath)
+  - script: make ensure-generated
+	  displayName: Check if generated code is up to date
```
mboersma: Bonus `\t` here.

docs/community/developer-guide.md
```diff
@@ -256,3 +256,18 @@ The following steps constitute the AKS Engine CI pipeline:
 1. The PR is code reviewed by the members of AKS Engine team
 1. Once the PR is approved and the end-to-end job has passed, the PR can now be merged into the master branch
 1. Once merged, another job is triggered to verify integrity of the master branch. This job is similar to the PR job.
+
+## Pull Requests and Generated Code
```
CecileRobertMichon: 👍

mboersma:
It's failing the `ensure-generated` check now. :-(

/lgtm cancel

mboersma:
Note that #1151 added the generated files to `make clean`, so this PR should undo that.

mboersma:
I think this is failing because Azure Pipelines runs `make ensure-generated` after the PR is merged to master. So if the PR isn't rebased against the master commit, there's an excellent chance the generated files will differ.

jackfrancis:
// TODO we need to add PR validation that confirms the hygiene of the source against the generated code within the fork/branch itself. Until then we can't move this forward.

andyliuliming:
@jackfrancis @mboersma  I took a look at the ensure-generated, I think that will protect us from the case that the generated go file not update to date. will generate the file again and pushing.

mboersma:
@andyliuliming I'm still worried that this will create a situation that requires PRs to be rebased often in order to pass `make ensure-generated`. 

Overall it seems that `/parts` is the most active area of change in the project, and any time something changes under that directory, the generated code needs to be updated. Once that's done, `make ensure-generate` will no longer pass for existing PRs: they will all need to be rebased against master. 

This situation has happened before and it is a frustrating cycle that repels contributors. We absolutely don't want to do that—AKS Engine is better served by making it easy for people to contribute even if it means our Go packaging is slightly clumsy.

I'm trying to think of ways around this but don't have any fresh ideas.

andyliuliming:
Hi @mboersma I suppose each PR-e2e build in VSTS will do 
pull the master and apply the commits in a seperate branch (just like the rebasing) and then run the tests, ours pipeline did not do that?

CecileRobertMichon:
Even if the pipeline does merge master before running the tests, developers will still run into merge conflicts (thus requiring a rebase) every time they change one of the files that requires an update to the generated code (eg. `parts/`). Running the pipeline tests on a rebased branch will only help `make ensure-generated` pass when no file that requires files to be generated was changed.

andyliuliming:
> Even if the pipeline does merge master before running the tests, developers will still run into merge conflicts (thus requiring a rebase) every time they change one of the files that requires an update to the generated code (eg. `parts/`). Running the pipeline tests on a rebased branch will only help `make ensure-generated` pass when no file that requires files to be generated was changed.

"developers will still run into merge conflicts (thus requiring a rebase) every time they change one of the files that requires an update to the generated code (eg. `parts/`)"

I think the git will do the auto-merge for us if there's no real code conflict after this code change.
we can do a test.

andyliuliming:
@mboersma @CecileRobertMichon any updates?
do you mind grant me the permission to trigger the pipeline, so I can do a test on if the "merge conflicts" will happen on a separate branch.

CecileRobertMichon:
@andyliuliming I've granted you write access you should now be able to trigger the pipeline using `/azp run pr-e2e`. (you don't need to trigger the pipeline to see if there are merge conflicts though, opening a PR is enough).

mboersma:
I do think `git` will merge changes in the generated files correctly in most cases. The bottleneck will be the `make ensure-generated` task.

Azure DevOps merges the branch into master before running tests, so any changes in master that haven't been incorporated into your branch will cause `make ensure-generated` to fail, since it will regenerate the files and there will be a diff. Then you need to rebase against master and force-push—and hope that no other changes to `parts/` have been committed to master in the meantime.

(I may be wrong: this chicken-egg situation is confusing.)

CecileRobertMichon:
https://github.com/Azure/aks-engine/pull/1255 seems to suggest that this is working

mboersma:
> #1255 seems to suggest that this is working

Yes...it does. I am happy to be wrong if so.

In that case I'm LGTM on this so long as we revert if it does somehow cause hassle for developers.

CecileRobertMichon:
> In that case I'm LGTM on this so long as we revert if it does somehow cause hassle for developers.

+ 1

jackfrancis:
/lgtm

we'll monitor this in the coming days to ensure it doesn't increase rate of master merge conflicts

PR #1090, opened 2019-04-18
feat: use all VMAS fault domains in a region

### LINKED ISSUE #671: Get FD count based on region
Use and assign the fault domain count based on region support as per:

https://github.com/MicrosoftDocs/azure-docs/blob/master/includes/managed-disks-common-fault-domain-region-list.md

mboersma:
**Reason for Change**:

Changes the hard-coded default of `2` for an availability set's `platformFaultDomainCount` to an ARM expression that will calculate the available number of fault domains in the deployment location. Also respects the current availability sets' value during scale and upgrade.

**Issue Fixed**:

Fixes #671

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
There is definitely some weirdness in here, so I'm looking for feedback. ~~There are no tests yet, but~~ I have tested it thoroughly by hand.

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -416,8 +416,10 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	return sc.saveAPIModel()
 }
 
+// setFaultDomainCount finds the platformFaultDomainCount for existing VM availability sets in
+// the cluster and updates corresponding fields in the the internal api model.
 func (sc *scaleCmd) setFaultDomainCount(ctx context.Context, resourceGroup string, vmasIds []string) error {
-	count := engine.PlatformFaultDomainCountNotSet
+	var count int
 
 	for _, id := range vmasIds {
 		// extract the last element of the id for VMAS name
@@ -427,14 +429,15 @@ func (sc *scaleCmd) setFaultDomainCount(ctx context.Context, resourceGroup strin
 		if err != nil {
 			return err
 		}
-		// TODO: return error if all VMASes don't agree on this value?
+
+		// Assume that all VMASes in the cluster share a value for platformFaultDomainCount
 		count = int(*vmas.AvailabilitySetProperties.PlatformFaultDomainCount)
+		sc.containerService.Properties.MasterProfile.PlatformFaultDomainCount = &count
+		for _, pool := range sc.containerService.Properties.AgentPoolProfiles {
+			pool.PlatformFaultDomainCount = &count
+		}
 		break
 	}
-
-	// Set the global value used to populate fault domain count.
-	// This assumes all VMASes in a cluster share the same value.
-	engine.PlatformFaultDomainCount = int32(count)
 
 	return nil
 }
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -266,9 +266,12 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 		}
 		highestUsedIndex = indexes[len(indexes)-1]
 
-		if err := sc.setFaultDomainCount(ctx, sc.resourceGroupName, availabilitySetIDs); err != nil {
+		// set the VMAS platformFaultDomainCount to match the existing value
+		fdCount, err := sc.client.GetAvailabilitySetFaultDomainCount(ctx, sc.resourceGroupName, availabilitySetIDs)
+		if err != nil {
 			return err
 		}
+		sc.containerService.SetPlatformFaultDomainCount(fdCount)
 
 		// Scale down Scenario
 		if currentNodeCount > sc.newDesiredAgentCount {
@@ -414,32 +417,6 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	}
 
 	return sc.saveAPIModel()
-}
-
-// setFaultDomainCount finds the platformFaultDomainCount for existing VM availability sets in
-// the cluster and updates corresponding fields in the the internal api model.
-func (sc *scaleCmd) setFaultDomainCount(ctx context.Context, resourceGroup string, vmasIds []string) error {
-	var count int
-
-	for _, id := range vmasIds {
-		// extract the last element of the id for VMAS name
-		ss := strings.Split(id, "/")
-		name := ss[len(ss)-1]
-		vmas, err := sc.client.GetAvailabilitySet(ctx, resourceGroup, name)
-		if err != nil {
-			return err
-		}
-
-		// Assume that all VMASes in the cluster share a value for platformFaultDomainCount
-		count = int(*vmas.AvailabilitySetProperties.PlatformFaultDomainCount)
-		sc.containerService.Properties.MasterProfile.PlatformFaultDomainCount = &count
-		for _, pool := range sc.containerService.Properties.AgentPoolProfiles {
-			pool.PlatformFaultDomainCount = &count
-		}
-		break
-	}
-
-	return nil
 }
 
 func (sc *scaleCmd) saveAPIModel() error {
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -223,6 +223,8 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 	indexes := make([]int, 0)
 	indexToVM := make(map[int]string)
 	if sc.agentPool.IsAvailabilitySets() {
+		availabilitySetIDs := []string{}
+
 		for vmsListPage, err := sc.client.ListVirtualMachines(ctx, sc.resourceGroupName); vmsListPage.NotDone(); err = vmsListPage.Next() {
 			if err != nil {
 				return errors.Wrap(err, "failed to get vms in the resource group")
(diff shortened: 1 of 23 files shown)


cmd/scale.go
```diff
@@ -404,6 +414,29 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
...
+
+	for _, id := range vmasIds {
+		// extract the last element of the id for VMAS name
+		ss := strings.Split(id, "/")
+		name := ss[len(ss)-1]
+		vmas, err := sc.client.GetAvailabilitySet(ctx, resourceGroup, name)
+		if err != nil {
+			return err
+		}
+		// TODO: return error if all VMASes don't agree on this value?
```
mboersma: An optimization here is to return as soon as one availability set's `platformFaultDomainCount` is found, on the assumption that all availability sets in the cluster will share that value. This should hold true for any cluster created with `aks-engine` or AKS, and it's not possible to change the value after creation (or this PR would have been quite a bit simpler).

My first approach was to build a map of `vmasName` -> `platformFaultDomainCount` so I could update each VMAS resource in the template specifically. But the names are referenced only very indirectly at template generation time, so the association could not easily be made, and in practice the values could always boil down to a single number.

But it's worth asking if this is a short-sighted optimization.

pkg/engine/armtype.go
```diff
@@ -60,6 +63,40 @@ type AvailabilitySetARM struct {
 	compute.AvailabilitySet
 }
 
+// MarshalJSON is the custom marshaler for an AvailabilitySetARM.
+// It acts as a decorator by replacing the JSON field "platformFaultDomainCount"
```
mboersma: Alternatives considered here:

- `json.Marshal` an anonymous struct that matches `AvailabilitySetARM` except that it has a string in the `platformFaultDomainField`. This is a common approach to mutating default output in a `MarshalJSON` function, but it's verbose and would break if a `compute.AvailabilitySet` ever changes.
- Decorate the `compute.AvailabilitySet` struct instead, or perhaps its `Properties` embedded struct where the field we care about actually resides. The default behavior for those structs is in vendored classes, and while it may be possible to type alias one of those structs (as we do below) to avoid blowing the stack with infinite recursion, I couldn't figure out how.
- Just change our `AvailabilitySetARM` to use a wrapper class around `compute.AvailabilitySet` that exposes `platformFaultDomainCount` as a string rather than an int32. Then have the default be our funky ARM expression and an actual value be a simple expression like `"[toString(3)]"` and set the value on the availability set in the apimodel before template generation. Simple...except availability sets aren't really in the model, they're really just template entities.

pkg/engine/availabilitysets.go
```diff
@@ -11,6 +11,13 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
+// PlatformFaultDomainCountUnset indicates an ARM expression should replace this value.
+const PlatformFaultDomainCountNotSet = 0
+
+// PlatformFaultDomainCount is the global value used to populate the fault domain count
+// for a VMAS. It may be changed when upgrade inspects an existing cluster.
+var PlatformFaultDomainCount int32 = PlatformFaultDomainCountNotSet
```
mboersma: This variable is being (ab)used as an API to tell template generation what value to set for all availability sets' `platformFaultDomainCount`. It is set by upgrade and scale: otherwise the default ARM expression might calculate a value that differs with an existing availability set's value, which is a deployment error.

It would be preferable to carry this context in the `api.ContainerService` itself if that is possible without polluting it.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -224,6 +231,33 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
...
 	}
 
+	if err := uc.setFaultDomainCount(ctx, resourceGroup, availabilitySetIDs); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (uc *UpgradeCluster) setFaultDomainCount(ctx context.Context, resourceGroup string, vmasIds []string) error {
```
mboersma: This function is copied with only slight modification from cmd/scale.go. I could not find a shared code location that did not cause a circular import, and it seems too specialized to add to `AzureClient` itself. But the duplication still needs to be addressed.

scripts/fault_domains_expr.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python3
+
+"""
+Generates an Azure Resource Manager (ARM) expression that will evaluate at
+deployment time to the number of fault domains available in a given location.
+Also generates Go code referencing the expression.
```
mboersma: This is a script that a human runs that emits copy-and-pasteable Go code. There is no automation, but the worst that can happen is an Azure location is upgraded from 2 to 3 fault domains, and there is a time period before we re-run the script and PR the changes where we fail to take advantage of that.

This could (should?) be rewritten in Go and added as a pre-flight check to `aks-engine deploy|generate`. Then we could regenerate the ARM expression just-in-time, although we would want to have a short timeout and fallback logic to ensure that never failed a deployment.

pkg/engine/availabilitysets.go
jackfrancis: Can we just add an additional `PlatformFaultDomainCount *int` property to the `AgentPoolProfile` type? Then we can infer "not set" by a `nil` value.

mboersma: Done.

pkg/operations/kubernetesupgrade/upgradecluster.go
CecileRobertMichon: Can this be unit tested? same for the one in scale.go

mboersma: I think so—I'm adding tests now—but I'd really like not to duplicate this first, if possible.

palma21:
@mboersma is there a lot of work required to merge this one or are we at the last mile? 

CC @craiglpeters

mboersma:
@palma21 I'd say last mile, pending real reviews. Let me re-re-rebase it and see if there are test coverage deficits. Sorry I let it drop!

palma21:
Btw @mboersma finally got this reply wrt to ARM API.

Something like `az vm list-skus -r availabilitySets -l westus` for example
gives you the MaximumPlatformFaultDomainCount for that region (the Aligned ones are the important ones)

jackfrancis:
@mboersma We know for sure that changing the FD value in a pre-existing ARM template / rg and "re-deploying" to the rg (which is essentially how scale and upgrade are implemented) will result in an ARM template validation error, right?

If so, does the code skip "find the FD value for this AvailabilitySet" logic and just re-use the existing value?

mboersma:
> changing the FD value in a pre-existing ARM template...will result in an ARM template validation error, right?

Indeed, which is basically what made this change difficult other than the just-in-time ARM formula. In upgrade, this PR respects the existing FD count [here](https://github.com/Azure/aks-engine/pull/1090/files#diff-95227b07e97ba7cbd04cfd8b2cdb6980R292).

jackfrancis:
Great, thank you. Next question ;) How do we handle applying a static FD value (probably 2) when the "introspect using manual foo what the FD value should be for this availability set"? Basically is there a "if anything fails just use 2" implementation?

mboersma:
> is there a "if anything fails just use 2" implementation?

How that works is that `AvailabilitySet.PlatformFaultDomainCount` defaults to nil. In upgrade it gets set to the existing value for the VMAS. Then when AKS Engine marshals the JSON for an `AvailabilitySetARM`, if it's set to a real value it gets passed through. If it's `nil` it gets replaced with an ARM expression that will evaluate the max `PlatformFaultDomainCount` based on the deployment's region.

(That last part basically has to be a lazy-evaluated ARM expression since `location` isn't available until template deployment time.)

jackfrancis:
@mboersma Got it. So the risk here is that if the ARM expression to tell the ARM API "choose the approprate FD for this region" changes and our code doesn't adapt quickly enough, we will be sending malformed data to ARM.

mboersma:
Yes, that's the risk, but since fault domains are only incremented, never decremented, the risk isn't major. The danger is that there is a certain amount of time where AKS Engine provisions new clusters with FDCount = 2 where it could have been 3 (for example), until we update our code.

That's much better than we do currently, although I'm open for suggestions on how to eliminate that gap. It's possible we could use the Azure API equivalent of what @palma21 mentioned above to craft the correct ARM formula at the time someone runs `aks-engine`, but that wouldn't be trivial work.

mboersma:
Also to clarify, we wouldn't send malformed data, we just might not take advantage of a recently added fault domain until we updated the code.

pkg/api/types.go
```diff
@@ -469,6 +469,7 @@ type MasterProfile struct {
 	ImageRef                 *ImageReference   `json:"imageReference,omitempty"`
 	CustomFiles              *[]CustomFile     `json:"customFiles,omitempty"`
 	AvailabilityProfile      string            `json:"availabilityProfile"`
+	PlatformFaultDomainCount *int              `json:"platformFaultDomainCount"`
```
jackfrancis: Can you explain why we don't want to carry this over to `vlabs` (and by extension, to persist it into the `apimodel.json`)?

mboersma: I guess I was considering it a hidden implementation detail. Would adding it in `vlabs` allow users to override it in a source apimodel? Because I think we don't want that.

jackfrancis: I was thinking if we put it in there during the first upgrade or scale (basically, once the cluster has been created the value can effectively be converted to a static int), then each successive upgrade or scale will have the value there conveniently without having to do the API calls.

jackfrancis: Consider the above comments as food for thought as this evolves.

PR #1102, opened 2019-04-19
chore: always ensure apt runs non-interactively

jackfrancis:
**Reason for Change**:

To prevent scripted apt operations from stalling waiting for interactive input, we should always ensure that `apt` executions have a `DEBIAN_FRONTEND=noninteractive` env context.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
I think `apt-get` (or maybe `dpkg` under the covers) is smart enough to know when it's running in a non-interactive terminal in general, but even if it doesn't stall it will print a lot of warnings. This silences that. :+1:

PR #1110, opened 2019-04-22
style: remove unneeded nil check

mboersma:
**Reason for Change**:

In master there is some (old) code that fails `govet` with golangci-lint v1.16.0. This removes a redundant nil check to satisfy it. Look at the code just above the diff to see why it's redundant.

```
==> Running static validations and linters <==
pkg/api/converterfromagentpoolonlyapi.go:148:8: nilness: tautological condition: non-nil != nil (govet)
	if kc != nil && kc.EnableRbac != nil && *kc.EnableRbac {
	      ^
Makefile:138: recipe for target 'test-style' failed
make: *** [test-style] Error 1
```

**Issue Fixed**:

Unblocks #1088 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I'm not sure how this made it into master without us noticing. The code hasn't been changed, and the linter version changed a while ago...

### DIFF
--- a/pkg/api/converterfromagentpoolonlyapi.go
+++ b/pkg/api/converterfromagentpoolonlyapi.go
@@ -145,7 +145,7 @@ func convertKubernetesConfigToEnableRBACV20180331AgentPoolOnly(kc *KubernetesCon
 	}
 	// We use KubernetesConfig.EnableRbac to convert to versioned api model
 	// The assumption here is KubernetesConfig.EnableSecureKubelet is set to be same
-	if kc != nil && kc.EnableRbac != nil && *kc.EnableRbac {
+	if kc.EnableRbac != nil && *kc.EnableRbac {
 		return to.BoolPtr(true)
 	}
 	return to.BoolPtr(false)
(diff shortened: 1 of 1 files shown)


pkg/api/converterfromagentpoolonlyapi.go
```diff
@@ -145,7 +145,7 @@ func convertKubernetesConfigToEnableRBACV20180331AgentPoolOnly(kc *KubernetesCon
 	}
 	// We use KubernetesConfig.EnableRbac to convert to versioned api model
 	// The assumption here is KubernetesConfig.EnableSecureKubelet is set to be same
-	if kc != nil && kc.EnableRbac != nil && *kc.EnableRbac {
```
CecileRobertMichon: slight preference for removing the above check instead so we can have one `return to.BoolPtr(false)` instead of two.

mboersma: Or even better:
```
return to.BoolPtr(kc != nil && kc.EnableRbac != nil && *kc.EnableRbac)
```

CecileRobertMichon: yes!! 😍

jackfrancis:
lgtm

PR #1113, opened 2019-04-22
chore: add two missing /var/log files for 18.04 in CIS script

CecileRobertMichon:
**Reason for Change**:

Adds `/var/log/landscape/sysinfo.log and /var/log/unattended-upgrades/unattended-upgrades-dpkg.log` to `assignFilePermissions`. Fixes E2E with Ubuntu 18.04-LTS distro.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cis.sh
+++ b/parts/k8s/cloud-init/artifacts/cis.sh
@@ -25,6 +25,7 @@ assignFilePermissions() {
     waagent.log
     syslog
     unattended-upgrades/unattended-upgrades.log
+    unattended-upgrades/unattended-upgrades-dpkg.log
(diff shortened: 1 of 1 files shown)


jackfrancis:
Is there any reason to add these two files *only* if Ubuntu 18.04?

CecileRobertMichon:
@jackfrancis we're not adding them only if 18.04 but they fix the 18.04 E2E case. It doesn't hurt the 16.04 case to have these extra files in `/var/log` so I don't think we need to add conditionals here.

jackfrancis:
lgtm

PR #1117, opened 2019-04-22
test: fallback google.com check for Windows outbound test

jackfrancis:
**Reason for Change**:

Adds the same "check google.com in case" functionality for Windows outbound test to match the Linux outbound test.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/pod/pod.go
+++ b/test/e2e/kubernetes/pod/pod.go
@@ -29,6 +29,8 @@ const (
 	deleteTimeout         = 5 * time.Minute
 )
 
+var externalURLs = []string{"www.bing.com", "google.com"}
+
 // List is a container that holds all pods returned from doing a kubectl get pods
 type List struct {
 	Pods []Pod `json:"items"`
@@ -606,7 +608,7 @@ func (l *List) CheckOutboundConnection(sleep, duration time.Duration, osType api
 			case api.Linux:
 				ready, err = localPod.CheckLinuxOutboundConnection(sleep, duration)
 			case api.Windows:
-				ready, err = localPod.CheckWindowsOutboundConnection("www.bing.com", sleep, duration)
+				ready, err = localPod.CheckWindowsOutboundConnection(sleep, duration)
 			default:
 				ready, err = false, errors.Errorf("Invalid osType for Pod (%s)", localPod.Metadata.Name)
 			}
@@ -692,12 +694,12 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
 					installedCurl = true
 				}
 				// if we can curl bing.com we have outbound internet access
-				out, err := p.Exec("--", "curl", "bing.com")
+				out, err := p.Exec("--", "curl", externalURLs[0])
 				if err == nil {
 					readyCh <- true
 				} else {
 					// in case bing.com is down let's hope google.com is also not down
-					_, err := p.Exec("--", "curl", "google.com")
+					_, err := p.Exec("--", "curl", externalURLs[2])
 					if err == nil {
 						readyCh <- true
 					} else {
@@ -793,7 +795,7 @@ func (p *Pod) ValidateOmsAgentLogs(execCmdString string, sleep, duration time.Du
 }
 
 // CheckWindowsOutboundConnection will keep retrying the check if an error is received until the timeout occurs or it passes. This helps us when DNS may not be available for some time after a pod starts.
-func (p *Pod) CheckWindowsOutboundConnection(url string, sleep, duration time.Duration) (bool, error) {
+func (p *Pod) CheckWindowsOutboundConnection(sleep, duration time.Duration) (bool, error) {
 	exp, err := regexp.Compile(`(StatusCode\s*:\s*200)`)
 	if err != nil {
 		log.Printf("Error while trying to create regex for windows outbound check:%s\n", err)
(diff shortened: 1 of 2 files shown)


jackfrancis:
@PatrickLang @craiglpeters FYI

test/e2e/kubernetes/pod/pod.go
```diff
@@ -692,12 +694,12 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
...
 				}
 				// if we can curl bing.com we have outbound internet access
-				out, err := p.Exec("--", "curl", "bing.com")
+				out, err := p.Exec("--", "curl", externalURLs[0])
 				if err == nil {
 					readyCh <- true
 				} else {
 					// in case bing.com is down let's hope google.com is also not down
-					_, err := p.Exec("--", "curl", "google.com")
+					_, err := p.Exec("--", "curl", externalURLs[2])
```
CecileRobertMichon: should this be `externalURLs[1]`?

jackfrancis: lol

test/e2e/kubernetes/pod/pod.go
```diff
@@ -818,8 +820,20 @@ func (p *Pod) CheckWindowsOutboundConnection(url string, sleep, duration time.Du
 						readyCh <- false
 					}
 				} else {
-					log.Printf("Error:%s\n", err)
-					log.Printf("Out:%s\n", out)
+					// in case bing.com is down let's hope google.com is also not down
```
CecileRobertMichon: I would remove this comment.

or make it something like "try with a second URL in the case the first one is down"

test/e2e/kubernetes/pod/pod.go
```diff
@@ -692,12 +694,12 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
 					installedCurl = true
 				}
 				// if we can curl bing.com we have outbound internet access
-				out, err := p.Exec("--", "curl", "bing.com")
+				out, err := p.Exec("--", "curl", externalURLs[0])
 				if err == nil {
 					readyCh <- true
 				} else {
 					// in case bing.com is down let's hope google.com is also not down
```
CecileRobertMichon: "let's hope google.com is not also down" or generalize comment

test/e2e/kubernetes/pod/pod.go
```diff
@@ -1026,3 +1045,20 @@ func (c *Container) getMemoryRequests() string {
 func (c *Container) getMemoryLimits() string {
 	return c.Resources.Limits.Memory
 }
+
+// getExternalURL returns an external URL that is not in the passed-in array of strings
+func getExternalURL(urls []string) string {
+	m := map[string]bool{}
+	for _, url := range externalURLs {
```
CecileRobertMichon: the `externalURLs` var is only used here so it seems like it doesn't need to be a global var

jackfrancis: fair, the only reason I put it up there would be for easier maintenance, but we can make it private for now

test/e2e/kubernetes/pod/pod.go
```diff
@@ -691,19 +691,17 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
...
+				// if we can curl an external URL we have outbound internet access
+				for i, url := range getExternalURLs() {
+					out, err := p.Exec("--", "curl", url)
 					if err == nil {
 						readyCh <- true
 					} else {
-						// if both bing.com and google.com are down let's say we don't have outbound internet access
-						log.Printf("Error:%s\n", err)
-						log.Printf("Out:%s\n", out)
+						if i > 0 {
```
CecileRobertMichon: `if i == len(getExternalURLs) - 1` would scale when we add new URLs

test/e2e/kubernetes/pod/pod.go
```diff
@@ -809,17 +807,19 @@ func (p *Pod) CheckWindowsOutboundConnection(url string, sleep, duration time.Du
...
+				// if we can curl an external URL we have outbound internet access
+				for i, url := range getExternalURLs() {
+					out, err := p.Exec("--", "powershell", "iwr", "-UseBasicParsing", "-TimeoutSec", "60", url)
 					matched := exp.MatchString(string(out))
-					if matched {
+					if err == nil && matched {
 						readyCh <- true
 					} else {
-						readyCh <- false
+						if i > 0 {
```
CecileRobertMichon: ditto

test/e2e/kubernetes/pod/pod.go
```diff
@@ -691,19 +691,17 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
...
+				for i, url := range getExternalURLs() {
+					out, err := p.Exec("--", "curl", url)
 					if err == nil {
 						readyCh <- true
 					} else {
-						// if both bing.com and google.com are down let's say we don't have outbound internet access
-						log.Printf("Error:%s\n", err)
-						log.Printf("Out:%s\n", out)
+						if i > 0 {
+							// if both are down let's say we don't have outbound internet access
```
CecileRobertMichon: if all* instead of both

test/e2e/kubernetes/pod/pod.go
```diff
@@ -809,17 +807,19 @@ func (p *Pod) CheckWindowsOutboundConnection(url string, sleep, duration time.Du
...
+				for i, url := range getExternalURLs() {
+					out, err := p.Exec("--", "powershell", "iwr", "-UseBasicParsing", "-TimeoutSec", "60", url)
 					matched := exp.MatchString(string(out))
-					if matched {
+					if err == nil && matched {
 						readyCh <- true
 					} else {
-						readyCh <- false
+						if i > 0 {
+							// if both are down let's say we don't have outbound internet access
```
CecileRobertMichon: same

test/e2e/kubernetes/pod/pod.go
```diff
@@ -691,19 +691,18 @@ func (p *Pod) CheckLinuxOutboundConnection(sleep, duration time.Duration) (bool,
...
+				urls := getExternalURLs()
+				for i, url := range urls {
+					out, err := p.Exec("--", "curl", url)
 					if err == nil {
 						readyCh <- true
 					} else {
-						// if both bing.com and google.com are down let's say we don't have outbound internet access
-						log.Printf("Error:%s\n", err)
-						log.Printf("Out:%s\n", out)
+						if i > (len(urls) - 1) {
```
CecileRobertMichon: this will never be true, i will never be > len of the array - 1.

it should be `==`

PR #1126, opened 2019-04-23
chore: disable --pod-max-pids by default

### LINKED ISSUE #1055: SupportPodPidsLimit becoming a beta feature breaks some 1.14 deployments
no
---

ISSUE
---

master
---

When deploying a new 1.14 cluster but especially when upgrading to 1.14 or higher from an earlier version the combination of `aks-engine` and the `SupportPodPidsLimit` feature breaks some deployments (rabbitmq for example).

PID hungry applications are running fine on 1.13 due to the fact that the `SupportPodPidsLimit` feature gate is still alpha and therefore disabled by default. In 1.14 the feature is enabled by default which would not be a real issue by itself. However `aks-engine` deploys with a `pids-per-pod` limit of 100 which is not used until the feature flag is turned on.

I don't know if we need to have a higher default value or none at all, but as-is this is not a sane default setup

**Kubernetes version**:
1.14

jackfrancis:
**Reason for Change**:

`--pod-max-pids` configuration became enabled without --feature-gates opt-in starting in 1.14. Because there are lots of valid workflows that fork a bunch of processes, we want to be careful about setting an upper bound by default. This PR disables the upper limit to be permissive for such workloads.

**Issue Fixed**:

Fixes #1055

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


docs/topics/clusterdefinitions.md
```diff
@@ -269,7 +269,7 @@ Below is a list of kubelet options that aks-engine will configure by default:
 | "--image-gc-low-threshold"          | "850"                                                                                                                                                         |
 | "--non-masquerade-cidr"             | "10.0.0.0/8"                                                                                                                                                  |
 | "--azure-container-registry-config" | "/etc/kubernetes/azure.json"                                                                                                                                  |
-| "--pod-max-pids"                    | "100" (need to activate the feature in --feature-gates=SupportPodPidsLimit=true)                                                                              |
+| "--pod-max-pids"                    | "-1" (need to activate the feature in --feature-gates=SupportPodPidsLimit=true)                                                                              |
```
CecileRobertMichon: We should track an issue to make this a sensible default as a follow up item (-1 doesn't really make sense as a default value).

PatrickLang: I agree, but that's what the work-in-progress docs say: https://github.com/kubernetes/website/pull/13806

PatrickLang: I also asked for clarification on what `0` means. If it disables the feature, then I think we may actually want `0` instead of `-1` (node enforceable limit) if we don't set a node enforceable limit today.

Sahasrara:
How do I get this change applied to my clusters?

jackfrancis:
@Sahasrara upgrading your cluster (`aks-engine upgrade`) using `v0.38.0` or later of aks-engine will apply these changes to a pre-existing cluster.

PR #1129, opened 2019-04-23
chore: clean up containerd and cc-runtime when unused

### LINKED ISSUE #1091: Remove clear container install at provisioning if the cluster doesn't need it
We pre-install clear containers on the AKS VM image, we should clean up if not needed by the cluster at provisioning.

CecileRobertMichon:
**Reason for Change**:

Remove the installs for containerd and clear containers during CSE when that cluster does not need them. These are pre-installed as part of the VHD.

**Issue Fixed**:

Fixes #1091 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -17,7 +17,7 @@ removeEtcd() {
 }
 
 removeMoby() {
-    sudo apt-get purge -y moby-engine moby-cli
+    apt-get purge -y moby-engine moby-cli
 }
 
 installEtcd() {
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -17,7 +17,7 @@ removeEtcd() {
 }
 
 removeMoby() {
-    sudo apt-get purge -y moby-engine moby-cli
+    apt-get purge -y moby-engine moby-cli
```
CecileRobertMichon: this is already running as root, no need for sudo

PR #1109, opened 2019-04-22
feat: upgrade metrics server to v0.3.4

andyzhangx:
**Reason for Change**:

This PR upgrades metrics server from `v0.2.1` to `v0.3.4`

BTW, we cannot disable `--read-only-port` now since AKS also depends on this port and metrics server on AKS is still v0.2.x

metric server(v0.2.1) is still using KubeletReadOnlyPort(10255), we cannot switch to KubeletPort(10250) now, after upgrade to metric server v0.3.4, it uses 10250 port instead, then we could move on to merge PR [fix: disable ReadOnlyPort ](https://github.com/Azure/aks-engine/pull/975)

This PR only applies for k8s v1.16.0:
 - upgrade metric server to v0.3.4

Tested on k8s v1.16.0-beta.1, test result:
```
$ kubectl top no
NAME                        CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%
6133k8s010                  56m          2%     445Mi           8%
k8s-linuxpool1-61339693-0   54m          2%     692Mi           11%
k8s-master-61339693-0       92m          4%     1065Mi          17%
```

**Issue Fixed**:

Fixes #1172

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

cc @feiskyer 
/azp run pr-e2e

PR #1140, opened 2019-04-24
feat: add support for Kubernetes v1.15.0-alpha.1 

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#v1150-alpha1

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/k8s_versions.go
```diff
@@ -13,6 +13,39 @@ import (
 // k8sComponentVersions defines "best known-working" component versions for each minor version of k8s
 // Don't delete these hash table keys as they are consumed independently by AKS!
 var k8sComponentVersions = map[string]map[string]string{
+	"1.15": {
+		"dockerEngine":                     "1.13.*",
+		"dashboard":                        "kubernetes-dashboard-amd64:v1.10.1",
+		"exechealthz":                      "exechealthz-amd64:1.2",
+		"addon-resizer":                    "addon-resizer:1.8.4",
+		"heapster":                         "heapster-amd64:v1.5.4",
+		"metrics-server":                   "metrics-server-amd64:v0.3.2",
```
mboersma: `metrics-server` was the only component that was bumped to a new version, according to [the release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1140).

zhiweiv:
Metrics server 0.3.x doesn’t work with Windows nodes, see https://github.com/kubernetes/kubernetes/issues/75934.

pkg/api/k8s_versions.go
mboersma: Reverting to v0.2.1 seems to get `kubectl top nodes` working again, so let's separate concerns and update `metrics-server` elsewhere. See #1172 and #1109.

mboersma:
> update the kubernetesmasteraddons-metrics-server-deployment.yaml

Let's do that as a separate task: #1172.

PR #1142, opened 2019-04-24
feat: static pods use custom hyperkube on az stack

### LINKED ISSUE #990: Pre-pull Azure Stack's dependencies

jadarsie:
**Reason for Change**:
Static pods are not using the pre-pulled Azure Stack's custom Hyperkube build

**Issue Fixed**:
Fixes #990

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
It is not clear to me how to update `params_k8s_test.go` to validate my changes.


jadarsie:
> This pull-request has been approved by: jadarsie

Not sure why `acs-bot` thinks I approved it.

pkg/api/k8s_versions.go
```diff
@@ -468,6 +468,7 @@ func getK8sVersionComponents(version string, overrides map[string]string) map[st
 	case "1.11":
 		ret = map[string]string{
 			"hyperkube":                        "hyperkube-amd64:v" + version,
+			"hyperkube-azs":                    "hyperkube-amd64:v" + version + "-azs",
```
honcao: Please add unit test to verify this

pkg/engine/params_k8s.go
```diff
@@ -44,7 +46,7 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 				addValue(parametersMap, "kubernetesCcmImageSpec", kubernetesCcmSpec)
 			}
 
-			kubernetesHyperkubeSpec := hyperKubeImageBase + k8sComponents["hyperkube"]
+			kubernetesHyperkubeSpec := hyperkubeImageBase + hyperkubeComponent
```
honcao: Please add unit test to cover this logic

pkg/api/k8s_versions.go
```diff
@@ -468,6 +468,7 @@ func getK8sVersionComponents(version string, overrides map[string]string) map[st
 	case "1.11":
```
honcao: how about 1.12, 1.13 and 1.14?
Please also ensure the images exists under docker account msazurestackdocker

pkg/api/k8s_versions.go
honcao: Please also try the scale and upgrade scenario for azure stack. 1.11.8 to 1.11.9 and 1.11.9 to 1.12.7

jadarsie: That should be already covered

pkg/api/k8s_versions.go
jadarsie: Only v1.12.7 is in docker hub.
However, I would rather have a separate PR for newer minor versions once we have a successful 1.11.X deployment on a disconnected azure stack

pkg/engine/params_k8s.go
jadarsie: The PR notes explain that I need help with that. This change does not add a new execution path though.

pkg/api/k8s_versions.go
CecileRobertMichon: I thought you were already adding the `-azs` suffix on the version? https://github.com/Azure/aks-engine/commit/39182cdc6579010b74ae8bf6df783a4f5868b9d2#diff-de95d09bcc65a2843f3bad212bdac8e8R80

jadarsie: That `version` value comes from `GetAllSupportedKubernetesVersions`. 
Let me know if it is preferred to add a new entry to that map.

pkg/engine/params_k8s.go
```diff
@@ -28,10 +28,12 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		k8sComponents := api.K8sComponentsByVersionMap[k8sVersion]
 		kubernetesConfig := orchestratorProfile.KubernetesConfig
 		kubernetesImageBase := kubernetesConfig.KubernetesImageBase
-		hyperKubeImageBase := kubernetesConfig.KubernetesImageBase
+		hyperkubeImageBase := kubernetesConfig.KubernetesImageBase
+		hyperkubeComponent := k8sComponents["hyperkube"]
 
 		if properties.IsAzureStackCloud() {
 			kubernetesImageBase = cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
+			hyperkubeComponent = k8sComponents["hyperkube-azs"]
```
CecileRobertMichon: this only exists for 1.11 right now so we should only override if the k8s version is 1.11.x

CecileRobertMichon:
> Not sure why acs-bot thinks I approved it.

the bot automatically makes the author an approver

pkg/api/k8s_versions.go
jadarsie: This seems to be working, adding 1.12

pkg/api/k8s_versions.go
jadarsie: these two operations worked on my dev env.

packer/install-dependencies.sh
```diff
@@ -350,7 +350,7 @@ K8S_VERSIONS="
 1.14.0
 1.13.5
 1.13.4
-1.12.7
+1.12.7-azs
```
CecileRobertMichon: this should be additive, we want both `1.12.7` and `1.12.7-azs`

pkg/api/k8s_versions.go
CecileRobertMichon: All good, I think I understand what's going on now.

pkg/engine/params_k8s.go
jadarsie: I'll wait for you to confirm that a new map entry for `hyperkube-azs` is the preferred approach before I make changes here.

If that's the case, I am thinking on adding the entry to all `K8sVersionComponents` maps.
However, that implies that only versions supported by AZS should be accepted.
This validation may tilt the scale for an approach that requires new entries for map `GetAllSupportedKubernetesVersions`.

pkg/engine/params_k8s.go
CecileRobertMichon: instead of needing a new `k8sComponents`  entry, you could also do:

```
if properties.IsAzureStackCloud() {
    kubernetesHyperkubeSpec := hyperKubeImageBase + k8sComponents["hyperkube"] + AzureStackSuffix
}
```

where `AzureStackSuffix` is a constant in https://github.com/Azure/aks-engine/blob/master/pkg/engine/const.go (and it can also be reused in armvariables.go)

packer/install-dependencies.sh
xdarsie: yeah, sorry

jadarsie:
Rebased

PR #1144, opened 2019-04-24
feat: pre-pull Pause from Azure Stack's docker repo

### LINKED ISSUE #990: Pre-pull Azure Stack's dependencies

jadarsie:
**Reason for Change**:
AKSe assumes that both `hyperkube` and `pause` share the same repository prefix.
In the past, the Azure Stack team re-tagged the official pause image and re-pushed to docker hub to satisfy this requirement.
For disconnected instances, we need to pre-pull or re-tag during the VHD provisioning process.

**Issue Fixed**:
Fixes #990

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -180,7 +180,7 @@ PAUSE_VERSIONS="3.1"
 for PAUSE_VERSION in ${PAUSE_VERSIONS}; do
     # Image 'msazurestackdocker/pause-amd64' is the same as 'k8s.gcr.io/pause-amd64'
     # At the time, re-tagging and pushing to docker hub seemed simpler than changing how `defaults-kubelet.go` sets `--pod-infra-container-image`
-    for IMAGE_BASE in "k8s.gcr.io msazurestackdocker"; do
+    for IMAGE_BASE in k8s.gcr.io msazurestackdocker; do
       CONTAINER_IMAGE="${IMAGE_BASE}/pause-amd64:${PAUSE_VERSION}"
       pullContainerImage "docker" ${CONTAINER_IMAGE}
       echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -178,14 +178,13 @@ done
 
 PAUSE_VERSIONS="3.1"
 for PAUSE_VERSION in ${PAUSE_VERSIONS}; do
-    CONTAINER_IMAGE="k8s.gcr.io/pause-amd64:${PAUSE_VERSION}"
-    pullContainerImage "docker" ${CONTAINER_IMAGE}
-    echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
     # Image 'msazurestackdocker/pause-amd64' is the same as 'k8s.gcr.io/pause-amd64'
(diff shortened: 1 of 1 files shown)


jackfrancis:
@jadarsie Will the custom config passed in via Azure Stack-specific template generation correctly reconcile the desired `msazurestackdocker` image base part of the URL? E.g., here's the spec:

```
"--pod-infra-container-image":         o.KubernetesConfig.KubernetesImageBase + K8sComponentsByVersionMap[o.OrchestratorVersion]["pause"],
```

packer/install-dependencies.sh
```diff
@@ -181,6 +181,11 @@ for PAUSE_VERSION in ${PAUSE_VERSIONS}; do
     CONTAINER_IMAGE="k8s.gcr.io/pause-amd64:${PAUSE_VERSION}"
     pullContainerImage "docker" ${CONTAINER_IMAGE}
     echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
+    # Image 'msazurestackdocker/pause-amd64' is the same as 'k8s.gcr.io/pause-amd64'
```
jackfrancis: how about this instead:

(inside the `for PAUSE_VERSION in ${PAUSE_VERSIONS}; do` loop)
```
for CONTAINER_IMAGE in "k8s.gcr.io/pause-amd64:${PAUSE_VERSION} msazurestackdocker/pause-amd64:${PAUSE_VERSION}"; do
done
```

jackfrancis: or

```
for IMAGE_BASE in "k8s.gcr.io/pause-amd64 msazurestackdocker/pause-amd64"; do
  pullContainerImage "docker" ${IMAGE_BASE}/pause-amd64:${CONTAINER_IMAGE}
done
```

packer/install-dependencies.sh
```diff
@@ -181,6 +181,11 @@ for PAUSE_VERSION in ${PAUSE_VERSIONS}; do
...
+    # Re-tagging seemed simpler than changing how `defaults-kubelet.go` sets `--pod-infra-container-image`
```
CecileRobertMichon: this comment implies we are doing a retag when in fact we're pulling the image twice.

xdarsie: Well, `msazurestackdocker/pause-amd64` is a re-tagged image, will clarify

xdarsie:
> @jadarsie Will the custom config passed in via Azure Stack-specific template generation correctly reconcile the desired `msazurestackdocker` image base part of the URL?

Yes, before this PR, K8s was pulling the pause image hosted on `msazurestackdocker`. The PR goal is to pre-pull it. I run a few tests (deploy, upgrade) and this seems to be working fine. Not sure if that's you concern though.

packer/install-dependencies.sh
xdarsie: updated

packer/install-dependencies.sh
xdarsie: hopefully that's better

packer/install-dependencies.sh
jackfrancis: Sorry, my bad, we need to lose the quotes around the image base vals

`for IMAGE_BASE in k8s.gcr.io/pause-amd64 msazurestackdocker/pause-amd64; do`

...

Issue #1099, opened 2019-04-19
No outbound UDP possible when using standard load balancer


sylr:
When using Azure Zones you need to use load balancers of the Standard SKU.

The Standard SKU does not work the same as the Basic SKU when it comes to SNAT'ing outbound connections.

In case of UDP connections the load balancer SNAT mechanism will only be fired up if there is at least one UDP load balancing rule. If not your requests will go out but you'll never get a response.

Source: https://stackoverflow.com/questions/50168334/public-lb-configuration-for-ntp-service-in-azure

jackfrancis:
@sylr is it acceptable to deliver a "dummy" UDP load balancing rule w/ the ARM template spec so that a user doesn't have to do so manually?

sylr:
That's what I did.

PR #1145, opened 2019-04-25
feat: Support existing load balancer backend address pool for agent nodes

xuto2:
**Reason for Change**:
Azure Standard Load Balancer(SLB) introduces some fundamental difference to Azure Basic Load Balancer(BLB). Specifically, vmas/vmss behind Azure SLB don't automatically get internet access like BLB, and it's Azure's recommendation that we "always use outbound rules on a Standard public Load Balancer" (see https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections for details).

To support Azure SLB in AKS, AKS needs to maintain the outbound rules of the SLB (and the load balancer rules as well to workaround upstream issue when using SLB https://github.com/kubernetes/kubernetes/pull/76691). So we need the ability to place AKS engine generated nodes (VMSS+VMAS) into an existing SLB's backend address pool.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


jackfrancis:
@sylr Does this address any symptoms described in #1099?

jackfrancis:
@xuto2 FYI I rebased on top of master/force-pushed to this branch, so you'll need to pull down for future dev

CecileRobertMichon:
Squash commit message needs to be updated to use semantic message (or if a second commit is pushed the title is enough).

/hold

CecileRobertMichon:
two commits will use PR title as squash message so we're good now.

/hold cancel

docs/topics/clusterdefinitions.md
```diff
@@ -569,7 +569,8 @@ A cluster can have 0 to 12 agent pool profiles. Agent Pool Profiles are used for
...
 | enableVMSSNodePublicIP | no                                                                   | Enable creation of public IP on VMSS nodes. This configuration is only valid on an agent pool with an `"availabilityProfile"` value of `"VirtualMachineScaleSets"`. Defaults to `false`                                                                                                                                                                                                                                                      |
-
+| LoadBalancerBackendAddressPoolIDs | no                                                                   | Enables automatic placement of the agent pool nodes into existing load balancer's backend address pools. Each element value of this string array is the corresponding load balancer backend address pool's Azure Resource Manager(ARM) resource ID.
```
CecileRobertMichon: Let's add the default value here

CecileRobertMichon: nevermind, this is a []string

jackfrancis: added a description of the default case

jackfrancis:
@sylr nevermind I understand what this does, I have another possible fix for the lack of UDP outbound incoming shortly...

xuto2:
Thanks @jackfrancis @CecileRobertMichon @feiskyer !

PR #1150, opened 2019-04-25
ci: return error if Go linter can't compile source code

mboersma:
**Reason for Change**:
The `golangci-lint` tool [doesn't return an error](https://github.com/golangci/golangci-lint/issues/276) if it can't do its job because the Go source code won't compile. This fixes that by returning an error if the command output includes `skipped due to error`. Hopefully this closes the loophole that has allowed a couple of linter errors to make it into master.

This also removes some dead Go code that was causing `govet` to complain, and adds the output of the `--version` command to the script for both the Go and shell linters to help users reproduce any failures.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/loadbalancers.go
+++ b/pkg/engine/loadbalancers.go
@@ -259,26 +259,3 @@ func CreateMasterInternalLoadBalancer(cs *api.ContainerService) LoadBalancerARM 
 
 	return loadBalancerARM
 }
-
-func getUDPLoadBalancingRule() network.LoadBalancingRule {
-	return network.LoadBalancingRule{
-		Name: to.StringPtr("LBRuleUDP"),
-		LoadBalancingRulePropertiesFormat: &network.LoadBalancingRulePropertiesFormat{
-			FrontendIPConfiguration: &network.SubResource{
-				ID: to.StringPtr("[variables('masterLbIPConfigID')]"),
-			},
-			BackendAddressPool: &network.SubResource{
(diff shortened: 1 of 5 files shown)


mboersma:
This gets curiouser and curiouser as I try to fix it. Let's consider this a WIP.

pkg/engine/loadbalancers.go
```diff
@@ -259,26 +259,3 @@ func CreateMasterInternalLoadBalancer(cs *api.ContainerService) LoadBalancerARM
 
 	return loadBalancerARM
 }
-
-func getUDPLoadBalancingRule() network.LoadBalancingRule {
```
jackfrancis: Is this dead code that my PR introduced?

jackfrancis:
lgtm

pkg/engine/loadbalancers.go
mboersma: Yes, I think so. But the specific error has been hidden by this problem.

jackfrancis: Verified. I recollect I scaffolded out this convenience so I could re-use it, then realized the re-use case was not as clean as I thought, and so abandonded the re-usable function path. Thanks for cleaning up the func for me! :)

autarch:
Wouldn't it be better to have `golangci-lint` actually exit non-zero when it cannot compile some code?

mboersma:
@autarch that would be preferable, but the [golangci-lint issue](golangci/golangci-lint#276) I linked to in the description indicates that's not going to be fixed any time soon. So I also grep the output to look for what should be an error.

PR #1157, opened 2019-04-26
fix: use cordon/drain timeout for pod eviction timeout during upgrade

### LINKED ISSUE #930: Upgrade Fails due to node drain timeout

zachpuck:
**Reason for Change**:

When testing changes from PR #938 I found that the [`WaitForDelete`](https://github.com/Azure/aks-engine/blob/c2229858c671013388bd2e14776d357fc9edc386/pkg/armhelpers/kubeclient.go#L144) call to delete/evict a pod was still using a 10s timeout. This caused some pods to not evict in time and the upgrade to fail. To quickly address this I have updated the getKubernetesClient to allow the `cordonDrainTimeout` to be passed in. 

**Issue Fixed**:

Fixes #930 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -319,7 +319,7 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
 		}
 
 		newCreatedVMs := []string{}
-		client, err := ku.getKubernetesClient()
+		client, err := ku.getKubernetesClient(10 * time.Second)
 		if err != nil {
 			ku.logger.Errorf("Error getting Kubernetes client: %v", err)
 			return err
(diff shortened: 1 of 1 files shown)


jackfrancis:
lgtm pending upgrade tests (in-progress)

PR #1159, opened 2019-04-26
style: add bash shebangs to scripts

mboersma:
**Reason for Change**:

Enforces the rule that shell scripts need a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) to be explicit about the interpreter used.
See https://github.com/koalaman/shellcheck/wiki/SC2148
and https://github.com/koalaman/shellcheck/wiki/SC2039

**Issue Fixed**:

Refs #953 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
mboersma: The other `cse_` scripts specified `bash`, and this script was using a comparison that isn't defined in POSIX `sh`.

CecileRobertMichon:
@mboersma can we make the bash shellcheck fixes `style: `?

PR #1167, opened 2019-04-27
style: use { cmd1; cmd2 } >> file instead of individual redirects

mboersma:
**Reason for Change**:
Enforces a simplification when multiple shell commands redirect to the same target.
See https://github.com/koalaman/shellcheck/wiki/SC2129
Also enables four `shellcheck` error codes that aren't being triggered in any scripts.

**Issue Fixed**:
Refs #953 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
The only subtle difference I see is when there is an initial `>` (as in, overwrite everything if the file already exists). Will this change not literally do the above? If the file already exists, will the output redirection append to the existing data?

If so, let's audit each instance of this change and make sure that's O.K.

mboersma:
The translation should be straightforward: if the original stanza created the file with `>`, then use that same operator to create the file in the new style. Otherwise use `>>`.

Thanks for pointing it out—I got one of them wrong, but it's fixed now.

PR #1169, opened 2019-04-29
fix: USER_ASSIGNED_IDENTITY_ID is empty in azure.json

### LINKED ISSUE #904: "userAssignedIdentityID" in cloud config file is empty, which can cause the cluster not work when multiple user assigned identity is associated to the VM
**Is this a request for help?**: yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): issue

---

**What version of aks-engine?**: latest

---

**Kubernetes version**: any version can use use assigend identity

**What happened**: in `azure.json` on agent nodes and master node, `userAssignedIdentityID` is an empty string. This works fine when there is only one user assigned identity associated to the VM, however, when user add another user assigend identities to the VM for some reason(for example, using aad-pod-identity), `cloud-provider-azure` will not be able to determine which identity to use and the authentication to Azure will fail. The `azure.json` file looks like:
```json
{
    "cloud":"AzurePublicCloud",
    "tenantId": "<redacted>",
    "subscriptionId": "<redacted>",
    "aadClientId": "msi",
    "aadClientSecret": "msi",
    "resourceGroup": "tosiAKSEngine",
    "location": "centralus",
    "vmType": "vmss",
    "subnetName": "k8s-subnet",
    "securityGroupName": "k8s-master-29857939-nsg",
    "vnetName": "k8s-vnet-29857939",
    "vnetResourceGroup": "",
    "routeTableName": "k8s-master-29857939-routetable",
    "primaryAvailabilitySetName": "",
    "primaryScaleSetName": "k8s-agentpool1-29857939-vmss",
    "cloudProviderBackoff": true,
    "cloudProviderBackoffRetries": 6,
    "cloudProviderBackoffExponent": 1.5,
    "cloudProviderBackoffDuration": 5,
    "cloudProviderBackoffJitter": 1,
    "cloudProviderRatelimit": true,
    "useManagedIdentityExtension": true,
    "userAssignedIdentityID": "",  ############## Empty string ################
    "useInstanceMetadata": true,
    "loadBalancerSku": "Basic",
    "excludeMasterFromStandardLB": false,
    "providerVaultName": "",
    "providerKeyName": "k8s",
    "providerKeyVersion": ""
}
```

**What you expected to happen**: `userAssignedIdentityID` in `azure.json` should be correctly filled.

Generate ARM template using following cluster definition and deploy the ARM template, in the deployed cluster, `userAssignedIdentityID` in `azure.json` will be an empty string:
```json
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorRelease": "1.12",
      "kubernetesConfig": {
         "useManagedIdentity": true,
         "userAssignedID": "AKSEngineMSIID"
      }
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "tosiaks3",
      "vmSize": "Standard_D2_v2"
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 2,
        "vmSize": "Standard_D2_v2"
      }
    ],
    "linuxProfile": {
      "adminUsername": "azureuser",
      "ssh": {
        "publicKeys": [
          {
            "keyData": "<redacted>"
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "<redacted>",
      "secret": "<redacted>"
    }
  }
}
```

Not sure whether [nested template](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-linked-templates#link-or-nest-a-template) can help. I have tested  following template can successfully output the client id of created user assigned identity. Maybe AKS-engine can make this template nested and uses its output to fill in `userAssignedIdentityID`?
```json
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "identityName": {
            "defaultValue": "AKS-MSI-test",
            "type": "String"
        }
    },
    "resources": [
        {
            "name": "[parameters('identityName')]",
            "apiVersion": "2018-11-30",
            "location": "eastus",
            "tags": {},
            "properties": {}
        }
    ],
    "outputs": {
        "clientid": {
            "type": "String",
            "value": "[reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', parameters('identityName')), '2018-11-30').clientId]"
        }
    }
}
```

norshtein:
**Reason for Change**:

This PR fixes #904, please refer to #904 for detailed background.

**Issue Fixed**:

Fixes #904 

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
To get the client id of the generated user assigned identity, we can use `reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID')), '2018-11-30').clientId`. But previously, in aks-engine, variables in `azure.json` are provided to provision script by a variable called `provisionScriptParametersCommon`. It is an ARM template variable, according to [ARM template manual](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-resource#reference):  
> The reference function can only be used in the properties of a resource definition and the outputs section of a template or deployment.

We can't use `reference` function in the definition of a variable. So in this PR, instead of using ARM template variable, I switched to use raw string to provide parameters to provision script. The raw string is returned by a helper function `GenerateProvisionScriptParameters`. The variable `provisionScriptParametersCommon` no longer has any effect, but I keep it as a variable in case it can be used later, and also avoid breaking anything.


norshtein:
I have manually tested to create a cluster using system assigned identity and a cluster using user assigned identity. For the cluster using system assigned identity,  `USER_ASSIGNED_IDENTITY_ID` in `azure.json` keeps being empty. For the cluster using user assigned identity,  `USER_ASSIGNED_IDENTITY_ID` in `azure.json` is successfully filled with the client id of created user assigned identity.

pkg/api/common/helper.go
```diff
@@ -226,3 +227,17 @@ func IsSgxEnabledSKU(vmSize string) bool {
...
...oviderConfig').cloudProviderBackoffJitter,' CLOUDPROVIDER_RATELIMIT=',toLower(string(parameters('cloudproviderConfig').cloudProviderRatelimit)),' CLOUDPROVIDER_RATELIMIT_QPS=',parameters('cloudproviderConfig').cloudProviderRatelimitQPS,' CLOUDPROVIDER_RATELIMIT_BUCKET=',parameters('cloudproviderConfig').cloudProviderRatelimitBucket,' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension'),' USER_ASSIGNED_IDENTITY_ID=\"\"',' USE_INSTANCE_METADATA=',variables('useInstanceMetadata'),' LOAD_BALANCER_SKU=',variables('loadBalancerSku'),' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB'),' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount'),' CONTAINER_RUNTIME=',parameters('containerRuntime'),' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase'),' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec'),' KMS_PROVIDER_VAULT_NAME=',variables('clusterKeyVaultName'),' IS_HOSTED_MASTER=%t',' PRIVATE_AZURE_REGISTRY_SERVER=',parameters('privateAzureRegistryServer'),' AUTHENTICATION_METHOD=',variables('customCloudAuthenticationMethod'),' IDENTITY_SYSTEM=',variables('customCloudIdentifySystem')",
```
andyzhangx: it looks like the two `return fmt.Sprintf( ...` statements are quite similar, can we combine these two to one common string, it's very hard to read and maintain.

norshtein: Combined them to one const vaule.

jackfrancis:
Thanks for fixing this @norshtein! Rather than offload the entire "common" params into a func, why not create a simple convenience func to get the value of just `USER_ASSIGNED_IDENTITY_ID`?

So we'd replace this inside the `provisionScriptParametersCommon` value in `pkg/engine/armvariables.go`:

`USER_ASSIGNED_IDENTITY_ID=',variables('userAssignedClientID')`

With something like this:

`
userAssignedID := ""
if cs.Properties.IsUserAssignedIdentityEnabled() {
    userAssignedID = "reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID')), '2018-11-30').clientId"
}
`

And then change the `USER_ASSIGNED_IDENTITY_ID` env var assignment to:

`USER_ASSIGNED_IDENTITY_ID=%s`

And then add `userAssignedID` to the end of the `fmt.Sprintf` statement so it populates the new `%s` properly.

Does that make sense? Something like this would seem to have less of an impact on the overall code flow. Thoughts @CecileRobertMichon?

CecileRobertMichon:
I agree that the refactoring isn't strictly necessary and that the approach Jack proposed above is the least impactful on existing code. That being said, I kind of like this move to refactor the extension script parameters to a function since that will allow us to do further breakdown later on as well as unit test them. The existing string variable is pretty hard to maintain because it's so long and any small typo could break CSE (and it's 0% unit tested).

jackfrancis:
This is arguably less maintainable as the function adds a level of code indirection, and the existing non-maintainability of the log string is intact (via the local const).

I agree this area can be improved for maintainability, but I'd argue that will require a focused effort towards *that* goal (and not simply making `USER_ASSIGNED_IDENTITY_ID` more dynamic).

pkg/engine/vmextensions.go
```diff
@@ -141,7 +141,7 @@ func createAgentVMASCustomScriptExtension(cs *api.ContainerService, profile *api
 		vmExtension.Publisher = to.StringPtr("Microsoft.Azure.Extensions")
 		vmExtension.VirtualMachineExtensionProperties.Type = to.StringPtr("CustomScript")
 		vmExtension.TypeHandlerVersion = to.StringPtr("2.0")
-		commandExec := fmt.Sprintf("[concat('retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; %s for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),' GPU_NODE=%s SGX_NODE=%s /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1%s\"')]", outBoundCmd, nVidiaEnabled, sgxEnabled, runInBackground)
```
CecileRobertMichon: if we're not using `variables('provisionScriptParametersCommon')` anywhere anymore we should be able to remove it from https://github.com/Azure/aks-engine/blob/master/pkg/engine/armvariables.go

norshtein:
> Thanks for fixing this @norshtein! Rather than offload the entire "common" params into a func, why not create a simple convenience func to get the value of just `USER_ASSIGNED_IDENTITY_ID`?
> 
> So we'd replace this inside the `provisionScriptParametersCommon` value in `pkg/engine/armvariables.go`:
> 
> `USER_ASSIGNED_IDENTITY_ID=',variables('userAssignedClientID')`
> 
> With something like this:
> 
> `userAssignedID := "" if cs.Properties.IsUserAssignedIdentityEnabled() { userAssignedID = "reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID')), '2018-11-30').clientId" }`
> 
> And then change the `USER_ASSIGNED_IDENTITY_ID` env var assignment to:
> 
> `USER_ASSIGNED_IDENTITY_ID=%s`
> 
> And then add `userAssignedID` to the end of the `fmt.Sprintf` statement so it populates the new `%s` properly.
> 
> Does that make sense? Something like this would seem to have less of an impact on the overall code flow. Thoughts @CecileRobertMichon?

Got your point. I also found it's not necessary to remove whole `provisionScriptParametersCommon`. I'll rewrite the code and resend the PR. @CecileRobertMichon @jackfrancis Thanks for reviewing and suggestions!

norshtein:
Closed and reopen a new PR.

PR #1173, opened 2019-04-29
fix: delete vm if node is not ready after upgrade.

JunSun17:
**Reason for Change**:

When upgrade, if the newly added node is not ready, it will fail the upgrade attempt but that node will be kept. However, in the next upgrade attempt, another working node will be deleted, which ends up with a cluster with less than specified number of working nodes.

Here the fix is to remove the newly added node if it reports not ready.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


jackfrancis:
lgtm

PR #1183, opened 2019-04-30
fix: kubeconfig enforcement via kubelet.sh

### LINKED ISSUE #1181: Issue with master VMSS and cosmosEtcd

---

ISSUE
---

v0.34.2
---

**Kubernetes version**:
1.10.13

**What happened**:
kubectl get nodes returns no masters and no workloads are running.

kubectl get nodes should return masters and the workload should run.

Use sampleMasterVmss.txt and create a kubernetes cluster. Then try to run kubectl get nodes or schedule a simple workload.
[sampleMasterVmss.txt](https://github.com/Azure/aks-engine/files/3132404/sampleMasterVmss.txt)

jackfrancis:
not setup-etcd.sh

**Reason for Change**:

For master vmss scenarios, kubelet configuration needs to be applied at runtime to direct apiserver communications to the local IP address. And this enforcement was incorrectly being done via etcd bootstrap script instead of general kubelet bootstrap script, which breaks cosmos etcd + master vmss scenarios

**Issue Fixed**:

Fixes #1181

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -383,7 +383,7 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     {{else}}
     PRIVATE_IP=$(hostname -i | cut -d" " -f1)
     {{end}}
-    ETCD_CLIENT_PORT={{WrapAsVariable "masterEtcdClientPort"}}
+    sed -i "s|<SERVERIP>|https://$PRIVATE_IP:443|g" "/var/lib/kubelet/kubeconfig"
   {{end}}
 {{if gt .MasterProfile.Count 1}}
     # Redirect ILB (4443) traffic to port 443 (ELB) in the prerouting chain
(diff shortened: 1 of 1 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -383,7 +383,7 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     {{else}}
     PRIVATE_IP=$(hostname -i | cut -d" " -f1)
     {{end}}
-    ETCD_CLIENT_PORT={{WrapAsVariable "masterEtcdClientPort"}}
```
jackfrancis: something funky in the state of Denmark, as this `ETCD_CLIENT_PORT` variable assignment shouldn't have been here. Probably legacy copy pasta.

mboersma:
Ha! I had a PR that's basically identical to this that I'm testing locally. I'll test this.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -383,7 +383,7 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
+    sed -i "s|<SERVERIP>|https://$PRIVATE_IP:443|g" "/var/lib/kubelet/kubeconfig"
```
CecileRobertMichon: Just to make sure I understand, do we only need to do this for vmss? Is this field not present in /var/lib/kubelet/kubeconfig for vmas?

It seems like we were doing this sed statement for both vmss and vmas before but now only for vmss.

CecileRobertMichon: Looks like that field isn't present for vmas

PR #1191, opened 2019-05-01
style: use 'grep -c' instead of 'grep|wc'

mboersma:
**Reason for Change**:
Suggests a simplification in a common usage of `grep` with pipes.

See https://github.com/koalaman/shellcheck/wiki/SC2126

**Issue Fixed**:

Refs #953

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
I don't see any `grep -c` usages

mboersma:
You're right—I took the advice lower down in the [SC2126](https://github.com/koalaman/shellcheck/wiki/SC2126) article and further simplified to `grep -q` since we were just testing to see if there was a match. The shellcheck rule that this enables suggests `grep -c` in general, however.

PR #1193, opened 2019-05-01
feat: add support for Kubernetes 1.11.10

mboersma:
**Reason for Change**:

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.11.md#changelog-since-v1119

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Some Azure-specific fixes in here and in cluster-autoscaler 1.3.7.


packer/install-dependencies.sh
```diff
@@ -357,6 +358,7 @@ K8S_VERSIONS="
 1.12.8
 1.12.7
 1.12.7-azs
+1.11.10
```
CecileRobertMichon: @jadarsie @honcao FYI you might want to add the azure stack equivalent

packer/install-dependencies.sh
```diff
@@ -357,6 +358,7 @@ K8S_VERSIONS="
...
 1.11.9
 1.11.9-azs
 1.11.8
```
CecileRobertMichon: why not remove 1.11.8?

mboersma: Because I forgot to. :-) Should I leave in `1.11.8-azs` until there is a `1.11.10-azs`, or remove it as well?

CecileRobertMichon: I'll leave that up to @jadarsie

jackfrancis:
FYI manually validated Windows 1.11.10

packer/install-dependencies.sh
CecileRobertMichon: /hold

pkg/api/common/versions.go
```diff
@@ -88,8 +88,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.11.5":         false,
 	"1.11.6":         false,
 	"1.11.7":         false,
-	"1.11.8":         true,
+	"1.11.8":         false,
```
jackfrancis: Does this change preclude 1.11.8 from Azure Stack as well?

jackfrancis: Preclude is the wrong verb. Will Azure Stack scenarios still be able to build 1.11.8 clusters (using their own VHD)?

packer/install-dependencies.sh
honcao: @CecileRobertMichon Thanks for the notification.
@rjaini FYI, we would need image for 1.11.10

CecileRobertMichon: any updates on getting 1.11.10 out for Azure Stack?

rjaini: Since this payload has new API version changes related to Azure cloud provider. We need to validate more thoroughly. We will update once done. Mostly by Monday we should be done.
@jadarsie told me that this is not blocking. Is that blocking ?.

CecileRobertMichon: not blocking, just trying to see if we should wait for it or go ahead and merge this PR without it and then do a follow up for Azure Stack.

jackfrancis: Let's not block. Ostensibly the extra docker image overhead is a no-op in terms of cleanup (i.e., all unneeded, extra images will be reliably cleaned up at provision time). Also, we have a storage alert in place, so we'll know if we run out of local storage to accommodate extra -azs images.

packer/install-dependencies.sh
CecileRobertMichon: let's keep 1.11.8-azs in there for now until we have a 1.11.10-azs

/hold cancel

pkg/api/common/versions.go
CecileRobertMichon: Good point. If we want to keep Azure Stack 1.11.8 support until 1.11.10 is available we shouldn't disable it.

CecileRobertMichon: (the answer to your question is no,  Azure Stack scenarios will not be able to build 1.11.8 clusters with this change).

jackfrancis: Great, yeah, let's undo this. In order to maintain non-Azure Stack velocity, let's be more permissive in allowing > 2 patch versions so that non-Azure Stack can move ahead quickly and not block on Azure Stack.

mboersma: I left 1.11.8 enabled for now and created #1219 so we remove it later when AzureStack support for 1.11.10 has been finalized.

.chglog/config.yml
```diff
@@ -18,9 +18,10 @@ options:
       style: Code Style 🎶
       test: Testing 💚
   header:
-    pattern: "^(\\w*)\\:\\s(.*)$"
+    pattern: "^(\\w*)(?:\\(([\\w\\$\\.\\-\\*\\s]*)\\))?\\:\\s(.*)$"
```
CecileRobertMichon: rebase error?

mboersma: Yes indeed: fixed now.

PR #1196, opened 2019-05-01
ci: permit optional scope field when generating changelog

mboersma:
**Reason for Change**:
Some commits didn't show up in the [v0.35.0](https://github.com/Azure/aks-engine/releases/tag/v0.35.0) changelog initially because their message contained a scope field. This was because the `git-chglog` config file was generated initially with the simplified option. Updating the regex to the `msg(scope):` form fixes it.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.chglog/config.yml
+++ b/.chglog/config.yml
@@ -18,9 +18,10 @@ options:
       style: Code Style 🎶
       test: Testing 💚
   header:
-    pattern: "^(\\w*)\\:\\s(.*)$"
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
@mboersma (sorry being picky) I would merge this with a "ci" or "chore" category since this isn't fixing any bugs from a user perspective.

PR #1197, opened 2019-05-01
style: don't use shell variables in printf format string

mboersma:
**Reason for Change**:
Enforces the use of arguments to `printf` rather than embedded variables, which can be problematic.

See https://github.com/koalaman/shellcheck/wiki/SC2059

**Issue Fixed**:

Refs #953 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@mboersma see https://github.com/Azure/aks-engine/pull/1199

PR #1200, opened 2019-05-01
style: remove quotes that accidentally unquote

mboersma:
**Reason for Change**:
Warns about shell scripts that contain apparently accidental quoting.

See https://github.com/koalaman/shellcheck/wiki/SC2027

**Issue Fixed**:

Refs #953 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/health-monitor.sh
```diff
@@ -75,5 +75,5 @@ if [[ "${component}" == "container-runtime" ]]; then
 elif [[ "${component}" == "kubelet" ]]; then
   kubelet_monitoring
 else
-  echo "Health monitoring for component "${component}" is not supported!"
+  echo "Health monitoring for component ${component} is not supported!"
```
jackfrancis: Borrowed directly from here https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/health-monitor.sh#L111 :)

mboersma: True—we didn't write this script originally. The mis-quoting is benign in this case since it's just an `echo` command, but the shellcheck rule seems useful.

PR #1202, opened 2019-05-02
style: replace unneeded 'cat' commands

mboersma:
**Reason for Change**:

Warns about useless uses of `cat` ([UUOC](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat)).

See https://github.com/koalaman/shellcheck/wiki/SC2002

**Issue Fixed**:

Refs #953 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -224,8 +224,8 @@ setKubeletOpts () {
 }
 
 ensureCCProxy() {
-    cat $CC_SERVICE_IN_TMP | sed 's#@libexecdir@#/usr/libexec#' > /etc/systemd/system/cc-proxy.service
-    cat $CC_SOCKET_IN_TMP sed 's#@localstatedir@#/var#' > /etc/systemd/system/cc-proxy.socket
```
mboersma: This looks like it was a bug in setting up the clear containers proxy.

parts/k8s/cloud-init/artifacts/cse_customcloud.sh
```diff
@@ -58,6 +58,7 @@ configureK8sCustomCloud() {
         KUBERNETES_FILE_DIR=$(dirname "${AZURE_JSON_PATH}")
         K8S_CLIENT_CERT_PATH="${KUBERNETES_FILE_DIR}/k8s_auth_certificate.pfx"
         echo $SERVICE_PRINCIPAL_CLIENT_SECRET_CERT | base64 --decode > $K8S_CLIENT_CERT_PATH
+        # shellcheck disable=SC2002
```
mboersma: This usage and the one below are piping `cat` in order to modify a single file in-place. I don't know of another way to do that with `jq` except by using a temp file.

PR #1208, opened 2019-05-02
fix: detect GPU VMs with _Promo SKU suffix

### LINKED ISSUE #1203: Add `Standard_NC6_Promo` to list of GPU virtual machines

No

---

Issue
---

v0.35.0
---

**Kubernetes version**:
1.14.1

**What happened**:
When deploying a cluster with a node pool of size `Standard_NC6_Promo`, the Nvidia driver is not installed, and thus the GPU is not exposed as an available resource.

GPU should be exposed as an available resource just like it is for regular `Standard_NC6` instances.

Deploy a cluster with a `Standard_NC6_Promo` pool.

I was able to fix it locally. Here's the diff:

```
diff --git a/pkg/api/common/helper.go b/pkg/api/common/helper.go
index baa3811fc..4524ff40e 100644
--- a/pkg/api/common/helper.go
+++ b/pkg/api/common/helper.go
@@ -77,6 +77,7 @@ func IsNvidiaEnabledSKU(vmSize string) bool {
        dm := map[string]bool{
                // K80
                "Standard_NC6":   true,
+               "Standard_NC6_Promo":   true,
                "Standard_NC12":  true,
                "Standard_NC24":  true,
                "Standard_NC24r": true,
@@ -121,6 +122,10 @@ func GetNSeriesVMCasesForTesting() []struct {
                        "Standard_NC6",
                        true,
                },
+               {
+                       "Standard_NC6_Promo",
+                       true,
+               },
                {
                        "Standard_NC12",
                        true,
```

I can confirm that this works for me but wasn't sure if there was more to it, thus not submitting a patch at this stage. Please let me know if you would like me to do so?

Regards,
Nicolas

ohthehugemanatee:
**Reason for Change**:
The `_Promo` SKU suffix is added to price-reduced variants with no difference in the underlying hardware. Without this patch, `Standard_NC6` is detected as a GPU VM (for example), but `Standard_NC6_Promo` is not. 

**Issue Fixed**:
Fixes #1203

**Requirements**:

- [ ] includes documentation (NA)
- [ ] adds unit tests (no, adds to an existing unit test)
- [ ] tested upgrade from previous version (NA)

**Notes**:

### DIFF
--- a/pkg/api/common/helper.go
+++ b/pkg/api/common/helper.go
@@ -101,6 +101,8 @@ func IsNvidiaEnabledSKU(vmSize string) bool {
 		"Standard_NC24s_v3":  true,
 		"Standard_NC24rs_v3": true,
 	}
+	// Trim the optional _Promo suffix.
+	vmSize = strings.TrimSuffix(vmSize, "_Promo")
 	if _, ok := dm[vmSize]; ok {
 		return dm[vmSize]
 	}
--- a/pkg/api/common/helper.go
+++ b/pkg/api/common/helper.go
@@ -122,6 +122,10 @@ func GetNSeriesVMCasesForTesting() []struct {
 			true,
 		},
 		{
+			"Standard_NC6_Promo",
+			true,
+		},
+		{
 			"Standard_NC12",
 			true,
 		},


ohthehugemanatee:
Looks like the issue was already resolved in another PR. Canceling this one...

PR #1213, opened 2019-05-02
style: sudo doesn't affect redirects

mboersma:
**Reason for Change**:
Points out in shell scripts when a `sudo` command is redirected, since that's likely to be a bug.

See https://github.com/koalaman/shellcheck/wiki/SC2024

**Issue Fixed**:
Refs #953 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
```diff
@@ -2,7 +2,7 @@
 set -x
 source /opt/azure/containers/provision_source.sh
 
-sudo echo "  dns-search <searchDomainName>" >> /etc/network/interfaces.d/50-cloud-init.cfg
```
mboersma: I think this script runs as root, so perhaps this wasn't causing an actual bug.

jackfrancis: yeah, let's enable the check but also get rid of the `sudo`

PR #1216, opened 2019-05-03
refactor: remove _Promo from static SKU lists

### LINKED ISSUE #1203: Add `Standard_NC6_Promo` to list of GPU virtual machines

### LINKED ISSUE #1209: Support _Promo SKU suffix on all SKUs
---
Nope

---
Issue

---
v0.35.0

**Kubernetes version**:
n/a

**What happened**:
Azure put a VM size on promotion, which added a `_Promo` suffix to the SKU. Now aks-engine doesn't like my SKUs or doesn't detect their capabilities (eg GPU). 

The VM should work identically to its' non-suffixed brethren.

This was recently fixed for current VM sizes in #1101 by adding the current crop of `_Promo` SKUs to static lists. This issue is about taking a different approach and actually ignoring the `_Promo` suffix whenever it is encountered.

ohthehugemanatee:
**Reason for Change**:
The `_Promo` SKU suffix is added to price-reduced variants with no difference in the underlying hardware. There's no reason to kick off a `chore: update SKUs` task every time Azure runs a different promotion. Also, GPU detection (currently done by SKU matching) should not depend on the promo status of an SKU. 

**Issue Fixed**:
Fixes #1203
Fixes #1209

**Requirements**:

- [ ] includes documentation (NA)
- [ ] adds unit tests (no, adds to an existing unit test)
- [ ] tested upgrade from previous version (NA)

**Notes**:

Identical to PR #1208 except this also updates `generate_azure_constants.py` and the azure constants files.


ohthehugemanatee:
Note: I don't understand what the constants lists are used for, so I'm not 100% that removing the `_Promo` SKUs is the right move here.. I can see them referenced in the template files, but they don't appear to be used for validation: I could successfully create ARM templates for a cluster using the `not_a_real_sku` SKU. 

Also: 
* Why do we have tests to validate that a statically declared constant is the value which was statically declared? `TestGetDCOSMasterAllowedSizes` `TestGetAzureLocations` etc are not testing anything.
* If the list of SKUs is dynamically generated from `az` anyway... why do it in python and save them in the repo? Wouldn't it be better to do it in go and cache the list locally? Or run the generator as a build step?

ohthehugemanatee:
/assign @CecileRobertMichon 
Would appreciate your feedback!

mboersma:
I think we shouldn't remove the `_Promo` types from `azureconst.go`. Eventually this ends up populating the allowed sizes in the ARM template, so removing them would indicate they can't be used. (This validation happens at template deployment time, not at the time of `aks-engine generate`, so that's probably why you didn't see errors with `not_a_real_sku`.)

Having `IsNvidiaEnabledSKU` accept the parallel `_Promo` types seems like a good fix, but removing them could cause other issues. I see your point about `_Promo` types changing often, though...not sure how to reconcile that with strict ARM validation in the template.

ohthehugemanatee:
What if:
* ditch the constants manual generation script and its' in-the-repo constants approach.
* rewrite the existing constants functions to query Azure for current "allowed SKUs".  It should keep a local cache with ~12hr soft expiry as a failover for connectionless operation (is that a requirement for `aks-engine generate`?).
* Ignore the `_Promo` suffix for GPU detection specifically (already done)

jackfrancis:
I think the value proposition of the local constants is to "fail fast" when a user fat fingers a VM SKU, rather than waiting up to several minutes for ARM to parse/process the deployment and fail itself.

Having said that, we could possily just add "suffix tolerance" to all SKUs. So let's say we ship `_Promo` as the first instance of that tolerance. When we get a SKU string input we first of all strip the `_Promo` substring from it before processing it for validation against the const allowed list.

Also we'd have to explicitly update the const generation script to no longer literally include `_Promo` SKUs.

Basically, as you say @ohthehugemanatee, we'll just assume that at any point, any VM SKU may be co-opted by this promotion mechanism and become a `_Promo` SKU. And we tolerate that.

This change should be relatively lightweight, and removes most of the friction from enabling folks to use dynamic SKU promotions in their aks-engine workflows.

Does the above make sense as a near-term path forward?

ohthehugemanatee:
@jackfrancis not only does that make perfect sense to me... it's how this PR is already written.

* all _Promo SKUs are removed from the constants list
* the constants file generation script ignores _Promo SKUs
* GPU detection ignores the _Promo suffix.

Looks like the constants are only otherwise in use in the template files. I tried test-generating a cluster with a gibberish SKU and got no validation errors, so I think this patch is complete.

CecileRobertMichon:
@ohthehugemanatee this is failing unit tests:
`--- FAIL: TestK8sVars (46.22s)
    armvariables_test.go:432: unexpected diff while expecting equal structs: root["vmSizesMap"]["Standard_D11_v2_Promo"]`

ohthehugemanatee:
gah. There were a bunch more _Promo SKUs hard coded into tests which I had apparently missed. Anyway now we're green to go. @CecileRobertMichon  may I please have the lgtm tag back? :)

pkg/helpers/helpers.go
```diff
@@ -117,25 +117,18 @@ func CreateSSH(rg io.Reader, s *i18n.Translator) (privateKey *rsa.PrivateKey, pu
 // AcceleratedNetworkingSupported check if the VmSKU support the Accelerated Networking
 func AcceleratedNetworkingSupported(sku string) bool {
 	switch sku {
```
jackfrancis: I think we want something like this here:

```
// Trim the optional _Promo suffix.
		sku = strings.TrimSuffix(sku, "_Promo")
```

Basically since we're tolerant to the `_Promo` suffix and assume it's the equivalent as the non-promo SKU we want to just strip it if it's present so we can mark equivalent promo SKUs appropriately for accelerated networking support.

jackfrancis:
This looks good! I added a note to ensure we don't mark promo SKUs as not being capable of accelerated networking. Thanks so much!

pkg/helpers/helpers.go
ohthehugemanatee: done

PR #223, opened 2019-01-03
feat: add support for Ubuntu 18.04-LTS

CecileRobertMichon:
**Reason for Change**:

Add support for Ubuntu 18.04-LTS. This includes a new VHD that is based on top of 18.04. The default is still 16.04-LTS so this will have to be specified explicitly for now. Upgrades and scale operations are unaffected.

**Issue Fixed**:

Fixes #222

**Requirements**:

- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
- [X] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1223, opened 2019-05-03
refactor: rename AKS VHD distros

### LINKED ISSUE #1246: rename "aks-16.04" and "aks-18.04" distros to "aks-ubuntu-<LTS>"
To anticipate future non-ubuntu distro, let's rename our sort of shorthand "aks-<lts>" distro values to more explicitly express that they are ubuntu.

CecileRobertMichon:
**Reason for Change**:

~Switch the default distro to `aks-ubuntu-18.04` (previously `aks`) which is the AKS image based on Ubuntu 18.04-LTS. Ubuntu 16.04-LTS will still be available as `"Distro": "aks-ubuntu-16.04"` (for VHD) or `"Distro": "ubuntu"` (vanilla Ubuntu image).~

Renames the `aks` distro to `aks-ubuntu-16.04` and `aks-18.04` to `aks-ubuntu-18.04` for consistency and clarity. New cluster deployments will need to use `"Distro": "aks-ubuntu-16.04"` to get the Ubuntu 16.04-LTS VHD. Existing clusters running a scale or an upgrade operation will automatically switch to use the `aks-ubuntu-16.04` distro if previously using `aks`. The base image remains exactly the same, this is just a distro rename.

**Issue Fixed**:

~1025~
Fixes #1246

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
BREAKING CHANGE: the `aks` and `aks-18.04` distro are now deprecated. Users can use the `aks-ubuntu-16.04` distro value to keep building new clusters based on Ubuntu-16.04-LTS and `aks-ubuntu-18.04` for Ubuntu-18.04-LTS. The default distro is now `aks-ubuntu-16.04` for all regions where the AKS VHD image is available.


jackfrancis:
Can we introduce a new distro `AKS1604` as part of this PR, which is essentially an alias to `AKS`?

pkg/api/defaults.go
```diff
@@ -308,7 +308,7 @@ func (p *Properties) setExtensionDefaults() {
 func (p *Properties) setMasterProfileDefaults(isUpgrade bool) {
 	if p.MasterProfile.Distro == "" {
 		if p.OrchestratorProfile.IsKubernetes() {
-			p.MasterProfile.Distro = AKS
+			p.MasterProfile.Distro = AKS1804
 		} else {
 			p.MasterProfile.Distro = Ubuntu
```
CecileRobertMichon: non-kubernetes clusters stay on Ubuntu 16.04

pkg/api/azenvtypes.go
```diff
@@ -100,8 +100,8 @@ var (
 		DockerComposeDownloadURL: "https://github.com/docker/compose/releases/download",
 	}
 
-	//DefaultUbuntuImageConfig is the default Linux distribution.
-	DefaultUbuntuImageConfig = AzureOSImageConfig{
+	//Ubuntu1604ImageConfig is the default Linux distribution.
+	Ubuntu1604ImageConfig = AzureOSImageConfig{
```
CecileRobertMichon: just a few renames for consistency, some image configs had the word "default" in their name, others didn't.

pkg/api/azenvtypes.go
```diff
@@ -190,12 +191,13 @@ var (
...
 			Ubuntu1804: Ubuntu1804ImageConfig,
-			RHEL:       DefaultRHELOSImageConfig,
-			CoreOS:     DefaultCoreOSImageConfig,
-			AKS:        DefaultUbuntuImageConfig,
-			AKS1804:    Ubuntu1804ImageConfig,
+			RHEL:       RHELOSImageConfig,
+			CoreOS:     CoreOSImageConfig,
+			AKS:        Ubuntu1604ImageConfig,
+			AKS1604:    Ubuntu1604ImageConfig,
+			AKS1804:    Ubuntu1604ImageConfig, // workaround for https://github.com/Azure/aks-engine/issues/761
```
CecileRobertMichon: this is an ugly hack to prevent Sovereign cloud users to run into #761 by default... Normally, this should have been `Ubuntu1804ImageConfig` but the `ubuntu-18.04` image causes OS provisioning timeouts so we keep using `ubuntu-16.04` for US Gov and German Cloud.

pkg/api/const.go
```diff
@@ -29,8 +29,9 @@ const (
 	Ubuntu1804      Distro = "ubuntu-18.04"
 	RHEL            Distro = "rhel"
 	CoreOS          Distro = "coreos"
-	AKS             Distro = "aks"
-	AKSDockerEngine Distro = "aks-docker-engine" // deprecated docker-engine distro
+	AKS             Distro = "aks"               // deprecated AKS 16.04 distro. Equivalent to aks-16.04.
```
jackfrancis: Rather than deprecating this, we could just call it a "mutable" reference. So at some point aks is redirected from 16.04 to 18.04

pkg/api/defaults-kubelet.go
```diff
@@ -143,10 +143,6 @@ func (cs *ContainerService) setKubeletConfig() {
 		setMissingKubeletValues(cs.Properties.MasterProfile.KubernetesConfig, o.KubernetesConfig.KubeletConfig)
 		addDefaultFeatureGates(cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig, o.OrchestratorVersion, "", "")
 
-		if cs.Properties.MasterProfile.IsUbuntu1804() {
```
jackfrancis: I'm a little confused why we'd remove these Ubuntu 18.04-specific defaults?

pkg/api/defaults.go
```diff
@@ -476,23 +488,39 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
 
+		// Override the --resolv-conf kubelet config value for Ubuntu 18.04 after the distro value is set.
+		if profile.IsUbuntu1804() {
+			if profile.KubernetesConfig == nil {
+				profile.KubernetesConfig = &KubernetesConfig{}
+			}
+			if profile.KubernetesConfig.KubeletConfig == nil {
+				profile.KubernetesConfig.KubeletConfig = map[string]string{}
+			}
+			profile.KubernetesConfig.KubeletConfig["--resolv-conf"] = "/run/systemd/resolve/resolv.conf"
```
jackfrancis: Is there no way to ensure that we set the distro before we set the kubelet defaults? That way we can move this default enforcement into the "set default kubelets configuration" flow.

pkg/api/const.go
jackfrancis: so I rescind this, it's better to be precise and expressive about which OS version a distro is based upon

pkg/api/const.go
```diff
@@ -29,8 +29,9 @@ const (
...
+	AKSDockerEngine Distro = "aks-docker-engine" // deprecated docker-engine distro.
+	AKS1604         Distro = "aks-16.04"
```
jackfrancis: Suggestion: lets' change this value to "aks-ubuntu-16.04" and "aks-ubuntu-18.04. Let's assume that non-Ubuntu Linux distros may come to AKS eventually, and so the simple "16.04" string is arguably not descriptive enough.

CecileRobertMichon: I agree that's better. This will break back compat though so we can't simply change the string value. If we want to do this for 18.04 we need a new distro with deprecation logic (or an alias distro that points to the same image config).

pkg/api/defaults-kubelet.go
CecileRobertMichon: not removed, moved to defaults.go

pkg/api/defaults.go
CecileRobertMichon: Not really. The kubelet defaults are part of `orchestratorProfile` defaults, which is before `masterProfile` / `agentPoolProfile` defaults. Changing that order seems like a riskier and more complex change to me than moving the kubelet distro-specific override to the profile defaults.

pkg/api/const.go
CecileRobertMichon: yeah that was my initial thinking as well but after digging into the code I decided it was more clean to enforce explicit OS versioned distros.

pkg/api/const.go
jackfrancis: Where's the back-compat breakage, for folks using the "aks-18.04" distro?

pkg/api/defaults.go
jackfrancis: If this is really a 16.04- and 18.04-specific setting, then we should also set the value for 16.04. We shouldn't assume some prior default enforcement has already been set with 16.04 defaults in mind.

pkg/api/const.go
CecileRobertMichon: if you have an existing "aks-18.04" and you try to upgrade or scale with this change, the "aks-18.04" distro won't be a valid distro and will return an error since it will have been replaced by "aks-ubuntu-18.04".

jackfrancis: https://github.com/Azure/aks-engine/issues/1246

pkg/api/defaults.go
CecileRobertMichon: Do you mean explicitly set it to the default (`"/etc/resolv.conf"`) for Ubuntu 16.04? This isn't really a 16.04-/18.04 setting, it's more of a workaround to fix container DNS that is broken with 18.04 due to a breaking change: https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1624320/comments/8

CecileRobertMichon: Btw this isn't a new change, I just moved it after finding out that kubelet defaults were set before the default distro was set. The original change was https://github.com/Azure/aks-engine/pull/223/files#r265284630

pkg/api/const.go
CecileRobertMichon: arguably it's easier and less risky to do before we move the default (since less clusters will be using the deprecated string) so doing the change in this PR might be easier, was just pointing out that this wasn't a trivial string rename.

pkg/api/defaults.go
jackfrancis: Got it. So basically it's 18.04-only (as of now) where we have to actually make this change.

jackfrancis: For consistency, can we also move the equivalent master profile distro-dependent config into the master profile defaults flow and out of the kubelet flow? At least then we have a consistent enforcement pattern when we reconcile kubeletConfig + nodepool.

CecileRobertMichon: that's already done, unless I'm misunderstanding

CecileRobertMichon: https://github.com/Azure/aks-engine/pull/1223/files#diff-fd43f080184ed15731e5d7e0d917d6e1R325

pkg/api/const.go
CecileRobertMichon: done

pkg/api/defaults-kubelet_test.go
```diff
@@ -544,57 +544,6 @@ func TestStaticWindowsConfig(t *testing.T) {
 	}
 }
 
-func TestUbuntu1804Flags(t *testing.T) {
```
CecileRobertMichon: moved to defaults_test.go

CecileRobertMichon:
@jadarsie could you please review this and its potential side effects on Azure Stack?

pkg/api/azenvtypes.go
```diff
@@ -116,40 +116,40 @@ var (
 		ImageVersion:   "latest",
 	}
 
-	//DefaultRHELOSImageConfig is the RHEL Linux distribution.
-	DefaultRHELOSImageConfig = AzureOSImageConfig{
+	//RHELOSImageConfig is the RHEL Linux distribution.
+	RHELOSImageConfig = AzureOSImageConfig{
```
jadarsie: I would also drop `OS` from `RHELOS`

jadarsie: Or add the `OS` suffix to `Ubuntu1604`

pkg/api/const.go
```diff
@@ -25,14 +25,16 @@ const (
...
-	Ubuntu1804      Distro = "ubuntu-18.04"
-	RHEL            Distro = "rhel"
-	CoreOS          Distro = "coreos"
-	AKS             Distro = "aks"
-	AKSDockerEngine Distro = "aks-docker-engine" // deprecated docker-engine distro
-	AKS1804         Distro = "aks-18.04"
-	ACC1604         Distro = "acc-16.04"
+	Ubuntu            Distro = "ubuntu"
+	Ubuntu1804        Distro = "ubuntu-18.04"
+	RHEL              Distro = "rhel"
```
jadarsie: Adding a version component to consts `rhel` and `coreos` may fit this PR

CecileRobertMichon: I don't think that's in the scope of this PR, I would rather leave the non-Ubuntu OSes untouched. Also, we use versions `latest` for both of these.

pkg/api/azenvtypes.go
CecileRobertMichon: done

jadarsie:
> @jadarsie could you please review this and its potential side effects on Azure Stack?

Thx @CecileRobertMichon. The full Azure marketplace is not available by default on Azure Stack, admins have to indicate which specific items should be available/syndicated for tenants use.

Therefore, this PR does not worsen the current situation for Azure Stack tenants. E.x.: deployments will currently fail if Ubuntu 16.04 is not syndicated.

However, if possible, I would like to keep the current default (Ubuntu 16.04) for Azure Stack deployments. We have have 0 mileage with Ubuntu 18.04. We can create a new PR to address this if preferred.

At some point, we should probably add an Azure Stack validation to fail fast and provide guidance if the default/selected distro is not yet syndicated on the stamp marketplace. We have public documentation that lists dependencies, but you know...

jadarsie:
@honcao feel free to chime in if something looks suspicious

CecileRobertMichon:
Pivot this PR to only to the refactoring part and table moving to Ubuntu 18.04 by default as testing has shown that timesync isn't reliable on 18.04 VMs https://github.com/Azure/aks-engine/issues/1306

pkg/api/defaults.go
```diff
@@ -308,7 +308,7 @@ func (p *Properties) setExtensionDefaults() {
 func (p *Properties) setMasterProfileDefaults(isUpgrade bool) {
```
jackfrancis: Do we need the conversion of deprecated distros used by master vms during upgrade in this func?

CecileRobertMichon: whoops yes you're right... I guess `AKSDockerEngine` didn't need it because it was an agent only distro

CecileRobertMichon: done

pkg/api/defaults_test.go
```diff
@@ -1077,45 +1077,94 @@ func TestAgentPoolProfile(t *testing.T) {
...
 
 	var tests = []struct {
-		name                string              // test case name
-		orchestratorProfile OrchestratorProfile // orchestrator to be tested
-		expectedDistro      Distro              // expected result default disto to be used
+		name                   string              // test case name
+		orchestratorProfile    OrchestratorProfile // orchestrator to be tested
+		masterProfileDistro    Distro
+		agentPoolProfileDistro Distro
+		expectedDistro         Distro // expected result default disto to be used
```
jackfrancis: Are we testing both agent profile and master profile distro results?

CecileRobertMichon: we are but the test was setup to have a common distro for both

CecileRobertMichon: I can change it to have distinct expected values

Issue #1158, opened 2019-04-26
Highly Available Private Cluster leveraging VMSS + AZ's


keyoke:
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): Issue/Question

---

**What version of aks-engine?**:

Version: v0.34.3
GitCommit: 3fbbf8225
GitTreeState: clean

---

**Kubernetes version**:
1.12

**What happened**:

$ az group deployment create --resource-group [RG] --template-file azuredeploy.json --parameters @azuredeploy.parameters.json
Azure Error: InvalidTemplate
Message: Deployment template validation failed: 'The template variable 'masterPublicIPAddressName' is not found. Please see https://aka.ms/arm-template/#variables for usage details.'.

**What you expected to happen**:
Cluster should be successfully created

**How to reproduce it** (as minimally and precisely as possible):
 "privateCluster": {
            "enabled": true,
}
...
 "masterProfile": {
        "count": 5,
        "vmSize": "Standard_DS2_v2",
       ...
        "availabilityProfile": "VirtualMachineScaleSets",
        "availabilityZones": [
            "1",
            "2"
        ]
      }

**Anything else we need to know**:
acs-engine tgithiub repro was tracking this issue but not clear after renaming to aks-engine - https://github.com/Azure/acs-engine/issues/3932

PR #1226, opened 2019-05-04
fix: private cluster with VMSS masters, jumpbox

alexeldeib:
**Reason for Change**:

Private clusters do not currently deploy due to mixing of jumpbox and master resources, leading to variables absent or resources duplicated in the ARM templates. This PR separates and dedupes in the appropriate places, and fixes the corresponding tests to account for the behavior.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/masterarmresources.go
+++ b/pkg/engine/masterarmresources.go
@@ -146,9 +146,11 @@ func createKubernetesMasterResourcesVMSS(cs *api.ContainerService) []interface{}
 		masterResources = append(masterResources, internalLb)
 	}
 
-	publicIPAddress := CreatePublicIPAddress()
-	loadBalancer := CreateLoadBalancer(cs.Properties, true)
-	masterResources = append(masterResources, publicIPAddress, loadBalancer)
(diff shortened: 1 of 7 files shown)


alexeldeib:
Related to https://github.com/Azure/aks-engine/issues/1158

alexeldeib:
One thing incomplete here is the networking setup when using private cluster + jumpbox. I wasn't sure what the desired output would be. I think it should be a single NSG with port 22 allowed from all sources and port 6443 (kube-tls) allowed inside VNET only? That should be okay because masters won't have any public IP?

alexeldeib:
Technically, I guess the current config is sufficient: jumpbox has public IP and is open on port 22, while master NSG *which has allow all on 443* doesn't expose any public IPs, only private. I'd probably prefer to make 443 vnet internal. WDYT?

jackfrancis:
lgtm

CecileRobertMichon:
@alexeldeib can you please add a small description of what the PR does for future context?

alexeldeib:
@CecileRobertMichon oops, how's that?

PR #999, opened 2019-04-10
chore(CIS): add protect-kernel-defaults

andyzhangx:
**Reason for Change**:

fix: add protect-kernel-defaults for CIS conformance

about `protect-kernel-defaults`: Default kubelet behaviour for kernel tuning. If set, kubelet errors if any of kernel tunables is different than kubelet defaults.

```
[FAIL] 2.1.6 Ensure that the --protect-kernel-defaults argument is set to true (Scored)
2.1.6 If using a Kubelet config file, edit the file to set protectKernelDefaults: true .
If using command line arguments, edit the kubelet service file
/etc/systemd/system/kubelet.service on each worker node and
set the below parameter in KUBELET_SYSTEM_PODS_ARGS variable.
--protect-kernel-defaults=true
Based on your system, restart the kubelet service. For example:
systemctl daemon-reload
systemctl restart kubelet.service
```

**Issue Fixed**:

Fixes #446

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

/assign @jackfrancis @CecileRobertMichon
cc  @feiskyer

PR #1227, opened 2019-05-06
chore(CIS):enable protect-kernel-defaults

andyzhangx:
**Reason for Change**:

chore(CIS):enable protect-kernel-defaults
PR(https://github.com/Azure/aks-engine/pull/999) does not enable `--protect-kernel-defaults` completely, this PR fixes this issue since below latest Ubuntu image has already enabled the required kernel flags in the VHD image:
```
    "agentpoolosImageSKU": {
      "value": "aks-ubuntu-1604-201904"
    },
    "agentpoolosImageVersion": {
      "value": "2019.04.24"
    },
```
/assign @jackfrancis 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/hold

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -81,11 +81,7 @@ func (cs *ContainerService) setKubeletConfig() {
 		"--streaming-connection-idle-timeout": "5m",
 	}
 
-	// "--protect-kernel-defaults" is true is currently only valid using base Ubuntu OS image
-	// until the changes are baked into a VHD
-	if cs.Properties.IsUbuntuDistroForAllNodes() {
-		defaultKubeletConfig["--protect-kernel-defaults"] = "true"
-	}
+	defaultKubeletConfig["--protect-kernel-defaults"] = "true"
 
 	// Set --non-masquerade-cidr if ip-masq-agent is disabled on AKS
 	if !cs.Properties.IsIPMasqAgentEnabled() {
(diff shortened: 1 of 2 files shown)


jackfrancis:
@andyzhangx should we expect this kubelet setting to work for Windows nodes?

andyzhangx:
> @andyzhangx should we expect this kubelet setting to work for Windows nodes?

I don't think so, so I only add this flag in linux node.

pkg/api/defaults-kubelet.go
```diff
@@ -58,6 +58,7 @@ func (cs *ContainerService) setKubeletConfig() {
 	staticWindowsKubeletConfig["--image-pull-progress-deadline"] = "20m"
 	staticWindowsKubeletConfig["--resolv-conf"] = "\"\"\"\""
 	staticWindowsKubeletConfig["--eviction-hard"] = "\"\"\"\""
+	staticWindowsKubeletConfig["--protect-kernel-defaults"] = ""
```
jackfrancis: @andyzhangx FYI the above pattern is the current way to explicitly remove kubelet flags for Windows nodes. See the `removeKubeletFlags` func below, which will eventually process `""` vals by removing the keys entirely.

pkg/api/defaults-kubelet_test.go
```diff
@@ -87,10 +88,11 @@ func TestKubeletConfigDefaults(t *testing.T) {
 	expected["--resolv-conf"] = "\"\"\"\""
 	expected["--eviction-hard"] = "\"\"\"\""
 	delete(expected, "--pod-manifest-path")
+	delete(expected, "--protect-kernel-defaults")
```
jackfrancis: and @andyzhangx this is the unit test that validates that `--protect-kernel-defaults` is not present as a kubelet flag for Windows nodes

PR #1228, opened 2019-05-06
feat: Auto-generate ssh pubkey

### LINKED ISSUE #689: Make aks-engine generate command automatically generate SSH key pairs when publicKeys is empty
**Is this a request for help?**: No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): FEATURE REQUEST

---

**What version of aks-engine?**: v0.31.1

---

**Kubernetes version**: N/A

**What happened**: aks-engine generate command currently requires publicKeys's keyData field to be non-empty.

**What you expected to happen**: aks-engine generate command should automatically generate SSH key pairs and populate it.

**How to reproduce it** (as minimally and precisely as possible): Just run aks-engine generate with a cluster definition with empty publicKeys keyData.

**Anything else we need to know**: This would be really helpful as it doesn't require user to manually generate SSH key pair.

ohthehugemanatee:
**Reason for Change**:
`generate` fails out if no public key is given in the template. This is inconsistent; in the same case `deploy` generates a new keypair on the fly.

**Issue Fixed**:
Fixes #689 

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Notes**:
Apart from the functionality described, also:
* removes validation step checking for empty pubkey. Validation happens before loading, so this seemed the most appropriate approach. Would be better to load and then validate the loaded template, of course...
* removes unit test for that removed validation step.

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -171,7 +171,7 @@ func (gc *generateCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
 		prop.CertificateProfile.CaPrivateKey = string(caKeyBytes)
 	}
 	// Create an SSH keypair if it's missing.
-	gc.containerService, err = api.CheckCreateKeyPair(gc.containerService, gc.locale, gc.outputDirectory)
+	err = api.CheckCreateKeyPair(gc.containerService, gc.locale, gc.outputDirectory)
 	if err != nil {
 		return errors.Wrap(err, "error creating missing SSH Keypair")
 	}
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/aks-engine/pkg/engine"
 	"github.com/Azure/aks-engine/pkg/engine/transform"
+	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/aks-engine/pkg/i18n"
 	"github.com/leonelquinteros/gotext"
 	"github.com/pkg/errors"
@@ -166,6 +167,20 @@ func (gc *generateCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
 		prop.CertificateProfile.CaPrivateKey = string(caKeyBytes)
 	}
 
+	// If no keypair has been supplied, generate one.
+	if gc.containerService.Properties.LinuxProfile != nil && (gc.containerService.Properties.LinuxProfile.SSH.PublicKeys == nil ||
+		len(gc.containerService.Properties.LinuxProfile.SSH.PublicKeys) == 0 ||
+		gc.containerService.Properties.LinuxProfile.SSH.PublicKeys[0].KeyData == "") {
+		translator := &i18n.Translator{
+			Locale: gc.locale,
+		}
+		var publicKey string
+		_, publicKey, err = helpers.CreateSaveSSH(gc.containerService.Properties.LinuxProfile.AdminUsername, gc.outputDirectory, translator)
+		if err != nil {
+			return errors.Wrap(err, "Failed to generate SSH Key")
+		}
+		gc.containerService.Properties.LinuxProfile.SSH.PublicKeys = []api.PublicKey{{KeyData: publicKey}}
+	}
 	return nil
 }
 
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -12,7 +12,6 @@ import (
 	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/aks-engine/pkg/engine"
 	"github.com/Azure/aks-engine/pkg/engine/transform"
-	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/aks-engine/pkg/i18n"
 	"github.com/leonelquinteros/gotext"
 	"github.com/pkg/errors"
@@ -171,20 +170,10 @@ func (gc *generateCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
 		prop.CertificateProfile.CaCertificate = string(caCertificateBytes)
 		prop.CertificateProfile.CaPrivateKey = string(caKeyBytes)
 	}
-
-	// If no keypair has been supplied, generate one.
-	if gc.containerService.Properties.LinuxProfile != nil && (gc.containerService.Properties.LinuxProfile.SSH.PublicKeys == nil ||
-		len(gc.containerService.Properties.LinuxProfile.SSH.PublicKeys) == 0 ||
-		gc.containerService.Properties.LinuxProfile.SSH.PublicKeys[0].KeyData == "") {
-		translator := &i18n.Translator{
-			Locale: gc.locale,
-		}
-		var publicKey string
-		_, publicKey, err = helpers.CreateSaveSSH(gc.containerService.Properties.LinuxProfile.AdminUsername, gc.outputDirectory, translator)
-		if err != nil {
-			return errors.Wrap(err, "Failed to generate SSH Key")
-		}
-		gc.containerService.Properties.LinuxProfile.SSH.PublicKeys = []api.PublicKey{{KeyData: publicKey}}
+	// Create an SSH keypair if it's missing.
+	gc.containerService, err = api.CheckCreateKeyPair(gc.containerService, gc.locale, gc.outputDirectory)
+	if err != nil {
+		return errors.Wrap(err, "error creating missing SSH Keypair")
 	}
 	return nil
 }
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -61,7 +61,7 @@ func newGenerateCmd() *cobra.Command {
 				return errors.Wrap(err, "loading API model in generateCmd")
 			}
 
-			if _, _, err := api.ValidateApiModel(gc.locale, gc.containerService, gc.apiVersion); err != nil {
+			if _, _, err := api.ValidateAPIModel(gc.locale, gc.containerService, gc.apiVersion); err != nil {
 				return errors.Wrap(err, "validating API model after populating values")
 			}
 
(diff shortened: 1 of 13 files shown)


ohthehugemanatee:
Code coverage decreases because I removed a failure condition (empty pubkey) without removing the containing function (`validateLinuxProfile`) . I left the function because it kicks off `validateKeyVaultSecrets`. But I removed `TestProperties_validateKeyVaultSecrets`, because the code it tests is now missing. So there is now a function without a test.

jackfrancis:
lgtm, thanks @ohthehugemanatee!

mboersma:
> So there is now a function without a test.

Makes sense. If we merge this as-is, let's create an issue so we don't forget to restore test coverage if possible.

pkg/api/vlabs/validate.go
```diff
@@ -509,11 +509,6 @@ func (a *Properties) validateZones() error {
 }
 
 func (a *Properties) validateLinuxProfile() error {
-	for _, publicKey := range a.LinuxProfile.SSH.PublicKeys {
```
CecileRobertMichon: my only concern with removing this validation is that we also use these functions for other aks-engine commands such as scale, upgrade, etc. and the public key is still a required value for those.

cmd/generate.go
```diff
@@ -166,6 +167,20 @@ func (gc *generateCmd) loadAPIModel(cmd *cobra.Command, args []string) error {
...
 
+	// If no keypair has been supplied, generate one.
+	if gc.containerService.Properties.LinuxProfile != nil && (gc.containerService.Properties.LinuxProfile.SSH.PublicKeys == nil ||
+		len(gc.containerService.Properties.LinuxProfile.SSH.PublicKeys) == 0 ||
+		gc.containerService.Properties.LinuxProfile.SSH.PublicKeys[0].KeyData == "") {
+		translator := &i18n.Translator{
+			Locale: gc.locale,
+		}
+		var publicKey string
+		_, publicKey, err = helpers.CreateSaveSSH(gc.containerService.Properties.LinuxProfile.AdminUsername, gc.outputDirectory, translator)
```
CecileRobertMichon: what if `gc.containerService.Properties.LinuxProfile` is nil?

CecileRobertMichon: then no validation would fail but no key would be created either which would make the cluster deployment fail

pkg/api/vlabs/validate.go
ohthehugemanatee: right - this is exactly why pubkey generation should happen on Load, and validation should happen afterwards. Then we can validate that there is a pubkey present before anything else tries to use it.

ohthehugemanatee:
Out of time today, so i couldn't run tests/linters locally. I'll let the bots do their job and resolve issues tomorrow.

Changes:

* restore validation of empty key/LinuxProfile
* `deploy` does exactly the delayed validation we want. I moved its (admittedly pretty janky) validation function into `api`
* delayed validation in `generate` the same way it's delayed in `deploy`

Given the similarity between these steps in `deploy` and `generate` there's probably more duplicate code we could refactor out. I tried to keep these changes focused on the issue, though.

CecileRobertMichon:
This is great, thanks so much for taking the time to do this the right way @ohthehugemanatee! The unit tests are failing on a `panic: runtime error: invalid memory address or nil pointer dereference`. 

The VHD pipeline shouldn't need to run for this PR as it doesn't touch any of the relevant files.

pkg/api/apiloader.go
```diff
@@ -474,3 +476,51 @@ func setContainerServiceDefaultsv20170131(c *v20170131.ContainerService) {
...
+	// This isn't terribly elegant, but it's the easiest way to go for now w/o duplicating a bunch of code
+	rawVersionedAPIModel, err := apiloader.SerializeContainerService(containerService, apiVersion)
+	if err != nil {
+		return nil, "", err
+	}
+	return apiloader.DeserializeContainerService(rawVersionedAPIModel, true, false, nil)
+}
+
+// Generates a missing pubkey pair.
+func CheckCreateKeyPair(cs *ContainerService, locale *gotext.Locale, outputDirectory string) (*ContainerService, error) {
```
CecileRobertMichon: no need to return `cs` here since it's a pointer

pkg/api/apiloader.go
```diff
@@ -474,3 +476,51 @@ func setContainerServiceDefaultsv20170131(c *v20170131.ContainerService) {
...
+	// If no ContainerService is supplied, don't bother.
+	if cs == nil {
+		return cs, nil
+	}
+	// If no keypair has been supplied, generate one.
+	if cs.Properties.LinuxProfile != nil && (cs.Properties.LinuxProfile.SSH.PublicKeys == nil ||
```
CecileRobertMichon: what if LinuxProfile.SSH is nil? should we create it for the user?

pkg/api/apiloader.go
ohthehugemanatee: golang noob here... I thought everything was pass-by-copy? Clearly misunderstood what the asterisk was for. 

Fixed anyway. Thanks!

pkg/api/apiloader.go
ohthehugemanatee: it can't be nil, because LinuxProfile includes SSH as a struct in its definition. I can't test if the struct == nil. Am I missing something? In a rough test without specifying SSH in my json, the object still arrives at this code point with the struct in place.

ohthehugemanatee:
todo:
* clean up the generated cert files after tests.

ohthehugemanatee:
It's really hard to be consistent here. Created #1281 to refactor generate and deploy to share more of the same code path... that will make them easier to maintain.

Meanwhile, added deferred RemoveAlls to clean up generated cert files. 

This is ready for review again.

ohthehugemanatee:
I'm not sure how to improve the code coverage of this patch. I just added a test for failure to create the SSH key, to make sure errors bubble up. Beyond that, I'm at a loss...

ohthehugemanatee:
@CecileRobertMichon can you help me out with this one? The e2e test is failing with something that looks totally unrelated. And I'm at a loss for codecov, too.

jackfrancis:
@ohthehugemanatee yeah, just an E2E flake, re-running

pkg/api/vlabs/validate.go
CecileRobertMichon: do we still need to remove this since you moved validation to after auto-fill?

pkg/api/v20170701/validate.go
```diff
@@ -111,9 +111,6 @@ func (a *AgentPoolProfile) Validate(orchestratorType string) error {
 func (l *LinuxProfile) Validate() error {
 	// Don't need to call validate.Struct(l)
 	// It is handled by Properties.Validate()
-	if e := validate.Var(l.SSH.PublicKeys[0].KeyData, "required"); e != nil {
```
CecileRobertMichon: we shouldn't touch this file, this apiversion is for ACS legacy support. This won't affect generate/deploy with vlabs in aks-engine.

pkg/api/agentPoolOnlyApi/vlabs/validate.go
```diff
@@ -29,9 +29,6 @@ func (a *AgentPoolProfile) Validate() error {
 func (l *LinuxProfile) Validate() error {
 	// Don't need to call validate.Struct(l)
 	// It is handled by Properties.Validate()
-	if e := validate.Var(l.SSH.PublicKeys[0].KeyData, "required"); e != nil {
```
CecileRobertMichon: same here

pkg/api/agentPoolOnlyApi/v20180331/validate.go
```diff
@@ -37,9 +37,6 @@ func (a *AgentPoolProfile) Validate() error {
 func (l *LinuxProfile) Validate() error {
 	// Don't need to call validate.Struct(l)
 	// It is handled by Properties.Validate()
-	if e := validate.Var(l.SSH.PublicKeys[0].KeyData, "required"); e != nil {
```
CecileRobertMichon: and here

pkg/api/vlabs/validate.go
ohthehugemanatee: I removed it because it's a condition that can never be fulfilled. If the key value was an empty string, a key was generated. Any failure in generating/saving the key throws a fatal error. So you can't get here with an empty key string.

I can put it back if you like, it's just not testing anything.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
ohthehugemanatee: See previous. It's not possible to get here with an empty string, without a fatal error. So it's not something we need to validate on the regular.

pkg/api/agentPoolOnlyApi/vlabs/validate.go
ohthehugemanatee: see previous

pkg/api/v20170701/validate.go
ohthehugemanatee: Reverted the file.

ohthehugemanatee:
@CecileRobertMichon thanks for the review, let me know if I should put the validation test back.

pkg/api/vlabs/validate.go
CecileRobertMichon: I think we still want it. We also run validate on upgrade/scale and this is also to guard against users making manual alterations to their apimodel before running upgrade/scale (like removing the key).

pkg/api/agentPoolOnlyApi/vlabs/validate.go
CecileRobertMichon: this is for `agentpoolonly` cluster A.K.A AKS so we definitely want to keep that one.

pkg/api/agentPoolOnlyApi/v20180331/validate.go
CecileRobertMichon: same comment, this one is also for AKS, we don't want to make any changes there (they don't use generate/deploy).

ohthehugemanatee:
@CecileRobertMichon OK I reverted the commit that removed validation. Now we have a bunch of tests which are failing because they try to generate with an empty pubkey... I'll go through and fix them.

jackfrancis:
Keep open

PR #1230, opened 2019-05-06
style: remove unneeded 'echo' commands

mboersma:
**Reason for Change**:
Warns about unneeded uses of 'echo'.

See https://github.com/koalaman/shellcheck/wiki/SC2005

**Issue Fixed**:

Refs #953

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_customcloud.sh
```diff
@@ -59,16 +59,16 @@ configureK8sCustomCloud() {
         K8S_CLIENT_CERT_PATH="${KUBERNETES_FILE_DIR}/k8s_auth_certificate.pfx"
         echo $SERVICE_PRINCIPAL_CLIENT_SECRET_CERT | base64 --decode > $K8S_CLIENT_CERT_PATH
         # shellcheck disable=SC2002
-        echo $(cat "${AZURE_JSON_PATH}" | \
+        cat "${AZURE_JSON_PATH}" | \
```
jackfrancis: Just curious how this passes our previously merged "eliminate unnecessary cat" operations PR?

mboersma: Because of the `# shellcheck disable=SC2002`, which makes `shellcheck` give that rule a pass on the subsequent line. 

This was one of two usages (the other is just below in the same file) where we're modifying a file in-place with `jq`, and the other approach of using a temp file seemed uglier.

Issue #1224, opened 2019-05-03
Disable addons not supported on Azure Stack


jadarsie:
**Is this a request for help?**:
No

**Is this an ISSUE or FEATURE REQUEST?**:
Feature

**What version of aks-engine?**:
Latest

**What happened**:
Azure Stack does not support all the addons that AKSe deploys by default.
Some of the optional addons are not supported either or were not yet validated.

**What you expected to happen**:
AKSe to ignore the input json and not deploy those addons

List of addons to block:
- AAD Pod Identity
- ACI Connector
- Cluster Autoscaler
- Blobfuse Flex Volume
- SMB Flex Volume
- KeyVault Flex Volume
- Rescheduler
- NVIDIA Device Plugin
- Container Monitoring

jadarsie:
> **Anything else we need to know:**
> How to document this difference in behavior between Azure / Azure Stack?

@mboersma I am considering creating a new page to explain what addons people should expect to work on Azure Stack. Would you rather add that info to an existing page? which one?

CecileRobertMichon:
@jadarsie I'll let you close this if it's fixed.

PR #1233, opened 2019-05-06
feat: disable unsupported addons on Azure Stack deployments

jadarsie:
**Reason for Change**:
Disable unsupported addons on Azure Stack deployments.

**Issue Fixed**:
Related to issue #1224 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests


jadarsie:
I could not find where/how `kubernetesContainerAddonSettingsInit` is currently unit tested. A pointer will be appreciated.

jadarsie:
I could not find where/how `kubernetesContainerAddonSettingsInit` is currently unit tested. A pointer will be appreciated.

pkg/engine/artifacts.go
```diff
@@ -43,38 +43,38 @@ func kubernetesContainerAddonSettingsInit(profile *api.Properties) map[string]ku
 		DefaultAADPodIdentityAddonName: {
 			"kubernetesmasteraddons-aad-pod-identity-deployment.yaml",
 			"aad-pod-identity-deployment.yaml",
-			profile.OrchestratorProfile.KubernetesConfig.IsAADPodIdentityEnabled(),
+			profile.OrchestratorProfile.KubernetesConfig.IsAADPodIdentityEnabled() && !profile.IsAzureStackCloud(),
```
CecileRobertMichon: @jadarsie you can actually move this one level up to `types.go` inside `IsAADPodIdentityEnabled()` which is unit tested (and add a UT case in types_test.go) if that makes sense.

pkg/engine/artifacts.go
```diff
@@ -43,38 +43,38 @@ func kubernetesContainerAddonSettingsInit(profile *api.Properties) map[string]ku
...
 			profile.OrchestratorProfile.KubernetesConfig.GetAddonScript(DefaultAADPodIdentityAddonName),
 		},
 		DefaultACIConnectorAddonName: {
 			"kubernetesmasteraddons-aci-connector-deployment.yaml",
 			"aci-connector-deployment.yaml",
-			profile.OrchestratorProfile.KubernetesConfig.IsACIConnectorEnabled(),
+			profile.OrchestratorProfile.KubernetesConfig.IsACIConnectorEnabled() && !profile.IsAzureStackCloud(),
```
CecileRobertMichon: same thing for all the other addons

pkg/engine/artifacts.go
jadarsie: Sounds right, I would rather increase coverage.
Is it OK to change the signature of the `IsXXXEnabled` methods? 
It looks like I would need a reference to `Properties` or pass the boolean result of `IsAzureStackCloud`.

CecileRobertMichon: yes that should be fine

CecileRobertMichon:
Does this fix #1224 ?

pkg/api/types.go
```diff
@@ -1510,43 +1510,64 @@ func (k *KubernetesConfig) IsAzureCNIMonitoringEnabled() bool {
...
 func (k *KubernetesConfig) IsTillerEnabled() bool {
 	return k.isAddonEnabled(DefaultTillerAddonName, DefaultTillerAddonEnabled)
 }
 
-// IsAADPodIdentityEnabled checks if the tiller addon is enabled
-func (k *KubernetesConfig) IsAADPodIdentityEnabled() bool {
-	return k.isAddonEnabled(DefaultAADPodIdentityAddonName, DefaultAADPodIdentityAddonEnabled)
+// IsAADPodIdentityEnabled checks if the AAD pod identity addon is enabled
+func (p *Properties) IsAADPodIdentityEnabled() bool {
+	if p.OrchestratorProfile == nil || p.OrchestratorProfile.KubernetesConfig == nil || p.IsAzureStackCloud() {
```
mboersma: We repeat this line several times. It could be a helper function to keep things more DRY:

```go
func (p *Properties) isKubernetesAndNotAzureStack() bool {
    return p.OrchestratorProfile != nil && p.OrchestratorProfile.KubernetesConfig != nil && !p.IsAzureStackCloud()
}

func (p *Properties) IsACIConnectedEnabled() bool {
    return p.isKubernetesAndNotAzureStack() &&
        p.OrchestratorProfile.KubernetesConfig.isAddonEnabled(DefaultACIConnectorAddonName, DefaultACIConnectorAddonEnabled)
}
```

Maybe we could look at that for a future refactoring.

jadarsie:
> 
> 
> Does this fix #1224 ?

It looks like it. Default addons are not deployed.

pkg/api/types.go
jadarsie: @mboersma comment addressed here => #1261

PR #1244, opened 2019-05-07
fix: default kubelet flags windows/linux reconciliation w/ unit tests

jackfrancis:
**Reason for Change**:

There are some open issues relating to kubelet flags defaults, let's get unit test coverage complete so that we can safely address these.

This PR uses empty string values to mark options as "not used", and then before rendering the final kubelet config map, we delete any keys that have `""` values. Unit tests should tell the story for expected default configuration functionality for all kubletConfig instances.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -32,10 +32,10 @@ func (cs *ContainerService) setKubeletConfig() {
 	for key, val := range staticLinuxKubeletConfig {
 		switch key {
 		case "--pod-manifest-path": // Don't add Linux-specific config
-			break
+			staticWindowsKubeletConfig[key] = ""
 		case "--anonymous-auth", "--client-ca-file":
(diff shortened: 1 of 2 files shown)


pkg/api/defaults-kubelet.go
```diff
@@ -161,13 +161,13 @@ func (cs *ContainerService) setKubeletConfig() {
...
 				profile.KubernetesConfig.KubeletConfig[key] = val
 			}
+		} else {
+			for key, val := range staticLinuxKubeletConfig {
+				profile.KubernetesConfig.KubeletConfig[key] = val
+			}
 		}
 
 		setMissingKubeletValues(profile.KubernetesConfig, o.KubernetesConfig.KubeletConfig)
-		// The previous call may have copied values from the orchestartor profile, which are not valid on Windows
```
jackfrancis: We don't need this deletion anymore as the `--pod-manifest-path` key is being set to `""` in the Windows static kublet config set, and those keys are now being deleted in the `removeKubeletFlags` convenience func

PR #1247, opened 2019-05-07
feat: add support for Kubernetes 1.15.0-alpha.3

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#v1150-alpha3

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/k8s_versions.go
```diff
@@ -22,7 +22,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"metrics-server":                   "metrics-server-amd64:v0.2.1",
 		"coredns":                          "coredns:1.3.1",
 		"kube-dns":                         "k8s-dns-kube-dns-amd64:1.15.0",
-		"addon-manager":                    "kube-addon-manager-amd64:v9.0",
+		"addon-manager":                    "kube-addon-manager-amd64:v9.0.1",
```
mboersma: I wasn't sure if this should be added to the VHD yet, so I didn't.

jackfrancis: I think this is the appropriate PR to do so, actually.

PR #1248, opened 2019-05-07
fix: Proper permission on the HOME dir for user on jumpbox VM

### LINKED ISSUE #725: Private cluster jumpbox user home owned by root
** Is this a request for help?: **
<br>

YES
<br>

**Is this an ISSUE or FEATURE REQUEST? (choose one):**

<br>
ISSUE
<br>

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**

<br>
Kubernetes 1.10
<br>

**What happened**:

<br>
When creating a private cluster (Kubernetes), the jumpbox (if enabled) user's home is owned by root and not writable for the user, e.g. `/home/azureuser` is owned by `root` and not by `azureuser`. 
<br>

<br>
I would expect `/home/azureuser` to be owned by `azureuser`
<br>

<br>
Deploy a basic Kubernetes cluster and enable private cluster and jumpbox. 
<br>

<br>
N/A
<br>

juergenpf:
**Reason for Change**:
The cloud init yml for the jumpbox VM in private clusters resulted in root permissions for the default user. Fixed this in a foolproof way to ensure correct permissions.

**Issue Fixed**:
Fixes #725

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
No unit test, as in general testing on the jumpbox in private clusters is currently lacking some testing and this PR is not going to cure that.

### DIFF
--- a/parts/k8s/cloud-init/jumpboxcustomdata.yml
+++ b/parts/k8s/cloud-init/jumpboxcustomdata.yml
@@ -11,7 +11,7 @@ write_files:
 
 - path: "/home/{{WrapAsParameter "jumpboxUsername"}}/.kube/config"
   permissions: "0644"
-  owner: "root"
+  owner: "{{WrapAsParameter "jumpboxUsername"}}"
   content: |
 {{WrapAsVariable "kubeconfig"}}
 
@@ -20,3 +20,7 @@ runcmd:
 - retrycmd_if_failure 10 5 10 curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{.OrchestratorProfile.OrchestratorVersion}}/bin/linux/amd64/kubectl
 - chmod +x ./kubectl
 - sudo mv ./kubectl /usr/local/bin/kubectl
+- chown -R "{{WrapAsParameter "jumpboxUsername"}}" "/home/{{WrapAsParameter "jumpboxUsername"}}"
+- chgrp -R "{{WrapAsParameter "jumpboxUsername"}}" "/home/{{WrapAsParameter "jumpboxUsername"}}"
+- chown -R root "/home/{{WrapAsParameter "jumpboxUsername"}}/.kube"
+- chgrp -R root "/home/{{WrapAsParameter "jumpboxUsername"}}/.kube"


parts/k8s/cloud-init/jumpboxcustomdata.yml
```diff
@@ -11,12 +11,16 @@ write_files:
...
 {{WrapAsVariable "kubeconfig"}}
 
 runcmd:
 - . /opt/azure/containers/provision_source.sh
 - retrycmd_if_failure 10 5 10 curl -LO https://storage.googleapis.com/kubernetes-release/release/v{{.OrchestratorProfile.OrchestratorVersion}}/bin/linux/amd64/kubectl
 - chmod +x ./kubectl
-- sudo mv ./kubectl /usr/local/bin/kubectl
\ No newline at end of file
+- sudo mv ./kubectl /usr/local/bin/kubectl
+- chown -R "{{WrapAsParameter "jumpboxUsername"}}" "/home/{{WrapAsParameter "jumpboxUsername"}}"
```
mboersma: You could save a couple commands by using `chown` with user and group:

```
- chown -R "{{WrapAsParameter "jumpboxUsername"}}":"{{WrapAsParameter "jumpboxUsername"}}" "/home/{{WrapAsParameter "jumpboxUsername"}}"
- chown -R root:root "/home/{{WrapAsParameter "jumpboxUsername"}}/.kube"
```

juergenpf: Except the fact, that not all implementations of chown support that, this depends on the tool chain. So my way is really portable.

juergenpf: @mboersma  If I've not twisted my brain, the first two lines set the default of the home subtree owned by the user, and the last two set the ~/.kube subtree owned by root. That was the minimum change to satisfy the request, I'm not sore whether or not the user must be able to modify the config manually, he can nevertheless read it. This is a defensive approach, and it appears to work properly.

parts/k8s/cloud-init/jumpboxcustomdata.yml
```diff
@@ -11,12 +11,16 @@ write_files:
...
+- chgrp -R "{{WrapAsParameter "jumpboxUsername"}}" "/home/{{WrapAsParameter "jumpboxUsername"}}"
+- chown -R root "/home/{{WrapAsParameter "jumpboxUsername"}}/.kube"
```
CecileRobertMichon: why do we want `~/.kube` to be owned by root?

juergenpf: @CecileRobertMichon see my reply to @mboersma review. It's a defensive minimal change and actually doesn't impose any restriction as the config can be read. If the user want's to change he can, because the default user usually has sudo rights.

CecileRobertMichon:
@juergenpf did you already test deploying a private cluster with a jumpbox from this branch?

juergenpf:
@CecileRobertMichon Of course I did several deployments.
See the output from one of the ssh sessions on a jumpbox:

$ ls -alR
.:
total 20
drwxr-xr-x 5 azureuser azureuser 4096 May  7 23:14 .
drwxr-xr-x 3 root      root      4096 May  7 23:07 ..
drwx------ 2 azureuser azureuser 4096 May  7 23:14 .cache
-rw-rw-r-- 1 azureuser azureuser    0 May  7 23:14 .cloud-locale-test.skip
drwxr-xr-x 2 root      root      4096 May  7 23:07 .kube
drwx------ 2 azureuser azureuser 4096 May  7 23:07 .ssh

./.cache:
total 8
drwx------ 2 azureuser azureuser 4096 May  7 23:14 .
drwxr-xr-x 5 azureuser azureuser 4096 May  7 23:14 ..
-rw-r--r-- 1 azureuser azureuser    0 May  7 23:14 motd.legal-displayed

./.kube:
total 20
drwxr-xr-x 2 root      root      4096 May  7 23:07 .
drwxr-xr-x 5 azureuser azureuser 4096 May  7 23:14 ..
-rw-r--r-- 1 root      root      9900 May  7 23:07 config

./.ssh:
total 12
drwx------ 2 azureuser azureuser 4096 May  7 23:07 .
drwxr-xr-x 5 azureuser azureuser 4096 May  7 23:14 ..
-rw------- 1 azureuser azureuser  725 May  7 23:07 authorized_keys

juergenpf:
@CecileRobertMichon  Please note, that before the fix for example .cache couldn't be created because the whole home subtree was owned by root.

CecileRobertMichon:
Am I missing something or does root still own `~/.kube/config`? 

from what you pasted:
```
-rw-r--r-- 1 root root 9900 May 7 23:07 config
```

juergenpf:
@CecileRobertMichon sure, as already explained: defensive change and no restriction because it's readable and via sudo changable. I was assuming that there was a reason in the original source to have ~/.kube owned by root. The problem was, that this directory was created in the home subtree before home existed, and thus the home got created with ~/.kube and inherited root permissions.
So the design decision for the fix was to leave ~/.kube with root permission (and readable by others), but fix the issue that root permission on home is simply wrong.

CecileRobertMichon:
Sounds good, thank you for explaining. 

/lgtm

PR #1253, opened 2019-05-08
style: Update syntax for var in Makefile

aramase:
**Reason for Change**:

Replaces `$(GOPATH)` with `${GOPATH}` in the makefile as GOPATH is a var and not a command.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -166,7 +166,7 @@ ifndef HAS_GIT
 	$(error You must install Git)
 endif
 ifndef HAS_GOLANGCI
-	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b $(GOPATH)/bin
+	curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh -s -- -b ${GOPATH}/bin
 endif
 ifndef HAS_GINKGO
 	go get -u github.com/onsi/ginkgo/ginkgo


mboersma:
Does this change anything? I thought `make` didn't care about the distinction, although shell does.

Also I see several similar examples elsewhere in the Makefile.

aramase:
@mboersma I was trying to build the binary on my mac and it kept failing for me until I made this change. (It still fails for other reasons now and @CecileRobertMichon told me it's a known issue. So building it in the docker container instead). `make` shouldn't care but I was surprised for some reason it just kept complaining for me. Also, since $() and ${} don't  make a difference is there a reason we have a mix of both in the makefile instead of just 1 standard one? Just curious.

mboersma:
> is there a reason we have a mix of both in the makefile?

I think it's just sloppiness. I even see one place where we use both styles in a single command, which is supposed to be problematic.

aramase:
> I think it's just sloppiness. I even see one place where we use both styles in a single command, which is supposed to be problematic.

Ah ok! I think we don't really need this PR then. Might just be a one of case. But if you think we should standardize the use of only 1 syntax in the makefile I can get a PR up anytime.

CecileRobertMichon:
@aramase I agree standardizing to one makes sense if you'd like to re-open a PR including the other usages!

aramase:
@CecileRobertMichon cool! I'll get a PR up soon.

PR #1256, opened 2019-05-08
style: check exit code directly, not indirectly with $?

mboersma:
**Reason for Change**:
Enforces checking shell command exit codes directly, which is safer and usually simpler.

See https://github.com/koalaman/shellcheck/wiki/SC2181

**Issue Fixed**:
Refs #953 

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/kubernetes/scripts/host-os-dns-validate.sh
```diff
@@ -13,11 +13,11 @@ HOSTS="${NODE_HOSTNAMES} www.bing.com google.com"
 for i in $(seq 1 $retries); do
   for host in $HOSTS; do
     set -x
-    [ -s $(dig +short +search +answer ${host}) ]
-      if [ $? -eq 0  ]; then
-        success="no"
-        break
-      fi
+    addrs=$(dig +short +search +answer ${host})
```
mboersma: This one was confusing: `dig` doesn't return 0 on a failure to find records, so I assume `-s` was attempting to see if the command returned a non-empty file...but that syntax didn't work for me in macOS `bash` or on Ubuntu.

I think assigning to `addrs` and testing with `-z` for zero-length is more readable and does the job, but maybe I misunderstood the original command.

PR #1261, opened 2019-05-08
style: DRY how addons are disabled on Azure Stack

jadarsie:
**Reason for Change**:
Addressing feedback from a previous PR regarding code organization.

**Issue Fixed**:
Related to issue #1224

**Requirements**:

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1511,7 +1511,7 @@ func (k *KubernetesConfig) IsAzureCNIMonitoringEnabled() bool {
 
 // IsContainerMonitoringEnabled checks if the container monitoring addon is enabled
 func (p *Properties) IsContainerMonitoringEnabled() bool {
-	return p.isKubernetesAndNotAzureStack() &&
+	return !p.isAzureStack() &&
 		p.OrchestratorProfile.KubernetesConfig.isAddonEnabled(ContainerMonitoringAddonName, DefaultContainerMonitoringAddonEnabled)
 }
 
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1829,3 +1811,7 @@ func FormatProdFQDNByLocation(fqdnPrefix string, location string, cloudName stri
 	FQDNFormat := AzureCloudSpecEnvMap[targetEnv].EndpointConfig.ResourceManagerVMDNSSuffix
 	return fmt.Sprintf("%s.%s."+FQDNFormat, fqdnPrefix, location)
 }
+
+func (p *Properties) isKubernetesAndNotAzureStack() bool {
+	return p.OrchestratorProfile != nil && p.OrchestratorProfile.KubernetesConfig != nil && !p.IsAzureStackCloud()
```
CecileRobertMichon: I assume this is already unit tested through the other functions' unit tests?

jadarsie: We should see failed tests if this function is not doing what is supposed to do

jackfrancis: Is there a reason not to just make a method with a simpler name like "`isAzureStack()`" ? Then we can employ `!isAzureStack()` if we want to validate the NOT condition.

(Also as AKS-Engine + Azure Stack is very unlikely to ever be anything other than Kubernetes, I think we can hide the "is kubernetes" logic behind the method, but we don't need to advertise that in the method name)

mboersma: I had suggested this explicit function name, but `isAzureStack` probably covers it.

CecileRobertMichon: We want to check for non-nil orcherstrator profile values regardless of Azure Stack so I don't think we should embed that in the IsAzureStack func. If later we want to re-enable some or all of these addons for Azure Stack, we will still want to keep the `isKubernetes` validation.

jadarsie: Renamed function and kept nil checks.

pkg/api/types.go
```diff
@@ -1829,3 +1811,7 @@ func FormatProdFQDNByLocation(fqdnPrefix string, location string, cloudName stri
 	FQDNFormat := AzureCloudSpecEnvMap[targetEnv].EndpointConfig.ResourceManagerVMDNSSuffix
 	return fmt.Sprintf("%s.%s."+FQDNFormat, fqdnPrefix, location)
 }
+
+func (p *Properties) isAzureStack() bool {
+	return p.OrchestratorProfile != nil && p.OrchestratorProfile.KubernetesConfig != nil && p.IsAzureStackCloud()
```
jackfrancis: I'm actually a little confused by this method now. What is the purpose of checking:

- non-nil OrchestratorProfile
- non-nil KubernetesConfig
- Azure Stack cloud definition

all together?

Why not just invoke the existing `p.IsAzureStackCloud()` method in the addons enabled methods?

jackfrancis: I see what's going on. Sigh. These "is addons enabled" methods should not be a method of the `Properties` object. Stand by.

CecileRobertMichon: they weren't. That was to accommodate using the `isAzureStackCloud` func in them.

jackfrancis: I'll make these changes in this PR: https://github.com/Azure/aks-engine/pull/1265

CecileRobertMichon: this PR is follow up from #1233

jackfrancis: @jadarsie see this commit:

https://github.com/Azure/aks-engine/pull/1265/commits/76411647e150d09487c412afde88a784846d454f

Basically, we should be gating addons in Azure Stack scenarios during the customData (cloud-init) string concatenation process, not in the "is this addon enabled?" method itself. That latter method we want to keep nice and lean, basically: "do I have this addon configured and enabled in my api model?".

jackfrancis: Also, as @CecileRobertMichon pointed out, I should have been more opinionated in the prior PR.

jadarsie: Sounds good @jackfrancis, just let me know your preferences.

The changes look OK, it looks similar to my first commit: https://github.com/Azure/aks-engine/pull/1233/commits/b8e70377ed7ebe403bd01ab82cab2da0b5691994

CecileRobertMichon: I think the reason we moved the `!isAzureStack()` condition one level up was to be able to unit test it.

jadarsie:
Fixed by #1265

PR #1268, opened 2019-05-08
chore: Allow pod to wireserver communication

tamilmani1989:
Snat wireserver traffic from pod

**Reason for Change**:

This PR allows pod to communicate with wireserver running in azure host.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -646,6 +646,14 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(successes).To(Equal(cfg.StabilityIterations))
 		})
 
+		It("should have stable pod to wireserver connectivity", func() {
+			name := fmt.Sprintf("alpine-%s", cfg.Name)
+			command := fmt.Sprintf("nc -vz 168.63.129.16 80")
+			successes, err := pod.RunCommandMultipleTimes(pod.RunLinuxPod, "alpine", name, command, cfg.StabilityIterations, 1*time.Second, retryCommandsTimeout, stabilityCommandTimeout)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(successes).To(Equal(cfg.StabilityIterations))
+		})
+
 		It("should have stable pod-to-pod networking", func() {
 			if eng.HasLinuxAgents() {
 				By("Creating a test php-apache deployment")
--- a/.pipelines/e2e-job-template.yaml
+++ b/.pipelines/e2e-job-template.yaml
@@ -13,9 +13,9 @@ jobs:
     maxParallel: 0
   pool:
     vmImage: ubuntu-16.04
-  
+
   container: dev1
-  
+
   variables:
     GOBIN:  '$(GOPATH)/bin' # Go binaries path
     GOROOT: '/usr/local/go' # Go installation path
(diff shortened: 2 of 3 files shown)


tamilmani1989:
@CecileRobertMichon lets hold this PR till @JunSun17 replies. It has to be changed in one other place ( for ipmasqagent enabled case). I'm not sure if the code is in aks-engine or somewhere else.

JunSun17:
> @CecileRobertMichon lets hold this PR till @JunSun17 replies. It has to be changed in one other place ( for ipmasqagent enabled case). I'm not sure if the code is in aks-engine or somewhere else.

Looks good to me.

jackfrancis:
@CecileRobertMichon FYI, I enabled 3 stability iterations in PR E2E so that @tamilmani1989's new test will actually run.

It is probably a good idea to keep this change to engage those tests generally at a minimal level (3 success attempts should not demonstrably increase flakes, and will tell us quickly when those tests become unstable generally).

Thoughts?

jackfrancis:
@tamilmani1989 does it make sense to you why only Ubuntu 18.04-LTS nodes work w/ these changes?

tamilmani1989:
I see the new test case I wrote was getting failed:

> nc: 168.63.129.16 (168.63.129.16:80): Operation timed out

Not sure why its getting failed. Can I get acces to VM debug?

tamilmani1989:
+@sharmasushant

tamilmani1989:
@jackfrancis Do you know if ipmasqagent pods running on the agent nodes? If so, they will rewrite the iptables rules.

jackfrancis:
Indeed, ip-masq-agent is enabled, though I believe that should be the case for 18.04 as well

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -646,6 +646,14 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(successes).To(Equal(cfg.StabilityIterations))
 		})
 
+		It("should have stable pod to wireserver connectivity", func() {
+			name := fmt.Sprintf("alpine-%s", cfg.Name)
```
jackfrancis: @tamilmani1989 you can wrap this in an if condition if this will only work for certain cluster configurations

e.g., if ip-masq-agent prevents container access to 168.63.129.16 on TCP 80:

`if !eng.ExpandedDefinition.Properties.IsIPMasqAgentEnabled()`

tamilmani1989: Once @JunSun17 PR is merged, it should work for all cases. For now, we can have this condition but it should would work in all cases

tamilmani1989:
> Indeed, ip-masq-agent is enabled, though I believe that should be the case for 18.04 as well

Not sure, how it worked there. Let me test it in ubuntu18 and update you.

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Which PR?

jackfrancis: Is there an AKS Engine change that this depends upon?

tamilmani1989: the one he shared in email. Its vsts link

tamilmani1989: ipmasqagent code is in AKS. we have dependency on that. Both PR has to be in to allow connection from pod to wireserver. otherwise ipmasqagent will rewrite those ipable rules if its enabled.

PR #1271, opened 2019-05-09
build: validate deps before 'make build', and some cleanup

mboersma:
**Reason for Change**:
Cleans up the `Makefile` in some minor ways:
- makes `validate-dependencies` a prerequisite target for `make build`
- adds separate `validate-go` and `validate-shell` targets for convenience and consistency
- add `validate-copyright-headers` to `make test-style`
- removes the `validate-commit-msg.sh` script that we punted on

The first bullet point in particular I think is important. I was bitten by this recently: my branch passed `make clean build test-style test` locally but failed in CI because `dep check` hadn't been run.

It's also worth noting that the `validate-dependencies` and `validate-copyright-headers` targets that this adds to the common developer workflow take negligible time.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Apologies for mixing concerns in one PR. I didn't think these changes were particularly controversial but I'm happy to separate them if that's preferable.


Makefile
```diff
@@ -134,9 +138,7 @@ test: generate
 	ginkgo -skipPackage test/e2e/dcos,test/e2e/kubernetes -failFast -r .
 
 .PHONY: test-style
-test-style:
-	@scripts/validate-shell.sh
-	@scripts/validate-go.sh
+test-style: validate-go validate-shell validate-copyright-headers
```
CecileRobertMichon: the pipeline (`pr-e2e.yaml`) calls these separately so we should update it as well

CecileRobertMichon: https://github.com/Azure/aks-engine/blob/master/.pipelines/pr-e2e.yaml#L55

mboersma: I thought about doing that, but `make` will actually sort out the dependent targets so that `make validate-copyright-headers` isn't called twice. But you're right that it's cleaner, so I'll change it.

PR #1272, opened 2019-05-09
docs: correct "--auth-method" arguments in help

mboersma:
**Reason for Change**:
Just a couple of typo fixes.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/scale.md
+++ b/docs/topics/scale.md
@@ -39,5 +39,5 @@ This command will re-use the `apimodel.json` file inside the output directory as
 |--node-pool|depends|Required if there is more than one node pool. Which node pool should be scaled.|
 |--new-node-count|yes|Desired number of nodes in the node pool.|
 |--master-FQDN|depends|When scaling down a kubernetes cluster this is required. The master FDQN so that the nodes can be cordoned and drained before removal. This should be output as part of the create template or it can be found by looking at the public ip addresses in the resource group.|
-|--auth-method|no|The authentication method used. Default value is 'client_secret'. Other supported values are: 'device' and 'client_certificate'.|
+|--auth-method|no|The authentication method used. Default value is `client_secret`. Other supported values are: `cli`, `client_certificate`, and `device`.|
 |--language|no|Language to return error message in. Default value is "en-us").|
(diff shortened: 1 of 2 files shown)


mboersma:
I'm actually noticing several other discrepancies in the docs around this. I'll try to clean them up in one PR.

/hold

mboersma:
Ok, there was just one other related docs fix that I found.

/hold cancel

docs/topics/scale.md
```diff
@@ -39,5 +39,5 @@ This command will re-use the `apimodel.json` file inside the output directory as
 |--node-pool|depends|Required if there is more than one node pool. Which node pool should be scaled.|
 |--new-node-count|yes|Desired number of nodes in the node pool.|
 |--master-FQDN|depends|When scaling down a kubernetes cluster this is required. The master FDQN so that the nodes can be cordoned and drained before removal. This should be output as part of the create template or it can be found by looking at the public ip addresses in the resource group.|
-|--auth-method|no|The authentication method used. Default value is 'client_secret'. Other supported values are: 'device' and 'client_certificate'.|
+|--auth-method|no|The authentication method used. Default value is `client_secret`. Other supported values are: `cli`, `client_certificate`, and `device`.|
```
CecileRobertMichon: is auth-method cli not supported?

CecileRobertMichon: I think it is https://github.com/Azure/aks-engine/blob/master/cmd/root.go#L222

mboersma: It is supported, but we didn't have it listed here.

CecileRobertMichon: I read the diff wrong... I'll see my self out ☕️

CecileRobertMichon: although thinking about it, I think it might be intentional that it's missing because it doesn't work for new users

CecileRobertMichon: https://github.com/Azure/aks-engine/pull/231

CecileRobertMichon: or is that device?

CecileRobertMichon: nevermind, it _is_ supported

PR #1273, opened 2019-05-09
chore: update client-go and k8s.io vendored code

mboersma:
**Reason for Change**:
Updates `k8s.io` vendored imports to more current versions. 

**Issue Fixed**:
Should fix an error upgrading or scaling an AAD-enabled cluster:
```
FATA[0377] Error upgrading cluster: No Auth Provider found for name "azure"
```
and unblock the OpenShift team trying to vendor in a RedHat plugin.

Closes #506 

See also Azure/acs-engine#4296 and #85.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
@skinny I tried to put a commit on top of #506 but ran into "filename too long errors" just trying to check out the branch and wasn't able to work around it (I think some Unix symlinks got checked in?) So I started fresh here, but if you'd rather rehab #506 let me know and we can work with that.


skinny:
As long as the lib is updated I'm fine :) 👍

PR #1276, opened 2019-05-10
feat: Make cordon drain timeout configurable with --upgrade

### LINKED ISSUE #950: Make cordon/drain timeout during upgrade configurable
We recently shipped a change to increase the cordon/drain timeout to 20 mins:

https://github.com/Azure/aks-engine/pull/938

Let's make this configurable, say:

`aks-engine upgrade --cordon-drain-timeout=60`

aramase:
**Reason for Change**:

Make the cordon drain timeout value configurable in `--upgrade` command.

**Issue Fixed**:

Fixes #950

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -72,7 +72,7 @@ func newUpgradeCmd() *cobra.Command {
 	f.StringVar(&uc.deploymentDirectory, "deployment-dir", "", "the location of the output from `generate`")
 	f.StringVarP(&uc.upgradeVersion, "upgrade-version", "k", "", "desired kubernetes version (required)")
 	f.IntVar(&uc.timeoutInMinutes, "vm-timeout", -1, "how long to wait for each vm to be upgraded in minutes")
-	f.IntVar(&uc.cordonDrainTimeoutInMinutes, "cordon-drain-timeout", 20, "how long to wait for each vm to be cordoned in minutes")
+	f.IntVar(&uc.cordonDrainTimeoutInMinutes, "cordon-drain-timeout", -1, "how long to wait for each vm to be cordoned in minutes")
 	f.BoolVarP(&uc.force, "force", "f", false, "force upgrading the cluster to desired version. Allows same version upgrades and downgrades.")
 	addAuthFlags(uc.getAuthArgs(), f)
 
@@ -105,8 +105,10 @@ func (uc *upgradeCmd) validate(cmd *cobra.Command) error {
 		uc.timeout = &timeout
 	}
 
-	cordonDrainTimeout := time.Duration(uc.cordonDrainTimeoutInMinutes) * time.Minute
-	uc.cordonDrainTimeout = &cordonDrainTimeout
+	if uc.cordonDrainTimeoutInMinutes != -1 {
+		cordonDrainTimeout := time.Duration(uc.cordonDrainTimeoutInMinutes) * time.Minute
+		uc.cordonDrainTimeout = &cordonDrainTimeout
+	}
 
 	if uc.upgradeVersion == "" {
 		cmd.Usage()
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -33,13 +33,14 @@ type upgradeCmd struct {
 	authProvider
 
 	// user input
-	resourceGroupName   string
-	apiModelPath        string
-	deploymentDirectory string
-	upgradeVersion      string
-	location            string
-	timeoutInMinutes    int
-	force               bool
+	resourceGroupName           string
+	apiModelPath                string
+	deploymentDirectory         string
+	upgradeVersion              string
+	location                    string
+	timeoutInMinutes            int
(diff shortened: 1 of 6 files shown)


cmd/upgrade.go
```diff
@@ -102,6 +105,9 @@ func (uc *upgradeCmd) validate(cmd *cobra.Command) error {
 		uc.timeout = &timeout
 	}
 
+	cordonDrainTimeout := time.Duration(uc.cordonDrainTimeoutInMinutes) * time.Minute
+	uc.cordonDrainTimeout = &cordonDrainTimeout
```
aramase: If both vm-timeout and cordon-drain-timeout are provided, I think we should validate to make sure vm-timeout >= cordon-drain-timeout. Otherwise I think it doesn't make sense. Want to know you'll thoughts before I add that check?

CecileRobertMichon: makes sense to me

aramase:
@CecileRobertMichon Had to fix unit tests. But looks good now. Let me know what you think.

cmd/upgrade.go
```diff
@@ -70,6 +72,7 @@ func newUpgradeCmd() *cobra.Command {
 	f.StringVar(&uc.deploymentDirectory, "deployment-dir", "", "the location of the output from `generate`")
 	f.StringVarP(&uc.upgradeVersion, "upgrade-version", "k", "", "desired kubernetes version (required)")
 	f.IntVar(&uc.timeoutInMinutes, "vm-timeout", -1, "how long to wait for each vm to be upgraded in minutes")
+	f.IntVar(&uc.cordonDrainTimeoutInMinutes, "cordon-drain-timeout", 20, "how long to wait for each vm to be cordoned in minutes")
```
CecileRobertMichon: let's move the default logic to the upgrader code (similar to how timeoutInMinutes is doing it) because AKS also calls upgrader so we want to make sure that the timeout isn't nil for them.

In other words, let's set the arg default to -1 and only set `uc.cordonDrainTimeout` if the timeout `!= -1`. Then in the upgrader, do something like https://github.com/Azure/aks-engine/blob/master/pkg/operations/kubernetesupgrade/upgrader.go#L261 and https://github.com/Azure/aks-engine/blob/master/pkg/operations/kubernetesupgrade/upgrader.go#L116.

We could arguably leave the `20` default here as well instead of setting it to `-1` but having a default timeout const in one place will make it easier to change the default in the future.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -263,6 +265,7 @@ func (ku *Upgrader) upgradeAgentPools(ctx context.Context) error {
 		} else {
 			upgradeAgentNode.timeout = *ku.stepTimeout
 		}
+		upgradeAgentNode.cordonDrainTimeout = *ku.cordonDrainTimeout
```
CecileRobertMichon: this is the place I'm suggesting we move the default, same as what stepTimeout is doing just above

cmd/upgrade.go
aramase: Yeah, I was wondering why the default was being set there instead of here for step vm upgrade timeout. Now I know 😊 I’ll make the change.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -498,7 +501,7 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
 			ku.logger.Infof("Successfully set capacity for VMSS %s", vmssToUpgrade.Name)
 
 			// Before we can delete the node we should safely and responsibly drain it
-			client, err := ku.getKubernetesClient(cordonDrainTimeout)
```
CecileRobertMichon: this used to be a const, we should be able to remove that const now that it's unused (or just turn it into the default)

aramase: Yeah, I've removed the const.

cmd/upgrade.go
aramase: I've made the changes to set the default value in upgrader.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -497,6 +505,11 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
 
 			ku.logger.Infof("Successfully set capacity for VMSS %s", vmssToUpgrade.Name)
 
+			cordonDrainTimeout := defaultCordonDrainTimeout
```
jackfrancis: Let's settle on a common "use default value if not provided" pattern. Lines 269-273 are doing the same thing, it seems, but in a sort of inverse way.

aramase: Done

jackfrancis:
lgtm pending upgrade validation

thanks @aramase!

PR #1277, opened 2019-05-10
feat: custom tags on VMs and scale sets

### LINKED ISSUE #1204: Add custom tags to the cluster vm's
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): FEATURE REQUEST

---

aks-engine 0.35.0
---

**Kubernetes version**:
1.13.5

**What happened**:
No option to add custom tags to azure VM's while creating the cluster

Add custom VM tags during the cluster installation, in the cluster definition file
 
customTags:
  project: ops
  team:    devops

Guilhermeslucas:
**Reason for Change**:
My PR fixes #1204 , making it possible to add custom tags for the VMs deployed at a cluster.

**Issue Fixed**:

Fixes #1204

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
Since the tags are already inside the structure ContainerService and we are not attributing them, I'll I did was a function that gets the new tags and add them as part of the tags on ARM Templates.

Please let me know if you have any suggestions, I'll be happy to fix them!
Thanks a lot!


jackfrancis:
I wonder what the `ContainerService` root property `Tags` is currently being used for. Is it meant to represent a general dictionary of tags that should be on all vms? I wonder if we should make this an agentpool-specific thing.

@mboersma @CecileRobertMichon are either of you familiar with this?

Thanks @Guilhermeslucas!

CecileRobertMichon:
+1 on creating a new vmTags property under agentpool profile (and potentially master profile as well?). I think there is a common use case for these tags to be per-pool, not per-cluster.

Also not sure what the existing `Tags` field under `ContainerService` represents but I don't think aks-engine uses it... Might be used by ACS/AKS though.

Guilhermeslucas:
So, as far as I saw on the repo, the tags are ment to declare orchestrator version, aksEngineVersion, creationSource and resourceNameSufix as ```virtualmachines.go``` file shows. 

Thanks for the suggestion everyone. So, to understand better the request, the desired state of this PR would be describing the cluster as something like this:

```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes"
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "guilherme-contoso-2",
      "vmSize": "Standard_D2_v2"
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 2,
        "vmSize": "Standard_D2_v2"
        "tags" : {
             "myAwesomeKey":"myAwesomeValue"
         }
      }
    ],
    "linuxProfile": {
      "adminUsername": "azureuser",
       "tags" : {
             "myAwesomeOtherKey":"myAwesomeOtherValue"
         }
      "ssh": {
        "publicKeys": [
          {
            "keyData": "s"
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "",
      "secret": ""
    }
  }
}
```

Or the way I implemented is closer to what we want? Thanks a lot! :)

CecileRobertMichon:
I don't think the `ContainerService.Tags` field is used to define the existing VM tags, the `Tags` field in https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachines.go#L50 is part of `compute.VirtualMachine`...  it's a bit confusing because they are named the same but they are different.

Regarding your proposed implementation above, we don't need a tags field in `linuxProfile`, only in `agentPoolProfile` and `masterProfile`. Also, I would prefer we make the naming a bit more explicit to avoid confusion with the existing `Tags` field in the `ContainerService` struct. How about `customVMTags`? This also encompasses the fact that there are also static VMs tags we define (orchestratorVersion, etc.). Thoughts @mboersma @jackfrancis ?

so it would look something like: 
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes"
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "guilherme-contoso-2",
      "vmSize": "Standard_D2_v2",
       "tags" : {
             "myAwesomeMasterKey":"myAwesomeMasterValue"
         }
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "tags" : {
             "myAwesomeKey1":"myAwesomeValue1"
         }
      }
      {
        "name": "agentpool2",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "tags" : {
             "myAwesomeKey2":"myAwesomeValue2"
         }
      }
    ],
    "linuxProfile": {
      "adminUsername": "azureuser",
      "ssh": {
        "publicKeys": [
          {
            "keyData": ""
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "",
      "secret": ""
    }
  }
}
```

mboersma:
Isn't ContainerService.Tags a way to actually add tags to a (deprecated) ACS resource itself? If not, I have no idea.

@Guilhermeslucas could you `git commit --amend` the commit message so it's something like:

`feat: custom tags on VMs and scale sets`

Otherwise this change won't show up in our CHANGELOG when we do a release. Thanks!

Guilhermeslucas:
Just did it @jackfrancis ! Thanks for the suggestions. So, I'll work on putting the tags as @CecileRobertMichon . Is that ok? Thanks a lot again :)

Guilhermeslucas:
Hello everyone, I have just added the feature of adding custom VM tags as @CecileRobertMichon suggested. Now, you have the option of adding tags to master and agent pools like this:
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes"
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "guilherme-contoso-2",
      "vmSize": "Standard_D2_v2",
       "customVMTags" : {
             "myAwesomeMasterKey":"myAwesomeMasterValue"
         }
    },
    "agentPoolProfiles": [
      {
        "name": "agentpool1",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "customVMTags" : {
             "myAwesomeKey1":"myAwesomeValue1"
         }
      }
      {
        "name": "agentpool2",
        "count": 2,
        "vmSize": "Standard_D2_v2",
        "customVMTags" : {
             "myAwesomeKey2":"myAwesomeValue2"
         }
      }
    ],
    "linuxProfile": {
      "adminUsername": "azureuser",
      "ssh": {
        "publicKeys": [
          {
            "keyData": ""
          }
        ]
      }
    },
    "servicePrincipalProfile": {
      "clientId": "",
      "secret": ""
    }
  }
}
```

Let me know what you think and what can I do to do better!
Thanks!

pkg/engine/virtualmachines.go
```diff
@@ -563,3 +565,9 @@ func getVaultSecretGroup(linuxProfile *api.LinuxProfile) []compute.VaultSecretGr
 	}
 	return vaultSecretGroups
 }
+
+func addCustomTagsToVM(cs *api.ContainerService, vm *compute.VirtualMachine) {
+	for key, value := range cs.Properties.MasterProfile.CustomVMTags {
```
CecileRobertMichon: where does the equivalent happen for agentpool VMs?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -718,3 +720,9 @@ func getVMSSDataDisks(profile *api.AgentPoolProfile) *[]compute.VirtualMachineSc
 	}
 	return &dataDisks
 }
+
+func addCustomTagsToVMScaleSets(cs *api.ContainerService, vm *compute.VirtualMachineScaleSet) {
+	for key, value := range cs.Properties.MasterProfile.CustomVMTags {
+		vm.Tags[key] = &value
```
CecileRobertMichon: can we add a check to only add this tag if it doesn't already exist? I think we want to avoid allowing users to override the  static tag such as `aksEngineVersion`

CecileRobertMichon:
```
if v, ok := vm.Tags[key]; !ok  {
    vm.Tags[key] = &value
}
```

CecileRobertMichon: same comment for addCustomTagsToVM

pkg/engine/virtualmachinescalesets.go
```diff
@@ -718,3 +720,9 @@ func getVMSSDataDisks(profile *api.AgentPoolProfile) *[]compute.VirtualMachineSc
 	}
 	return &dataDisks
 }
+
+func addCustomTagsToVMScaleSets(cs *api.ContainerService, vm *compute.VirtualMachineScaleSet) {
```
Guilhermeslucas: It is happening here, @CecileRobertMichon :)

pkg/engine/virtualmachines.go
```diff
@@ -563,3 +565,9 @@ func getVaultSecretGroup(linuxProfile *api.LinuxProfile) []compute.VaultSecretGr
 	}
 	return vaultSecretGroups
 }
+
+func addCustomTagsToVM(cs *api.ContainerService, vm *compute.VirtualMachine) {
```
CecileRobertMichon: let's unit test this (and the VMSS equivalent)

pkg/engine/virtualmachines.go
CecileRobertMichon: here you could pass in the tags instead of passing in the full container service

`func addCustomTagsToVM(tags map[string]string, vm *compute.VirtualMachine) {`

that way it's reusable for agents

pkg/engine/virtualmachinescalesets.go
CecileRobertMichon: this is still for MasterProfile

CecileRobertMichon: we're missing the agentPoolProfile tags

CecileRobertMichon: https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachines.go#L361 for VMAS

CecileRobertMichon: https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachinescalesets.go#L379 for VMSS

pkg/engine/virtualmachines.go
Guilhermeslucas: I'll create the tests :)

pkg/engine/virtualmachinescalesets.go
Guilhermeslucas: Sure! Makes total sense not allowing someone to override a tag that already exists! Thanks for the comments!

Guilhermeslucas: Done!

pkg/engine/virtualmachines.go
Guilhermeslucas: Sure, I'll fix that!

Guilhermeslucas: Done!

pkg/engine/virtualmachinescalesets.go
Guilhermeslucas: Oh, I see! Sorry about that! I'll fix that!

Guilhermeslucas: Also done!

pkg/engine/virtualmachines.go
```diff
@@ -361,6 +363,8 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 		Tags: tags,
 	}
 
+	addCustomTagsToVM(cs.Properties.MasterProfile.CustomVMTags, &virtualMachine)
```
CecileRobertMichon: this should be `addCustomTagsToVM(profile.CustomVMTags, &virtualMachine)`

pkg/engine/virtualmachinescalesets.go
```diff
@@ -375,6 +377,8 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		Tags: tags,
 	}
 
+	addCustomTagsToVMScaleSets(cs.Properties.MasterProfile.CustomVMTags, &virtualMachineScaleSet)
```
CecileRobertMichon: same here `addCustomTagsToVM(profile.CustomVMTags, &virtualMachine)`

pkg/engine/virtualmachines.go
Guilhermeslucas: Sure! I didn't noticed this was already been instantiated! Thanks for the suggestion!

Guilhermeslucas: Hey @CecileRobertMichon , can you explain why I have to use the profile here, instead of the cs? I'm just curious cause I'm getting the right arm templates when I use the cs parameter, but not getting the tags when I change it to the profile. Thanks a lot!

CecileRobertMichon: This specific function is to add tags to the agent VMs, so you want the tags in the agent pool profile part of the apimodel. `cs.Properties.MasterProfile` is the master profile of the cluster, so that should only be used for master VMs. In this context, `profile` is the agent pool profile that we're looping on so that's what we want.

If you're not getting the tags while testing maybe that's because you didn't add tags to the agent pool profiles or maybe you're checking the wrong VM?

pkg/engine/virtualmachinescalesets.go
Guilhermeslucas: Yeah, makes total sense! I got this in the wrong way and I just fixed this, lets see how the unit tests goes.

pkg/engine/virtualmachines.go
Guilhermeslucas: I was not converting to vlabs correctly. Thanks a lots for the comments

pkg/api/converterfromapi.go
```diff
@@ -998,6 +998,7 @@ func convertAgentPoolProfileToVLabs(api *AgentPoolProfile, p *vlabs.AgentPoolPro
 	p.Name = api.Name
 	p.Count = api.Count
 	p.VMSize = api.VMSize
+	p.CustomVMTags = api.CustomVMTags
```
CecileRobertMichon: You're missing the `convertMasterProfileToVLabs` line

pkg/engine/virtualmachines.go
```diff
@@ -563,3 +567,12 @@ func getVaultSecretGroup(linuxProfile *api.LinuxProfile) []compute.VaultSecretGr
 	}
 	return vaultSecretGroups
 }
+
+func addCustomTagsToVM(tags map[string]string, vm *compute.VirtualMachine) {
```
CecileRobertMichon: let's add a small unit test for this

pkg/engine/virtualmachinescalesets.go
```diff
@@ -718,3 +722,12 @@ func getVMSSDataDisks(profile *api.AgentPoolProfile) *[]compute.VirtualMachineSc
 	}
 	return &dataDisks
 }
+
+func addCustomTagsToVMScaleSets(tags map[string]string, vm *compute.VirtualMachineScaleSet) {
```
CecileRobertMichon: and for this

pkg/engine/virtualmachines.go
CecileRobertMichon: let me know if you need any pointers on where/how

Guilhermeslucas: Just did some tests, please let me know what you think about them! Thanks a lot for offering help :)

pkg/api/converterfromapi.go
Guilhermeslucas: You are right! Just did this too. Thanks!

Guilhermeslucas:
Thanks everyone for the help on that, you are awesome :)

PR #1279, opened 2019-05-10
Update hyperkube command in manifests for k8s components.

adelina-t:
In k8s 1.15 this PR : https://github.com/kubernetes/kubernetes/pull/76953
removed aliases for k8s binaries in hyperkube ( i.e proxy alias for kube-proxy ).

This PR changes the commands in all manifests that invoke hyperkube
to use the full name of the binaries.

**Reason for Change**:

PR https://github.com/kubernetes/kubernetes/pull/76953 in kubernetes/kubernetes removed aliases for hyperkube binaries ( apiserver, proxy etc. ), thus the binaries have to be invoked by their full name ( kube-apiserver, kube-proxy etc. ).

This issue has been observed in upstream windows testing for k8s ( https://testgrid.k8s.io/sig-windows#aks-engine-azure-windows-master ).

NOTE: 
- It does not affect aks-engine as it is now, as it only supports k8s 1.15.0-alpha.3 ( Change merged after this ) however it may pose a problem in future releases.
- the full binary name,  kube- , is supported in all k8s versions thus there shouldn't be any backwards compatibility issues.

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
@@ -27,7 +27,7 @@ spec:
       containers:
       - command:
         - /hyperkube
-        - proxy
+        - kube-proxy
         - --kubeconfig=/var/lib/kubelet/kubeconfig
         - --cluster-cidr=<CIDR>
         - --feature-gates=ExperimentalCriticalPodAnnotation=true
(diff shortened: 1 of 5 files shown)


mboersma:
Could you `git commit --amend` the commit message so it's something like:

`fix: update short hyperkube commands in manifests for k8s components`

Otherwise this change won't show up in our CHANGELOG when we do a release. Thanks!

jackfrancis:
Ditto, thanks @adelina-t! (fixed the commit message via squash commit)

Issue #1258, opened 2019-05-08
Walkthrough not up to date, doesn't work


srudin:
**Is this a request for help?**: Yes

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of aks-engine?**: Latest

---

**Kubernetes version**: 1.14

**What happened**: There is a recent blog post saying that Kubernetes now supports Windows containers: https://cloudblogs.microsoft.com/opensource/2019/03/25/windows-server-containers-now-supported-kubernetes/
Indeed the documentation here was updated a few days after the post: https://github.com/Azure/aks-engine/blob/master/docs/topics/windows.md
However the examples that are used inside the tutorial have not been updated for a while: https://github.com/Azure/aks-engine/tree/master/examples/windows
Also the yaml example in the tutorial still referrs to "beta.kubernetes.io/os" which feels wrong.
I tried to follow the tutorial but couldn't make it work. The way I see it the whole thing cannot work as long as the yaml examples are not updated. In fact I wonder if the tutorial is reflecting the latest changes at all as it still feels very comlicated to get Windows containers to run. I mean it was already possible to get Windows container to run on Kubernetes before the latest update, but it involved a couple of tweaks (https://anthonychu.ca/post/hybrid-kubernetes-linux-windows-cluster-easy-steps/). I was hoping that everything would work more "out of the box" now.

**What you expected to happen**:
Please verify and update all required resources for the tutorial. I would expect that I can follow and finish the tutorial without running into errors at some point in the middle.

**How to reproduce it** (as minimally and precisely as possible):
Follow the tutorial...

**Anything else we need to know**:
Thank you.

munishm:
The beta in example is not outdated, it is yet to graduate. Please check this- 
https://github.com/kubernetes/enhancements/blob/master/keps/sig-windows/20190103-windows-node-support.md

munishm:
@srudin - Can you please let us know what did not work in the example?

srudin:
@munishm Thanks for the response. I got stuck after generating the template, when trying to deploy it (Deploy the cluster). I don't recall the error and I didn't investigate too much because I was confused that the samples were several months old and I assumed that they are outdated.

[...]

PR #1280, opened 2019-05-10
docs: update group name and corresponding output folder names

munishm:
**Reason for Change**:
More accurate with the names even though hard coded. Would be easy to understand the example and relate. 

**Issue Fixed**:
#1258 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/windows.md
+++ b/docs/topics/windows.md
@@ -397,7 +397,7 @@ After several minutes, it will return the list of resources created in JSON. Loo
       },
 ```
 
-A dnsPrefix which forms part of the the hostname for your cluster (e.g. staging, prodwest, blueberry). The DNS prefix must be unique so pick a random name. In case you have not modified the dns prefix, it might error out.
+The DNS prefix provided is used to generate the hostname of the cluster (e.g. staging, prodwest, blueberry) and must be unique for each cluster deployment.
 
 #### Check that the cluster is up
 
--- a/docs/topics/windows.md
+++ b/docs/topics/windows.md
@@ -278,7 +278,7 @@ Now that the group is created, create a service principal with Contributor acces
 
 ```powershell
 # Get the group id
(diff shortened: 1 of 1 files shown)


docs/topics/windows.md
```diff
@@ -385,30 +385,32 @@ It will also create a working Kubernetes client config file in `_output/<dnspref
...
 
 ```json
       "masterFQDN": {
         "type": "String",
-        "value": "plangk8swin1.westus2.cloudapp.azure.com"
+        "value": "wink8s1.westus2.cloudapp.azure.com"
       },
 ```
 
+A dnsPrefix which forms part of the the hostname for your cluster (e.g. staging, prodwest, blueberry). The DNS prefix must be unique so pick a random name. In case you have not modified the dns prefix, it might error out.
```
mboersma: This isn't a full sentence, and I'm not sure what it relates to. Is this necessary here?

munishm: DnsPrefix has to be unique as this is global resource and can get into errors if some other user has it under his/her name.
It is more about letting the user know to make it unique.  Wasn't sure how to add this so took a cue from the other documentation.

CecileRobertMichon:
```suggestion
The DNS prefix provided is used to generate the hostname of the cluster (e.g. staging, prodwest, blueberry), and must be unique for each cluster deployment.
```

CecileRobertMichon: @mboersma thoughts?

mboersma: That fixes it. 👍

mboersma:
@PatrickLang could you take a look at this Windows doc change?

docs/topics/windows.md
mboersma: @munishm if you're ok with Cécile's change above, we're ok with this PR. Thanks again!

PR #1283, opened 2019-05-10
fix: USER_ASSIGNED_IDENTITY_ID is empty in azure.json

### LINKED ISSUE #904: "userAssignedIdentityID" in cloud config file is empty, which can cause the cluster not work when multiple user assigned identity is associated to the VM

norshtein:
**Reason for Change**:

This PR is a re-implementation of #1169 and fixes #904.

**Issue Fixed**:

Fixes #904.

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:
This PR reimplements #1169. Instead of removing `provisionScriptParametersCommon`, this PR keeps `provisionScriptParametersCommon` and remove `USER_ASSIGNED_IDENTITY_ID` from `provisionScriptParametersCommon`. When generating envrionment variable, add `USER_ASSIGNED_IDENTITY_ID` section after variable `provisionScriptParametersCommon`.


norshtein:
Manually tested to create a cluster using system assigned identity and a cluster using user assigned identity. For the cluster using system assigned identity, USER_ASSIGNED_IDENTITY_ID in azure.json keeps being empty. For the cluster using user assigned identity, USER_ASSIGNED_IDENTITY_ID in azure.json is successfully filled with the client id of created user assigned identity.

pkg/api/common/helper.go
```diff
@@ -226,3 +226,10 @@ func IsSgxEnabledSKU(vmSize string) bool {
 	}
 	return false
 }
+
+func GenerateUserAssignedIdentityClientIDParameter(isUserAssignedIdentity bool) string {
```
jackfrancis: Because we're only invoking this func from the `engine` package under `pkg/engine` we don't need to define it in the `common` package as an exportable func.

A good place for this might be `pkg/engine/template_generator.go`.

pkg/api/types.go
```diff
@@ -925,6 +925,15 @@ func (p *Properties) IsHostedMasterProfile() bool {
 	return p.HostedMasterProfile != nil
 }
 
+// IsUserAssignedIdentityEnabled returns true if the cluster uses managed identity
```
jackfrancis: We actually already have a method called `UserAssignedIDEnabled` (part of the `KubernetesConfig` type spec) so we don't need this new one.

pkg/engine/armvariables.go
```diff
@@ -110,7 +110,7 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
...offJitter,' CLOUDPROVIDER_RATELIMIT=',toLower(string(parameters('cloudproviderConfig').cloudProviderRatelimit)),' CLOUDPROVIDER_RATELIMIT_QPS=',parameters('cloudproviderConfig').cloudProviderRatelimitQPS,' CLOUDPROVIDER_RATELIMIT_BUCKET=',parameters('cloudproviderConfig').cloudProviderRatelimitBucket,' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension'),' USER_ASSIGNED_IDENTITY_ID=',variables('userAssignedClientID'),' USE_INSTANCE_METADATA=',variables('useInstanceMetadata'),' LOAD_BALANCER_SKU=',variables('loadBalancerSku'),' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB'),' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount'),' CONTAINER_RUNTIME=',parameters('containerRuntime'),' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase'),' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec'),' KMS_PROVIDER_VAULT_NAME=',variables('clusterKeyVaultName'),' IS_HOSTED_MASTER=%t',' PRIVATE_AZURE_REGISTRY_SERVER=',parameters('privateAzureRegistryServer'),' AUTHENTICATION_METHOD=',variables('customCloudAuthenticationMethod'),' IDENTITY_SYSTEM=',variables('customCloudIdentifySystem'))]",
```
jackfrancis: Is there any reason we can't just add this new param into the `provisionScriptParametersCommon` definition explicitly? That way all the other usages of `provisionScriptParametersCommon` will get the proper `USER_ASSIGNED_IDENTITY_ID` env var for free.

jackfrancis:
Thanks again @norshtein!

pkg/engine/armvariables.go
norshtein: The reason is `provisionScriptParametersCommon` is an ARM variable, in [ARM manual](https://docs.microsoft.com/en-us/azure/azure-resource-manager/resource-group-template-functions-resource#reference):

>The reference function can only be used in the properties of a resource definition and the outputs section of a template or deployment.

We can't use function `reference` in an ARM variable. And `USER_ASSIGNED_IDENTITY_ID` will be assigned with `reference(concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID')), '2018-11-30').clientId` , which uses function `reference`, so we can't put it directly in `provisionScriptParametersCommon`

pkg/api/common/helper.go
norshtein: Thanks! Updated.

pkg/api/types.go
norshtein: I've updated to use `UserAssignedIDEnabled` and I feel it's a little bit awkward: I have to judge whether `OrchestratorProfile` and `KubernetesConfig` is `nil` every time before I call this method.

pkg/engine/armvariables.go
jackfrancis: thx for the explanation!

jackfrancis:
One more comment!

Thanks for all the nil dereference protection you've added :) Instead of passing a long command such as:

`generateUserAssignedIdentityClientIDParameter(cs.Properties != nil && cs.Properties.OrchestratorProfile != nil && cs.Properties.OrchestratorProfile.KubernetesConfig != nil && cs.Properties.OrchestratorProfile.KubernetesConfig.UserAssignedIDEnabled())`

Let's do something like this. Prior to the invocation of `generateUserAssignedIdentityClientIDParameter`, get the bool arg that we will pass into it, e.g.:

```
var userAssignedIDEnabled bool
if cs.Properties.OrchestratorProfile != nil && cs.Properties.OrchestratorProfile.KubernetesConfig != nil {
    userAssignedIDEnabled = cs.Properties.OrchestratorProfile.KubernetesConfig.UserAssignedIDEnabled()
} else {
    userAssignedIDEnabled = false
}
```

Then whenever we invoke `generateUserAssignedIdentityClientIDParameter` we'll re-use the `userAssignedIDEnabled` var.

I'd argue something like above is more readable/maintainable.

(btw you don't need to check for cs nil, we'll assume the instance object for these methods is not nil)

norshtein:
@jackfrancis thanks for tips, updated!

PR #1265, opened 2019-05-08
chore: rename addons-related types and consts

jackfrancis:
**Reason for Change**:

While reviewing some addons-related PRs it occured to me that the readability of these areas could be greatly improved with more expressive naming.

Also added unit test coverage to the addons-specific logic that determines the addon payloads to add to cloud-init.

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1290, opened 2019-05-11
fix: PSP using GetAddonScript func

allantargino:
Following the refactor on https://github.com/Azure/aks-engine/pull/1265:
* Now that PodSecurityPolicy manifests live in addons folders (instead of manifests), we can use the func `GetAddonScript` to retrieve custom/raw PSP definitions.
* Changing PSP manifests to use `addonmanager.kubernetes.io/mode: EnsureExists`, so the `addons-manager` will create it only once, allowing the users to create/edit their own PSPs.

**Reason for Change**:

it standardized addons folders so every addon can retrieve custom definitions the same way (using `GetAddonScript` func

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/addons/kubernetesmasteraddons-pod-security-policy.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   annotations:
     seccomp.security.alpha.kubernetes.io/allowedProfileNames: "*"
   labels:
-    addonmanager.kubernetes.io/mode: Reconcile
+    addonmanager.kubernetes.io/mode: EnsureExists
```
jackfrancis: After further testing did you determine that `EnsureExists` is a better option in practice for psp?

pkg/api/types.go
```diff
@@ -379,7 +379,6 @@ type KubernetesConfig struct {
 	CloudControllerManagerConfig     map[string]string `json:"cloudControllerManagerConfig,omitempty"`
 	APIServerConfig                  map[string]string `json:"apiServerConfig,omitempty"`
 	SchedulerConfig                  map[string]string `json:"schedulerConfig,omitempty"`
-	PodSecurityPolicyConfig          map[string]string `json:"podSecurityPolicyConfig,omitempty"`
```
jackfrancis: To ensure back-compat, we can't remove these type definition properties. See `DockerEngineVersion` above: add a `// Deprecated` comment instead of actually deleting the property

pkg/api/vlabs/types.go
```diff
@@ -318,7 +318,6 @@ type KubernetesConfig struct {
 	CloudControllerManagerConfig    map[string]string `json:"cloudControllerManagerConfig,omitempty"`
 	APIServerConfig                 map[string]string `json:"apiServerConfig,omitempty"`
 	SchedulerConfig                 map[string]string `json:"schedulerConfig,omitempty"`
-	PodSecurityPolicyConfig         map[string]string `json:"podSecurityPolicyConfig,omitempty"`
```
jackfrancis: ditto above about deprecation instead of deletion

parts/k8s/addons/kubernetesmasteraddons-pod-security-policy.yaml
allantargino: Yes. The original behavior on the old func https://github.com/Azure/acs-engine/pull/2048/files#diff-f2f5404d19a8ead71fce9bc62331f682R616 only create the PSPs once - without reconciliating them in case of updates/deletes - which makes sense. With `EnsureExists`, the addon manager will create only once (instead of `Reconcile` option)

pkg/api/types.go
allantargino: I see @jackfrancis! My bad :) I forget that this change would break the public API.
Even marking this as `Deprecated` - should we keep both implementations of receiving Raw PSP Manifests? Or just this new one (using addons raw definitions) ? Or just mark this property as deprecated and use the new approach

jackfrancis: We will need to eliminate the old implementation and then add back-compat tolerance for existing clusters. There is an `isUpgrade` flag on the defaults funcs in defaults.go. Basically, we want to check `if isUpgrade`, and if so, if `PodSecurityPolicyConfig` is not nil, then do the appropriate composition into the new addons implementation based on the data inside the "old" map[string]string

allantargino: Cool! Thanks for the detailed instructions 👍

allantargino:
Hi @jackfrancis, 
Can you please check if this was a good way to handle the back compatibility you described?
Also, it is my first time sending tests to the project :) I'm really open to feedbacks.
Thanks!

jackfrancis:
/lgtm

This was a pretty specific back-compat scenario, what you did will work!

Thanks so much @allantargino!

PR #1299, opened 2019-05-14
fix: Get VMSS node version from K8s API instead of tags for old instances

xizhamsft:
Get VMSS node version from K8S API for instances with old model. Also adding back a missed lower case conversion

**Reason for Change**:
All instances in Azure VMSS share the same tags. So in case of partial agent pool upgrade failures, we will end up with nodes on both new model and old model. Yet, the tags of all nodes are of the new model. So retry upgrade in this situation ends up doing nothing. With this fix we will check whether each instance has the latest model applied before checking the tags. If not, we will use K8s API to get node version.

Also, adding back toLower to newnode name when copying node properties. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradecluster.go
+++ b/pkg/operations/kubernetesupgrade/upgradecluster.go
@@ -5,7 +5,6 @@ package kubernetesupgrade
 
 import (
 	"context"
-	"fmt"
 	"strings"
 	"time"
 
@@ -153,7 +152,7 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 					uc.Logger.Infof("Set isWindows flag for vmss %s.", *vmScaleSet.Name)
 				}
 				for _, vm := range vmScaleSetVMsPage.Values() {
-					currentVersion := uc.getNodeVersion(kubeClient, strings.ToLower(*vm.Name), vm.Tags)
+					currentVersion := uc.getNodeVersion(kubeClient, strings.ToLower(*vm.VirtualMachineScaleSetVMProperties.OsProfile.ComputerName), vm.Tags, *vm.VirtualMachineScaleSetVMProperties.LatestModelApplied)
 					if uc.Force {
 						if currentVersion == "" {
 							currentVersion = "Unknown"
@@ -198,7 +197,7 @@ func (uc *UpgradeCluster) getClusterNodeStatus(az armhelpers.AKSEngineClient, re
 					*vm.Name, uc.NameSuffix)
 				continue
 			}
-			currentVersion := uc.getNodeVersion(kubeClient, strings.ToLower(*vm.Name), vm.Tags)
+			currentVersion := uc.getNodeVersion(kubeClient, strings.ToLower(*vm.Name), vm.Tags, true)
 
 			if uc.Force {
 				if currentVersion == "" {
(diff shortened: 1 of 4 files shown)


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -558,7 +558,7 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
 				}
 
 				ku.logger.Infof("Copying custom annotations, labels, taints from old node %s to new node %s...", vmToUpgrade.Name, newNodeName)
-				err = ku.copyCustomPropertiesToNewNode(client, vmToUpgrade.Name, newNodeName)
+				err = ku.copyCustomPropertiesToNewNode(client, strings.ToLower(vmToUpgrade.Name), strings.ToLower(newNodeName))
```
CecileRobertMichon: Good catch

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -248,6 +248,26 @@ func (uc *UpgradeCluster) upgradable(currentVersion string) error {
 	return errors.Errorf("%s cannot be upgraded to %s", currentVersion, targetVersion)
 }
 
+// getVMSSNodeVersion returns a VMSS node's current Kubernetes version via Kubernetes API.
+// We can't rely on tags because VMSS does not support tags for individual instances.
+func (uc *UpgradeCluster) getVMSSNodeVersion(client armhelpers.KubernetesClient, name string, tags map[string]*string, latestModelApplied bool) string {
+	if latestModelApplied && tags != nil && tags["orchestrator"] != nil {
+		parts := strings.Split(*tags["orchestrator"], ":")
+		if len(parts) == 2 {
+			return parts[1]
```
CecileRobertMichon: this seems to me like it does rely on the tag as the first attempt to get the k8s version which seems to be the opposite of what the PR description and function comment indicate

xizhamsft: updated the comments and the PR title.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -248,6 +248,28 @@ func (uc *UpgradeCluster) upgradable(currentVersion string) error {
...
+	}
+
+	if client != nil {
+		node, err := client.GetNode(name)
+		if err == nil {
+			return strings.TrimPrefix(node.Status.NodeInfo.KubeletVersion, "v")
+		}
+		uc.Logger.Warnf("Failed to get node %s: %v", name, err)
+	}
+	return ""
```
CecileRobertMichon: See https://github.com/Azure/aks-engine/pull/1299/files#diff-95227b07e97ba7cbd04cfd8b2cdb6980R288

The VMSS case probably needs to either be applied here or removed from the other function if we don't think it's needed anymore since `getNodeVersion` doesn't get used for VMSS anymore.

Also `getVMSSNodeVersion` has a lot of duplicated code with `getNodeVersion` so we can probably reuse some of the code in one common function.

xizhamsft: Yeap. Thought about doing this but would like to get confirmation first. Removed the VMSS part from getNodeVersions now. Changed the flag name to getVersionFromTags.

PR #1310, opened 2019-05-15
fix: don’t add etcd data disk in cosmos etcd scenario

### LINKED ISSUE #1311: etcd data disk present on cosmos etcd scenario
Cosmos etcd scenario does not use local etcd, and so master vms should not have an etcd-specific data disk attached to the vm.

jackfrancis:
**Reason for Change**:

In cosmos etcd scenario we should not add a data disk for local etcd.

**Issue Fixed**:

Fixes #1311

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -14,7 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
-func CreateVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
+func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 	hasAvailabilityZones := cs.Properties.MasterProfile.HasAvailabilityZones()
 	isStorageAccount := cs.Properties.MasterProfile.IsStorageAccount()
 	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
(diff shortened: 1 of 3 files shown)


pkg/engine/virtualmachines.go
```diff
@@ -14,7 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
-func CreateVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
+func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
```
jackfrancis: Renamed this func because it is (currently) only used to create the master vm. As long as that's the case having this func have a name that reflects that makes code maintenance easier.

pkg/engine/virtualmachines_test.go
```diff
@@ -110,14 +110,28 @@ func TestCreateVirtualMachines(t *testing.T) {
 		t.Errorf("unexpected diff while expecting equal structs: %s", diff)
 	}
 
+	// Validate cosmosetcd
+	cs.Properties.MasterProfile.CosmosEtcd = to.BoolPtr(true)
+	actualVM = CreateMasterVM(cs)
+	expectedVM.StorageProfile.DataDisks = nil
```
jackfrancis: This is the key unit test validation for cosmos etcd scenario

pkg/engine/virtualmachines.go
```diff
@@ -149,19 +149,21 @@ func CreateVirtualMachine(cs *api.ContainerService) VirtualMachineARM {
...
-	dataDisk := compute.DataDisk{
-		CreateOption: compute.DiskCreateOptionTypesEmpty,
-		DiskSizeGB:   to.Int32Ptr(int32(etcdSizeGB)),
-		Lun:          to.Int32Ptr(0),
-		Name:         to.StringPtr("[concat(variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')),'-etcddisk')]"),
-	}
-	if cs.Properties.MasterProfile.IsStorageAccount() {
-		dataDisk.Vhd = &compute.VirtualHardDisk{
-			URI: to.StringPtr("[concat(reference(concat('Microsoft.Storage/storageAccounts/',variables('masterStorageAccountName')),variables('apiVersionStorage')).primaryEndpoints.blob,'vhds/', variables('masterVMNamePrefix'),copyIndex(variables('masterOffset')),'-etcddisk.vhd')]"),
+	if !to.Bool(cs.Properties.MasterProfile.CosmosEtcd) {
```
jackfrancis: Only add data disk if we're not using cosmos etcd

PR #1312, opened 2019-05-16
test: unit tests for kubernetesManifestSettingsInit

jackfrancis:
**Reason for Change**:

Add unit test coverage for convenience funcs that contribute to k8s manifest data payload composition

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/artifacts.go
+++ b/pkg/engine/artifacts.go
@@ -254,6 +254,18 @@ func kubernetesManifestSettingsInit(p *api.Properties) []kubernetesComponentFile
 	}
 	o := p.OrchestratorProfile
 	k := o.KubernetesConfig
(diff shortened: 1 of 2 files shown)


pkg/engine/artifacts.go
```diff
@@ -254,6 +254,18 @@ func kubernetesManifestSettingsInit(p *api.Properties) []kubernetesComponentFile
 	}
 	o := p.OrchestratorProfile
 	k := o.KubernetesConfig
+	if k.SchedulerConfig == nil {
```
jackfrancis: I didn't like that we were dereferencing these maps inside the func but not protecting against nil

PR #1323, opened 2019-05-17
style: check for unused parameters

mboersma:
**Reason for Change**:
Enables the `unparam` linter to check that Go functions don't have unused parameters. This led to interesting results.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/orchestrators.go
+++ b/pkg/api/orchestrators.go
@@ -226,10 +226,7 @@ func dcosInfo(csOrch *OrchestratorProfile, hasWindows bool) ([]*OrchestratorVers
 	if csOrch.OrchestratorVersion == "" {
 		// get info for all supported versions
 		for _, ver := range common.AllDCOSSupportedVersions {
-			upgrades, err := dcosUpgrades(&OrchestratorProfile{OrchestratorVersion: ver})
-			if err != nil {
-				return nil, err
-			}
+			upgrades := dcosUpgrades(&OrchestratorProfile{OrchestratorVersion: ver})
 			orchs = append(orchs,
 				&OrchestratorVersionProfile{
 					OrchestratorProfile: OrchestratorProfile{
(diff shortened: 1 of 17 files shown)


pkg/api/orchestrators.go
```diff
@@ -272,7 +266,7 @@ func dcosUpgrades(csOrch *OrchestratorProfile) ([]*OrchestratorProfile, error) {
 			OrchestratorVersion: common.DCOSVersion1Dot11Dot2,
 		})
 	}
-	return ret, nil
+	return ret
```
mboersma: This always returns `nil` for the error, so I removed that from the function signature.

pkg/armhelpers/azurestack/converter.go
```diff
@@ -25,11 +25,11 @@ func DeepCopy(dst, src interface{}) error {
 	if dstValue.Type().String() != srcValue.Type().String() {
 		return fmt.Errorf("the dst type (%q) and src type (%q) are not the same", dstValue.Type().String(), srcValue.Type().String())
 	}
-	deepCopyInternal(dstValue, srcValue, 0, "")
+	deepCopyInternal(dstValue, srcValue, 0)
 	return nil
 }
 
-func deepCopyInternal(dstValue, srcValue reflect.Value, depth int, path string) {
+func deepCopyInternal(dstValue, srcValue reflect.Value, depth int) {
```
mboersma: This function was only ever called with `path=""` and `path` was otherwise unused.

pkg/engine/engine.go
```diff
@@ -181,28 +181,20 @@ func getStorageAccountType(sizeName string) (string, error) {
...
-			cs.Properties.ExtensionProfiles, copyIndex)
+			cs.Properties.ExtensionProfiles)
 	}
 	return makeExtensionScriptCommands(profile.PreprovisionExtension,
-		cs.Properties.ExtensionProfiles, copyIndex)
+		cs.Properties.ExtensionProfiles)
 }
 
-func makeExtensionScriptCommands(extension *api.Extension, extensionProfiles []*api.ExtensionProfile, copyIndex string) string {
+func makeExtensionScriptCommands(extension *api.Extension, extensionProfiles []*api.ExtensionProfile) string {
```
mboersma: This *feels* like a bug, but `copyIndex` appears to be unused in these functions.

pkg/engine/params.go
```diff
@@ -252,5 +252,5 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
 		}
 	}
 
-	return parametersMap, nil
+	return parametersMap
```
mboersma: This only returned `nil` for error, so that was removed from the function signature.

PR #1333, opened 2019-05-20
test: add prototype for getTemplateFuncMap unit tests

jackfrancis:
**Reason for Change**:

Prior to implementing this change generally, let's determine if this is the right approach. This PR suggests a practical solution to the lack of unit tests in the heavily used `getTemplateFuncMap` template generation function:

- standardize these functions to be exclusively thin wrapping around native go funcs, with the business logic implemented in the native funcs, and not inside the `interface{}`-typed function that the `template.FuncMap` actually returns
- unit test the business logic in the golang native func
- validate the expected funcmap by checking for the existence of expected keys

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/template_generator.go
```diff
@@ -258,8 +258,8 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"GetCustomCloudIdentitySystem": func() string {
 			return cs.Properties.GetCustomCloudIdentitySystem()
 		},
-		"IsMultipleMasters": func() bool {
-			return cs.Properties.MasterProfile != nil && cs.Properties.MasterProfile.Count > 1
+		"IsMultiMasterCluster": func() bool {
+			return cs.Properties.MasterProfile.HasMultipleNodes()
```
jackfrancis: This is an example of the business logic being moved out of the funcmap itself and into the unit testable `MasterProfile.HasMultipleNodes()` method.

pkg/api/defaults_test.go
```diff
@@ -1514,7 +1514,7 @@ func TestSetCertDefaults(t *testing.T) {
 			t.Errorf("expected last IP of master vm from SetDefaultCerts %d, actual %d", expectedLastIP, ips[len(ips)-2])
 		}
 
-		if cs.Properties.MasterProfile.Count > 1 {
+		if cs.Properties.MasterProfile.HasMultipleNodes() {
```
jackfrancis: Replaced a bunch of manual "is count > 1" implementations with the new method

pkg/api/types.go
```diff
@@ -1263,6 +1263,11 @@ func (m *MasterProfile) IsUbuntuNonVHD() bool {
 	return m.IsUbuntu() && !m.IsVHDDistro()
 }
 
+// HasMultipleNodes returns true if there are more than one master nodes
+func (m *MasterProfile) HasMultipleNodes() bool {
```
jackfrancis: Here's the new method

pkg/api/types_test.go
```diff
@@ -387,6 +387,49 @@ func TestMasterProfileIsUbuntuNonVHD(t *testing.T) {
 	}
 }
 
+func TestMasterProfileHasMultipleNodes(t *testing.T) {
```
jackfrancis: Here's how we would add functional unit test coverage, to the golang native func that implements the business logic

pkg/engine/template_generator_test.go
```diff
@@ -28,3 +28,35 @@ func TestGenerateTemplateV2(t *testing.T) {
...
+	apiModelStr := getAPIModelString()
+	cs := &api.ContainerService{}
+	err = json.Unmarshal([]byte(apiModelStr), &cs)
+	if err != nil {
+		t.Errorf("unexpected error while unmarshalling the apiModel JSON: %s", err.Error())
+	}
+	funcmap := tg.getTemplateFuncMap(cs)
+	cases := []struct {
+		key string
+	}{
```
jackfrancis: This is the simple "I know my pre-generated templates will depend upon this function, please validate it's there" unit test.

pkg/engine/template_generator.go
CecileRobertMichon: this is called in nodecustomdata so we still need to check for nil Master profile

CecileRobertMichon: (and in general I think it's always better to nil check MasterProfile everywhere we dereference it since new uses could be added later in code paths that are used by hosted masters)

jackfrancis: I almost think for defensiveness we should move all `MasterProfile` methods into `Properties` so we can do that nil check in once place.

CecileRobertMichon: yes also arguably nil master profile (which returns false in this case) is part of the logic and is one of the cases that should be tested for in the unit test and moving the nil check to the function itself would allow us to do that.

jackfrancis: Yeah, will do that as a discrete PR (apart from these unit test PRs)

PR #1344, opened 2019-05-21
test: getTemplateFuncMap unit tests part 3

jackfrancis:
**Reason for Change**:

Ensures that the following `getTemplateFuncMap` function keys are thinly wrapped go functions with unit test coverage:

```
		"GetProbes",
		"GetSecurityRules",
		"GetUniqueNameSuffix",
		"GetVNETAddressPrefixes",
		"GetVNETSubnets",
		"GetDataDisks",
		"HasBootstrap",
```

Also removed the entirely unnecessary `HasBootstrapPublicIP` and `IsHostedBootstrap` static `return false` template functions.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/dcos/bstrap/bootstrapparams.t
+++ b/parts/dcos/bstrap/bootstrapparams.t
@@ -12,22 +12,6 @@
       "type": "securestring"
       },
     {{end}}
-{{if IsHostedBootstrap}}
-    "bootstrapSubnet": {
-      "defaultValue": "{{.HostedBootstrapProfile.Subnet}}",
-      "metadata": {
-        "description": "Sets the subnet for the VMs in the cluster."
-      },
-      "type": "string"
-    },
-    "bootstrapEndpoint": {
-      "defaultValue": "{{.HostedBootstrapProfile.FQDN}}",
-      "metadata": {
-        "description": "Sets the static IP of the first bootstrap"
-      },
-      "type": "string"
-    },
-{{else}}
     "bootstrapStaticIP": {
       "metadata": {
         "description": "Sets the static IP of the first bootstrap"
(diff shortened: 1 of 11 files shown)


pkg/engine/engine_test.go
```diff
@@ -672,28 +697,6 @@ func TestGenerateKubeConfig(t *testing.T) {
 	}
 }
 
-func TestGetDataDisks(t *testing.T) {
```
jackfrancis: busted this was not an actual test :)

pkg/engine/template_generator.go
```diff
@@ -355,13 +355,7 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			return getDataDisks(profile)
 		},
 		"HasBootstrap": func() bool {
-			return cs.Properties.OrchestratorProfile.DcosConfig != nil && cs.Properties.OrchestratorProfile.DcosConfig.BootstrapProfile != nil
-		},
-		"HasBootstrapPublicIP": func() bool {
-			return false
```
CecileRobertMichon: good one

parts/dcos/bstrap/bootstrapparams.t
```diff
@@ -12,22 +12,6 @@
       "type": "securestring"
       },
     {{end}}
-{{if IsHostedBootstrap}}
```
CecileRobertMichon: I agree this change is pretty straight forward but I still think we should hold ourselves to not changing any legacy DCOS code without having an e2e test.

jackfrancis: I think this is just unacceptable cruft we should get rid of in order to streamline the code surface area.

PR #1351, opened 2019-05-22
refactor: triage untestable DCOS code

jackfrancis:
**Reason for Change**:

Move untestable DC/OS code to a distinct go file marked as "skip unit tests here". Also includes a refactor of the `generateIPList` func (to `generateConsecutiveIPsList`) to include basic edge case error checking.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/engine.go
```diff
@@ -119,16 +118,25 @@ func validateDistro(cs *api.ContainerService) bool {
 	return true
 }
 
-func generateIPList(count int, firstAddr string) []string {
+// generateConsecutiveIPsList takes a starting IP address and returns a string slice of length "count" of subsequent, consecutive IP addresses
+func generateConsecutiveIPsList(count int, firstAddr string) ([]string, error) {
```
jackfrancis: I actually made the effort to refactor + test this function because it was appalling in its current state

pkg/engine/engine_dcos.go
```diff
@@ -0,0 +1,250 @@
+//+build !test
```
jackfrancis: @mboersma I was hoping that this pragma would remove this file from unit test coverage calculation, which should have moved the needle on our % coverage by a wee bit. Do you know of any better ideas?

mboersma: You could add it to the `ignore:` section in .codecov/codecov.yml.

CecileRobertMichon: does `//+build !test` do anything?

pkg/engine/template_generator.go
```diff
@@ -10,7 +10,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"runtime/debug"
-	"strconv"
```
CecileRobertMichon: did you run `make generate` on these changes?

jackfrancis: Yes, does something look off?

pkg/engine/engine_dcos.go
jackfrancis: It tells the compiler not to include this when running unit tests

CecileRobertMichon: but does it actually work? I understood your comment above to mean that it doesn't

jackfrancis: It didn't seem to influence our current test coverage toolchain, but I'm not sure that our current test coverage calculation implementation is vanilla go-idiomatic. So basically: I left this in place as a way of formally declaring intent for these funcs in a go-idiomatic fashion. So at minimum it's basically a code comment at this point.

pkg/engine/template_generator.go
CecileRobertMichon: I don't see a templates_generated.go diff

jackfrancis: What part of this changeset do we expect to see reflected in `templates_generated.go`?

CecileRobertMichon: I guess none, I thought I saw a parts/ change earlier but it's actually all pkg/

CecileRobertMichon:
@jackfrancis is this ready for final review?

pkg/engine/engine.go
```diff
@@ -119,16 +118,25 @@ func validateDistro(cs *api.ContainerService) bool {
...
 	ipaddr := net.ParseIP(firstAddr).To4()
 	if ipaddr == nil {
-		panic(fmt.Sprintf("IPAddr '%s' is an invalid IP address", firstAddr))
```
mboersma: Every time you remove a `panic()`, you gain karma. :-)

PR #1354, opened 2019-05-23
chore: ensure all Linux-bound files are LF (not CRLF)

jackfrancis:
**Reason for Change**:

In particular, apt operations are not able to upgrade + modify configuration files that have CRLF newlines.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon:
is there an easy way to add a check to our lint step in CI to ensure that new files with CRLF do not get added?

mboersma:
Maybe instead of a CI step, we could change our [`.gitattributes`](https://help.github.com/en/articles/dealing-with-line-endings) file to declare *.yaml and *.json as always requiring LF line endings.

jackfrancis:
I don't think LF line endings will help us all the time, as a file could have both

Issue #1291, opened 2019-05-12
Cluster creation in US DoD Central


oronboni:
**Is this a request for help?**:

Yes

**Is this an ISSUE or FEATURE REQUEST?** (choose one):

ISSUE

**What version of aks-engine?**:

0.35.1

{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorVersion": "1.12.8",
      "kubernetesConfig": {
        "apiServerConfig": {
           "--enable-admission-plugins": "NamespaceLifecycle,LimitRanger,ServiceAccount,DefaultStorageClass,DefaultTolerationSeconds,MutatingAdmissionWebhook,ValidatingAdmissionWebhook,ResourceQuota,ExtendedResourceToleration"
          },
        "useManagedIdentity": true,
		"userAssignedID": "@__userAssignedName__@",
        "addons": [
          {
            "name": "cluster-autoscaler",
            "enabled": true,
            "config": {
              "minNodes": "4",
              "maxNodes": "50"
            }
          }
        ]
      }
    },
    "masterProfile": {
      "count": 3,
      "dnsPrefix": "@__Master_dnsPrefix__@",
      "vmSize": "Standard_D2_v2",
      "vnetSubnetId": "/subscriptions/@__subscriptionID__@/resourceGroups/@__resourceGroupName__@/providers/Microsoft.Network/virtualNetworks/@__clusterVnetName__@/subnets/@__Master_vnetSubnetName__@",
      "firstConsecutiveStaticIP": "10.10.0.230",
      "vnetCidr": "10.10.0.0/16",
      "subjectAltNames": [
        "@__Master_dnsPrefix__@.@__Master_DomainSuffix__@"
      ]
    },
    "agentPoolProfiles": [
      {
        "name": "wdatpagents",
        "count": 4,
        "vmSize": "Standard_D4_v2",
        "vnetSubnetId": "/subscriptions/@__subscriptionID__@/resourceGroups/@__resourceGroupName__@/providers/Microsoft.Network/virtualNetworks/@__clusterVnetName__@/subnets/@__agent_vnetSubnetName__@",
        "availabilityProfile": "VirtualMachineScaleSets",
        "storageProfile": "ManagedDisks",
        "osDiskSizeGB":  400,
        "acceleratedNetworkingEnabled": true
      }
    ],
    "linuxProfile": {
      "adminUsername": "wdatpuser",
[...]

PR #1357, opened 2019-05-23
fix: do not set distro to VHD for US Gov and German cloud

CecileRobertMichon:
**Reason for Change**:

The distro value needs to be set to `Ubuntu` for Azure clouds that don't support the VHD distros yet. Overriding the image value does not suffice since we do introspection on the distro value in the code to determine other cluster configurations (for example --protect-kernel-defaults kubelet flag).

**Issue Fixed**:

#1291 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -1582,7 +1582,7 @@ func TestSetCertDefaults(t *testing.T) {
 	}
 
 	cs.setOrchestratorDefaults(false)
-	cs.Properties.setMasterProfileDefaults(false, false)
+	cs.Properties.setMasterProfileDefaults(false, false, AzurePublicCloud)
 	result, ips, err := cs.SetDefaultCerts()
 
 	if !result {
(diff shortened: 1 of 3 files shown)


jackfrancis:
see the `TestSetComponentsNetworkDefaults` unit test under `defaults_test.go` for an example of how we will want to unit test these cloud spec-influenced `Distro` default assignments

CecileRobertMichon:
@jackfrancis `TestSetComponentsNetworkDefaults` doesn't exist anymore but I assume you mean `TestDistroDefaults`

jackfrancis:
I guess old branch :)

pkg/api/defaults_test.go
```diff
@@ -1109,6 +1126,21 @@ func TestDistroDefaults(t *testing.T) {
...
 			true,
+			false,
+			AzurePublicCloud,
+		},
+		{
+			"default_kubernetes_usgov",
+			OrchestratorProfile{
+				OrchestratorType: Kubernetes,
+			},
+			AKS1604Deprecated,
```
jackfrancis: Why does this scenario get the deprecated aks distro?

jackfrancis: oh this is because this test is too cute and doesn't declare the keys

CecileRobertMichon: yeah that's not the expected output, that's the input

Issue #1355, opened 2019-05-23
[bug] scale and upgrade the cluster fails due to moby installation failure


mo-saeed:
**Is this a request for help?**: YES

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): ISSUE

---

**What version of aks-engine?**: 0.36.0

---

**Kubernetes version**: 1.13.5 

**What happened**:
cluster was installed by aks-engine (0.33.2) and trying to scale one new node with aks-engine (0.36.0) but failed with this error:

```
"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/arm-debug for usage details." Details=[{"code":"Conflict","message":"{\r\n  \"status\": \"Failed\",\r\n  \"error\": {\r\n    \"code\": \"ResourceDeploymentFailure\",\r\n    \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n    \"details\": [\r\n      {\r\n        \"code\": \"VMExtensionProvisioningError\",\r\n        \"message\": \"VM has reported a failure when processing extension 'cse-master-0'. Error message: \\\"Enable failed: failed to execute command: command terminated with exit status=27\\n[stdout]\\n\\n[stderr]\\nConnection to k8s.gcr.io 443 port [tcp/https] succeeded!\\nConnection to gcr.io 443 port [tcp/https] succeeded!\\nConnection to docker.io 443 port [tcp/https] succeeded!\\n\\\".\"\r\n      }\r\n    ]\r\n  }\r\n}"}
```
status=27: ERR_MOBY_INSTALL_TIMEOUT=27 # Timeout waiting for moby install

same error raised when trying to upgrade the cluster from 1.13.5 to 1.13.6

**What you expected to happen**:
- Scale up or upgrade the cluster should work with latest aks-engine version 
- scale up/upgrade with latest aks-engine should automatically update apimode.json (e.g. with moby version 3.0.5) but apimodel.json kept the same with the old moby version 3.0.4

**How to reproduce it** (as minimally and precisely as possible):
- install cluster with aks-engine 0.33.2
- scale up new node with aks-engine 0.36.0

**Anything else we need to know**:
what's the best practice to upgrade aks-engine version ? should the apimodel.json be changed manually to adapt new changes of the new version or ?

CecileRobertMichon:
@mo-saeed I was able to repro, thank you for reporting this. I'm working on finding root cause + fix.

[...]

PR #1359, opened 2019-05-23
fix: never use 3.0.4 for moby-cli and update moby on upgrade

CecileRobertMichon:
**Reason for Change**:

Backwards compatibility is broken in master because the moby install function is trying to install moby-cli 3.0.4 for clusters that were deployed with previous versions of aks-engine that had mobyVersion `3.0.4`.

**Issue Fixed**:

#1355 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -772,7 +772,43 @@ func TestContainerRuntime(t *testing.T) {
 	mockCS := getMockBaseContainerService("1.10.13")
 	properties := mockCS.Properties
 	properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	properties.OrchestratorProfile.KubernetesConfig.MobyVersion = "3.0.4"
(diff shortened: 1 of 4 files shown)


pkg/api/defaults_test.go
```diff
@@ -772,6 +772,7 @@ func TestContainerRuntime(t *testing.T) {
 	mockCS := getMockBaseContainerService("1.10.13")
 	properties := mockCS.Properties
 	properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	properties.OrchestratorProfile.KubernetesConfig.MobyVersion = "3.0.4"
```
jackfrancis: Can we do a for loop with at least two < 3.0.5 versions as input? Just to make sure that we don't accidentally regress the test if the static string actually matches the value of `DefaultMobyVersion`

PR #1360, opened 2019-05-23
fix: update etcd and containerd versions during upgrade

CecileRobertMichon:
**Reason for Change**:

When a cluster is upgraded or scaled currently, the etcd version and containerd version in. the apimodel get persisted instead of getting bumped to the latest. This PR changes the behavior to match other components by updating the version of etcd and containerd during update operations.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -118,7 +118,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpdate bool) {
 		if o.KubernetesConfig.KubernetesImageBase == "" {
 			o.KubernetesConfig.KubernetesImageBase = cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase
 		}
-		if o.KubernetesConfig.EtcdVersion == "" {
+		if o.KubernetesConfig.EtcdVersion == "" || isUpdate {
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
@mboersma https://github.com/Azure/aks-engine/blob/master/pkg/api/addons.go#L360 makes me think that we're already updating addon images on updates.

pkg/api/defaults.go
```diff
@@ -138,7 +138,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpdate bool) {
 				o.KubernetesConfig.MobyVersion = DefaultMobyVersion
 			}
 		case Containerd, ClearContainers, KataContainers:
-			if o.KubernetesConfig.ContainerdVersion == "" {
+			if o.KubernetesConfig.ContainerdVersion == "" || isUpdate {
```
jackfrancis: Can we get unit tests for this as well?

CecileRobertMichon: it's already tested

CecileRobertMichon: oh wait let me rebase and fix the log messages

Issue #1306, opened 2019-05-15
Ubuntu 18.04: Timesync stays Idle for a while after boot


CecileRobertMichon:
Our E2E tests revealed that timesync isn't always Synchronized on all VMs when the cluster initially comes up on Ubuntu 18.04 cluster. The same test passes 100% of the time with Ubuntu 16.04.

Soak tests show that eventually all the VMs have Synchronized `systemd-timesyncd` and that once Synchronized it stays Synchronized without any unreliability. 

```
systemctl status systemd-timesyncd● systemd-timesyncd.service - Network Time Synchronization
   Loaded: loaded (/lib/systemd/system/systemd-timesyncd.service; enabled; vendor preset: enabled)
   Active: active (running) since Wed 2019-05-08 17:32:43 UTC; 7h ago
     Docs: man:systemd-timesyncd.service(8)
 Main PID: 771 (systemd-timesyn)
   Status: "Idle."
    Tasks: 2 (limit: 4915)
   CGroup: /system.slice/systemd-timesyncd.service
           └─771 /lib/systemd/systemd-timesyncd
```

E2E output showing failure:
```
+ [[ 18.04 == \1\8\.\0\4 ]]
+ sudo timedatectl status
+ grep 'systemd-timesyncd.service active: yes'
systemd-timesyncd.service active: yes
+ grep 'System clock synchronized: yes'
+ sudo timedatectl status
       System clock synchronized: yes
+ grep 'RTC in local TZ: no'
+ sudo timedatectl status
                 RTC in local TZ: no
+ grep 'Active: active'
+ sudo systemctl status systemd-timesyncd
   Active: active (running) since Tue 2019-05-14 17:52:27 UTC; 2min 52s ago
+ grep 'Status: "Synchronized to time server'
+ sudo systemctl status systemd-timesyncd
+ exit 1"
```

How to reproduce:
Run E2E with `GINKGO_FOCUS=should have healthy time synchronization` on a cluster built with `aks-18.04` distro.

```
[...]

PR #1362, opened 2019-05-24
fix: start Docker service before the Kubelet service on Windows node

### LINKED ISSUE #351: Kubelet service causes CPU 100% when docker service is not started in Windows nodes
No
---

ISSUE
---

master
---

**What happened**:
Due to timeout issue, docker service was not started after Windows started, I found these in event logs,
```
The Docker service failed to start due to the following error: %%1053
A timeout was reached (30000 milliseconds) while waiting for the Docker service to connect.
```
Kubelet service was causing CPU 100% in this situation.
aks-engine uses nssm to run kubelet service, kubeletstart.ps1 ran into exception because it faild to run c:\k\kubelet.exe because of stopped docker service, nssm will re-execute kubeletstart.ps1 indefinitely which causing CPU 100%.

Kubelet service starts successfully.
Is it possible to ensure docker service before call c:\k\kubelet.exe in kubeletstart.ps1?

**How to reproduce it**:
remote into k8s Windows node, execute these cmdlets,
Stop-Service docker -Force
Stop-Service kubelet -Force
Start-Service kubelet

sanderaernouts:
**Reason for Change**:
When the `Docker` service is not running on a Windows node the `Kubelet` service will fail to start. However `nssm.exe` will try indefinitely to restart the `Kubelet` service. This causes the CPU on the node to spike to 100%.

This PR contains 2 changes:
- add a restart delay of 5 seconds so that the CPU does not go to 100% when the `Kubelet` service repeatedly fails to start.
- make the `Kubelet` service depended on the `Docker` service so that the `Docker` service will be started when the `Kubelet` service is restarted. 

This should allow a Windows node to recover automatically from failure when the Docker service crashes or shuts down.

**Issue Fixed**:
Fixes #351

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
Tested these settings on our current AKS engine 0.34.0 cluster.
As you can see below after applying `nssm.exe set Kubelet AppRestartDelay 5000` the CPU does not spike to 100% anymore:
826E0C26-7DEC-4520-8096-E8F20D8B9A8F

As you can see below after applying `nssm.exe set Kubelet DependOnService docker` the `Docker` service is started when the `Kubelet` service is started:
BDE19046-8FCF-4F58-9606-93222BCA1111


sanderaernouts:
@PatrickLang these are the fixes you mentioned in https://github.com/Azure/aks-engine/issues/351#issuecomment-457256310

jackfrancis:
Thanks for this @sanderaernouts! What would be an appropriate E2E test to simulate this additional resilience? Log into a windows node and kill the docker process? Would we expect the running kubelet process to detect that and restart docker automatically with these changes?

sanderaernouts:
Yes that should be the case. I tested it on one of our nodes by running these commands:
Stop-Service docker -Force
Stop-Service kubelet -Force
Start-Service kubelet

The 2nd GIF I posted in my PR shows what happened when I ran: Start-Service kubelet.

Is it possible/desirable to automate that test? Or to add it to an existing automated e2e test?

sanderaernouts:
We had one node in our cluster failing overnight for days in a row now because the `Docker` service stopped. I already applied the changes I made in this PR to our production cluster and it seems it is now at least able to recover from this:
image

Before the node would stay down until I logged on and manually restarted the `Docker` service.  I haven't checked the logs from the `Kubelet` yet though to verify this was the `Docker` service crashing or shutting down.

sanderaernouts:
@CecileRobertMichon I see the `pr-e2e` tests are failing on the `should have healthy time sychronization` test. Is this related to my change or is this something else?
image

CecileRobertMichon:
> @CecileRobertMichon I see the pr-e2e tests are failing on the should have healthy time sychronization test. Is this related to my change or is this something else?

Not related, it's a known intermittent issue see https://github.com/Azure/aks-engine/issues/1306.

I've restarted the failed job.

jackfrancis:
@sanderaernouts what happens in this PR branch when you simply stop docker? Based on the description, I'd expect the kubelet service dependency definition to automatically detect that and reconcile accordingly. That would be my vote for what to add to an E2E test btw: at some point during the run, force stop docker on all nodes, and validate that the nodes become operational again automatically within an appropriate timeout.

sanderaernouts:
@jackfrancis I'll have a look at adding that test to the e2e tests.

sanderaernouts:
@jackfrancis @CecileRobertMichon I have added an e2e test for this, but I cannot get it to run locally:
```
root@LAPTOP-TB37VQUB:/mnt/c/Users/sande/go/src/github.com/Azure/aks-engine# make test-kubernetes 
237 
23270
23525
23752
29283
29621
29844
# github.com/Azure/aks-engine/vendor/github.com/influxdata/platform/pkg/escape 
runtime: failed to create new OS thread (have
2019/05/30 16:01:14 Picking Random Region from list [eastus southcentralus southeastasia westus2 westeurope] 
2019/05/30 16:01:14 Picked Random Region:southeastasia

$ which timeout

$ timeout 60 az account set --subscription df2453f2-4198-4eca-bc30-b87e9388c523 
2019/05/30 16:01:16  
Kubeconfig:/mnt/c/Users/sande/go/src/github.com/Azure/aks-engine/_output/LAPTOP-TB37VQUB/kubeconfig/kubeconfig.southeastasia.json
2019/05/30 16:01:16 Error while trying to read cluster definition at (/mnt/c/Users/sande/go/src/github.com/Azure/aks-engine/_output/LAPTOP-TB37VQUB.json):open /mnt/c/Users/sande/go/src/github.com/Azure/aks-engine/_output/LAPTOP-TB37VQUB.json: no such file or directory
2019/05/30 16:01:16 cliProvisioner.FetchProvisioningMetrics error: open /mnt/c/Users/sande/go/src/github.com/Azure/aks-engine/_output/LAPTOP-TB37VQUB-ssh: no such file or directory
2019/05/30 16:01:16 Error trying to parse engine template into memory:open /mnt/c/Users/sande/go/src/github.com/Azure/aks-engine/_output/LAPTOP-TB37VQUB.json: no such file or directory
exit status 1 
test.mk:28: recipe for target 'test-kubernetes' failed 
make: *** [test-kubernetes] Error 1
```

I cannot find any guidance as to how to generate the `apimodel.json` or cluster files to run these tests. How do I fix this?

CecileRobertMichon:
Here's what I use to run tests locally:

```
CLUSTER_DEFINITION=examples/kubernetes.json SUBSCRIPTION_ID="<YOUR_SUB_ID>" CLIENT_ID="<YOUR_CLIENT_ID" CLIENT_SECRET="<YOUR_CLIENT_SECRET>" TENANT_ID="<YOUR_TENANT_ID>" LOCATION=<REGION> CLEANUP_ON_EXIT=true make test-kubernetes
```

sanderaernouts:
Thanks, that `CLUSTER_DEFINITION` variable seems to do the trick. It is missing from this list though in the developer guid though: https://github.com/Azure/aks-engine/blob/master/docs/community/developer-guide.md#end-to-end-tests.

CecileRobertMichon:
@sanderaernouts good catch. Could you please add it in the PR? 

Something like:
```
CLUSTER_DEFINITION: Input apimodel. Defaults to `examples/kubernetes.json`.
```

Note that it shouldn't be required since it defaults to `examples/kubernetes.json`.

sanderaernouts:
done, I also added the `LOCATION` env variable and the example command you posted in your comment to the e2e test section.

FYI for me the tests do not work without me settings `CLUSTER_DEFINITION` as per your example so it does not seem to default to `examples/kubernetes.json`

CecileRobertMichon:
@sanderaernouts I see 3 changes to vendor directory files but no changes to gopkg.toml, are these changes intentional?

sanderaernouts:
@CecileRobertMichon no that was not intentional. I have overwritten that commit to exclude the changes in the `vendor/` directory.

sanderaernouts:
@CecileRobertMichon still struggling to get the e2e to run locally. Can you start the e2e job on this PR again?

sanderaernouts:
@CecileRobertMichon I have the test running locally now, but I get a `permission denied` error when I try to execute a command on the node. Could you please review my test?

examples/windows/kubernetes.json
```diff
@@ -3,7 +3,7 @@
   "properties": {
     "orchestratorProfile": {
       "orchestratorType": "Kubernetes",
-      "orchestratorRelease": "1.13"
+      "orchestratorVersion": "1.14.1"
```
CecileRobertMichon: let's not pin to a specific k8s version, this is hard to maintain overtime as it will need to be updated everytime we introduce a new patch. Why is this change relevant to this PR?

test/e2e/kubernetes/node/node.go
```diff
@@ -100,6 +100,11 @@ func (n *Node) IsLinux() bool {
 	return n.Status.NodeInfo.OperatingSystem == "linux"
 }
 
+// IsWindows checks for a Windows node
+func (n *Node) IsWindows() bool {
```
CecileRobertMichon: I don't see this being used anywhere, is it intentional?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -92,6 +92,37 @@ var _ = BeforeSuite(func() {
...
 var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", func() {
+	Describe("when docker service is stopped", func() {
+		It("node should become ready again in a reasonable amount of time", func() {
+			kubeConfig, err := GetConfig()
+			Expect(err).NotTo(HaveOccurred())
+			master := fmt.Sprintf("%s@%s", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, kubeConfig.GetServerName())
+			nodeList, err := node.GetReady()
+			Expect(err).NotTo(HaveOccurred())
+
+			for _, node := range nodeList.Nodes {
```
CecileRobertMichon: should this be for all nodes? ie both Linux and Windows

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -92,6 +92,37 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", func() {
+	Describe("when docker service is stopped", func() {
```
CecileRobertMichon: you'll need to move this test down to after the SSH key forwarding is set up in order for SSH to work

test/e2e/kubernetes/kubernetes_test.go
sanderaernouts: I originally intended to only do it for Windows nodes but I changed my mind on that, I think the Kubelet should be able to recover for both Windows and Linux node even if the docker is service is stopped. If the Linux nodes are giving trouble I'll switch it back to Windows only for now.

examples/windows/kubernetes.json
sanderaernouts: This was the only way I could get it to work with the Windows setup. I kept getting errors like "version x does not support windows nodes, use one of these versions instead: ....."

sanderaernouts: I changed it to `orchestratorRelease: 1.14`, that seems to work

test/e2e/kubernetes/kubernetes_test.go
sanderaernouts: Done, the test is running (locally) now.

sanderaernouts:
@CecileRobertMichon thanks for your review, I have processed your comments and am now able to run the tests locally as well. Can you have another look based on the changes I made? 

Locally the new e2e is still failing though. It looks like the node is not getting back up after I stop the docker service. I'll have to go onto one of the nodes and manually try to reproduce the results but that will have to wait till later this week.

sanderaernouts:
@CecileRobertMichon I have updated the test and they now pass locally. It does seem that the `kubelet` does not crash (or it takes a while) when the `docker` service is stopped. So I had to change the test to the following flow:
1. stop docker service on each Windows node
2. stop kubelet service on each Windows node
3. wait 1 minute (giving the nodes some time to report a `NotReady` state)
4. start the kubelet service on each Windows node
5. wait for all nodes to become ready again

Can you preform a review again and maybe @jackfrancis as well

CecileRobertMichon:
@sanderaernouts thanks. Could you please rebase to fix the merge conflict and run `make generate` again to ensure the generated files are up to date?

sanderaernouts:
@CecileRobertMichon done :+1:

pkg/engine/template_generator.go
```diff
@@ -536,6 +536,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 			}
 			return false
 		},
+		"HasWindowsAgents": func() bool {
```
CecileRobertMichon: there is already an existing function for this https://github.com/Azure/aks-engine/blob/master/pkg/api/types.go#L767

CecileRobertMichon: and it doesn't look like you're using this anywhere right now, the e2e tests are using https://github.com/Azure/aks-engine/blob/master/test/e2e/engine/template.go#L198

examples/windows/kubernetes.json
```diff
@@ -3,7 +3,7 @@
   "properties": {
     "orchestratorProfile": {
       "orchestratorType": "Kubernetes",
-      "orchestratorRelease": "1.13"
+      "orchestratorRelease": "1.14"
```
CecileRobertMichon: @PatrickLang are you okay with this change?

sanderaernouts:
@CecileRobertMichon I fixed the linting error but I'm not allowed to start the Azure Pipeline, can you run it again?

sanderaernouts:
@PatrickLang looks like this PR is awaiting a review from you, can you have a look?

sanderaernouts:
@CecileRobertMichon I rebased the branch again. Can we get the PR reviewed and merged?

examples/windows/kubernetes.json
PatrickLang: yes looks great. we should probably trim down the number of examples here if they're out of date

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -238,6 +238,8 @@ New-NSSMService {
     & "$KubeDir\nssm.exe" set Kubelet AppDirectory $KubeDir
     & "$KubeDir\nssm.exe" set Kubelet AppParameters $KubeletStartFile
     & "$KubeDir\nssm.exe" set Kubelet DisplayName Kubelet
+    & "$KubeDir\nssm.exe" set Kubelet AppRestartDelay 5000
+    & "$KubeDir\nssm.exe" set Kubelet DependOnService docker
```
PatrickLang: this will need to change later once containerd is done, but ETA on that is still late this year. I'll update it then

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -782,6 +782,69 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+	Describe("with a windows agent pool", func() {
+		It("kubelet service should be able to recover when the docker service is stopped", func() {
+			if eng.HasWindowsAgents() {
+				kubeConfig, err := GetConfig()
+				Expect(err).NotTo(HaveOccurred())
+				master := fmt.Sprintf("%s@%s", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, kubeConfig.GetServerName())
+				nodeList, err := node.GetReady()
+				Expect(err).NotTo(HaveOccurred())
+
+				simulateDockerdCrashScript := "simulate-dockerd-crash.cmd"
```
PatrickLang: rather than scp & ssh, I think it might be cleaner to just run `ssh ... cmd /c 'net stop docker & ping localhost -t 1 -w 5 & net start docker'`

PatrickLang: I'll go ahead and do that in a followup PR since this is passing though

PatrickLang:
/lgtm
I restarted the 2 failed tests in the pipeline. if they pass, go ahead and merge. I'm making some other test changes so I may simplify the service restart code slightly

test/e2e/kubernetes/kubernetes_test.go
sanderaernouts: @PatrickLang was just going to have a look a this comment. But I see the PR is already merged 👍 

Not sure if `ssh ... cmd /c 'net stop docker & ping localhost -t 1 -w 5 & net start docker'` will do the trick though. The situation that I wanted to create is:
1. `docker` is stopped
2. `kubelet` is restarted

That is what caused the original issue at least. I did notice during testing that the `kubelet` itself (not the service but the executable) does not crash/exit in reasonable amount of time when you crash the docker service (`taskkill /F /FI "SERVICES eq yourservice"`). That is why the `kubelet` is restarted over SSH, in the end the goal was to verify that the `kubelet` service could start when the `docker` service was stopped.

sanderaernouts:
@CecileRobertMichon is it possible to release this fix? I want to update our cluster to AKS Engine 0.36.x and the latest Azure CNI version but if I do that right now I will have to manually go onto the Windows nodes and make this fix.

CecileRobertMichon:
@sanderaernouts it will be included in 0.37.0 which should drop by tomorrow.

PR #1364, opened 2019-05-24
docs: update prow docs

CecileRobertMichon:
**Reason for Change**:

Refresh prow docs.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.prowci/plugins.yaml
+++ b/.prowci/plugins.yaml
@@ -1,8 +1,8 @@
 approve:
 - repos:
   - Azure/aks-engine
-  implicit_self_approve: true
   lgtm_acts_as_approve: true
+  require_self_approval: false
 
 config_updater:
   maps:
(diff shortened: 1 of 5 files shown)


.prowci/plugins.yaml
```diff
@@ -1,8 +1,8 @@
 approve:
 - repos:
   - Azure/aks-engine
-  implicit_self_approve: true
   lgtm_acts_as_approve: true
+  require_self_approval: false
```
mboersma: What behavior does this actually change? No "approve" label required?

CecileRobertMichon: `implicit_self_approve: true` is deprecated and replaced by `require_self_approve: false`. No behavior change compared to before. The author does not need to approve their own PR.

https://github.com/kubernetes/test-infra/blob/master/prow/plugins/config.go#L233

PR #1385, opened 2019-05-28
fix: ensure cluster-autoscaler image gets updated during upgrades

### LINKED ISSUE #1372: upgrade command doesn't update addon images
**Is this a request for help?**: Yes

---

---

**What version of aks-engine?**: 0.33.2

---

**Kubernetes version**: 1.13.5

**What happened**:
- existing cluster with aks-engine 0.33.2
- upgrading cluster with aks-engine 0.36.1
- addons section in api-model.json wasn't updated with the new version of the addons images included in 0.36.1 version 
- addons like cluster-autoscaler kept with the same old version "1.13.2" not the new one "1.13.4"

upgrade command should update all addons images

- existing cluster with aks-engine 0.33.2
- upgrade cluster with aks-engine 0.36.1

CecileRobertMichon:
**Reason for Change**:

- change cluster-autoscaler `addonmanager.kubernetes.io/mode`to `Reconcile` to ensure that cluster-autoscaler spec gets updated and use the latest yaml definition in `/etc/kubernetes/addons` after a cluster upgrade, including new cluster-autoscaler versions.
- Add unit test for addon default images with update to test that addon images get updated during upgrades and scales.

**Issue Fixed**:

Fixes #1372 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
+++ b/parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
@@ -5,7 +5,7 @@ metadata:
     k8s-addon: cluster-autoscaler.addons.k8s.io
     k8s-app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
   name: cluster-autoscaler
   namespace: kube-system
 ---
@@ -17,7 +17,7 @@ metadata:
     k8s-addon: cluster-autoscaler.addons.k8s.io
     k8s-app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 rules:
 - apiGroups: [""]
   resources: ["events","endpoints"]
@@ -78,7 +78,7 @@ metadata:
     k8s-addon: cluster-autoscaler.addons.k8s.io
     k8s-app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
@@ -97,7 +97,7 @@ metadata:
     k8s-addon: cluster-autoscaler.addons.k8s.io
     k8s-app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: Role
@@ -121,7 +121,7 @@ metadata:
   namespace: kube-system
   labels:
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
--- a/parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
+++ b/parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
@@ -129,7 +129,7 @@ metadata:
   labels:
     app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
   name: cluster-autoscaler
   namespace: kube-system
 spec:
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -284,6 +284,20 @@ func TestAssignDefaultAddonImages(t *testing.T) {
 			t.Errorf("expected setDefaults to set Image %s in addon %s, but got %s", expected, addon.Name, actual)
 		}
 	}
+
+	mockCS = getMockBaseContainerService("1.10.8")
+	mockCS.Properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	mockCS.Properties.OrchestratorProfile.KubernetesConfig.Addons = addons
+	mockCS.SetPropertiesDefaults(false, true)
+	modifiedAddons = mockCS.Properties.OrchestratorProfile.KubernetesConfig.Addons
+
+	for _, addon := range modifiedAddons {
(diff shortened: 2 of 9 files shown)


parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
```diff
@@ -129,7 +129,7 @@ metadata:
   labels:
     app: cluster-autoscaler
     kubernetes.io/cluster-service: "true"
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
```
CecileRobertMichon: @feiskyer any reasons that you know of why we would NOT want to do this?

pkg/api/defaults_test.go
```diff
@@ -249,6 +250,54 @@ func TestAssignDefaultAddonImages(t *testing.T) {
...
+		actual := addon.Containers[0].Image
+		if actual != expected {
+			t.Errorf("expected setDefaults to set Image %s in addon %s, but got %s", expected, addon.Name, actual)
+		}
+	}
+
+	// Image should be overridden in update scenarios.
+	mockCS = getMockBaseContainerService("1.10.8")
+	mockCS.Properties.OrchestratorProfile.OrchestratorType = Kubernetes
+	mockCS.Properties.OrchestratorProfile.KubernetesConfig.Addons = addons
```
jackfrancis: Can we make the `addons` object here a singleton, like the return value from a func? I'm a little concerned that as we progress through these tests we're overriding this reference unintentionally. I assume there is meant to be a sort of immutable canonical "baseline default addons spec" that we want to validate against (in the case of create, we want to validate that we can overwrite it, in the case of upgrade, we want to validate that overwriting it has no effect)

CecileRobertMichon: I'm not sure I understand your comment. https://github.com/Azure/aks-engine/pull/1385/files#diff-daf624fb49e7cf74e670db52d41e949aL214 is where addons is defined.

Are you talking about `mockCS.Properties.OrchestratorProfile.KubernetesConfig.Addons` or `addons`?

parts/k8s/containeraddons/kubernetesmasteraddons-cluster-autoscaler-deployment.yaml
feiskyer: With EnsureExists, users could update the config options later, e.g. change the node pools and scan intervals.

CecileRobertMichon: Users will need to change the spec on the disk, ie. in /etc/kubernetes/addons in order to change the config. If this is set to EnsureExists the spec won't get updated during upgrades unless the deployment is deleted.

jackfrancis: To speak to @feiskyer's point:

It's true that the current cluster-autoscaler spec/implementation is artificially limited to "VMSS" + "one VMSS pool" (the first pool, to be specific). So, to use a concrete example, if I build a VMSS AKS Engine cluster with 3 pools, and I enable the cluster-autoscaler addon, I will have to (after the cluster is deployed) manually re-configure my cluster-autoscaler Deployment spec to add the add'l 2 node pools. So, changing this to `Reconcile` means that when the user upgrades his/her cluster that custom configuration will be overridden, and effectively those other 2 pools won't be enabled for cluster-autoscaler.

However, I think the root cause of this issue is the artificial limitation of the current cluster-autoscaler addon. It's not too much work to make it generally useful for all VMSS pools in the api model.

So, my proposal is that we move this to Reconcile to address the issue of the autoscaler implementation (especially the deployment image) becoming unexpectedly stale across cluster upgrades. And also, to add validation to ensure that customers with > 1 VMSS node pool aren't allowed to use this addon. In practice, there isn't much value here for those folks anyways compared to following the online docs (IMO).

And then medium term we actually make this spec useful for clusters with arbitrary numbers of node pools.

@CecileRobertMichon @mboersma thoughts?

mo-saeed: @jackfrancis I think cluster-autoscaler shouldn't be an addon in the current state, in our cluster we disabled the addon and used helm to install the cluster-autoscaler as we need always to add/modify some custom configs (e.g. scale down delay interval). if we use the addon then we need will every time to change the addon file on the master node which is not a good way of managing the configuration changes.
 
I would say addons should be only for components which rarely need configuration changes from the user, like azure npm, cni, etc..

jackfrancis: @mo-saeed I agree 100%. I do, however, think there is *some* convenience value in automatically enabling various componentry (e.g., cluster-autoscaler) with a simple api model flag. Which componentry we choose to enable as addons (which as you point out does not expose any production-useful configuration interfaces like helm) should really be up to our users to decide. (especially users who aren't sophisticated enough to curate their own helm charts/repositories)

But yeah, if more folks tell us that they don't use this addon then I agree we should move our investments along this vector elsewhere.

CecileRobertMichon: It shouldn't be too hard to evolve the addon to allow for more than one scale set (and also make that configurable potentially). 

Agreed with the short term fix plan of setting to `Reconcile` (hence the PR :) ) . However, I'm not sure we need to add validation to disallow more that one scale set as it might be useful to keep that as a possibility (for testing or for quickly deploying the autoscaler and then changing its config on the cluster) in the meantime. Our docs do mention that this addon shouldn't be used with more than one agent pool. https://github.com/Azure/aks-engine/blob/master/examples/addons/cluster-autoscaler/README.md#L16

CecileRobertMichon: btw in case you don't already know this @mo-saeed (you probably do), the cluster-autoscaler addon can be configured to change scan-interval, etc.:

```
"addons": [
          {
            "name": "cluster-autoscaler",
            "enabled": true,
            "config": {
              "min-nodes": "1",
              "max-nodes": "5",
              "scan-interval": "10s"
            }
          }
        ]
```

mo-saeed: @CecileRobertMichon Thank you ! that was just an example and it's nice to have this configurable but my point is configuration management after first installation, which i would love some day to be supported by aks-engine command (sth like `aks-engine update`. for example changing api-server, kubelet or even addons flags/config so i don't need to go the master nodes and change files manually which can lead sometimes to problems.
I also understand @jackfrancis and your point for the value of having it as addon for some users :)

CecileRobertMichon: Definitely, agree 💯 %

Issue #1084, opened 2019-04-17
Deterministic Windows VHD


craiglpeters:
**Is this a request for help?**: No

---

**Is this an ISSUE or FEATURE REQUEST?** (choose one): Feature Request

---

**What version of aks-engine?**: Going forward

---

**Kubernetes version**: 1.14 and later

**What happened**:
Build Windows VHDs with

- [x] Windows Server 2019 and latest cumulative update
- [x] (optional) Additional KBs installed
- [ ] container runtime(s)
- [ ] Kubernetes
- [ ] CNI 
- [ ] Log collection scripts

preinstalled for faster deployments. Need to create a process for managing the creation of VHD for supported Windows Server releases, and curating the configuration and patches that corresponds to the way AKS-Engine does this for Linux.

If log paths change, update https://github.com/Azure/azure-diskinspect-service/pull/95/files

**How to reproduce it** (as minimally and precisely as possible):
Pulling the latest Windows version isn't deterministic

**Anything else we need to know**:

jackfrancis:
I think we can close this. @marosset kindly re-open if I'm wrong.

marosset:
> I think we can close this. @marosset kindly re-open if I'm wrong.

You are not wrong. Thanks for doing some house-cleaning here.

PR #1386, opened 2019-05-28
Starting a design for custom Windows VHDs

PatrickLang:
**Reason for Change**:

This is capturing the requirements for an AKS-tailored Windows VHD into a doc. This is aiming to bring more determinism and speed to Windows deployments, and improve our ability to test new Windows  and container runtime patches and versions.

**Issue Fixed**:
Design for #1084

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/hold

### DIFF
--- a/docs/design/windows-images.md
+++ b/docs/design/windows-images.md
@@ -1,0 +1,164 @@
+# Building Azure marketplace images for Windows AKS support
+
+## Motivation
+
+Initially, AKS-Engine's codebase on Linux installed all of the needed stuff from package repos or container registries. This provided the best flexibility as any packages could be easily updated and tested with a simple generate+deploy. However, it had a few major downsides. Network or repo downtime would cause failed deployments that were not recoverable. It also took additional time to do the downloads, making the overall deployments slower. To mitigate this while preserving deployment flexibility, AKS-engine switched to using custom VM images [built with Packer](../../packer). 
+
+Windows is in a halfway state today were the Windows Server team publishes `Windows Server version ___ with Containers` that has Docker preinstalled and some container images `nanoserver` and `windowsservercore` pre-pulled. This helps save some deployment time, but means that making changes to it depends on another team, and not having automation makes it much more difficult to pre-test changes to what's installed in these images. We need the flexibility to validate new OS versions, patches, and other dependencies. Moving to an image-based solution would align with the Linux workflow and allow us to control and pre-test all changes. It also leaves the option to make other customizations specific to AKS that will help with reliability or deployment time.
+
+The other challenge with the Windows deployment workflow today is that all of the Kubernetes node binaries, CNI plugin, and default configuration files are in a single ZIP published to the acsmirror CDN. It's built + packaged in [build-windows-k8s.sh](../../scripts/build-windows-k8s.sh). This was written when the upstream kubernetes/kubernetes code was in an alpha state, downstream patches were required, and public builds were not yet available. This is no longer needed and makes it harder to revise and test those components individually.
+
+All of this needs to be cleaned up so that we can get avoid deployment time penalties due to downloads and keep all changes in dependencies in source code so that they can be thoroughly tested and reproduced. 
+
+## Use Cases
+
+In general, most code build for AKS-Engine deployments should go through a VHD-based workflow to match what will be used in production and avoid having extra implementations for the same scenario. Ones that are intentionally overlapping should be called out.
+
+### Testing new Windows versions - VHD
+
+Testing new Windows versions is particularly difficult for a few reasons:
+
+- Windows Server Insider builds are published frequently, but not all are available on Azure. Starting from an ISO+Packer works around that limitation.
+- The base containers are different -  `nanoserver-insider` vs `nanoserver`, and are incompatible with the released versions. This means that images Kubernetes depends on such as `pause` need to be built and used as well. You can't build these on an older OS, but they can be done within the VM that's running as part of the VHD build process.
+  - As an aside, matching test containers are required too. If these aren't built from Kubernetes official builds, they could be built and pushed to a private registry using the same VM. Leaving them on the VHD will actually improve test time by avoiding pulls later.
+
+With the VHD-based workflow, adding (or removing) support for a new version could be reduced to a few parameters: `WindowsIsoUrl`, `NanoServerImage`. If these are in a pull request, automation can run the full vhd build followed by AKS-Engine E2E test pass.
+
+If a private build needed to be tested, `WindowsIsoUrl` could be a path requiring an Azure login, and the resulting image could be kept in the private store for that subscription.
+
+
+### Testing new Windows patches - VHD
+
+This is similar, but simpler than the new Windows version case above. It could use a new `WindowsIsoUrl`, `WindowsPatchList`, or `WindowsPrivatePatchUrl`. There are no changes needed to the container images. 
+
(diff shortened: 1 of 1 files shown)


PatrickLang:
cc @craiglpeters @MikkelHegn @jsturtevant @CecileRobertMichon @vyta

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+### Testing new Windows patches - VHD
+
+This is similar, but simpler than the new Windows version case above. It could use a new `WindowsIsoUrl`, `WindowsPatchList`, or `WindowsPrivatePatchUrl`. There are no changes needed to the container images. 
+
+Teams needing to create and test prerelease patches will need to use a private branch and Azure DevOps Pipeline to avoid unintentional security disclosures. This should be kept in-sync with the public repo but use a different Azure subscription and accounts.
+
+### AKS-Users testing new Windows patches
+
+During the VHD build, `WindowsPatchList` and `WindowsPrivatePatchList` control what cumulative updates and/or patches are installed. However, there may be cases where a private patch has been given to a customer who cannot build a new VHD. The existing [windows-patches](C:\Users\patri\Source\aks-engine\extensions\windows-patches) extension should still work to do this.
+
```
MikkelHegn: Is this AKS-Engine users? I don't see how AKS users can test this without falling back to AKS-Engine.

MikkelHegn:
LGTM

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+The other challenge with the Windows deployment workflow today is that all of the Kubernetes node binaries, CNI plugin, and default configuration files are in a single ZIP published to the acsmirror CDN. It's built + packaged in [build-windows-k8s.sh](../../scripts/build-windows-k8s.sh). This was written when the upstream kubernetes/kubernetes code was in an alpha state, downstream patches were required, and public builds were not yet available. This is no longer needed and makes it harder to revise and test those components individually.
+
+All of this needs to be cleaned up so that we can get avoid deployment time penalties due to downloads and keep all changes in dependencies in source code so that they can be thoroughly tested and reproduced.
```
CecileRobertMichon: For Linux, our base image is not available in all sovereign clouds (I assume it will be the same for Windows). When modifying the code to work with the VHD, we should always maintain a working installation path for clusters not using the VHD (by choice or by necessity). Also, keeping an install path for all components allows for faster development speed because we don't need to wait for a new image to be published (which can take a few days) to be able to test changes to the download code flows (eg. new component versions).

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+These configurations drive the VHD build process. As such, they should be version controlled in source code for reproducability. They will be set as variables for a Packer build. Some of these have intentional overlap between the VHD build steps and node deployment to fulfill different use cases.
+
+1. (optional, only if installing from ISO) Run Packer with an unattend file to get a working Windows setup from an ISO image
+  * Install optional features: `Containers`, `Hyper-V`
+  * Enable remote access with WinRM & Remote Desktop
+  * Install the [Azure Agent](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-windows#manual-installation)
+  * Reboot
+1. Build a cache of all needed dependencies - Docker EE, Moby, ContainerD, Azure-CNI, SdnBridge, and Kubernetes
+1. Install the default version of Docker EE-basic, Moby, or ContainerD
+1. Pull container images needed such as core-dns, and also list of containers in `ContainerTags`
```
jsturtevant: Will we be able to build a cache for the base images like windows server core or other custom images via the `ContainerTags` flag?

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+`windowsProfile.windowsSku` | Old default: `Datacenter-Core-1809-with-Containers-smalldisk`. New values `Datacenter-Core-2019` (default), additional values: `Datacenter-Core-19H1`. **1809 will be ignored** as there is no reason to use it instead of 2019. All images will be 30Gb and expanded at deployment time if needed.
+`windowsProfile.imageVersion` | Default: `latest`. Specific values will follow format `YYYYMMDD.commithash`
+`orchestratorProfile.orchestratorRelease` + `orchestratorProfile.orchestratorVersion` | These together make up the Kubernetes version. Multiple releases+versions need to be cached on the node.
+`kubernetesConfig.containerRuntime` | Windows will need to support `docker` and `containerd` and download/cache each.
+`WindowsProfile.WindowsDockerVersion` | Version of Docker EE-basic to use on Windows
+`kubernetesConfig.mobyVersion` | For future use. Linux nodes use [moby](https://github.com/moby/moby) builds from `packages.microsoft.com`. If Windows builds are published too, they should be supported. As of May 2019, this isn't needed. Only Docker EE-basic and ContainerD are implemented.
```
jsturtevant: Should this be added now or when it is needed?

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+It will:
+
+1. Finish installing the desired Kubernetes, Moby/Docker/ContainerD, and network plugin versions. The user will provide these through the AKS-Engine [Cluster Definition](docs\topics\clusterdefinitions.md)
+2. Create or update configuration files for the above. Many of these have interdependencies based on file paths, binary names, or other configurations such as kubelet parameters that aren't known until deployment time.
+3. Start the necessary services
+4. Join the Kubernetes cluster
+
+## AKS-Engine Apimodel Changes
+
+These parameters are defined in [Cluster Definition](docs\topics\clusterdefinitions.md), and will have impact on what needs to be put in the Windows images and the configuration steps needed as they're deployed. Some of these are not used as-is, but instead are used to compute other values such as a path to a tarball on the acsmirror CDN. These ones should be checked for cached downloads before trying the acsmirror CDN since they may be in the VHD already.
```
jsturtevant: checking for the cached downloads before trying the mirror sounds like it could be complicated to debug if something when wrong.  Is it possible to say make parameters override (if provided)?

docs/design/windows-images.md
PatrickLang: exactly, that's the intent. if we get feedback that people also want something like aspnet, we can include it too.

docs/design/windows-images.md
PatrickLang: sorry yes that should be aks-engine

docs/design/windows-images.md
```diff
@@ -0,0 +1,164 @@
...
+### VHD Build
+
+These configurations drive the VHD build process. As such, they should be version controlled in source code for reproducability. They will be set as variables for a Packer build. Some of these have intentional overlap between the VHD build steps and node deployment to fulfill different use cases.
+
+1. (optional, only if installing from ISO) Run Packer with an unattend file to get a working Windows setup from an ISO image
+  * Install optional features: `Containers`, `Hyper-V`
+  * Enable remote access with WinRM & Remote Desktop
+  * Install the [Azure Agent](https://docs.microsoft.com/en-us/azure/virtual-machines/extensions/agent-windows#manual-installation)
+  * Reboot
+1. Build a cache of all needed dependencies - Docker EE, Moby, ContainerD, Azure-CNI, SdnBridge, and Kubernetes
```
rjaini: Since on Azure Stack we don't have Azure-CNI, We might need to package Kubenet/flannel.

PR #1390, opened 2019-05-29
fix: Update Kata download URL

### LINKED ISSUE #1387: Kata URL has changed and needs to be updated in the code

Yes

---

ISSUE
---

0.34.3
---

**Kubernetes version**:
1.14

**What happened**:
Not able to use kata-container runtime using aks-engine.

kata-container runtime should have been provisioned.

Just create a cluster using aks-engine and in the ARM template:
        "containerRuntime": "kata-containers",

Jukie:
**Reason for Change**:

Fixes #1387

**Issue Fixed**:

Fixes #1387

**Requirements**:

- [ x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ n/a] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -137,13 +137,15 @@ installMoby() {
 installKataContainersRuntime() {
     # TODO incorporate this into packer CI so that it is pre-baked into the VHD image
     echo "Adding Kata Containers repository key..."
+    ARCH=$(arch)
+    BRANCH=stable-1.7
     KATA_RELEASE_KEY_TMP=/tmp/kata-containers-release.key
-    KATA_URL=http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_${UBUNTU_RELEASE}/Release.key
+    KATA_URL=http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/${BRANCH}/xUbuntu_${UBUNTU_RELEASE}/Release.key
     retrycmd_if_failure_no_stats 120 5 25 curl -fsSL $KATA_URL > $KATA_RELEASE_KEY_TMP || exit $ERR_KATA_KEY_DOWNLOAD_TIMEOUT
     wait_for_apt_locks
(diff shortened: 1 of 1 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -137,13 +137,15 @@ installMoby() {
 installKataContainersRuntime() {
     # TODO incorporate this into packer CI so that it is pre-baked into the VHD image
     echo "Adding Kata Containers repository key..."
+    ARCH=x86_64
```
CecileRobertMichon: can we do `ARCH=$(arch)` to make this more reusable?

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -137,13 +137,15 @@ installMoby() {
...
+    BRANCH=stable-1.7
```
CecileRobertMichon: where does stable-1.7 come from?

CecileRobertMichon:
@Jukie thanks for the PR!

Could you please `git commit --amend` when you rebase so the commit message has the `fix: ` prefix?

fix: Update Kata download URL

Otherwise our release tools won't include this in the CHANGELOG.

parts/k8s/cloud-init/artifacts/cse_install.sh
Jukie: Good point, I'll update this now.

parts/k8s/cloud-init/artifacts/cse_install.sh
Jukie: From http://download.opensuse.org/repositories/home:/katacontainers:/releases:/x86_64:/
Is there a better place to track the version in use? I assumed pinning to master isn't ideal

CecileRobertMichon: nope, just curious. I agree, let's not pin to master :)

Jukie:
Sure, I had already amended it before your comment. Does it need to be lowercase as well?

CecileRobertMichon:
> Sure, I had already amended it before your comment. Does it need to be lowercase as well?

Yeah, it does. It should be `fix: `

Jukie:
Okay, please take another look when you can. Should be good to go

PR #1394, opened 2019-05-29
feat: add support for Kubernetes 1.15.0-beta.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1150-alpha3

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
~~Requires #1369 to merge before this will pass.~~ *Merged*
CoreDNS was bumped to 1.5.0, but I think this may require configuration changes. I'll take a look at that next.


mboersma:
I have the CoreDNS changes ready, but I'd like to separate that into a follow-up PR.

Issue #275, opened 2018-10-28
Heapster keeps restarting every minute in newly deployed k8s cluster


chreichert:
**Is this a request for help?**:
yes
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
ISSUE
---

**What version of acs-engine?**:
0.24.1
---

**Orchestrator and version (e.g. Kubernetes, DC/OS, Swarm)**
Kubernetes 1.10.9

**What happened**:
In a newly deployed k8s cluster the heapster component is is being redeployed every minute or so.

**What you expected to happen**:
Heapster running without problems.

**How to reproduce it** (as minimally and precisely as possible):
Initial setup of Cluster with acs-engine 0.24.1:

- Kubernetes 1.10.9
- Private cluster
- RBAC enabled
- Calico Networkpolicy
- Disable addons:
  - blobfuse-flexvolume
  - smb-flexvolume
- 1 Master
- 3 Agentpools as vmss

**Anything else we need to know**:
I found a question in Stackoverflow (https://stackoverflow.com/questions/46110856/kubernetes-monitoring-service-heapster-keeps-restarting) which describes the very similar behaviour. There a fix was suggested to modify the heapster deployment to contain a label "addonmanager.kubernetes.io/mode: EnsureExists".

I edited the deployment manifest for heapster on all master nodes and changed that label for the heapster deployment from "Reconcile" to "EnsureExists", which fixed the problem for me.

Doing some research, I found the following PR https://github.com/Azure/acs-engine/pull/3918, which changed that label from "EnsureExists" to "Reconcile", so you might check this PR again? See also PR https://github.com/Azure/acs-engine/pull/1133, which introduced that label and was referenced in that above recommended fix.

PR #1401, opened 2019-05-29
fix: use reconcile mode in addon specs

CecileRobertMichon:
**Reason for Change**:

Addon specs delivered via aks-engine should use `Reconcile` mode in order for clusters to use the latest spec after an upgrade or a scale (ie. whenever the addon spec gets rewritten on disk in /etc/kubernetes/addons). Otherwise, new values in the addon spec such as new image versions won't be applied after an upgrade.

**Issue Fixed**:

Related to #1372 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Omitted Heapster purposefully because of https://github.com/Azure/aks-engine/issues/275

### DIFF
--- a/parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
+++ b/parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
@@ -52,7 +52,7 @@ metadata:
   name: kube-dns
   namespace: kube-system
   labels:
-    addonmanager.kubernetes.io/mode: Reconcile
+    addonmanager.kubernetes.io/mode: EnsureExists
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
--- a/parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
+++ b/parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
@@ -52,7 +52,7 @@ metadata:
   name: kube-dns
   namespace: kube-system
   labels:
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
 ---
 apiVersion: extensions/v1beta1
 kind: Deployment
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -56,7 +56,7 @@ metadata:
   name: coredns
   namespace: kube-system
   labels:
-      addonmanager.kubernetes.io/mode: Reconcile
+      addonmanager.kubernetes.io/mode: EnsureExists
 data:
   Corefile: |
     .:53 {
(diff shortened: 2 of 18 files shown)


parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
```diff
@@ -52,7 +52,7 @@ metadata:
   name: kube-dns
   namespace: kube-system
   labels:
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
```
jackfrancis: I think we probably want to keep this as `EnsureExists` as a `Reconcile` might blow away a pre-existing `ConfigMap` with data in it.

Ref:

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/kube-dns/kube-dns.yaml.in

parts/k8s/addons/coredns.yaml
```diff
@@ -35,7 +35,7 @@ metadata:
     rbac.authorization.kubernetes.io/autoupdate: "true"
   labels:
     kubernetes.io/bootstrapping: rbac-defaults
-    addonmanager.kubernetes.io/mode: EnsureExists
```
jackfrancis: So, same comment about `ConfigMap` resources everywhere, but I'm torn on this particular `ClusterRoleBinding` as it is set to `EnsureExists` in the reference yaml:

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dns/coredns/coredns.yaml.in

But I don't know why you wouldn't want to set this to `Reconcile` to allow the spec to adapt over time. @lachie83 @khenidak Any idea why you would bootstrap an addon spec w/ a `ClusterRoleBinding` set to `EnsureExists`? (That means essentially that the next time the addon manager picks up a change to this resource from `/etc/kubernetes/addons/*` it will not automatically reconcile it in the runtime)

parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
CecileRobertMichon: updated all ConfigMaps to use `EnsureExists`

parts/k8s/addons/coredns.yaml
CecileRobertMichon: I left this as `EnsureExists` for now

parts/k8s/addons/1.10/kubernetesmasteraddons-kube-dns-deployment.yaml
CecileRobertMichon: I'm torn about the omsagent ConfigMap since it data that is part of the ConfigMap resource definition, wondering I'm thinking we want to leave it to `Reconcile` so changes to that data get picked up by updates (it's currently set to `Reconcile`): 

https://github.com/Azure/aks-engine/blob/403a11c1ea84c0fbadeb07e81b1a7c97d500fd59/parts/k8s/containeraddons/kubernetesmasteraddons-omsagent-daemonset.yaml#L51

jackfrancis:
lgtm

PR #1402, opened 2019-05-29
fix: prune addon defaults if the addon is disabled

CecileRobertMichon:
**Reason for Change**:

Currently, we only set default addon values for addons when they are enabled but we still populate the defaults if the addon itself was omitted from the apimodel and is disabled by default. Although this has no functional impact, it is counter intuitive. As a user, when upgrading I would expect my addon Images to be refreshed, even if an addon is disabled or wasn't included in my input apimodel (since it will be included in the expanded apimodel). See #1372 for more details.

Steps to repro
- Deploy a cluster without specifying an addon that is disabled by default (eg. cluster-autoscaler)
- Note that the addon is added to the Addons section of the generated apimodel with `Enabled: false`.
- Upgrade the cluster with a more recent aks-engine version where the default addon image has changed.
The apimodel after the upgrade will still have the old image. This is purely cosmetic and has no effect on the cluster functionality since the addon is disabled. However, it might be confusing to users getting different images on different clusters using the same aks-engine version (one with the addon enabled, the other with the same addon disabled).

Before:
```
          {
            "name": "cluster-autoscaler",
            "enabled": false,
            "containers": [
              {
                "name": "cluster-autoscaler",
                "image": "k8s.gcr.io/cluster-autoscaler:v1.12.3",
                "cpuRequests": "100m",
                "memoryRequests": "300Mi",
                "cpuLimits": "100m",
                "memoryLimits": "300Mi"
              }
            ],
            "config": {
              "max-nodes": "5",
              "min-nodes": "1",
              "scan-interval": "10s"
            }
          }
```

After:

```
          {
            "name": "cluster-autoscaler",
            "enabled": false
          }
```

**Issue Fixed**:

Related to #1372 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -352,6 +352,12 @@ func assignDefaultAddonVals(addon, defaults KubernetesAddon, isUpdate bool) Kube
 	if addon.Enabled == nil {
 		addon.Enabled = defaults.Enabled
 	}
+	if !to.Bool(addon.Enabled) {
+		return KubernetesAddon{
+			Name:    addon.Name,
+			Enabled: addon.Enabled,
+		}
+	}
 	for i := range defaults.Containers {
 		c := addon.GetAddonContainersIndexByName(defaults.Containers[i].Name)
 		if c < 0 {
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -316,7 +316,7 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 	}
 
 	for _, addon := range defaultAddons {
-		synthesizeAddonsConfig(o.KubernetesConfig.Addons, addon, false, isUpdate)
+		synthesizeAddonsConfig(o.KubernetesConfig.Addons, addon, isUpdate)
 	}
 
 	if len(o.KubernetesConfig.PodSecurityPolicyConfig) > 0 && isUpdate {
@@ -385,12 +385,10 @@ func assignDefaultAddonVals(addon, defaults KubernetesAddon, isUpdate bool) Kube
 	return addon
 }
 
-func synthesizeAddonsConfig(addons []KubernetesAddon, addon KubernetesAddon, enableIfNil bool, isUpdate bool) {
+func synthesizeAddonsConfig(addons []KubernetesAddon, addon KubernetesAddon, isUpdate bool) {
 	i := getAddonsIndexByName(addons, addon.Name)
 	if i >= 0 {
-		if addons[i].IsEnabled(enableIfNil) {
-			addons[i] = assignDefaultAddonVals(addons[i], addon, isUpdate)
-		}
+		addons[i] = assignDefaultAddonVals(addons[i], addon, isUpdate)
 	}
 }
 
--- a/pkg/api/addons_test.go
+++ b/pkg/api/addons_test.go
@@ -4,6 +4,7 @@
 package api
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/Azure/go-autorest/autorest/to"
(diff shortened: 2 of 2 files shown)


CecileRobertMichon:
@mo-saeed FYI

jackfrancis:
Just to offer a counter-idea: what about pruning the `containers` array and/or `config` object entirely from the finally rendered apimodel for disabled addons? So basically, instead of this:

```
          {
            "name": "cluster-autoscaler",
            "enabled": false,
            "containers": [
              {
                "name": "cluster-autoscaler",
                "image": "k8s.gcr.io/cluster-autoscaler:v1.12.3",
                "cpuRequests": "100m",
                "memoryRequests": "300Mi",
                "cpuLimits": "100m",
                "memoryLimits": "300Mi"
              }
            ],
            "config": {
              "max-nodes": "5",
              "min-nodes": "1",
              "scan-interval": "10s"
            }
          }
```

We render this:

```
          {
            "name": "cluster-autoscaler",
            "enabled": false
          }
```

CecileRobertMichon:
@jackfrancis I like that idea... it makes the apimodel nicer to look at :) 

Only thing is we need to make sure that there wouldn't be side effects, for example: some part of the code using the config to generate templates even for disabled addons (if that's the case it can be fixed), or upgrade not writing the config if the addon becomes enabled by default. I'm just think "out loud" here, these are all fixable if needed.

jackfrancis:
@CecileRobertMichon I think we can encapsulate/isolate the pruning during the addons composition business logic itself. I'm not sure if the current implementation supports just a few lines of code, but basically the output addons object should be a thing that we conditionally construct in an additive way, one addon at a time, only if the addon is enabled.

CecileRobertMichon:
@jackfrancis updated according to your suggestion, let me know what you think

pkg/api/addons.go
```diff
@@ -385,12 +391,10 @@ func assignDefaultAddonVals(addon, defaults KubernetesAddon, isUpdate bool) Kube
 	return addon
 }
 
-func synthesizeAddonsConfig(addons []KubernetesAddon, addon KubernetesAddon, enableIfNil bool, isUpdate bool) {
```
jackfrancis: I supposed `enableIfNil` was once used but is leftover cruft at this point (being that is only called above w/ static `false`.

pkg/api/addons.go
```diff
@@ -352,6 +352,12 @@ func assignDefaultAddonVals(addon, defaults KubernetesAddon, isUpdate bool) Kube
 	if addon.Enabled == nil {
 		addon.Enabled = defaults.Enabled
 	}
+	if !to.Bool(addon.Enabled) {
+		return KubernetesAddon{
```
jackfrancis: Should we add something like this:

```
if addon.Containers != nil || addon.Config != nil || addon.Data != nil {
    log.Warn("configuration data for disabled addon <addon> will be ignored and removed from generated api model")
}
```

Something like that?

CecileRobertMichon: hmm the thing is that at this point the config/containers/etc won't be nil if the addon wasn't specified by the user in their apimodel. ie. If I omit autoscaler from my apimodel, the `appendAddonIfNotPresent` function will append the whole addon default struct. This means that the warning will be logged for every single addon that was omitted. The only way that warning message would be accurate would be if I specified autoscaler in my apimodel but set it to disabled. That or we need to refactor the way default addons are set when they are omitted from the apimodel which IMO is more effort than this small UX improvement was meant to take.

jackfrancis: Got it

Issue #1318, opened 2019-05-16
generated code files are falling out of sync


mboersma:
**Is this an ISSUE or FEATURE REQUEST?** (choose one):

ISSUE

---

**What version of aks-engine?**:

master

---

**Kubernetes version**:

n/a

**What happened**:

PRs that changed files involved in code generation still pass the `make ensure-generated` task in Azure DevOps CI, and therefore `templates_generated.go` falls out of sync with the sources used to generate it.

See #1088, #1295, #1317

**What you expected to happen**:

The CI pipeline should have failed until `templates_generated.go` was updated to match the changes in the PR.

**How to reproduce it** (as minimally and precisely as possible):

Create a PR that changes scripts under `/parts` and ensure they imply a change to generated code by running `make generate`. Then check in the non-generated code and create a PR and watch the `Check if generated code is up to date` task to see it (incorrectly) succeed.

**Anything else we need to know**:

cc: @andyliuliming

PR #1406, opened 2019-05-29
debugging 'make ensure-generated' false positives in CI

mboersma:
**Reason for Change**:
A way to test why `make ensure-generated` succeeds when it shouldn't in CI, as described in #1318.

It uses [`shfmt`](https://github.com/mvdan/sh) to raise an error if shell scripts aren't formatted consistently, and reformatting all the shell scripts causes a massive and obvious delta to `templates_generated.go` which *should* fail the PR since that's not included here.

```shell
$ make ensure-generated
...
< echo "<searchDomainRealmPassword>" | realm join -U <searchDomainRealmUser>@$(echo "<searchDomainName>" | tr /a-z/ /A-Z/) $(echo "<searchDomainName>" | tr /a-z/ /A-Z/)
---
> echo "<searchDomainRealmPassword>" | realm join -U $(echo "<searchDomainName>" | tr /a-z/ /A-Z/) <searchDomainRealmUser >@$(echo "<searchDomainName>" | tr /a-z/ /A-Z/)
+ echo 'go generate produced changes that were not already present.'
go generate produced changes that were not already present.
+ echo 'Make sure you include generated assets in your commit.'
Make sure you include generated assets in your commit.
+ exit 1
+ rm -rf /var/folders/tq/2cdy0tqn6g556ny243rx3p840000gn/T/tmp.M5TYhoHW
make: *** [ensure-generated] Error 1
```

(We could merge the `shfmt` change if we like it, but that's not the current goal.)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Both the `ensure-generated.sh` script and my attempt at a `git diff` replacement **succeeded** in the `make ensure-generated` step in Azure DevOps. Both strategies correctly fail when run locally against this commit. I'm not seeing what the difference is.

CecileRobertMichon:
@mboersma could it have anything to do with how devops handles checking out the changes? Does it merge the changes with master before running the build?

mboersma:
It does merge with master before running the build. But even in that case, `templates_generated.go` wouldn't be up-to-date so `make ensure-generated` should fail...unless I'm confused, which is entirely possible.

mboersma:
This served its purpose: to highlight the underlying problem in #1318 and to work out the fix in #1501.

PR #949, opened 2019-04-04
fix: mount tmp emptyDir to stop CoreDNS restarts

mboersma:
**Reason for Change**:

Mounts an emptyDir at /tmp in the coredns container and reenables the dns-liveness test, in hopes that this works around a container restart issue as suggested [here](https://github.com/kubernetes/kubernetes/issues/75414#issuecomment-479905336).

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1411, opened 2019-05-30
chore: use CoreDNS 1.5.0 for k8s >= 1.12.0

mboersma:
**Reason for Change**:
This follows up on #1394 to upgrade CoreDNS as specified in Kubernetes 1.15.0-beta.1 [release notes](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1150-alpha3).

**Issue Fixed**:
See `/tmp` dir workaround in #949, which this reverts.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I copied `coredns.yaml` into a 1.15 directory and then updated it based on all recent changes in the [reference manifest](https://github.com/coredns/deployment/blob/master/kubernetes/coredns.yaml.sed).


jackfrancis:
@mboersma Strictly speaking this doesn't revert that change, as < 1.15 clusters using coredns will still get it (and they should, right?)

We could consider backporting this new version of coredns on older versions of k8s if that's what coredns says we should do...

jackfrancis:
@mboersma Sorry I didn't notice this earlier, but this new 1.15-specific configuration should go in a new file under a new `parts/k8s/addons/1.15/` directory. Unless we're sure the changes to the yaml spec are recommended for all versions of coredns that we deliver across the k8s versions that use coredns.

mboersma:
@jackfrancis that's exactly how the PR was until I acted on your comment and backported CoreDNS 1.15 to all supported k8s versions. I can drop that commit if we want to be conservative and just make this change for k8s 1.15. The project guidance wasn't very clear: nothing like a support matrix.

jackfrancis:
@mboersma Sorry for the confusion. On your guidance this PR looks good as an implementation of "update coredns to 1.5 for all versions of k8s".

If that's what we want lgtm

mboersma:
> if that's what we want

I think so. DNS is one of the truly pluggable components in the system, it passes tests, and upgrading across the board keeps things simpler. I can't find any guidance against it, so 🤞.

parts/k8s/addons/coredns.yaml
```diff
@@ -35,7 +39,7 @@ metadata:
     rbac.authorization.kubernetes.io/autoupdate: "true"
   labels:
     kubernetes.io/bootstrapping: rbac-defaults
-    addonmanager.kubernetes.io/mode: EnsureExists
+    addonmanager.kubernetes.io/mode: Reconcile
```
CecileRobertMichon: I think we might want to keep this as EnsureExists, see https://github.com/Azure/aks-engine/pull/1401/files/37bb24f4666a6d4e4db36edb28dc29027dcade8d#r288814213

parts/k8s/addons/coredns.yaml
```diff
@@ -52,35 +56,35 @@ metadata:
   name: coredns
   namespace: kube-system
   labels:
-      addonmanager.kubernetes.io/mode: EnsureExists
+      addonmanager.kubernetes.io/mode: Reconcile
```
CecileRobertMichon: same here https://github.com/Azure/aks-engine/pull/1401/files/37bb24f4666a6d4e4db36edb28dc29027dcade8d#r288813426

PR #1416, opened 2019-05-31
chore(CIS): generate kubelet server cert in aks-engine

andyzhangx:
**Reason for Change**:

This PR fixed the following CIS failure:
```
[FAIL] 2.1.10 Ensure that the --tls-cert-file and --tls-private-key-file arguments are set as appropriate (Scored)

== Remediations ==
2.1.10 If using a Kubelet config file, edit the file to set tlsCertFile to the location of the certificate
file to use to identify this Kubelet, and tlsPrivateKeyFile to the location of the
corresponding private key file.
/etc/systemd/system/kubelet.service.d/10-kubeadm.conf on each worker node and
set the below parameters in KUBELET_CERTIFICATE_ARGS variable.
--tls-cert-file=<path/to/tls-certificate-file>
file=<path/to/tls-key-file>
systemctl daemon-reload
```

If `--tls-cert-file` and `--tls-private-key-file` parameters are not set, kubelet will generate self signed cert with one year valid, this PR generate the certs in aks-engine node, would be only applied in k8s v1.15.0+

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -143,6 +143,14 @@ generateAggregatedAPICerts() {
     $AGGREGATED_API_CERTS_SETUP_FILE
 }
 
+configureKubeletServerCert() {
+    KUBELET_SERVER_PRIVATE_KEY_PATH="/etc/kubernetes/certs/kubeletserver.key"
+    KUBELET_SERVER_CERT_PATH="/etc/kubernetes/certs/kubeletserver.crt"
+
+    openssl genrsa -out $KUBELET_SERVER_PRIVATE_KEY_PATH 2048
+    openssl req -new -x509 -days 7300 -key $KUBELET_SERVER_PRIVATE_KEY_PATH -out $KUBELET_SERVER_CERT_PATH -subj "/CN=${NODE_NAME}"
+}
+
 configureK8s() {
     KUBELET_PRIVATE_KEY_PATH="/etc/kubernetes/certs/client.key"
     touch "${KUBELET_PRIVATE_KEY_PATH}"
@@ -207,6 +215,8 @@ EOF
             generateAggregatedAPICerts
         fi
     fi
+
+    configureKubeletServerCert
 }
 
 configureCNI() {
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -25,13 +25,15 @@ func (cs *ContainerService) setKubeletConfig() {
 		"--cgroups-per-qos":             "true",
 		"--kubeconfig":                  "/var/lib/kubelet/kubeconfig",
 		"--keep-terminated-pod-volumes": "false",
+		"--tls-cert-file":               "/etc/kubernetes/certs/kubeletserver.crt",
+		"--tls-private-key-file":        "/etc/kubernetes/certs/kubeletserver.key",
 	}
 
 	// Start with copy of Linux config
 	staticWindowsKubeletConfig := make(map[string]string)
 	for key, val := range staticLinuxKubeletConfig {
 		switch key {
-		case "--pod-manifest-path": // Don't add Linux-specific config
+		case "--pod-manifest-path", "--tls-cert-file", "--tls-private-key-file": // Don't add Linux-specific config
 			staticWindowsKubeletConfig[key] = ""
 		case "--anonymous-auth", "--client-ca-file":
 			if !to.Bool(o.KubernetesConfig.EnableSecureKubelet) { // Don't add if EnableSecureKubelet is disabled
--- a/pkg/api/defaults-kubelet_test.go
+++ b/pkg/api/defaults-kubelet_test.go
@@ -54,6 +54,8 @@ func TestKubeletConfigDefaults(t *testing.T) {
 		"--streaming-connection-idle-timeout": "5m",
 		"--feature-gates":                     "PodPriority=true,RotateKubeletServerCertificate=true",
 		"--tls-cipher-suites":                 TLSStrongCipherSuitesKubelet,
+		"--tls-cert-file":                     "/etc/kubernetes/certs/kubeletserver.crt",
+		"--tls-private-key-file":              "/etc/kubernetes/certs/kubeletserver.key",
 	}
 	for key, val := range kubeletConfig {
 		if expected[key] != val {
@@ -90,6 +92,8 @@ func TestKubeletConfigDefaults(t *testing.T) {
 	expected["--eviction-hard"] = "\"\"\"\""
 	delete(expected, "--pod-manifest-path")
 	delete(expected, "--protect-kernel-defaults")
+	delete(expected, "--tls-cert-file")
+	delete(expected, "--tls-private-key-file")
 	for key, val := range windowsProfileKubeletConfig {
 		if expected[key] != val {
 			t.Fatalf("got unexpected Windows agent profile kubelet config value for %s: %s, expected %s",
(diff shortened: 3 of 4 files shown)


pkg/api/defaults-kubelet_test.go
```diff
@@ -145,6 +145,17 @@ func TestKubeletConfigDefaultsRemovals(t *testing.T) {
 				key)
 		}
 	}
+
+	expected := map[string]string{
```
CecileRobertMichon: let's also add a check to make sure that these values aren't set for k8s < 1.15.0-beta.1. You can use the cs above since it's 1.12 https://github.com/Azure/aks-engine/pull/1416/files#diff-c116fab93edb004accebd12e5742e80cR116.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -143,6 +143,14 @@ generateAggregatedAPICerts() {
     $AGGREGATED_API_CERTS_SETUP_FILE
 }
 
+configureKubeletServerCert() {
+    KUBELET_SERVER_PRIVATE_KEY_PATH="/etc/kubernetes/certs/kubeletserver.key"
```
CecileRobertMichon: this will get written for every k8s version even if it only gets used only for k8s 1.15+? Is there a good reason we'd want to only generate the certs for clusters 1.15 and up?

andyzhangx: the `--tls-cert-file and --tls-private-key-file arguments` should be used in all kubelet versions, I have changed the PR, PTAL.

andyzhangx:
@CecileRobertMichon there are one e2e case failed on windows 1.13 and 1.15, it looks like there is outbound connection, not related to this PR:

```
• Failure [436.164 seconds]‌
&#x2F;__w&#x2F;1&#x2F;s&#x2F;gopath&#x2F;src&#x2F;github.com&#x2F;Azure&#x2F;aks-engine&#x2F;test&#x2F;e2e&#x2F;kubernetes&#x2F;kubernetes_test.go:94‌
  with a windows agent pool
  ‌&#x2F;__w&#x2F;1&#x2F;s&#x2F;gopath&#x2F;src&#x2F;github.com&#x2F;Azure&#x2F;aks-engine&#x2F;test&#x2F;e2e&#x2F;kubernetes&#x2F;kubernetes_test.go:1308‌
    ‌should be able to deploy and scale an iis webserver [It]‌
    ‌&#x2F;__w&#x2F;1&#x2F;s&#x2F;gopath&#x2F;src&#x2F;github.com&#x2F;Azure&#x2F;aks-engine&#x2F;test&#x2F;e2e&#x2F;kubernetes&#x2F;kubernetes_test.go:1309‌

    ‌Expected error:‌
        <*errors.fundamental | 0xc000579b80>: {
            msg: "Timeout exceeded (1m0s) while waiting for Pod (iis-kubernetes-eastus-88833-96633-5bf4fcb86c-msqlx) to check outbound internet connection",
            stack: [0x9cfd44, 0x45e8c1],
        }
        Timeout exceeded (1m0s) while waiting for Pod (iis-kubernetes-eastus-88833-96633-5bf4fcb86c-msqlx) to check outbound internet connection
    not to have occurred‌

    /__w/1/s/gopath/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:1372
------------------------------‌
S‌S‌S‌S‌S‌S‌

Summarizing 1 Failure:‌

[Fail] ‌Azure Container Cluster using the Kubernetes Orchestrator ‌with a windows agent pool ‌[It] should be able to deploy and scale an iis webserver ‌
&#x2F;__w&#x2F;1&#x2F;s&#x2F;gopath&#x2F;src&#x2F;github.com&#x2F;Azure&#x2F;aks-engine&#x2F;test&#x2F;e2e&#x2F;kubernetes&#x2F;kubernetes_test.go:1372‌

Ran 25 of 44 Specs in 1587.467 seconds‌
FAIL!‌ -- ‌24 Passed‌ | ‌1 Failed‌ | ‌0 Pending‌ | ‌19 Skipped‌
--- FAIL: TestKubernetes (1587.47s)
FAIL
```

parts/k8s/cloud-init/artifacts/cse_config.sh
CecileRobertMichon: is it supported for all k8s versions?

andyzhangx: yes, here is the k8s v1.7 parameters: https://github.com/kubernetes/kubernetes/blob/release-1.7/cmd/kubelet/app/options/options.go#L185-L189

pkg/api/defaults-kubelet.go
```diff
@@ -199,6 +201,12 @@ func removeKubeletFlags(k map[string]string, v string) {
 			delete(k, key)
 		}
 	}
+
+	if strings.ToLower(osType) == "windows" {
```
CecileRobertMichon: let's make this the `Windows` constant

pkg/api/defaults-kubelet.go
```diff
@@ -126,7 +128,7 @@ func (cs *ContainerService) setKubeletConfig() {
 		}
 	}
 
-	removeKubeletFlags(o.KubernetesConfig.KubeletConfig, o.OrchestratorVersion)
+	removeKubeletFlags(o.KubernetesConfig.KubeletConfig, o.OrchestratorVersion, "linux")
```
CecileRobertMichon: let's make this the `Linux` constant

pkg/api/defaults-kubelet.go
```diff
@@ -137,7 +139,7 @@ func (cs *ContainerService) setKubeletConfig() {
 		setMissingKubeletValues(cs.Properties.MasterProfile.KubernetesConfig, o.KubernetesConfig.KubeletConfig)
 		addDefaultFeatureGates(cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig, o.OrchestratorVersion, "", "")
 
-		removeKubeletFlags(cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig, o.OrchestratorVersion)
+		removeKubeletFlags(cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig, o.OrchestratorVersion, "linux")
```
CecileRobertMichon: same here

pkg/api/defaults-kubelet.go
```diff
@@ -199,6 +201,12 @@ func removeKubeletFlags(k map[string]string, v string) {
 			delete(k, key)
 		}
 	}
+
+	if strings.ToLower(osType) == "windows" {
+		for _, key := range []string{"--tls-cert-file", "--tls-private-key-file"} {
```
CecileRobertMichon: instead of doing this, you could also add these values as empty `""` to `staticWindowsKubeletConfig ` like we do for  `--pod-manifest-path` and they would get deleted by https://github.com/Azure/aks-engine/pull/1416/files#diff-ac6b80e7a914ca77b31a38a10c77826bR198

pkg/api/defaults-kubelet.go
andyzhangx: fixed

pkg/api/defaults-kubelet.go
andyzhangx: fixed

pkg/api/defaults-kubelet.go
andyzhangx: that's a good suggestion, it would make the code logic more clear, I have followed your guide, thanks, PTAL.

andyzhangx:
@CecileRobertMichon rebased, PTAL

PR #1420, opened 2019-05-31
revert: add support for Kubernetes 1.12.9

CecileRobertMichon:
This reverts commit 23b016462f26a0d02f7fed07a2d63d067022e18b.

**Reason for Change**:

Upgrade from 1.12.9 clusters is broken. Revert until we understand what's going on.

**Issue Fixed**:

addresses #1421 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/versions.go
```diff
@@ -104,7 +104,6 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.12.6":         false,
 	"1.12.7":         true,
 	"1.12.8":         true,
-	"1.12.9":         true,
```
jackfrancis: Can we re-add it back in as explicit `"false", and with a comment as to why?

CecileRobertMichon: this was never released so a simple commit revert made most sense to me

jackfrancis: I kind of mean as a reminder to ourselves so when 1.12.10 arrives we run extra upgrade validation on it to see if things have been fixed

CecileRobertMichon: how about we keep track of that in an issue? we can put it in the backlog too so we don't forget

CecileRobertMichon: we also need to do more investigation to understand what's going on... it's still unclear what the exact root cause is.

jackfrancis: Let's do both!

PR #1424, opened 2019-06-03
feat: IPv6dualStack feature flag

aramase:
**Reason for Change**:

Enables testing of clusters with ipv6 dual stack feature. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@aramase Does this PR work given the fact that AKS Engine node (including master nodes) VMs ship with this setting?:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/default-grub#L13

parts/k8s/cloud-init/artifacts/enable-dhcpv6.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
```
khenidak: set -e
set -o pipefail
set -u

pkg/api/const.go
```diff
@@ -326,13 +326,18 @@ const (
...
+	DefaultKubernetesMasterSubnetIPv6 = "2001:1234:5678:9abc::/64"
 	// DefaultAgentSubnetTemplate specifies a default agent subnet
 	DefaultAgentSubnetTemplate = "10.%d.0.0/16"
 	// DefaultKubernetesSubnet specifies the default subnet used for all masters, agents and pods
 	// when VNET integration is enabled.
 	DefaultKubernetesSubnet = "10.240.0.0/12"
 	// DefaultVNETCIDR is the default CIDR block for the VNET
 	DefaultVNETCIDR = "10.0.0.0/8"
+	// DefaultVNETCIDRIPv6 is the default IPv6 CIDR block for the VNET
+	DefaultVNETCIDRIPv6 = "2001:1234:5678:9abc::/64"
```
khenidak: shouldn't `DefaultKubernetesMasterSubnetIPv6` and `DefaultVNETCIDRIPv6` be different?

pkg/api/const.go
```diff
@@ -414,6 +419,8 @@ const (
 	DefaultKubernetesAgentSubnetVMSS = "10.248.0.0/13"
 	// DefaultKubernetesClusterSubnet specifies the default subnet for pods.
 	DefaultKubernetesClusterSubnet = "10.244.0.0/16"
+	// DefaultKubernetesClusterSubnetIPv6 specifies the IPv6 default subnet for pods.
+	DefaultKubernetesClusterSubnetIPv6 = "ace:cab:deca::/8"
```
khenidak: can we use one of `fec0::/7` addresses instead?

pkg/engine/ipaddresses.go
```diff
@@ -51,3 +51,50 @@ func createJumpboxPublicIPAddress() PublicIPAddressARM {
 		},
 	}
 }
+
+// CreateClusterPublicIPAddress returns public ipv4 address resource for cluster
```
khenidak: I think we need a *big* note some where stating why this is needed and when it should be removed (along with the horrible hack we had to do with `dhcp6`). So when the platform improves and removes the need for these things we remove them.

pkg/engine/networkinterfaces.go
```diff
@@ -91,6 +91,22 @@ func CreateNetworkInterfaces(cs *api.ContainerService) NetworkInterfaceARM {
 		}
 	}
 
+	// add ipv6 nic config for dual stack
+	if cs.Properties.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack") {
```
khenidak: This is only Linux right?

aramase:
api model used for testing - https://gist.github.com/aramase/9431e9701a916e782406005f57d6f20d

pkg/api/defaults-kubelet.go
```diff
@@ -187,7 +187,7 @@ func removeKubeletFlags(k map[string]string, v string) {
 	}
 
 	// Get rid of values not supported in v1.15 and up
-	if common.IsKubernetesVersionGe(v, "1.15.0-beta.1") {
+	if common.IsKubernetesVersionGe(v, "1.15.0-alpha.3") {
```
jackfrancis: Why this change?

pkg/api/types.go
```diff
@@ -458,6 +459,7 @@ type MasterProfile struct {
 	AgentVnetSubnetID        string            `json:"agentVnetSubnetID,omitempty"`
 	FirstConsecutiveStaticIP string            `json:"firstConsecutiveStaticIP,omitempty"`
 	Subnet                   string            `json:"subnet"`
+	SubnetIPv6               string            `json:"subnetIPv6"`
```
jackfrancis: We need to enable this in the vlabs API type as well (including conversion to the base api version in both directions)

pkg/api/vlabs/validate.go
```diff
@@ -1041,20 +1046,35 @@ func validatePasswordComplexity(name string, password string) (out bool) {
...
+	if ipv6DualStackEnabled && k.NetworkPlugin != "kubenet" {
+		return errors.Errorf("OrchestratorProfile.KubernetesConfig.NetworkPlugin '%s' is invalid. IPv6 dual stack supported only with kubenet.", k.NetworkPlugin)
+	}
+
 	if k.ClusterSubnet != "" {
-		_, subnet, err := net.ParseCIDR(k.ClusterSubnet)
-		if err != nil {
-			return errors.Errorf("OrchestratorProfile.KubernetesConfig.ClusterSubnet '%s' is an invalid subnet", k.ClusterSubnet)
+		clusterSubnets := strings.Split(k.ClusterSubnet, ",")
+		if !ipv6DualStackEnabled && len(clusterSubnets) > 1 {
```
jackfrancis: I think these validation errors should complain about `OrchestratorProfile.KubernetesConfig.ClusterSubnet` instead of `OrchestratorProfile.KubernetesConfig.NetworkPlugin`

pkg/api/vlabs/validate_test.go
```diff
@@ -340,7 +340,7 @@ func Test_KubernetesConfig_Validate(t *testing.T) {
 	// Tests that should pass across all versions
 	for _, k8sVersion := range common.GetAllSupportedKubernetesVersions(true, false) {
 		c := KubernetesConfig{}
-		if err := c.Validate(k8sVersion, false); err != nil {
```
jackfrancis: Let's add unit test coverage here for the `true` `ipv6DualStackEnabled` scenarios.

pkg/api/defaults-kubelet.go
aramase: @jackfrancis Was using that for testing with custom hyperkube until 1.15 beta support was added. Removed it now.

pkg/api/const.go
aramase: @khenidak tried using that subnet and creating the cluster. The controller panics and fails to start because the cidr is too big - 

```
E0605 21:25:49.614408       1 controllermanager.go:520] Error starting "nodeipam"
F0605 21:25:49.614635       1 controllermanager.go:238] error starting controllers: New CIDR set failed; the node CIDR size is too big
goroutine 119 [running]:
```

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -98,6 +98,24 @@ write_files:
   content: !!binary |
     {{CloudInitData "aptPreferences"}}
 
+{{if not .IsCoreOS}}
```
CecileRobertMichon: if CoreOS + IPv6 isn't supported, can we add validation in https://github.com/Azure/aks-engine/pull/1424/files#diff-f45af0b14c127459636440a5981a6934R405 to fail fast when IPv6 is enabled with CoreOS? Then no need to check for CoreOS here.

parts/masterparams.t
```diff
@@ -62,6 +69,13 @@
       },
       "type": "string"
     },
+    "masterSubnetIPv6": {
```
CecileRobertMichon: why is this param defined twice? If it's inside an if/else statement let's take it out of it and define it once outside of the {{if}}

pkg/api/vlabs/types.go
```diff
@@ -809,3 +822,16 @@ func (k *KubernetesConfig) IsRBACEnabled() bool {
 	}
 	return false
 }
+
+// IsFeatureEnabled returns true if a feature flag is on for the provided feature
+func (f *FeatureFlags) IsFeatureEnabled(feature string) bool {
+	if f != nil {
+		switch feature {
+		case "EnableIPv6DualStack":
```
CecileRobertMichon: let's define the whole function here, not just the `EnableIPv6DualStack` scenario. Even if only the IPv6 flag is being used right now, this could lead to future bugs if developers assume that the function checks for all existing feature flags. Another option would be to rename the function to something more specific like `IsIPv6DualStackEnabled` to avoid confusion.

CecileRobertMichon: Also, passing in the string is a leftover implementation from before the template refactor when we needed to pass in the arg in go templating. Since you're only using this function in `*.go` files, it could just be:

```
func(f *FeatureFlags) IsIPv6DualStackEnabled() bool {
    return f != nil && f. EnableIPv6DualStack
}
```

aramase:
@jackfrancis @CecileRobertMichon addressed the comments. Let me know if anything else needs to be changed.

CecileRobertMichon:
@aramase can you try to add unit tests as much as possible to avoid decreasing coverage? https://codecov.io/gh/Azure/aks-engine/compare/be990964527eca3bab7c3af0c0a7c281fb955e32...cd6ddc49ed5345d5f65c197da428b2f4f6c12f23/diff gives a good overview of where coverage is missing.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -104,6 +104,24 @@ write_files:
   content: !!binary |
     {{CloudInitData "aptPreferences"}}
 
+{{if not .MasterProfile.IsCoreOS}}
```
CecileRobertMichon: you don't need this since `IsIPv6DualStackFeatureEnabled` can only be true if distro != CoreOS

pkg/api/vlabs/validate.go
```diff
@@ -401,6 +401,16 @@ func (a *Properties) validateAgentPoolProfiles(isUpdate bool) error {
 			return e
 		}
 
+		// validate os type is linux if dual stack feature is enabled
+		if a.FeatureFlags.IsIPv6DualStackEnabled() {
+			if agentPoolProfile.OSType == Windows {
+				return errors.Errorf("Dual stack feature is supported only with Linux, but agent pool '%s' is of os type %s", agentPoolProfile.Name, agentPoolProfile.OSType)
+			}
+			if agentPoolProfile.Distro == CoreOS {
```
CecileRobertMichon: let's add the same to MasterProfile validation (Distro only, masters don't support windows).

CecileRobertMichon:
Nice UT coverage :green_heart:

CecileRobertMichon:
@jackfrancis and @khenidak for additional reviews

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -154,6 +154,14 @@ if [[ -n "${MASTER_NODE}" && "${KMS_PROVIDER_VAULT_NAME}" != "" ]]; then
     ensureKMS
 fi
 
+# configure and enable dhcpv6 for dual stack feature
+DHCPV6_SYSTEMD_SERVICE=/etc/systemd/system/dhcpv6.service
+DHCPV6_CONFIGURATION_SCRIPT=/opt/azure/containers/enable-dhcpv6.sh
+
+if [[ -f ${DHCPV6_SYSTEMD_SERVICE} ]] && [[ -f ${DHCPV6_CONFIGURATION_SCRIPT} ]];then
```
jackfrancis: There is a possible race condition here. If cloud-init fails to pave files before CustomScriptExtension code execution reaches this line, then we will skip this step, even if the cluster configuration declares that we *should* pave these files.

CecileRobertMichon: Ooh yeah good catch... +1

Instead of checking if the file exist, you could add a variable for IS_IPV6 (with a better name) and use `wait_for_file` to wait for these two files to exist before preceding, similar to how we wait for the Azure Stack script https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_main.sh#L34

aramase: Done

CecileRobertMichon:
@aramase did you answer Jack's question earlier?

https://github.com/Azure/aks-engine/pull/1424#issuecomment-498328517

aramase:
@CecileRobertMichon I think we discussed about this on a thread in teams. Yeah, this PR works even if VM ships with that setting. As part of the PR I setup and enable dhcpv6 which ensure the nodes get ipv6 ips.

jackfrancis:
lgtm

PR #1426, opened 2019-06-03
test: quicker “defaults addons image” unit test

jackfrancis:
**Reason for Change**:

Parallelize and reduce code surface overhead of validating defaults addons image for create/update.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -219,54 +219,46 @@ func TestAssignDefaultAddonImages(t *testing.T) {
 	}
 
 	cases := []struct {
+		name           string
 		myAddons       []KubernetesAddon
-		isUpgrade      bool
-		isScale        bool
+		isUpdate       bool
 		expectedImages map[string]string
 	}{
 		{
+			name:           "default",
 			myAddons:       getFakeAddons(defaultAddonImages, ""),
-			isUpgrade:      false,
-			isScale:        false,
+			isUpdate:       false,
 			expectedImages: defaultAddonImages,
 		},
 		{
+			name:           "create scenario",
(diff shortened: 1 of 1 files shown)


pkg/api/defaults_test.go
```diff
@@ -219,54 +219,46 @@ func TestAssignDefaultAddonImages(t *testing.T) {
...
-		for _, result := range resultAddons {
-			if result.Containers[0].Image != c.expectedImages[result.Name] {
-				t.Errorf("expected setDefaults to set Image to \"%s\" in addon %s, but got \"%s\"", c.expectedImages[result.Name], result.Name, result.Containers[0].Image)
+		c := c
+		t.Run(c.name, func(t *testing.T) {
+			t.Parallel()
+			mockCS := getMockBaseContainerService("1.10.8")
+			mockCS.Properties.OrchestratorProfile.OrchestratorType = Kubernetes
+			mockCS.Properties.OrchestratorProfile.KubernetesConfig.Addons = c.myAddons
+			mockCS.setOrchestratorDefaults(c.isUpdate)
```
jackfrancis: @CecileRobertMichon to validate that calling `setOrchestratorDefaults` covers the equivalent functional code surface area for this test as compared to `SetPropertiesDefaults` (with 30 seconds savings!)

CecileRobertMichon: It does. I guess the only subtlety that the previous implementation covered that this one does not is that `isUpdate` in the setOrchestratorDefaults context is equivalent to `isScale || isUpgrade` in the setPropertiesDefault context (which isn't really what this test is about).

PR #1435, opened 2019-06-04
chore: add go mod support

mboersma:
**Reason for Change**:
Adds [go module](https://github.com/golang/go/wiki/Modules) support for dependency vendoring, but doesn't remove `dep` stuff yet.

All credit goes to @andyliuliming, I just polished this up.

**Issue Fixed**:
Replaces #1077 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
@andyliuliming is this important for AKS vendoring, or just something to future-proof us?


mboersma:
There's no particular urgency to this, so let's put it on the back burner again until there is an actual need. I'll keep the branch around.

PR #1438, opened 2019-06-06
feat: add support for Kubernetes 1.15.0-beta.2

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1150-beta1
and kubernetes/kubernetes#78242

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Is it ok to add this new `ClusterRole[Binding]` in general, or should it be scoped to k8s 1.15+?

### DIFF
--- a/parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
+++ b/parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
@@ -31,7 +31,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: ClusterRole
 metadata:
-  name: system:azure-persistent-volume-binder 
+  name: system:azure-persistent-volume-binder
   labels:
     kubernetes.io/cluster-service: "true"
 rules:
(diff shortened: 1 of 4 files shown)


parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
```diff
@@ -48,8 +48,35 @@ metadata:
...
   kind: ClusterRole
   apiGroup: rbac.authorization.k8s.io
-  name: system:azure-persistent-volume-binder 
+  name: system:azure-persistent-volume-binder
 subjects:
 - kind: ServiceAccount
   name: persistent-volume-binder
   namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1beta1
```
jackfrancis: Are these additional `ClusterRole` and `ClusterRoleBinding` definitions compatible w/ < 1.15 k8s?

jackfrancis: I posed that question to @feiskyer who introduced this into upstream k8s. Depending on the answer we may want to move this modified spec into `parts/k8s/addons/1.15/`

PatrickLang:
LGTM from Windows perspective. v1.15.0-beta.2 has a fix that's needed - https://github.com/kubernetes/kubernetes/issues/78628

parts/k8s/addons/kubernetesmasteraddons-azure-cloud-provider-deployment.yaml
feiskyer: yep, they're compatible with old versions, moving to 1.15 dir also works

PR #1439, opened 2019-06-06
feat: add support for Kubernetes 1.14.3

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#changelog-since-v1142

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -361,10 +361,8 @@ echo "  - busybox" >> ${RELEASE_NOTES_FILEPATH}
 # TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
 K8S_VERSIONS="
 1.14.3
-1.14.2
 1.14.1
 1.13.7
-1.13.6
 1.13.5
 1.13.5-azs
 1.12.9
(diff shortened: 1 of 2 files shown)


packer/install-dependencies.sh
```diff
@@ -360,6 +360,7 @@ echo "  - busybox" >> ${RELEASE_NOTES_FILEPATH}
 
 # TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
 K8S_VERSIONS="
+1.14.3
 1.14.2
```
CecileRobertMichon: Maybe we can remove 1.14.2 since it's disabled (and we most likely won't re-enable it)

packer/install-dependencies.sh
```diff
@@ -360,9 +360,8 @@ echo "  - busybox" >> ${RELEASE_NOTES_FILEPATH}
 
 # TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
 K8S_VERSIONS="
-1.14.2
+1.14.3
 1.14.1
-1.13.6
```
mboersma: Or I can remove this in the upcoming 1.13.7 PR if we prefer.

CecileRobertMichon: I'm fine with either, the 1.13.7 PR makes more logical sense but it's not a big deal.

PR #1442, opened 2019-06-06
feat: support for Application Gateway Ingress Controller

akshaysngupta:
**Reason for Change**:
This PR adds support to deploy `Application Gateway` with the Kubernetes cluster to provide `ingress` support.
1) Azure `Application Gateway` is Azure's native layer-7 loadbalancer which additionally provides support for SSL Offload, WAF, Autoscaling, Private Ip, Redirects, etc.
2) [**Skipping adding ingress controller section, as discussed with Sean**] Ingress Controller is a controller that looks for `Ingress/Service/Endpoint/Pod` related changes on the cluster and configures the loadbalancer with the updated routing rules.
Ref: [Application Gateway Ingress Controller](https://github.com/Azure/application-gateway-kubernetes-ingress).

**Add-on Description**
This add-on will deploy an Application Gateway and dependency resources with your new Kubernetes cluster.

Supported Add-on `Config` options:

| Option | Required | Default | Description |
|--|--|--|--|
| `appgw-subnet` | true | N/A | CIDR of the Application Gateway subnet. This should not overlap with master/agent subnets. |
| `appgw-sku` | false | `WAF_v2` | SKU of the Application Gateway. (`Standard_v2`/`WAF_v2`) |
| `appgw-private-ip` | false | null | Private IP assigned to the Application Gateway from subnet. |

Once, the infrastructure is deployed, please follow the instructions to deploy the [Application Gateway  Ingress controller](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/install-new.md#setting-up-application-gateway-ingress-controller-on-aks)

***Additions to ARM resources:***
1) Azure Virtual Network with a subnet specifically for `AppGw`.
2) Azure Application Gateway
3) User Assigned Identity to initialize the aad-pod-identity service and ingress controller.
4) Set required RBACs.

Result
image

[**Skipping adding ingress controller section, as discussed with Sean**] ***Additions to Kubernetes deployment:***
1) Deploy Ingress Controller `deployment/pod`
2) Deploy `AAD-Pod-indentity` for token required by the controller to talk to ARM

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -599,6 +599,7 @@
       "type": "string"
     }
  {{end}}
+ {{if .OrchestratorProfile.KubernetesConfig.IsAppGWIngressEnabled}}
     ,"appGwSubnet": {
       "metadata": {
         "description": "Sets the subnet of the Application Gateway"
@@ -611,3 +612,4 @@
       },
       "type": "string"
     }
+  {{end}}
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -599,3 +599,15 @@
       "type": "string"
     }
  {{end}}
+    ,"appGwSubnet": {
+      "metadata": {
+        "description": "Sets the subnet of the Application Gateway"
+      },
+      "type": "string"
+    }
+    ,"appGwSku": {
+      "metadata": {
+        "description": "Sets the subnet of the Application Gateway"
+      },
+      "type": "string"
+    }
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -612,4 +612,4 @@
       },
       "type": "string"
     }
-  {{end}}
+ {{end}}
--- a/pkg/engine/roleassignments.go
+++ b/pkg/engine/roleassignments.go
@@ -4,6 +4,7 @@
 package engine
 
 import (
+	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/azure-sdk-for-go/services/preview/authorization/mgmt/2018-09-01-preview/authorization"
 	"github.com/Azure/go-autorest/autorest/to"
 )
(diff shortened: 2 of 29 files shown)


jackfrancis:
@akshaysngupta is this PR feature-complete? Specifically:

- User Assigned Identity to initialize the aad-pod-identity service and ingress controller.
- Set required RBACs

jackfrancis:
Also, this is a nice, elegant re-use of the addons configuration interface. Depending on how the actual k8s specs land it may be appropriate to use this as the configuration interface for the entire app gateway functionality set. (Rather than having discrete configuration objects for this functionality)

That's a way of saying that I like this approach so far, looking forward to seeing it evolve!

akshaysngupta:
@jackfrancis Can you PTAL ?

pkg/engine/roleassignments.go
```diff
@@ -37,3 +38,74 @@ func createMSIRoleAssignment(identityRoleDefinition IdentityRoleDefinition) Role
 		},
 	}
 }
+
+func createKubernetesSpAppGIdentityOperatorAccessRoleAssignment(prop *api.Properties) RoleAssignmentARM {
```
jackfrancis: Looks like these new funcs need UT

jackfrancis:
FYI unit test failures at present are `gofmt` lint errors. `gofmt -s -w <filename>` will address those.

akshaysngupta:
@jackfrancis I have addressed the comments.
Can you PTAL ?

jackfrancis:
lgtm

jackfrancis:
@akshaysngupta we'd love to protect this new feature w/ E2E tests, if you have a set of `kubectl` commands that we could run against clusters with the `appgw-ingress` addon enabled, let us know!

jackfrancis:
to address unit test failures:

```
gofmt -s -w pkg/engine/applicationgateway_test.go
gofmt -s -w pkg/engine/armoutputs_test.go
```

akshaysngupta:
@jackfrancis e2e would require `helm` usage.

jackfrancis:
@akshaysngupta there's a unit test named `TestAssignDefaultAddonImages` in `pkg/api/defaults_test.go` that will need updating. Specifically add the name of the new addon as the key and the default image string as the value to the `defaultAddonImages` dictionary, should make things pass. (Also feel free to run that test to validate before pushing a commit)

parts/k8s/kubernetesparams.t
```diff
@@ -599,3 +599,15 @@
       "type": "string"
     }
  {{end}}
+    ,"appGwSubnet": {
```
jackfrancis: Add a `"defaultValue": ""` property (see params definitions above) to fix the ARM errors. What's happening is that the ARM parameters spec is not being fulfilled by the template in cluster scenarios that don't assign values to these params (`pkg/engine/params_k8s.go`).

CecileRobertMichon: instead of adding a default value, you can conditionally define those variables (see IsAzureCNI case above). This will avoid defining extra parameters when they are unused.

akshaysngupta:
@CecileRobertMichon looks like the issue was with spacing.

david-jarman:
@akshaysngupta very cool! Do you know if this will work with Windows Containers?

akshaysngupta:
@david-jarman It will work for anything with an IP address. Only requirement is AKS network plugin should be `azure` (advanced networking).

PR #1452, opened 2019-06-10
test: Switch Windows test passes to VMSS by default to match AKS

PatrickLang:
**Reason for Change**:
`VirtualMachineScaleSets` has been the recommended default for k8s > 1.10 per https://github.com/Azure/aks-engine/blob/windows-vmss/docs/topics/clusterdefinitions.md#L530 , but the PR test passes are using `AvailabilitySet`


PatrickLang:
1.14 failure looked unrelated - failed in some add-ons on Linux. Rerunning

PR #1454, opened 2019-06-10
feat: enable PSP w/ privileged ClusterRoleBinding for 1.15

### LINKED ISSUE #1453: Pod security policy should be enabled by default for > 1.15.0-beta.1
**Describe the request**
Enable pod security policy by default for cluster versions > `1.15.0-beta.1`

`--allow-privileged` flag is no longer supported in kubelet with `1.15.0-beta.1`. The changes to remove that usage is already merged - https://github.com/Azure/aks-engine/pull/1369. This means none of the container addons  (kube-proxy, ip-masq-agent) get created as they have `privileged: true`.

- Need to enable psp by default based on cluster version.
- Need to document the changes so the user is aware of it.

**Additional context**
The psp currently defined in the repo creates a privileged and restricted policy. The rolebinding allows pods in kube-system to have `privileged: true` and blocks using that in other namespaces. I think the default behavior should be to allow `privileged: true` in all ns as was allowed with `--allow-privileged` flag in kubelet in previous versions.

jackfrancis:
**Reason for Change**:

Use PodSecurityPolicy to implement privileged auth for 1.15.0-beta.1 and greater.

**Issue Fixed**:

Fixes #1453 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
+++ b/parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
@@ -127,3 +127,6 @@ subjects:
 - kind: Group
   name: system:authenticated
   apiGroup: rbac.authorization.k8s.io
+- kind: Group
+  name: system:nodes
+  apiGroup: rbac.authorization.k8s.io
--- a/parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
+++ b/parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
@@ -1,0 +1,129 @@
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: privileged
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: "*"
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+spec:
+  privileged: true
+  allowPrivilegeEscalation: true
+  allowedCapabilities:
+  - "*"
+  volumes:
+  - "*"
+  hostNetwork: true
+  hostPorts:
+  - min: 0
+    max: 65535
+  hostIPC: true
+  hostPID: true
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+---
+apiVersion: extensions/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: restricted
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: docker/default
+    apparmor.security.beta.kubernetes.io/allowedProfileNames: runtime/default
+    seccomp.security.alpha.kubernetes.io/defaultProfileName:  docker/default
+    apparmor.security.beta.kubernetes.io/defaultProfileName:  runtime/default
+  labels:
+    addonmanager.kubernetes.io/mode: Reconcile
+spec:
+  privileged: false
+  allowPrivilegeEscalation: false
+  requiredDropCapabilities:
+    - ALL
+  volumes:
+    - configMap
+    - emptyDir
+    - projected
+    - secret
+    - downwardAPI
+    - persistentVolumeClaim
+  hostNetwork: false
+  hostIPC: false
+  hostPID: false
+  runAsUser:
+    rule: MustRunAsNonRoot
+  seLinux:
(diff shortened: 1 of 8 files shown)


parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
```diff
@@ -0,0 +1,129 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: psp:restricted
+subjects:
+- kind: Group
+  name: system:authenticated
+  apiGroup: rbac.authorization.k8s.io
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
jackfrancis: @aramase This `ClusterRoleBinding` definition is what's different compared to the existing PSP spec. Is this what we want?

pkg/engine/artifacts.go
```diff
@@ -211,7 +211,7 @@ func kubernetesAddonSettingsInit(p *api.Properties) []kubernetesComponentFileSpe
 			sourceFile:      "kubernetesmasteraddons-pod-security-policy.yaml",
 			base64Data:      k.GetAddonScript(PodSecurityPolicyAddonName),
 			destinationFile: "pod-security-policy.yaml",
-			isEnabled:       to.Bool(p.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy),
+			isEnabled:       to.Bool(p.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy) || common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.15.0-beta.1"),
```
aramase: @jackfrancis If the user sets EnablePodSecurityPolicy to false in the api model, then the cluster deployment will fail. Do we want add a validation check and fail if the user sets kube version > 1.15.0-beta.1 and also set enable psp to false?

parts/k8s/addons/1.15/kubernetesmasteraddons-pod-security-policy.yaml
aramase: Yeah, this should allow `privileged: true` for all namespaces. Also, I think we need to add `system:nodes`  to the subjects here, otherwise the docker containers start, but the pods don't show up in kube-system. This is the reason I've this change in my PR - https://github.com/Azure/aks-engine/pull/1424/files#diff-2f29451dcd8553e749b786b06ac1ed42. But that's with role binding, so we can check what the result is with user cluster role binding instead.

aramase: cc @khenidak ^^

jackfrancis: Done, please review

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -351,9 +351,6 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     sed -i "s|<img>|{{WrapAsParameter "kubernetesCcmImageSpec"}}|g" /etc/kubernetes/manifests/cloud-controller-manager.yaml
     sed -i "s|<config>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.CloudControllerManagerConfig}}|g" /etc/kubernetes/manifests/cloud-controller-manager.yaml
 {{end}}
-{{if not EnablePodSecurityPolicy}}
```
jackfrancis: These sed lines are no longer needed, as the `apparmor_parser` references were removed when we implemented kubelet running on the host OS:

https://github.com/Azure/acs-engine/pull/2584/files#diff-42c513b221406469e43868cefdc5ad55L30

pkg/api/defaults.go
```diff
@@ -268,6 +268,10 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			a.OrchestratorProfile.KubernetesConfig.ExcludeMasterFromStandardLB = to.BoolPtr(DefaultExcludeMasterFromStandardLB)
 		}
 
+		if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.15.0-beta.1") {
```
jackfrancis: Statically enabling AKS Engine PSP implementation if >= 1.15.0-beta.1

jackfrancis:
@aramase Both the 1.15 tests in the pipeline jobs list above will always run the latest supported patch version for the k8s versions they're testing. In this case they ran against `v1.15.0-beta.2` clusters, so we're good here.

PR #1456, opened 2019-06-11
feat: enable deployment of windows nodes on Azure Stack

### LINKED ISSUE #1458: feat: support deployment of windows nodes on Azure Stack
**Describe the request**
Currently deployment of windows nodes on Azure Stack fails. The kubelet tries to connect to azure but fails since it is being provisioned on azure stack. AKS Engine needs to configure the kubelet to connect to azure stack instead. 

In order for customers to use kubernetes on windows on azure stack this change needs to be made. 

`cloud: AzureStackCloud` needs to be added to the azure config file (`c:\k\azure.json`). This instructs the kubelet to look for a configuration file with the azure stack environment information. This configuration file already gets injected into the master node, it just needs to be placed on the windows node as well. By default the kubelet looks for the file at `/etc/kubernetes/azurestackcloud.json` but this can be overriden with an environment variable: `AZURE_ENVIRONMENT_FILEPATH="c:\k\azurestackcloud.json"`. 

Also when azure stack is using adfs the tenant ID needs to be changed in the azure config file (`c:\k\azure.json`) to be `tenantId: adfs`. 

**Additional context**

t-chappl:
**Reason for Change**:

Enables deployment of kubernetes with windows nodes on Azure Stack

**Issue Fixed**:

Fixes #1458 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jadarsie:
@PatrickLang, @t-chappl he will create an issue explaining what's the issue he is trying to solve and briefly explain his approach.

t-chappl:
This PR fixes #1458

pkg/engine/template_generator.go
```diff
@@ -621,6 +621,14 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
 		"IsCustomVNET": func() bool {
 			return cs.Properties.AreAgentProfilesCustomVNET()
 		},
+		"GetEscapedEnvironmentJSON": func() string {
+			customEnvironmentJSON, _ := cs.Properties.GetCustomEnvironmentJSON(false)
+			customEnvironmentJSON = strings.Replace(customEnvironmentJSON, `"`, `""`, -1)
+			return customEnvironmentJSON
+		},
+		"IsIdentitySystemADFS": func() bool {
+			return cs.Properties.IsAzureStackCloud() && cs.Properties.CustomCloudProfile.IdentitySystem == api.ADFSIdentitySystem
```
jadarsie: Should we nil check `CustomCloudProfile` for Azure deployments?

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -410,35 +413,35 @@ if (`$hnsNetwork)
...
-    }
-    `$cnilock = [io.path]::Combine("$KubeDir", "azure-vnet.json.lock")
-    if ((Test-Path `$cnilock))
-    {
-        Remove-Item `$cnilock
-    }
-}
-
-# Restart Kubeproxy, which would wait, until the network is created
+    Remove-HnsNetwork `$hnsNetwork
```
jadarsie: I do not see what changed here. Make sure you did not change EOL

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -40,7 +40,11 @@ param(
 
     [parameter(Mandatory=$true)]
     [ValidateNotNullOrEmpty()]
-    $AADClientSecret # base64
+    $AADClientSecret, # base64
+
+    [parameter(Mandatory=$true)]
+    [ValidateNotNullOrEmpty()]
+    $TargetEnvironment
```
jadarsie: Should this be required for Azure deployments as well?

pkg/engine/template_generator.go
t-chappl: That is what `IsAzureStackCloud()` does. 
```
func (p *Properties) IsAzureStackCloud() bool {
	return p.CustomCloudProfile != nil
}
```
Would it be better to directly check if it is nil?

jadarsie: no, it is fine, I forgot the function was already nit checking, thx

parts/k8s/kuberneteswindowssetup.ps1
t-chappl: The linux deploy process always puts the target environment in the config file. If it is left blank it defaults to AzurePublicCloud, so providing it doesn't make a difference for azure deployments.

parts/k8s/windowskubeletfunc.ps1
t-chappl: My ide must have changed these. The file is half CRLF and half LF. Which should I be using?

jadarsie: Probably better to leave it as it was before. You or someone else can push a subsequent clean up PR.

parts/k8s/kuberneteswindowssetup.ps1
jadarsie: It is OK with me if this change does not break or alter behavior on Azure.
What does the `ValidateNotNullOrEmpty` attribute do if left black as you mention?

t-chappl: The arguments for this script are contained in the ARM template in pkg/engine/vmextensions.go. TargetEnvironment is already a parameter in the template so on line 408 I pass it as an argument to this script. That means it will always be set to whatever the target environment is, be it AzureStackCloud or AzurePublicCloud

t-chappl: Also, from what I can see leaving it blank could cause failures in China or Germany or the government cloud. It should be set to AzureChinaCloud or whichever so that the kubelet can connect correctly in those locations. It gets set every time when we deploy on a linux host but never gets set on windows hosts

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -523,6 +526,8 @@ Update-CNIConfig(`$podCIDR, `$masterSubnetGW)
 
 try
 {
+    `$env:AZURE_ENVIRONMENT_FILEPATH="c:\k\azurestackcloud.json"
```
PatrickLang: what process uses `AZURE_ENVIRONMENT_FILEPATH` ? I'm wondering why it's set here as an environment variable when I don't see it referenced anywhere

t-chappl: The kubelet uses it

t-chappl: https://github.com/kubernetes/kubernetes/blob/master/vendor/github.com/Azure/go-autorest/autorest/azure/environments.go#L27

t-chappl: Kubernetes has a dependency on Azure's goautorest. In the azure cloud provider code it checks the config file to see which environment it should connect to, and when it realizes it is azure stack cloud it looks for the configuration file stored in the location specified by that env var

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -192,7 +200,13 @@ try
             -UserAssignedClientID $global:UserAssignedClientID `
             -UseInstanceMetadata $global:UseInstanceMetadata `
             -LoadBalancerSku $global:LoadBalancerSku `
-            -ExcludeMasterFromStandardLB $global:ExcludeMasterFromStandardLB
+            -ExcludeMasterFromStandardLB $global:ExcludeMasterFromStandardLB `
+            -TargetEnvironment $TargetEnvironment
+
+        {{if IsAzureStackCloud}}
+        $azureStackConfigFile = [io.path]::Combine($global:KubeDir, "azurestackcloud.json")
+        "{{ GetEscapedEnvironmentJSON }}" | Out-File -encoding ASCII -filepath "$azureStackConfigFile"
```
PatrickLang: Can you share what the contents of this file are? Since this file will be included in the contents of a custom script extension, that means it has to be escaped for both powershell and ARM. Single quotes break ARM

My preference would be to split this up into separate variables up at the top, then have function like `Write-AzureConfig` above that handles creating the file. I'm planning to delete kuberneteswindowssetup.ps1 as it exists today. If that code is in another function we can still reuse it after the refactor.

PatrickLang: The list of "stuff" needed to configure a VM will be going into an API or config file, and the code applying the config will be put in the VM image instead of being passed through a custom script extension.

pkg/engine/vmextensions.go
```diff
@@ -152,7 +152,7 @@ func createAgentVMASCustomScriptExtension(cs *api.ContainerService, profile *api
...
+			"commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('kubernetesAPIServerIP'),' -KubeDnsServiceIp ',parameters('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -TargetEnvironment ',parameters('targetEnvironment'),' -AgentKey ',parameters('clientPrivateKey'),' -AADClientId ',variables('servicePrincipalClientId'),' -AADClientSecret ',variables('singleQuote'),variables('singleQuote'),base64(variables('servicePrincipalClientSecret')),variables('singleQuote'),variables('singleQuote'), ' ',variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]",
```
PatrickLang: Also - be sure to update the same commandline for VM scale sets:
https://github.com/Azure/aks-engine/blob/0a2ded2dfbb8d504fab0567f9f241775faabfc96/pkg/engine/virtualmachinescalesets.go#L641

AKS uses VMSS by default and could break if only vmextensions.go is updated

parts/k8s/kuberneteswindowssetup.ps1
t-chappl:
```
{
    "name": "AzureStackCloud",
    "managementPortalURL": "https://portal.local.azurestack.external/",
    "publishSettingsURL": "",
    "serviceManagementEndpoint": "https://management.azurestack.onmicrosoft.com/00000000-0000-0000-0000-0000000000",
    "resourceManagerEndpoint": "https://management.local.azurestack.external/",
    "activeDirectoryEndpoint": "https://login.microsoftonline.com/",
    "galleryEndpoint": "https://galleryartifacts.hosting.local.azurestack.external/galleryartifacts/",
    "keyVaultEndpoint": "",
    "graphEndpoint": "https://graph.windows.net/",
    "serviceBusEndpoint": "",
    "batchManagementEndpoint": "",
    "storageEndpointSuffix": "local.azurestack.external",
    "sqlDatabaseDNSSuffix": "",
    "trafficManagerDNSSuffix": "",
    "keyVaultDNSSuffix": "vault.local.azurestack.external",
    "serviceBusEndpointSuffix": "",
    "serviceManagementVMDNSSuffix": "",
    "resourceManagerVMDNSSuffix": "cloudapp.azurestack.external",
    "containerRegistryDNSSuffix": "",
    "tokenAudience": ""
}
```

t-chappl:
```
{
    "name": "AzureStackCloud",
    "managementPortalURL": "https://portal.local.azurestack.external/",
    "publishSettingsURL": "",
    "serviceManagementEndpoint": "https://management.azurestack.onmicrosoft.com/00000000-0000-0000-0000-0000000000",
    "resourceManagerEndpoint": "https://management.local.azurestack.external/",
    "activeDirectoryEndpoint": "https://login.microsoftonline.com/",
    "galleryEndpoint": "https://galleryartifacts.hosting.local.azurestack.external/galleryartifacts/",
    "keyVaultEndpoint": "",
    "graphEndpoint": "https://graph.windows.net/",
    "serviceBusEndpoint": "",
    "batchManagementEndpoint": "",
    "storageEndpointSuffix": "local.azurestack.external",
    "sqlDatabaseDNSSuffix": "",
    "trafficManagerDNSSuffix": "",
    "keyVaultDNSSuffix": "vault.local.azurestack.external",
    "serviceBusEndpointSuffix": "",
    "serviceManagementVMDNSSuffix": "",
    "resourceManagerVMDNSSuffix": "cloudapp.azurestack.external",
    "containerRegistryDNSSuffix": "",
    "tokenAudience": ""
}
```

t-chappl: ^^^ Here is an example of the file as it would be found on the machine. 
The file is different depending on which azure stack instance the user is trying to connect to, so it can't be on the VM image.

t-chappl: There is an example of the escaped file in the unit test https://github.com/Azure/aks-engine/blob/2e50b6343b7775e5da87c0c6b7be518c04dcb598/pkg/engine/template_generator_test.go#L175

I escape the `"` with `""` in this file https://github.com/Azure/aks-engine/blob/2e50b6343b7775e5da87c0c6b7be518c04dcb598/pkg/engine/template_generator.go#L626

I should probably escape single quotes as well so that if the input has them they don't break the template. 

I could make separate variables at the top, but there would be around 20 of them given the file structure above. Should I add all 20 variables?

t-chappl: Also, with linux we just use cloud-init to place the file on the host during provisioning. Is there a way to add multiple custom data files to a windows host while provisioning?

PatrickLang: unfortunately there isn't a way to add multiple files during provisioning, they need to be written from the PowerShell scripts

t-chappl: That's too bad. So would you rather I put it in a function with a bunch of new variables or pass the escaped string?

PatrickLang: Initially I thought some of those might be needed elsewhere, but it looks like that JSON never really needs to be parsed as input, just placed. From that perspective I think we should just get it onto disk rather than passing it across functions.

What I was initially asking would mean you couldn't reuse this between Linux & Windows nodes. We shouldn't reimplement it in PowerShell.
https://github.com/azure/aks-engine/blob/c68c2a94b63ff401241b88fbca104f000a9481e7/pkg/api/types.go#L1698

I think the cleanest thing here, and the closest to dropping a file with cloud-init is to:

1. Change template_generator.go to use `base64.StdEncoding.EncodeToString`  to encode the output from `GetCustomEnvironmentJSON`

2. write it out like this https://github.com/azure/aks-engine/blob/407207ba834c2fea9ea93d8a23d36f67519e2f96/parts/k8s/kuberneteswindowssetup.ps1#L119

PatrickLang: base64 avoids all the differences in escapes. once that's done it looks ready to merge

PatrickLang:
LGTM - @CecileRobertMichon will take a look too

CecileRobertMichon:
@t-chappl unit tests are failing, see https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=1523&view=logs for failure logs

PR #1471, opened 2019-06-12
fix: fix race condition between cloud init walinuxagent hold and CSE

CecileRobertMichon:
**Reason for Change**:

We apt-mark hold walinuxagent in cloud init but there is no guarantee that that code will execute before apt-get upgrade and apt-get dist-upgrade run in CSE. 

Partly reverts https://github.com/Azure/aks-engine/pull/778.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -189,6 +189,7 @@ apt_get_dist_upgrade() {
     export DEBIAN_FRONTEND=noninteractive
     dpkg --configure -a
     apt-get -f -y install
+    apt-mark showhold
     ! (apt-get dist-upgrade -y 2>&1 | tee $apt_dist_upgrade_output | grep -E "^([WE]:.*)|([eE]rr.*)$") && \
     cat $apt_dist_upgrade_output && break || \
     cat $apt_dist_upgrade_output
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -189,6 +189,7 @@ apt_get_dist_upgrade() {
     export DEBIAN_FRONTEND=noninteractive
     dpkg --configure -a
     apt-get -f -y install
+    apt-mark showhold
```
CecileRobertMichon: this doesn't have any functional behavior change but it's nice to have for the logs

jackfrancis:
Are we sure that the `aptmarkWALinuxAgent` func is quick and idempotent? I.e., if a hold already exists (because it was held firstly via cloud-init) this invocation will be a speedy no-op?

CecileRobertMichon:
> Are we sure that the aptmarkWALinuxAgent func is quick and idempotent? I.e., if a hold already exists (because it was held firstly via cloud-init) this invocation will be a speedy no-op?

@jackfrancis correct:

```
azureuser@k8s-master-85834828-0:~$ sudo apt-mark hold walinuxagent
walinuxagent set on hold.
azureuser@k8s-master-85834828-0:~$ apt-mark showhold
walinuxagent
azureuser@k8s-master-85834828-0:~$ sudo apt-mark hold walinuxagent
walinuxagent was already set on hold.
```

jackfrancis:
Should we remove the `wait_for_apt_locks` line in `aptmarkWALinuxAgent`? I'm not sure including it was perfectly thought out. The idea is that if apt is already doing something in the background that a hold might not actually have the intended effect (i.e., it's too late). However, we're not actually erroring out or actually checking that the thing being done in the background (which the presence of locks would indicate) is related to the hold we're trying to make.

So, keeping it in there means that we might be adding lots of blocking time during CSE, as this new hold will block for the release of apt locks.

Does that make sense?

CecileRobertMichon:
@jackfrancis I think we still want the lock there. Otherwise, the hold might fail. 

We can explore time optimization and potentially removing it later on, I would rather keep this PR as is because it will be released in a patch release and this bug is blocking for dod.

PR #1473, opened 2019-06-12
fix: remove walinuxagent version pin

CecileRobertMichon:
**Reason for Change**:

Why remove this?
- This was added in https://github.com/Azure/aks-engine/pull/778 to "to further protect against runtime changes to the walinuxagent". #1471 proved that this does not actually work. 
- Different clouds/regions might different walinuxagent versions rolled out so pinning to one version might break sovereign clouds that don't have the latest version available.
- Having this setting introduces friction to have the latest walinuxagent version as we are not actively monitoring agent releases, which might lead us to run into bugs in the future that would have otherwise been fixed in the latest version. I would argue we should delegate walinuxagent version updates to the linux agent team and not override it.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/apt-preferences
+++ b/parts/k8s/cloud-init/artifacts/apt-preferences
@@ -1,3 +1,0 @@
-Package: walinuxagent
-Pin: version 2.2.32.2
-Pin-Priority: 550
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -10002,9 +10002,7 @@ func k8sArmparametersT() (*asset, error) {
 	return a, nil
 }
 
-var _k8sCloudInitArtifactsAptPreferences = []byte(`Package: walinuxagent
-Pin: version 2.2.32.2
-Pin-Priority: 550`)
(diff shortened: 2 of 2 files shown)


parts/k8s/cloud-init/artifacts/apt-preferences
```diff
@@ -1,3 +0,0 @@
-Package: walinuxagent
```
CecileRobertMichon: Please description for why we want this change

CecileRobertMichon: why leave this file instead of deleting it completely?
- it it not taking significant space
- leaving it intact will make it easier to quickly re-add a version pin (for this package or another) later on.
- the file is referenced in many places in the code so removing it would make this PR a significantly bigger code change

jackfrancis:
I assume this is a WIP, we'll need to remove all `aptPreferences` references in various code flows.

CecileRobertMichon:
@jackfrancis it's not, see my comment above.

PR #1479, opened 2019-06-13
refactor: kube-dns as as api model "addon"

jackfrancis:
**Reason for Change**:

This PR moves the kube-dns addon implementations to the user-configurable "addons" array in the api model.

Advantages:

- more explicit configuration
- eliminate need for "last mile sed"
- gets us closer to standardizing upon a single Kubernetes addon implementation

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@CecileRobertMichon @mboersma This WIP label will remain until we have consensus that we want to move forward with this light refactor. Defer to you for that go-ahead.

Additionally, if we're good to go ahead, should we include all existing "addons" in-scope for transition to "containerAddons"? If so, I'll leave the WIP in place until the PR has a full migration.

Or, if we'd prefer to merge just one addon at at time, we can convert this into a kube-dns-specific PR.

CecileRobertMichon:
My vote: 
- we want to move this forward (I believe it's an improvement)
- we separate PRs per addon (so we make this one kube-dns only). Looking at the diff it's already really big for only one addon so I don't think we want to change all of them at once because 1) they are independent, nothing prevents us from splitting them technically 2) smaller PRs means easier reviews + less merge conflicts --> faster merging

mboersma:
I agree 💯 with @CecileRobertMichon's comment. This looks much cleaner and more consistent, but let's keep it in digestible pieces as it moves forward.

jackfrancis:
This is ready for reviews as a kube-dns-specific change

pkg/api/addons.go
```diff
@@ -432,3 +446,51 @@ func synthesizeAddonsConfig(addons []KubernetesAddon, addon KubernetesAddon, isU
...
+	return spec
+}
+
+func getKubeDNSClusterDomain(k *KubernetesConfig) string {
+	if k.KubeletConfig != nil {
+		if val, ok := k.KubeletConfig["--cluster-domain"]; ok {
+			return val
+		}
+	}
+	return DefaultKubernetesClusterDomain
```
CecileRobertMichon: I assume the reason we need to do this is because kubelet defaults happen after addon defaults? if so we need to be careful about using this logic since it's a duplicate, if the default logic in kubelet defaults changes we'll need to make sure to update this as well. Could you add a code comment to explain why we need to do this here?

pkg/api/vlabs/validate.go
```diff
@@ -601,6 +601,16 @@ func (a *Properties) validateAddons() error {
 				if to.Bool(addon.Enabled) && a.HasCoreOS() {
 					return errors.New("flexvolume add-ons not currently supported on coreos distro. Please use Ubuntu")
 				}
+			case "kube-dns":
+				if to.Bool(addon.Enabled) {
+					maxVersion, err := semver.Make("1.12.0")
+					if err != nil {
+						return errors.New("could not validate version")
+					}
+					if sv.GT(maxVersion) {
```
CecileRobertMichon: this should be is Greater or Equal (`.GE`), or is not Less Than (`! .LT`) since 1.12.0 is included in "not supported" cases.

pkg/api/addons.go
jackfrancis: The coupling here is between the kubelet runtime `--cluster-domain` configuration, and the `--domain=` value of the `kube-dns-deployment` container runtime in the cluster. The only possible duplication is the "fallback to default" implementation, which would not be necessary if the kubeletConfig "set default if empty" implementation happened prior to this (in that case we'd simply take the value of kubelet's `--cluster-domain` configuration, trusting that it would be valid).

In any event, because we're pulling from a common, known-working default `DefaultKubernetesClusterDomain`, I think this is actually maintenance-safe going forward. The only risk would be carelessly undoing this default value inheritance, say, changing it to a string literal in one place that was different from the const. I think that's unlikely.

pkg/api/vlabs/validate.go
```diff
@@ -601,6 +601,16 @@ func (a *Properties) validateAddons() error {
 				if to.Bool(addon.Enabled) && a.HasCoreOS() {
 					return errors.New("flexvolume add-ons not currently supported on coreos distro. Please use Ubuntu")
 				}
+			case "kube-dns":
+				if to.Bool(addon.Enabled) {
+					maxVersion, err := semver.Make("1.12.0-alpha.1")
```
jackfrancis: @CecileRobertMichon To be extra paranoid I changed this to `1.12.0-alpha.1` as it's the earliest 1.12 version ever supported by AKS Engine

pkg/api/vlabs/validate.go
```diff
@@ -617,6 +617,16 @@ func (a *Properties) validateAddons() error {
...
+					if err != nil {
+						return errors.New("could not validate version")
+					}
+					if !sv.LT(maxVersion) {
```
CecileRobertMichon: how does this work on upgrades to 1.12 if kube-dns was previously enabled?

jackfrancis: The current code suggests there's no way to get kube-dns (via aks-engine) w/ 1.12 or greater at present, so this validation is meant to carry over the current "non-user-configurable" functionality into the "user-configurable" area. Check out the delta in the `artifacts.go` implementation to validate whether or not I'm correct about that.

CecileRobertMichon: the previous code didn't use the isEnabled apimodel value to determine if the addon was enabled so that would work but the way I understand your changes it the behavior is now to use `isEnabled` as source of truth which means it _is_ configurable now and if it is enabled previously upgrade won't change the value since the apimodel values don't get overwritten unless the addon is missing. I believe this breaks back compat.

CecileRobertMichon: also if we don't want to make it user-configurable we also need to validate that it's enabled for < 1.12

pkg/api/addons.go
```diff
@@ -321,6 +324,16 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 		},
 	}
 
+	defaultKubeDNSAddonsConfig := KubernetesAddon{
+		Name:    KubeDNSAddonName,
+		Enabled: to.BoolPtr(!common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.12.0")),
```
CecileRobertMichon: this is only a default, I don't see the Enabled value getting overriden anywhere for > 1.12 upgrades and/or < 1.12 clusters that have a kube-dns addon defined in the apimodel with Enabled set to `false`.

pkg/api/vlabs/validate.go
jackfrancis: The current code (before this PR) as I read it didn't expose `kube-dns` as an api model addon at all. It was implemented as a k8s addon, but it wasn't implemented as an `aks-engine` addon.

CecileRobertMichon: we could run two tests to prove this works:
1) create 1.11 cluster from previous aks-engine (master branch for eg.) and upgrade to 1.12 with this branch
2) create 1.11 cluster with this branch, upgrade to 1.12 with this branch
3) create 1.11 cluster with this branch with apimodel that has "kube-dns" addon set to `Enabled: false`

pkg/api/addons.go
```diff
@@ -363,6 +377,13 @@ func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
 			o.KubernetesConfig.Addons = appendAddonIfNotPresent(o.KubernetesConfig.Addons, pspAddonsConfig)
 		}
 	}
+
+	if !isKubeDNSSupported(o) && isUpdate {
```
CecileRobertMichon: that's what was missing before, I think this will fix the upgrade scenario.

jackfrancis:
/hold

deploy + upgrade from this branch doesn't actually upgrade agent pools (root cause unknown)

PR #1480, opened 2019-06-13
test: HPA scale down config only works if k8s >= 1.12

jackfrancis:
**Reason for Change**:

**Issue Fixed**:

Two things:

- `--horizontal-pod-autoscaler-downscale-stabilization` and `--horizontal-pod-autoscaler-cpu-initialization-period` were introduced in k8s 1.12, so we can only set that value if the cluster is >= 1.12
- the way the runner is setting the config is only if the api model `KubernetesConfig` is empty, which in practice it rarely is

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/engine/template.go
+++ b/test/e2e/engine/template.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/aks-engine/pkg/api"
+	"github.com/Azure/aks-engine/pkg/api/common"
 	"github.com/Azure/aks-engine/pkg/api/vlabs"
 	"github.com/Azure/aks-engine/pkg/helpers"
 	"github.com/Azure/aks-engine/pkg/i18n"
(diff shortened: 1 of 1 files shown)


test/e2e/engine/template.go
```diff
@@ -161,8 +162,16 @@ func Build(cfg *config.Config, masterSubnetID string, agentSubnetIDs []string, i
...
-	if prop.OrchestratorProfile.KubernetesConfig == nil {
-		prop.OrchestratorProfile.KubernetesConfig = &vlabs.KubernetesConfig{}
+	var version string
+	if prop.OrchestratorProfile.OrchestratorRelease != "" {
+		version = prop.OrchestratorProfile.OrchestratorRelease + ".0"
+	} else if prop.OrchestratorProfile.OrchestratorVersion != "" {
+		version = prop.OrchestratorProfile.OrchestratorVersion
+	}
+	if common.IsKubernetesVersionGe(version, "1.12.0") {
+		if prop.OrchestratorProfile.KubernetesConfig == nil {
```
CecileRobertMichon: why do we only want to set a previously nil `KubernetesConfig` when k8s version >= 1.12.0?

jackfrancis: Because only in the >= 1.12.0 scenario do we want the E2E runner to explicitly inject the below `controllerManagerConfig` values into the api model before we generate/deploy. The nil check is just to anticipate that the input api model may have a nil `kubernetesConfig`, and so we have to create it from scratch in such cases in order to define the child `controllerManagerConfig` object.

jackfrancis: The high level purpose here is to optimize the E2E test time without having to explicitly update all of our test api models in order to do so. The default "wait until deciding to HPA scale down" config is 5 minutes, which basically means our "validate that HPA scaled down after load stopped" test takes at least that long. This (valid) configuration just means that our test clusters are configured for a more responsive/aggressive HPA scale down behavior, which will speed up the test runs.

PR #1486, opened 2019-06-17
chore: remove support for Kubernetes 1.9.x

mboersma:
**Reason for Change**:
Removes support for Kubernetes 1.9.x, which stopped getting updates a while ago and has now been removed from AKS. Also removes the few VHD components that were used only by 1.9.x.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Codecov didn't report back in this case, but I'd like to know the numbers. I just removed a few test cases that related to 1.9.x, but not any underlying logic that may still test for that. So tl:dr this probably drops unit test coverage and I want to see how much.

mboersma:
Code coverage [dropped by 0.03%](https://codecov.io/gh/Azure/aks-engine/pull/1486?src=pr&el=desc). I think that's acceptable, but I can go spelunking for `>= 1.9` logic in the code if that is preferable.

jackfrancis:
I vote merge now

PR #1488, opened 2019-06-17
refactor: add HyperkubeImageBase, api model cloud context

jackfrancis:
**Reason for Change**:

Add a `HyperkubeImageBase` configuration property to the `KubernetesSpecConfig` cloud context configuration object to allow for hyperkube-specific base image URI configuration.

Modify `ContainerService.GetCloudSpecConfig()` method to enable `ContainerService`-contextual responses. In particular, we want to enable ACR-hosted hyperkube images starting with a particular k8s version.

To minimize the change surface area for Azure Stack, the `HyperkubeImageBase` property inherits from the existing `KubernetesImageBase` property; Azure Stack scenarios will not deviate from their existing implementation, which uses a common "base image URI" across all Kubernetes container images, including hyperkube.

This PR proposes to deprecate the `KubernetesConfig.KubernetesImageBase` user-driven api model configuration property in order to simplify the data merge overhead of reconciling multiple entry points for the `KubernetesImageBase` configuration, as well as tolerating those configuration combinations as relates to backwards compatibility.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jchauncey:
Where are we on being able to get this merged? 

I guess I could build from this branch and use it as a base for testing my release pipelines... If that is what we would like to try.

jackfrancis:
@jchauncey What image builds specifically are you blocked on at this time?

jchauncey:
Any of them. I have the latest builds for every image we utilize in aks-engine (as far as i know). So we can start here with hyperkube/pause and then move on from there.

pkg/api/vlabs/types.go
```diff
@@ -272,7 +272,7 @@ const (
 // KubernetesConfig contains the Kubernetes config structure, containing
 // Kubernetes specific configuration
 type KubernetesConfig struct {
-	KubernetesImageBase             string            `json:"kubernetesImageBase,omitempty"`
+	KubernetesImageBase             string            `json:"kubernetesImageBase,omitempty"` // Deprecated
```
CecileRobertMichon: leaving this here will cause apimodels using this field to not return an error but the value of `KubernetesImageBase` will be ignored silently which might cause error confusion. Let's add a validation: if `KubernetesImageBase` is present, either:
- print a warning that `KubernetesImageBase` is deprecated and will be ignored, use `KubernetesSpecConfig .HyperkubeImageBase` instead or
- return an error explaining that `KubernetesImageBase` is deprecated, use `KubernetesSpecConfig .HyperkubeImageBase`

pkg/api/vlabs/azenvtypes.go
```diff
@@ -35,6 +35,7 @@ type DCOSSpecConfig struct {
 //KubernetesSpecConfig is the kubernetes container images used.
 type KubernetesSpecConfig struct {
 	KubernetesImageBase              string `json:"kubernetesImageBase,omitempty"`
+	HyperkubeImageBase               string `json:"hyperkubeImageBase,omitempty"`
```
CecileRobertMichon: should this also be added to https://github.com/Azure/aks-engine/blob/master/pkg/engine/types.go#L49 ?

docs/topics/clusterdefinitions.md
```diff
@@ -59,7 +59,6 @@ $ aks-engine get-versions
...
 | kubeletConfig                   | no       | Configure various runtime configuration for kubelet. See `kubeletConfig` [below](#feat-kubelet-config)                                                                                                                                                                                                                                                                                                        |
-| kubernetesImageBase             | no       | Specifies the default image base URL (everything preceding the actual image filename) to be used for all kubernetes-related containers such as hyperkube, cloud-controller-manager, pause, addon-manager, heapster, exechealthz etc. e.g., `k8s.gcr.io/`                                                                                                                                                                                                                                     |
```
CecileRobertMichon: This is also referenced in the Azure Stack docs

CecileRobertMichon: So this change effectively removes the ability to customize Kubernetes components container images that aren't hyperkube and ccm?

pkg/api/types.go
```diff
@@ -1849,8 +1849,23 @@ func (f *FeatureFlags) IsFeatureEnabled(feature string) bool {
...
-	return AzureCloudSpecEnvMap[targetEnv]
+	var customCloudName string
+	if cs.Properties != nil {
+		customCloudName = cs.Properties.GetCustomCloudName()
+	}
+	targetEnv := helpers.GetTargetEnv(cs.Location, customCloudName)
+	spec := AzureCloudSpecEnvMap[targetEnv]
+	// TODO Implement ACR-specific overrides below for public cloud
+	/*
+		if spec.CloudName == AzurePublicCloud {
```
CecileRobertMichon: let's remove this sample code since we most likely don't want to do it this way as discussed in the meeting about this PR (and instead use a function or map).

CecileRobertMichon: (or change the sample commented code to something like: 

`spec.KubernetesSpecConfig.HyperkubeImageBase = cs.Properties.GetHyperkubeImageBase()`

and then implement the image logic in its own func.

)

pkg/api/addons.go
```diff
@@ -11,9 +11,8 @@ import (
 	"github.com/Azure/aks-engine/pkg/api/common"
 )
 
-func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
+func (cs *ContainerService) setAddonsConfig(isUpdate bool, cloudSpecConfig AzureEnvironmentSpecConfig) {
 	o := cs.Properties.OrchestratorProfile
-	cloudSpecConfig := cs.GetCloudSpecConfig()
```
CecileRobertMichon: why this change? I would argue it's cleaner to get the CloudSpecConfig from cs than passing it in as a separate arg.

pkg/api/defaults-kubelet.go
```diff
@@ -79,7 +79,7 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 	defaultKubeletConfig := map[string]string{
 		"--cluster-domain":                    "cluster.local",
 		"--network-plugin":                    "cni",
-		"--pod-infra-container-image":         o.KubernetesConfig.KubernetesImageBase + K8sComponentsByVersionMap[o.OrchestratorVersion]["pause"],
+		"--pod-infra-container-image":         cloudSpecConfig.KubernetesSpecConfig.KubernetesImageBase + K8sComponentsByVersionMap[o.OrchestratorVersion]["pause"],
```
CecileRobertMichon: why not get the cloudSpecConfig from `cs` here? no need to pass in an extra arg, it should be available from this context no?

pkg/api/addons.go
CecileRobertMichon: hmm reading into this more I think I understand why you did this. If we don't want to "init" the cloudSpecConfig multiple times, would a variable work better? `GetCloudSpecConfig()` would check if `cloudSpecConfig` is already defined, if not call `SetCloudSpecConfig` (which would be the logic your current `GetCloudSpecConfig` func has, and then return it. ie:

```
func (cs *ContainerService) GetCloudSpecConfig() AzureEnvironmentSpecConfig {
    if cloudSpecConfig == nil {
        setCloudSpecConfig()
   }
    return cloudSpecConfig

} 
```

CecileRobertMichon: Also, the current code is already calling `GetCloudSpecConfig()` in multiple times in other places so if the idea is to only set the config once you'll need to change all these calls as well.

CecileRobertMichon:
How are we handling back compat? If I deploy a cluster with master before this branch, my `KubernetesConfig.KubernetesImageBase` will be set (either from the default value or from a custom value I set). Will that value simply be ignored? Have you tested upgrade from this branch?

pkg/api/vlabs/types.go
CecileRobertMichon: Probably the first option if we don't want to break upgrade...

PR #1489, opened 2019-06-17
chore: Change Windows default image to Jun 2019

PatrickLang:
**Reason for Change**:

This updates the last-known-good image to the latest Windows patch release. There were some issues that could break DNS in April/May cumulative updates. 

**Issue Fixed**:

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/hold
I was expecting this to have a date on or after the 2nd Tuesday which is June 10, not June 4. I need to make sure this is the right one before merge.

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -247,11 +247,11 @@ const (
 	// DefaultWindowsPublisher sets the default WindowsPublisher value in WindowsProfile
 	DefaultWindowsPublisher = "MicrosoftWindowsServer"
 	// DefaultWindowsOffer sets the default WindowsOffer value in WindowsProfile
-	DefaultWindowsOffer = "WindowsServerSemiAnnual"
+	DefaultWindowsOffer = "WindowsServer"
 	// DefaultWindowsSku sets the default WindowsSku value in WindowsProfile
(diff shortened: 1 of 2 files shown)


pkg/api/const.go
```diff
@@ -247,11 +247,11 @@ const (
 	// DefaultWindowsPublisher sets the default WindowsPublisher value in WindowsProfile
 	DefaultWindowsPublisher = "MicrosoftWindowsServer"
 	// DefaultWindowsOffer sets the default WindowsOffer value in WindowsProfile
-	DefaultWindowsOffer = "WindowsServerSemiAnnual"
+	DefaultWindowsOffer = "WindowsServer"
```
PatrickLang: I don't think this should have changed. Checking with team publishing the images

PatrickLang:
cc @adelina-t 
once this merges, will need to move upstream testing to this image

PatrickLang:
Will need to continue hold. This isn't the right image :(

pkg/api/const.go
PatrickLang: The publishing team intended to change this, so I'm updating the docs to match.

PR #1493, opened 2019-06-18
chore: CoreDNS configmap is not updated during upgrades

skinny:
Upgrading clusters from previous versions to the current CoreDNS version break when this ConfigMap isn't also updated

/cc @mboersma

### DIFF
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -56,7 +56,7 @@ metadata:
   name: coredns
   namespace: kube-system
   labels:
-      addonmanager.kubernetes.io/mode: EnsureExists
+      addonmanager.kubernetes.io/mode: Reconcile
 data:
   Corefile: |
     .:53 {


CecileRobertMichon:
should we consider changing this for other addons as well? The reason I didn't do this in the original PR was https://github.com/Azure/aks-engine/pull/1401#discussion_r288813426 but I think using "EnsureExists" for all config maps might have more unintended side effects... @mboersma thoughts?

skinny:
> should we consider changing this for other addons as well? The reason I didn't do this in the original PR was [#1401 (comment)](https://github.com/Azure/aks-engine/pull/1401#discussion_r288813426) but I think using "EnsureExists" for all config maps might have more unintended side effects... @mboersma thoughts?

I sort of agree with the potential side effects, but the current side effect is that the default deployment of the "critical" CoreDNS stops working after the upgrade. I don't know how many deployments use a custom CoreDNS ConfigMap but this might be a "smaller" issue as with other addons ?

Is there a way to add a "custom script" to execute after a `aks-engine upgrade` command ? That might be handy for these kind of scenarios

mboersma:
I can think of two "potential side effects" from changing **all** the ConfigMaps to `Reconcile`:
- causes a battle with a nanny pod that tries to update the same ConfigMap (we've seen this once)
- removes a user-configurable surface that someone relied upon

The first issue I think was specific to `heapster-nanny`, but there may be other ConfigMaps that are not considered read-only by some component.

The second issue is fuzzier. AFAICT AKS Engine has always tried to be as tolerant as possible of users making changes to their clusters after provisioning, but there's no guarantee in this area. So if someone was in the habit of making site-specific CoreDNS config changes after provisioning (for example), that no longer works.

I think that's acceptable here. It's probably acceptable in general too, but we should evaluate each change if we make a PR to change the others to `Reconcile`.

mo-saeed:
For my point of view, core kubernetes components should always have reconciled configs and just documented so the user is aware of this fact and to be very clear that user can't change these configs. what i see now under the addons besides coredns is "ip-masq-agent" which i believe should be also with reconciled config map.

fazdamoa:
Does coredns-custom work in aks-engine clusters? I can't seem to get it to work (while it works in an aks cluster)

We rely on customized coredns configmap - after this update we can't use custom dns proxies due to reconcile.

fsigalov:
Other reasons not to use Reconcile: we routinely have to update the memory limits of tiller, oms, heapster, etc. If you install the add-on as Reconcile, we cannot update the limits any more.

mo-saeed:
@mboersma @CecileRobertMichon I am not sure if it makes sense to use normal kubernetes deployment instead of add-ons ? I am not sure what are the main reasons of using add-ons not normal deployment ?

PR #1474, opened 2019-06-13
fix(auth): Explicitly import the AAD library to allow cluster upgrades

skinny:
After https://github.com/Azure/aks-engine/pull/506 was closed/merged it's still not possible to upgrade clusters with AAD support enabled. During the upgrade or scaling operation the error

`FATA[0377] Error upgrading cluster: No Auth Provider found for name "azure"`

still appears. This PR fixes that.

PR #1512, opened 2019-06-21
revert: "chore: update client-go and k8s.io vendored code (#1273)"

mboersma:
**Reason for Change**:
Updating k8s.io/client-go in AKS Engine eventually had negative ripple effects when vendoring it into AKS. One path forward may be to revert these updated libraries.

**Issue Fixed**:
See #1273, #506, #1474

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
`make build` won't pass unless I replace the v7.0.0 `azure.go` plugin file with the v10.0.0 version, although the rest of client-go is at v7.0.0. Then of course `dep check` fails, so I had to add `k8s.io/client-go` to the `noverify` list. I've gone through this several times, including starting fresh off master as opposed to `git revert`, and this was the only recipe I could find that works. But something smells bad.

cc: @skinny


Gopkg.toml
```diff
@@ -27,7 +27,13 @@
 required = [
   "github.com/go-bindata/go-bindata/go-bindata",
   # Apply workaround from https://github.com/golang/dep/issues/1799
-  "gopkg.in/fsnotify.v1"
+  "gopkg.in/fsnotify.v1",
+  "k8s.io/client-go/plugin/pkg/client/auth/azure",
```
mboersma: This is redundant with the `import` in engine.go, but it seems more appropriate here.

Gopkg.toml
```diff
@@ -27,7 +27,13 @@
...
+  "k8s.io/klog"
+]
+
+noverify = [
+  "k8s.io/client-go"
```
mboersma: This hack makes `dep check` pass since we have a rogue file under this package.

mboersma:
/hold

We're seeing a similar test failure even with this change, so more investigation is required.

mboersma:
As far as we can tell at this point, the client-go update isn't responsible for the issues we see in AKS testing.

PR #1514, opened 2019-06-21
fix: Set encoding:gzip on cilium BPFFS mount unit

### LINKED ISSUE #1393: Cilium BPFFS mount unit file is incorrectly gzipped
https://github.com/Azure/aks-engine/pull/508 introduced, via node custom data, the systemd unit `/etc/systemd/system/sys-fs-bpf.mount`. This works on the master nodes but on a agent/worker node:

(the following on a cluster built with v0.35.1)
```
azureuser@k8s-main-40037667-vmss000008:~$ cat /etc/systemd/system/sys-fs-bpf.mount
41j0E9/VPRR$0.W,GB3"a=Q!Ƣ1})7V1՝|S-/ƬKOZ}X4

## It's gzippped still!
azureuser@k8s-main-40037667-vmss000008:~$ cat /etc/systemd/system/sys-fs-bpf.mount | gunzip 
[Unit]
Description=Cilium BPF mounts
Documentation=http://docs.cilium.io/
DefaultDependencies=no
Before=local-fs.target umount.target
After=swap.target

[Mount]
What=bpffs
Where=/sys/fs/bpf
Type=bpf

[Install]
WantedBy=multi-user.target
```

It looks like when some refactoring/reorganizing got done in https://github.com/Azure/aks-engine/commit/206328a0270c91b8b3dfa6d859eeff56b2f9c51c a small mistake was introduced.
* The unit file itself is a `getBase64EncodedGzippedCustomScript`  [original-patch](https://github.com/Azure/aks-engine/commit/206328a0270c91b8b3dfa6d859eeff56b2f9c51c#diff-af8d8862e75162bcadadbb574ada9e8bR86); it has now moved [on master to here](https://github.com/Azure/aks-engine/blob/master/pkg/engine/armvariables.go#L167)
* In the `masternodecustomdata.yml` it is marked as being `encoding: gzip` [in 206328a](https://github.com/Azure/aks-engine/commit/206328a0270c91b8b3dfa6d859eeff56b2f9c51c#diff-607db9f9c89e7874bd0ef216c92f7e02R163), still present [on master](https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/masternodecustomdata.yml#L144)
* In the `nodecustomdata.yml` the `encoding: gzip` is missing even though it is the same variable reference. [in 206328a](https://github.com/Azure/aks-engine/commit/206328a0270c91b8b3dfa6d859eeff56b2f9c51c#diff-db62d66d8529cd5128f96f21dae13279R158); still missing [on master](https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/nodecustomdata.yml#L145)

UnwashedMeme:
**Reason for Change**:

Since the mount content that we're including here is gzipped we need
this flag to cloud-init to gunzip it as it is being put in
place. Without this the file is still gzipped and systemd ignores it.

**Issue Fixed**:

Fixes #1393

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/nodecustomdata.yml
+++ b/parts/k8s/cloud-init/nodecustomdata.yml
@@ -160,6 +160,7 @@ write_files:
 {{if HasCiliumNetworkPlugin }}
 - path: /etc/systemd/system/sys-fs-bpf.mount
   permissions: "0644"
+  encoding: gzip
   owner: root
   content: !!binary |
     {{WrapAsVariable "systemdBPFMount"}}
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -13625,6 +13625,7 @@ write_files:
 {{if HasCiliumNetworkPlugin }}
 - path: /etc/systemd/system/sys-fs-bpf.mount
   permissions: "0644"
+  encoding: gzip
   owner: root
   content: !!binary |
     {{WrapAsVariable "systemdBPFMount"}}


UnwashedMeme:
This fixes a small bug that was introduced during refactoring (See #1393 for gory details) where a systemd unit file is gzipped in the cloud-init but not gunzipped when put in place.

/assign @jackfrancis

jackfrancis:
And thanks for enduring all the gore @UnwashedMeme :|

PR #1524, opened 2019-06-25
test: validate expected nodes version during E2E

jackfrancis:
**Reason for Change**:

A PR in-flight has an odd issue w/ nodes not getting upgraded to the expected version. Unfortunately that didn't generate an E2E error.

This PR expects the api model k8s version will be reflected in the node version, for every node in every agent pool, including master nodes.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/node/node.go
+++ b/test/e2e/kubernetes/node/node.go
@@ -65,7 +65,7 @@ type Address struct {
 type Info struct {
 	ContainerRuntimeVersion string `json:"containerRuntimeVersion"`
 	KubeProxyVersion        string `json:"kubeProxyVersion"`
-	KubeletProxyVersion     string `json:"kubeletVersion"`
+	KubeletVersion          string `json:"kubeletVersion"`
 	OperatingSystem         string `json:"operatingSystem"`
 	OSImage                 string `json:"osImage"`
 }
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/node/node.go
```diff
@@ -65,7 +65,7 @@ type Address struct {
 type Info struct {
 	ContainerRuntimeVersion string `json:"containerRuntimeVersion"`
 	KubeProxyVersion        string `json:"kubeProxyVersion"`
-	KubeletProxyVersion     string `json:"kubeletVersion"`
```
jackfrancis: Not sure what happened here, it doesn't make sense for this property to be named `KubeletProxyVersion`, not to mention it deviates from the k8s type def (and it's not being used in the E2E code anywhere).

CecileRobertMichon:
lgtm pending upgrade e2e test to validate this works

mboersma:
/hold

pending confirmation from upgrade e2e test

jackfrancis:
This needs more work as re-running E2E on an existing cluster after upgrade is not testing against the upgraded version, and so the upgrade test fails.

jackfrancis:
validated upgrade

PR #1526, opened 2019-06-25
feat: Add vscode containerized development config

ohthehugemanatee:
**Reason for Change**:
VSCode Insiders has a new (awesome) feature which allows you to develop inside a container, with all of the benefits of local development. Here's [a great explanation](https://www.hanselman.com/blog/VisualStudioCodeRemoteDevelopmentMayChangeEverything.aspx). This PR adds the necessary config to use the same docker container as `make dev`.

**Requirements**:

- [ ] includes documentation (is this necessary? If you have vscode it will prompt you. If you don't, it's invisible.
- [ ] adds unit tests (N/A)
- [ ] tested upgrade from previous version (N/A)

**Notes**:

This is my first time doing this, and I partially did it for learning.! If there's better practice to follow, please let me know.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -9,8 +9,7 @@
 		"--security-opt",
 		"seccomp=unconfined"
 	],
-	"workspaceMount": "type=bind,src=.,dst=/go/src/github.com/Azure/aks-engine",
-  "workspaceFolder": "/go/src/github.com/Azure/aks-engine",
+  "postCreateCommand": "mkdir -p /go/src/github.com/Azure && ln -sf /workspaces/aks-engine /go/src/github.com/Azure/aks-engine",
 	"settings": {
 		"go.gopath": "/go"
 	}
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,6 +1,6 @@
 {
 	"name": "Go for aks-engine",
-	"image": "deis/go-dev",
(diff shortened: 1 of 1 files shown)


.devcontainer.json
```diff
@@ -0,0 +1,16 @@
+{
+	"name": "Go for aks-engine",
+	"image": "deis/go-dev",
```
mboersma: We should use the same image as `make dev` and our CI pipeline do: `quay.io/deis/go-dev:v1.22.3`. It's a pain to have to update the tag, but it avoids surprises.

mboersma:
If you have the "Remote Development" plugin installed, VS Code recognizes the presence of the `.devcontainer.json` file and asks if you want to work inside a go-dev container. Otherwise things are silent.

Should we add a paragraph to the developer documentation about how to activate this, or is that good enough?

ohthehugemanatee:
That's a project policy question. Given that it's an insiders-only feature, I don't know if it's appropriate to say "The easiest way to get started developing with aks-engine is to use VSCode Insiders edition and the remote development plugin". But when the feature makes it to stable, we could definitely say "If you're using Visual Studio Code with the Remote Development extension, you're already set up!"

My 2c.

mboersma:
That makes sense to me given that VS Code Remote Debugging is an insiders preview release.

PR #1527, opened 2019-06-25
chore: replace deprecated extensions/v1beta APIs for 1.16

### LINKED ISSUE #1522: Addons are not started for K8s > 1.16-alpha.0.
K8s 1.16 is deprecating some beta workload APIs [1] , thus all common resources like DaemonSets, Deployments, ReplicaSets will be moved from `extensions/v1beta1` to `apps/v1` by default and the `extensions/v1beta1 `endpoints will not be enabled by default anymore ( See linked PR for all resources changed ).

Since most of the manifests for the addons used by aks-engine use v1beta1 api version ( example [2] ), these resources will not be able to start when using k8s versions newer than 1.16-alpha.0. 

We noticed this issue in the upstream Windows K8s CI [3]. For anyone that might be affected by this ( sig-azure CI ? https://testgrid.k8s.io/sig-azure-master ), the quick workaround is to just reenable the endpoints in the api-server by passing `--runtime-config` flag. The longterm option is of course to change the manifests for the addons.

[1] https://github.com/kubernetes/kubernetes/pull/70672
[2] https://github.com/Azure/aks-engine/blob/master/parts/k8s/addons/kubernetesmasteraddons-kube-proxy-daemonset.yaml
[3] https://testgrid.k8s.io/sig-windows#aks-engine-azure-windows-master-staging

mboersma:
**Reason for Change**:
Future-proofs the addons manifests not to use deprecated beta workload APIs. 

The new API names should work from Kubernetes 1.14 onward, ~~so that's where this change is made~~. In Kubernetes 1.16, the old `extensions/v1beta1` names will be unsupported.

* All resources under `apps/v1beta1` and `apps/v1beta2` - use `apps/v1` instead
* `daemonsets`, `deployments`, `replicasets` resources under `extensions/v1beta1` - use `apps/v1` instead
* `podsecuritypolicies` resources under `extensions/v1beta1` - use `policy/v1beta1` instead
* `networkpolicies` resources under `extensions/v1beta1` - use `networking.k8s.io/v1` instead (**NOTE**: there weren't any `networkpolicies` to convert in AKS Engine, just including here for completeness. See kubernetes/kubernetes#70672.)

**Issue Fixed**:
Fixes #1522

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
So, the issue with putting all of these into the 1.14/ directory is that they will be interpreted as 1.14-only. The implementation as-is for "which manifest to choose" is "if there is a version-specific manifest that matches my desired k8s version, use it, otherwise use the one in the root dir".

Does that match the way we've organized these changes across files?

mboersma:
> they will be interpreted as 1.14-only.

Whoops, so how would we segregate this change into before- and after-1.14 manifests? I can make them the default, but then how would I say "use the older versions for Kubernetes <= 1.13"?

mboersma:
I suppose a more conservative approach would be only to make this change for 1.16 and onward, but that begs the same question eventually.

jackfrancis:
It would be to create a pre-1.14 directory for each manifest.

Alternatively, because this change is apparently valid for all relevant k8s versions, we could just change the existing manifests.

mboersma:
After refactoring, this should now represent the minimum set of changes to keep just Kubernetes 1.16 from choking on `extensions/v1beta1` in addon manifests.

Is that worth merging in advance, or should I keep it ready until there is a testable alpha release of Kubernetes 1.16?

jackfrancis:
I vote for merging this in advance. @adelina-t if we build hyperkube from upstream HEAD right now does that have the relevant 1.16 changes?

@CecileRobertMichon FYI and thoughts welcome

CecileRobertMichon:
Does this mean we'll have to make a 1.x dir for every new minor k8s version after 1.16?

+1 on merging in advance

jackfrancis:
@CecileRobertMichon either that or before 1.17 lands we'd create pre-1.14 directories for every component.

I have a preference towards a per-version manifest definition. It will require more boilerplate when we introduce a new version, and additional overhead for making general changes to manifests that span versions; but it is arguably more expressive and encourages per-version curation, which is arguably the right way to maintain these manifests.

mboersma:
> a per-version manifest definition

As in duplicating the entire set of addon manifests for each new version, even if they're the same as before?

That obviously requires more maintenance, but removes any confusion about the source of a manifest. I'm in favor of that tradeoff.

PR #1529, opened 2019-06-25
feat: Set PlatformFaultDomainCount value as 3 for Azure Stack

### LINKED ISSUE #1528: feat: Set PlatformFaultDomainCount as 3 for Azure Stack case
Currently PlatformFaultDomainCount variable is not set causing default value as 2 but it should be 3

rjaini:
**Reason for Change**:
Currently fault domain count value is set using location condition based on the region. 
But in Azure Stack location value is not constant so we can't derive based on that.
Hence we setting PlatformFaultDomainCount value in SetAzureStackCloudSpec

**Issue Fixed**:
Fixes #1528 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -549,7 +549,7 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName 
 			faultDomainCount := 3
 			profile.PlatformFaultDomainCount = &faultDomainCount
 		}
-		
+
 		// Accelerated Networking is supported on most general purpose and compute-optimized instance sizes with 2 or more vCPUs.
 		// These supported series are: D/DSv2 and F/Fs // All the others are not supported
 		// On instances that support hyperthreading, Accelerated Networking is supported on VM instances with 4 or more vCPUs.
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -483,6 +483,12 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName
 	if nil == p.MasterProfile.CosmosEtcd {
 		p.MasterProfile.CosmosEtcd = to.BoolPtr(DefaultUseCosmos)
 	}
+
+	// Update fault domain value to 3 for Azure Stack
+	if p.IsAzureStackCloud() && p.MasterProfile.PlatformFaultDomainCount == nil {
+		faultDomainCount := 3
+		p.MasterProfile.PlatformFaultDomainCount = &faultDomainCount
+	}
 }
 
 // setVMSSDefaultsForMasters
@@ -538,6 +544,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName 
 			profile.OSType = Linux
 		}
 
+		// Update fault domain value to 3 for Azure Stack
+		if p.IsAzureStackCloud() && profile.PlatformFaultDomainCount == nil {
+			faultDomainCount := 3
+			profile.PlatformFaultDomainCount = &faultDomainCount
+		}
+		
 		// Accelerated Networking is supported on most general purpose and compute-optimized instance sizes with 2 or more vCPUs.
 		// These supported series are: D/DSv2 and F/Fs // All the others are not supported
 		// On instances that support hyperthreading, Accelerated Networking is supported on VM instances with 4 or more vCPUs.
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -2272,7 +2272,7 @@ func TestSetCustomCloudProfileEnvironmentDefaults(t *testing.T) {
 		},
 	}
 
-	//test setCustomCloudProfileDefaults with protal url
+	//test setCustomCloudProfileDefaults with portal url
 	mockCS := getMockBaseContainerService("1.11.6")
 	mockCS.Properties.CustomCloudProfile = &CustomCloudProfile{
 		PortalURL: "https://portal.testlocation.contoso.com",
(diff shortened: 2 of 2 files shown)


pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -157,6 +157,19 @@ func (p *Properties) SetAzureStackCloudSpec() error {
 			p.CustomCloudProfile.AzureEnvironmentSpecConfig = &azureStackCloudSpec
 		}
 		AzureCloudSpecEnvMap[AzureStackCloud] = azureStackCloudSpec
+
+		// Update Fault domain value to 3 for AzureStack Case
+		// for both Master and Agent pools.
+
+		faultDomaincount := 3
```
jackfrancis: Will these changes break upgrade scenarios for existing Azure Stack clusters? I don't think you can re-deploy an ARM template that has, say, a FD count of 2, and change it to 3.

rjaini: Yes. Good point. Do you have answer how did you handle similar situation ?. 
As if I recall correctly previously faultdomain value was hardcoded to 2 and now its dynamic based on the region (in some region now this value became 3 compared to 2 previously).

mboersma: Now in aks-engine, when you provision a new VMAS cluster it uses that ARM formula that calculates the FD count for the region. When upgrading, the existing FD count is looked up and used, because the count can never be changed after creation.

mboersma: So if this is just setting defaults for an Azure Stack cluster, and the upgrade code follows the same code path, then this change should be fine. New clusters will start using 3 instead of 2 (I assume Azure Stack's VMAS implementation is always ok with 3), and existing clusters will respect 2 if that's how they were provisioned.

Is there some pattern to the `location` field in Azure Stack? It's also possible we could modify the ARM formula to take Azure Stack into account.

rjaini: Thx for the update @mboersma. Looks like on upgrade flow Azure Stack is not looking up existing value. This can cause failure. Let me take a look on Azure Stack upgrade flow.

I looked at the code where ARM formula was getting modify but in Azure Stack based on location we cant say what is the fault domain value. Let me check with team if their is any better way to derive fault domain value.

rjaini: @mboersma/ @jackfrancis : After adding check extra check in updated PR things are working for upgrade scenario as well for existing customer :). 
Please let me know now you are fine with new changes.

@honcao @jadarsie: Please do review and let me know your comments.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -157,6 +157,22 @@ func (p *Properties) SetAzureStackCloudSpec() error {
 			p.CustomCloudProfile.AzureEnvironmentSpecConfig = &azureStackCloudSpec
 		}
 		AzureCloudSpecEnvMap[AzureStackCloud] = azureStackCloudSpec
+
+		// Update fault domain value to 3 for Azure Stack
+		// on both Master and Agent pools.
+		faultDomaincount := 3
+		if p.MasterProfile != nil && p.MasterProfile.PlatformFaultDomainCount == nil {
+			p.MasterProfile.PlatformFaultDomainCount = &faultDomaincount
```
honcao: since this change is specific for Azure Stack, please add the condition that only set the value when it is for Azure Stack.
You can use p.IsAzureStackCloud()

pkg/api/defaults.go
```diff
@@ -538,6 +544,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName
 			profile.OSType = Linux
 		}
 
+		// Update fault domain value to 3 for Azure Stack
+		if p.IsAzureStackCloud() && profile.PlatformFaultDomainCount == nil {
```
honcao: could you please explain me why do we need check the nil? 
in my mind, as long as it is for azure stack, we will set the fault domain count.

pkg/api/defaults.go
```diff
@@ -483,6 +483,12 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName
 	if nil == p.MasterProfile.CosmosEtcd {
 		p.MasterProfile.CosmosEtcd = to.BoolPtr(DefaultUseCosmos)
 	}
+
+	// Update fault domain value to 3 for Azure Stack
+	if p.IsAzureStackCloud() && p.MasterProfile.PlatformFaultDomainCount == nil {
+		faultDomainCount := 3
```
honcao: please add unit test to verify this.

pkg/api/defaults.go
```diff
@@ -538,6 +544,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName
 			profile.OSType = Linux
 		}
 
+		// Update fault domain value to 3 for Azure Stack
+		if p.IsAzureStackCloud() && profile.PlatformFaultDomainCount == nil {
+			faultDomainCount := 3
```
honcao: please add unit test to verify this.

pkg/api/defaults.go
rjaini: This is needed to support upgrade of old deployments. Since we are in preview we can the check.

pkg/api/defaults.go
rjaini: Sure.

pkg/api/defaults.go
rjaini: Sure.

rjaini:
@honcao @mboersma: Please review unit test case as well.

PR #1533, opened 2019-06-26
fix: --anonymous-auth and --client-ca-file in insecure kubelet config

jackfrancis:
**Reason for Change**:

--anonymous-auth and --client-ca-file kubelet configuration enforcement is not responding to disabled EnableSecureKubelet properly

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -29,17 +29,32 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		"--tls-private-key-file":        "/etc/kubernetes/certs/kubeletserver.key",
 	}
 
+	for key := range staticLinuxKubeletConfig {
+		switch key {
+		case "--anonymous-auth", "--client-ca-file":
+			if !to.Bool(o.KubernetesConfig.EnableSecureKubelet) { // Don't add if EnableSecureKubelet is disabled
+				staticLinuxKubeletConfig[key] = ""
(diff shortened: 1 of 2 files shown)


pkg/api/defaults-kubelet.go
```diff
@@ -55,7 +70,6 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 	staticWindowsKubeletConfig["--cgroups-per-qos"] = "false"
 	staticWindowsKubeletConfig["--enforce-node-allocatable"] = "\"\"\"\""
 	staticWindowsKubeletConfig["--system-reserved"] = "memory=2Gi"
-	staticWindowsKubeletConfig["--client-ca-file"] = "c:\\k\\ca.crt"
```
CecileRobertMichon: was this a bug? ie. previous behavior was to always override `--client-ca-file` to ` "c:\\k\\ca.crt"` for Windows, now behavior is to only set `--client-ca-file` to ` "c:\\k\\ca.crt"` if `EnableSecureKubelet` is true for Windows.

jackfrancis: Yes, the `--client-ca-file` has always been delivered in concert w/ the `EnableSecureKubelet` configuration option since the beggining:

Reference:

https://github.com/Azure/acs-engine/pull/2050

PR #1544, opened 2019-06-27
chore: include https://fqdn:443 for AKS upgrade

jackfrancis:
**Reason for Change**:

Tests suggest that AKS upgrade implementation is failing after client-go update to v10 because the k8s api client object isn't aware that its apiserver is listening on SSL + TCP 443

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgradeagentnode.go
+++ b/pkg/operations/kubernetesupgrade/upgradeagentnode.go
@@ -52,7 +52,7 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 	var kubeAPIServerURL string
 
 	if kan.UpgradeContainerService.Properties.HostedMasterProfile != nil {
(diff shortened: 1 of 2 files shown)


pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -52,7 +52,8 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 	var kubeAPIServerURL string
 
 	if kan.UpgradeContainerService.Properties.HostedMasterProfile != nil {
```
CecileRobertMichon: Can you add unit test coverage for this and the below if/else as well?

jackfrancis: Not easily, no

PR #1545, opened 2019-06-27
docs: make build before make clean in release docs

jackfrancis:
**Reason for Change**:

We should run `make build && make clean` before cutting a release to validate that the release branch we're cutting from has rationalized generated files.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/community/release-checklist.md
+++ b/docs/community/release-checklist.md
@@ -87,7 +87,7 @@ Save the markdown that it prints so it can be pasted into the GitHub release.
 Make sure your repository has no local changes, then build the aks-engine distribution archives:
 
 ```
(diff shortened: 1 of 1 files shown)


docs/community/release-checklist.md
```diff
@@ -87,7 +87,7 @@ Save the markdown that it prints so it can be pasted into the GitHub release.
 Make sure your repository has no local changes, then build the aks-engine distribution archives:
 
 ```
-$ make info  # check that the git tree state is clean and the tag is correct
+$ make build && make info  # check that the git tree state is clean after a build, and that the tag is correct
```
CecileRobertMichon: make generate is less heavy and does the job

PR #1548, opened 2019-06-28
[WIP] fix: don't use deprecated kubernetes.io node labels 

### LINKED ISSUE #1546: kubernetes.io labels set via --node-label are unsupported for 1.16

mboersma:
**Reason for Change**:
Removes and replaces uses of deprecated `kubernetes.io` labels, some of which will cause errors in Kubernetes 1.16.

**Issue Fixed**:
Fixes #1546

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
The [legal names spec](https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/0000-20170814-bounding-self-labeling-kubelets.md#implementation-timeline) seems to preclude the use of `node-role.kubernetes.io/master` as well, but `kubeadm` is [still using it](https://github.com/kubernetes/kubernetes/blob/6a277e0c4dac8cce4a69cad91fcc7c65de32688c/cmd/kubeadm/app/phases/markcontrolplane/markcontrolplane.go#L29)... Should we switch this to `node.kubernetes.io/role=[master|agent]`?


pkg/api/common/helper.go
```diff
@@ -239,8 +239,7 @@ func IsSgxEnabledSKU(vmSize string) bool {
 // GetMasterKubernetesLabels returns a k8s API-compliant labels string
 func GetMasterKubernetesLabels(rg string) string {
 	var buf bytes.Buffer
-	buf.WriteString("kubernetes.io/role=master")
-	buf.WriteString(",node-role.kubernetes.io/master=")
+	buf.WriteString("node-role.kubernetes.io/master=")
```
adelina-t: I believe node-role is not an allowed prefix for kubernetes.io namespace. https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/well_known_labels.go#L30-L32

pkg/api/types.go
```diff
@@ -1418,7 +1418,7 @@ func (a *AgentPoolProfile) IsUbuntuNonVHD() bool {
 func (a *AgentPoolProfile) GetKubernetesLabels(rg string) string {
 	var buf bytes.Buffer
 	buf.WriteString("node-role.kubernetes.io/agent=")
-	buf.WriteString(fmt.Sprintf(",kubernetes.io/role=agent,agentpool=%s", a.Name))
+	buf.WriteString(fmt.Sprintf(",agentpool=%s", a.Name))
```
adelina-t: ditto

pkg/api/common/helper.go
mboersma: Right–I asked a question in the PR summary about how we wanted to proceed here, given that `kubeadm` is still using `node-role.kubernetes.io`. I'm inclined to use `node.kubernetes.io/role=master`.

mboersma:
I reworked this PR to address only the `node-role.kubernetes.io` and `kubernetes.io/role` changes to `node.kubernetes.io/role=`. Assuming that works, I'll add on the more straightforward changes for `beta.kubernetes.io/[os|arch]`.

pkg/api/common/helper.go
khenidak: `node.kubernetes.io` is the correct prefix

https://github.com/kubernetes/kubernetes/commit/36ee154243669ae78e7894bf3633282a8a924dca#diff-497c4bfcc48105a7294fd421d66049cfL27

justinsb: I don't believe it is, sadly.  node-role.kubernetes.io was the replacement for node.kubernetes.io/role.   Moreover the intent of the kubelet label lockdown is to prevent labels that steer workloads from being set by the kubelet.

The KEP does not define _how_ to achieve this in a secure way.  But I would say that node.kubernetes.io/role is not what was intended...

feiskyer:
@mboersma any updates on the fix?

pkg/api/common/helper.go
mboersma: I understand that the intent was to disallow this type of label in general, but it would be highly preferable from this project's point of view if there were a transitional replacement label we could use for now. AKS Engine still has to support Kubernetes 1.10+ and every change may have ripple effects into the AKS service, so the fewer conditional behaviors and tests we can write, the better.

`node.kubernetes.io/role` seems to be allowed. If I can work through the remaining issues in this PR, would it be legit to use that, then create a work item to replace labels in this context altogether? If not, I think I would make the label creation and test expectations all skip it if `kubernetesVersion >= 1.16.0`.

mboersma:
@feiskyer I just got back from vacation this morning ⛰ 🌞 😄, so the status is unfortunately the same: it's failing on an ILB E2E test, waiting for a public IP address to be assigned.

Based on the conversation above, I don't know if we should pursue this change as-is. The other approach might be not to apply these labels if Kubernetes is 1.16+. But I don't know how essential this master/agent scheduling functionality is to our cluster provisioning and tests–we might have to replace it with something other than labels rather than drop it.

pkg/api/common/helper.go
mboersma: And it's possible I misunderstand, but one of the goals of the KEP is:

> - provide a place under the kubernetes.io label namespace for kubelets to self-label with kubelet and node-specific labels

And `[*.]node.kubernetes.io/*` is specifically allowed. That seems to approve of `node.kubernetes.io/role`,  but I don't have much other context here so I'd be grateful for more guidance.

justinsb: You can still add the labels - you just can't add them from kubelet.  The idea AIUI is that kubelet can only add "unimportant" labels, and not labels that could be used to steer important pods.

I believe we need to write a controller that watches Nodes, and labels them in a secure way by consulting an external source of truth i.e. the cloud APIs, or by some other verification of the node.

Is this right @liggitt?

liggitt: Correct. If you want to add `node-role.kubernetes.io/...` labels to nodes, that needs to be done by something other than the kubelet itself. Something that checks metadata attributes (which the nodes cannot modify) and adds appropriate labels to the nodes would make sense to me. Running that centrally as a controller with specific permission to patch nodes would also make sense.

mboersma:

Issue #1553, opened 2019-07-01
UAE North and UAE Central regions are missing from certificate host names


hkamel:
**Describe the bug**
When trying to provision a new cluster on UAE regions, I'm getting an invalid certificate as uaenorth and uaecentral regions DNS prefix are not added to the certificate allowed hosts
**Steps To Reproduce**
Provision a cluster in UAE North and try to check cluster info using kubectl cluster-info

You will get an error 

[...]

PR #1554, opened 2019-07-01
fix: UAE North and UAE Central regions are missing from certificate host names

hkamel:
Adding UAE regions to the list to fix issue #1553

**Reason for Change**:
Adding newly GA regions UAE regions (north and central) to the list of regions

**Issue Fixed**:
#1553

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


jackfrancis:
@mboersma Is this something that needs to be changed via automated code?

hkamel:
> @mboersma Is this something that needs to be changed via automated code?

I was under the assumption that we are using Azure ARM APIs to retrieve a list of region but found all of them defined in the const file. dynamically pull the regions will be great

pkg/helpers/azureconst.go
```diff
@@ -61,6 +61,8 @@ func GetAzureLocations() []string {
 		"usgovarizona",
 		"usgovtexas",
 		"francecentral",
+		"uaenorth",
```
CecileRobertMichon: This file is auto generated and shouldn't be edited manually, you can add the new regions at  https://github.com/Azure/aks-engine/blob/ce3f9884c0d9c579f5a8ef78749b2d100d322298/pkg/helpers/generate_azure_constants.py#L107 instead and run `make generate`

mboersma:
@jackfrancis running `./pkg/helpers/generate_azure_constants.py` adds `uaenorth` and `ueacentral` to `azureconst.go`. So I think it's sane still, we just need to run it regularly to see if there's a delta.

It only modifies that file, so the test additions needed to be written manually. Thanks @hkamel!

pkg/helpers/azureconst.go
CecileRobertMichon: My bad, `generate_azure_constants.py` shouldn't be changed manually -- see @mboersma's comment above he got it right. 

In any case, the regions shouldn't be added manually into the file either, but running `make generate-azure-constants` should add them to azureconst.go and generate the same delta you have (if that's not already what you did).

mboersma: I don't think we need to modify the python script: it will pick up the two new regions as-is. That's a list of special regions that might not be visible depending on subscription. This PR already has the correct changes the python script would produce (except for some new VM types, but that doesn't need to be part of this addition).

hkamel:
@CecileRobertMichon I'm fixing the format issue which break unit test, please give me a minute

mboersma:
> dynamically pull the regions will be great

That implies replacing the code constants with the API equivalent of `az account list-locations`, which may not be ok in all cases. For example, I don't think `aks-engine generate` makes any network connections in the current implementation (and maybe some users expect that behavior).

hkamel:
> > dynamically pull the regions will be great
> 
> That implies replacing the code constants with the API equivalent of `az account list-locations`, which may not be ok in all cases. For example, I don't think `aks-engine generate` makes any network connections in the current implementation (and maybe some users expect that behavior).

That make sense ... thanks for the clarifications

Issue #923, opened 2019-04-01
Question about size of etcd disk default size


yaron-idan:
**Is this a request for help?**:
No
---

**Is this an ISSUE or FEATURE REQUEST?** (choose one):
ISSUE
---

**What version of aks-engine?**:
0.32.3
---

**Kubernetes version**:
1.11.8

**What happened**:
We realised clusters are turning up with 1024GB etcd disks, and then found out this is expected behaviour according to the docs, even though this seems like a very large disk for ~20 nodes.

**What you expected to happen**:
According to the [etcd docs](https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware.md#xlarge-cluster), 60GB is an adequate size for very large clusters. So I'm wondering what makes aks-enging provision such a large disk by default.
Furthermore, the etcd docs stress that the critical factor for these disks is speed, so I was also looking for the possibility of using SSD drives for etcd, but couldn't find this config in aks-engine api-model.

**How to reproduce it** (as minimally and precisely as possible):
Spin up a cluster with default values.

CecileRobertMichon:
For context, here's the PR that set those defaults: https://github.com/Azure/acs-engine/pull/2435

@khenidak to see if he has any additional guidance.

yaron-idan:
Not sure I'm following...
The PR seems to indicate that the bigger disks are allocated to provide better IOPS ratio, but looking at the [pricing page for managed disks](https://azure.microsoft.com/en-us/pricing/details/managed-disks/), it seems only SSD are getting better IOPS as their size increase - while HDD are always marked as `up to 500`.
Am I missine something here?

zhiweiv:
In my k8s cluster,  there are 7 nodes and about 60 pods, I allocated only 60G for ectd disk. In azure monitor, Avg Data Disk Queue Depth/Avg Data Disk Write Operations/Sec of ectd disk are very low. 

Back to the doc https://github.com/etcd-io/etcd/blob/master/Documentation/op-guide/hardware.md, 1500~3000 IOPS could handle up to 50 nodes, that fits my case, 500 IOPS is enough for 5 nodes.

In https://github.com/Azure/acs-engine/issues/2510, it used etcd 2.x, aks-e uses 3.x now, maybe it improved performance, I think aks-e may allocated too aggressive ectd disk size now.

yaron-idan:
Thanks @zhiweiv, this is valuable input.
[...]

PR #1558, opened 2019-07-02
feat: Set EtcdDiskSizeGB max default size to 1023 if target platform is Azure Stack

### LINKED ISSUE #1557: feat: Set EtcdDiskSizeGB max default size to 1023 if target platform is Azure Stack
Set EtcdDiskSizeGB max default size to 1023 if target platform is Azure Stack as the max managed disk size supported on Azs is 1023GB.

rjaini:
**Reason for Change**:
Azure Stack only supports max disk size to be 1023GB. So in case when the number of nodes increases
we cant increase the size to 1024 or 2048GB.

**Issue Fixed**:
Fixes #1557 

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -252,6 +252,9 @@ const (
 
 	// DefaultAzureStackFaultDomainCount set to 3 as Azure Stack today has minimum 4 node deployment.
 	DefaultAzureStackFaultDomainCount = 3
+
+	// DefaultAzureStackEtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
+	DefaultAzureStackEtcdDiskSizeGT10Nodes = "1023"
 )
 
 // WindowsProfile defaults
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -253,8 +253,8 @@ const (
 	// DefaultAzureStackFaultDomainCount set to 3 as Azure Stack today has minimum 4 node deployment.
 	DefaultAzureStackFaultDomainCount = 3
 
-	// DefaultAzureStackEtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
-	DefaultAzureStackEtcdDiskSizeGT10Nodes = "1023"
+	// MaxAzureStackManagedDiskSize = size for Kubernetes master etcd disk volumes in GB if > 10 nodes as this is max what Azure Stack supports today.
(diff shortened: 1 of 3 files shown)


pkg/api/defaults.go
```diff
@@ -233,9 +233,19 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		if "" == a.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB {
 			switch {
 			case a.TotalNodes() > 20:
-				a.OrchestratorProfile.KubernetesConfig.EtcdDiskSizeGB = DefaultEtcdDiskSizeGT20Nodes
+				if a.IsAzureStackCloud() {
+					// Currently on Azure Stack max size of managed disk size is 1023GB.
```
honcao: @rjaini do we know how much disk is required for 20 + nodes?  The cluster will be not usable if the disc is out of space.
Another alternative is to limit the total count of agent nodes.
And we should raise the request to Disk RP on Azure Stack to support larger size of disk.

rjaini: I am not sure how much disk space requirements. But from https://github.com/Azure/aks-engine/issues/923 thread it looks like the 1024 size itself is very high.
I will check with Disk RP team to understand the reason of 1023 limit.

@mboersma : Do you know why we have such large disk size for higher number of nodes in AKS Engine ?.

pkg/api/const.go
```diff
@@ -252,6 +252,9 @@ const (
 
 	// DefaultAzureStackFaultDomainCount set to 3 as Azure Stack today has minimum 4 node deployment.
 	DefaultAzureStackFaultDomainCount = 3
+
+	// DefaultAzureStackEtcdDiskSizeGT10Nodes = size for Kubernetes master etcd disk volumes in GB if > 10 nodes
+	DefaultAzureStackEtcdDiskSizeGT10Nodes = "1023"
```
CecileRobertMichon: the naming "GT10Nodes" doesn't really apply here since there's only one default size

pkg/api/defaults.go
CecileRobertMichon: Matt is out on vacation but see https://github.com/Azure/acs-engine/pull/2435 for context.

pkg/api/const.go
rjaini: My Bad. Updating name to MaxAzureStackManagedDiskSize to be consistent to another PR.

PR #1560, opened 2019-07-02
feat: support CoreOS-like Linux distributions

### LINKED ISSUE #1559: Support CoreOS-like Linux distributions
**Describe the request**

I would like AKS Engine to support Linux distros, which are similar to CoreOS (which set `ID_LIKE=coreos` in `/etc/os-release`).

Flatcar Linux is compatible with CoreOS, so CoreOS provisioning method works just fine on it.

As what will happen with CoreOS by the end of the year is uncertain, it would be great to keep it's support in AKS Engine and make it supports similar alternatives, like [Flatcar Linux](https://www.flatcar-linux.org/).

Since AKS Engine supports CoreOS already, only OS detection part needs to be adjusted to respect `ID_LIKE` variable in `/etc/os-release` file.

**Additional context**

Similar issue: hashicorp/vagrant#9600.

invidian:
**Reason for Change**:

Distributions like Flatcar Linux sets `ID_LIKE=coreos` in `/etc/os-release` file.

This PR changes the way we set OS variable.

`sort -r` is used to always prefer ID_LIKE over ID, since `gawk` only
prints first line.

`/^ID(|_LIKE)=(.*)/` matches both `ID=` and `ID_LIKE=` and extracts the
value after `=` sign.

`print toupper(a[2]); exit` prints uppercased value and exits, so only
first line is always printed.

**Issue Fixed**:

Closes #1559 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -53,7 +53,7 @@ ERR_CIS_ASSIGN_FILE_PERMISSION=112 # Error assigning permission to a file in CIS
 ERR_CIS_COPY_FILE=113 # Error writing a file to disk for CIS enforcement
 ERR_CIS_APPLY_PASSWORD_CONFIG=115 # Error applying CIS-recommended passwd configuration
 
-OS=$(cat /etc/*-release | grep ^ID= | tr -d 'ID="' | awk '{print toupper($0)}')
+OS=$(sort -r /etc/*-release | gawk 'match($0, /^ID(|_LIKE)=(.*)/, a) { print toupper(a[2]); exit }')
 UBUNTU_OS_NAME="UBUNTU"
 RHEL_OS_NAME="RHEL"
 COREOS_OS_NAME="COREOS"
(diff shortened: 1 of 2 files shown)


alexeldeib:
(not an approver, just a curious contributor)

@invidian did you test building a cluster? was curious about flatcar after coreos => rhel stuff

invidian:
>@invidian did you test building a cluster?

Do you mean if I tested the deployment? Yes, I did, on Ubuntu, Coreos and Flatcar and everything works as expected.

> was curious about flatcar after coreos => rhel stuff

What do you mean exactly?

alexeldeib:
I meant the Project Atomic/CoreOS merger under Red Hat. Anyway,  very cool :)

invidian:
>I meant the Project Atomic/CoreOS merger under Red Hat, and the proliferation of CoreOS-like OS. 

Ah, so our company, Kinvolk will continue to support and develop Flatcar Linux, regardless what happens with CoreOS. For now we follow upstream with patches + we have our own Edge channel with experimental software.

invidian:
Thanks for quick review and merge @CecileRobertMichon!

PR #1561, opened 2019-07-02
fix: old VMSS AKS cluster cannot be upgraded after April 2

chengliangli0918:
**Reason for Change**:
Old VMSS AKS cluster cannot be upgraded after April 2

AKS-Engine changed AKS billing extension type from Compute.AKS-Engine.Linux.Billing to Compute.AKS.Linux.Billing in this PR (https://github.com/Azure/aks-engine/commit/bc462f1c49d5fc9bd93b9b7910504019853e4db9). The change was vendored into AKS on Apr. 2nd.  So any VMSS cluster created before the change will hit this problem during upgrade. New clusters should be fine.

**Issue Fixed**:
Rename the billing extension to make the ARM deployment happy

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -688,6 +688,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		}
 
 		if cs.Properties.IsHostedMasterProfile() {
+			aksBillingExtension.Name = to.StringPtr(fmt.Sprintf("[concat(variables('%sVMNamePrefix'), '-AksLinuxBilling')]", profile.Name))
 			if profile.IsWindows() {
 				aksBillingExtension.Type = to.StringPtr("Compute.AKS.Windows.Billing")
 			} else {
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -688,10 +688,11 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		}
 
 		if cs.Properties.IsHostedMasterProfile() {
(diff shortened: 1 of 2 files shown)


pkg/engine/virtualmachinescalesets.go
```diff
@@ -688,6 +688,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		}
 
 		if cs.Properties.IsHostedMasterProfile() {
+			aksBillingExtension.Name = to.StringPtr(fmt.Sprintf("[concat(variables('%sVMNamePrefix'), '-AksLinuxBilling')]", profile.Name))
```
CecileRobertMichon: is this going to affect how we track billing data in any way?

`Aks` should be `AKS`

What about Windows? This name contains "linux" and will be applied to all extensions, even Windows ones so it might be confusing.

CecileRobertMichon: I see, before the extension name was `computeAksLinuxBilling`. This might break how we track these extensions if we query by name `computeAksLinuxBilling`.

chengliangli0918: Actually before the extension name was set to <vmssname>-computeLinuxBilling, e.g. aks-nodepool1-40022152-vmss-computeAksLinuxBilling. Don't think this name that varies on different VMSS is used to track billing data.

CecileRobertMichon:
/hold

verifying the impact of the change on the billing tracking

pkg/engine/virtualmachinescalesets.go
yangl900: good point. @CecileRobertMichon we'd like to get this merged since this broke existing VMSS clusters. If billing tracking is broken, we take the risk.

PR #1564, opened 2019-07-02
feat: Add checks for Metadata, AcceleratedNetworking and etcdDiskSizeGB on Azure Stack to fail fast.

### LINKED ISSUE #1563: feat: Add checks for Metadata, AcceleratedNetworking and etcdDiskSizeGB which are not supported on Azure Stack to fail fast.
feat: Add checks for Metadata, AcceleratedNetworking and etcdDiskSizeGB which are not supported on Azure Stack to fail fast.

rjaini:
**Reason for Change**:
On Azure Stack we dont support certain flags and expected to be in certain value or range.
This change add validation so that proper error is returned early and fail fast.

**Issue Fixed**:
Fixes #1563 

**Requirements**:
- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/vlabs/const.go
+++ b/pkg/api/vlabs/const.go
@@ -134,6 +134,8 @@ const (
 const (
 	// AzureStackCloud is a const string reference identifier for Azure Stack cloud
 	AzureStackCloud = "AzureStackCloud"
+	// MaxAzureStackManagedDiskSize is max etcd disk size supported on AzureStackCloud
+	MaxAzureStackManagedDiskSize = 1023
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/api/vlabs/const.go
```diff
@@ -134,6 +134,8 @@ const (
 const (
 	// AzureStackCloud is a const string reference identifier for Azure Stack cloud
 	AzureStackCloud = "AzureStackCloud"
+	// MaxAzureStackManagedDiskSize is max etcd disk size supported on AzureStackCloud
+	MaxAzureStackManagedDiskSize = 1023
```
CecileRobertMichon: how does this change relate to https://github.com/Azure/aks-engine/pull/1558/files?

should this be `MaxAzureStackEtcdDiskSize` ?

CecileRobertMichon: or I guess it's just the general limit imposed by disk RP, in which case the name makes sense

rjaini: Yes this is the general limit imposed by disk RP.

Issue #1550, opened 2019-06-30
Windows nodes don't join cluster if using custom net


msbecker:
**Describe the bug**
I am creating a custom vnet and two subnets for an AKS mixed-node cluster, with one subnet for the master and another for the workers. When I run "kubectl get nodes", it returns the Linux worker nodes, but not the Windows worker nodes. 
I am able to ping the Linux worker node IP addresses from the master node, but not the Windows worker node IP addresses, even though all worker nodes are on the same subnet.

If I deploy without the custom vnet and let aks-engine configure the networking, both the Linux and Windows nodes are returned by the kubectl command, and I am able to deploy workloads to the Linux and Windows nodes.

**Steps To Reproduce**

{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes"
    },
    "masterProfile": {
      "count": 1,
      "dnsPrefix": "",
      "vmSize": "Standard_D2_v3",
      "vnetSubnetId": "/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Network/virtualNetworks/CustomVNET/subnets/MasterSubnet",
      "firstConsecutiveStaticIP": "10.100.0.5"
    },
    "agentPoolProfiles": [
      {
        "name": "linuxpool1",
        "count": 2,
        "vmSize": "Standard_D2_v3",
        "availabilityProfile": "AvailabilitySet",
        "vnetSubnetId": "/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Network/virtualNetworks/CustomVNET/subnets/SUBNET"
      },
      {
        "name": "windowspool2",
        "count": 2,
        "vmSize": "Standard_D2_v3",
        "availabilityProfile": "AvailabilitySet",
        "osType": "Windows",
        "vnetSubnetId": "/subscriptions/SUBSCRIPTION_ID/resourceGroups/RESOURCE_GROUP_NAME/providers/Microsoft.Network/virtualNetworks/CustomVNET/subnets/SUBNET"
      }
    ],
    "windowsProfile": {
      "adminUsername": "<username>",
      "adminPassword": "<password>"
    },
    "linuxProfile": {
      "adminUsername": "<username>",
      "ssh": {
        "publicKeys": [
          {
            "keyData": ""
          }
        ]
      }
    },
[...]

PR #1570, opened 2019-07-03
fix: do not override vnetCidr param when empty string

CecileRobertMichon:
**Reason for Change**:

azuredeploy.json

    "vnetCidr": {
      "defaultValue": "10.0.0.0/8",
      "metadata": {
        "description": "Cluster vnet cidr"
      },
      "type": "string"
    },
azuredeploy.parameters.json overrides this with the empty string which results in an empty vnetCidr.

**Issue Fixed**:

#1550 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/params.go
+++ b/pkg/engine/params.go
@@ -65,7 +65,7 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
 			if properties.MasterProfile.IsVirtualMachineScaleSets() {
 				addValue(parametersMap, "agentVnetSubnetID", properties.MasterProfile.AgentVnetSubnetID)
 			}
-			if properties.OrchestratorProfile.IsKubernetes() {
+			if properties.OrchestratorProfile.IsKubernetes() && properties.MasterProfile.VnetCidr != "" {
 				addValue(parametersMap, "vnetCidr", properties.MasterProfile.VnetCidr)
 			}
 		} else {


CecileRobertMichon:
@PatrickLang @Alwandy @msbecker

CecileRobertMichon:
I have [another attempt](https://github.com/Azure/aks-engine/compare/master...CecileRobertMichon:fix-default-vnet-cidr) at rationalizing the vnetCidr defaults and using apimodel defaults instead of ARM param defaults but it's more complex so I'd rather start with this PR as an easy, straightforward fix.

Alwandy:
@CecileRobertMichon first fix looks good to me, 

2nd fix looks OK, assuming that it will merge with the 1st fix? Or any reason why you won't be setting: 
`if properties.OrchestratorProfile.IsKubernetes() && properties.MasterProfile.VnetCidr != ""` on 2nd fix?

PatrickLang:
I think this fix makes the most sense - we shouldn't set an empty parameter that overrides the variables default.

For the second approach - I don't see why there's an if (windows) type decision tree. We should be passing the same networking parameters down to both Windows & Linux nodes and I would rather aim to get those consistent.

PatrickLang:
this is included in https://github.com/Azure/aks-engine/releases/tag/v0.38.0 :)

PR #1571, opened 2019-07-03
revert: Revert "feat: support CoreOS-like Linux distributions (#1560)"

CecileRobertMichon:
This reverts commit 31e804520e2839b823485ec8975a7a7da8b5722e.

**Reason for Change**:

#1560 broke Ubuntu distro deployments.

Repro steps:

- deploy cluster with `"distro": "ubuntu"`: `{   "apiVersion": "vlabs",   "properties": {     "orchestratorProfile": {       "orchestratorType": "Kubernetes",       "orchestratorRelease": "1.14",   "kubernetesConfig": {         "networkPlugin": "azure", "addons": [           {             "name": "tiller",             "enabled": false           },           {             "name": "aci-connector",             "enabled": false           },           {             "name": "blobfuse-flexvolume",             "enabled": false           },           {             "name": "smb-flexvolume",             "enabled": false           },           {             "name": "keyvault-flexvolume",             "enabled": false           },           {             "name": "kubernetes-dashboard",             "enabled": false           },           {             "name": "azure-cni-networkmonitor",             "enabled": false           }         ]       }     },     "masterProfile": {       "count": 1,       "dnsPrefix": "",       "vmSize": "Standard_D2_v2",       "distro": "ubuntu"     },     "agentPoolProfiles": [       {         "name": "agentpool1",         "count": 2,         "vmSize": "Standard_D2_v2",         "distro": "ubuntu"       }     ],     "linuxProfile": {       "adminUsername": "azureuser",       "ssh": {         "publicKeys": [           {             "keyData": ""           }         ]       }     },     "servicePrincipalProfile": {       "clientId": "",       "secret": ""     }   } }`

The deployment fails with error:

```
Error from deployment for kubernetes-westus2-xxx in resource group kubernetes-westus2-xxx:exit status 1
2019/07/03 17:42:29 Command Output: ERROR: Deployment failed. Correlation ID: xxx. {
  "status": "Failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "VMExtensionProvisioningError",
[...]

PR #1572, opened 2019-07-03
feat: support CoreOS-like Linux distributions

### LINKED ISSUE #1559: Support CoreOS-like Linux distributions

invidian:
**Reason for Change**:

This PR is a follow up to #1571 and #1560.

Distributions like Flatcar Linux sets `ID_LIKE=coreos` in `/etc/os-release` file.

With this PR, if `/etc/os-release` has set `ID_LIKE=coreos` it will be
preferred over `ID=`.

`sort -r` is used to always prefer ID_LIKE over ID, since `gawk` only
prints first line.

`/^(ID_LIKE=(coreos)|ID=(.*))$/` matches either `ID_LIKE=coreos` or regular `ID=` and
extracts the value after '=' sign.

`print toupper(a[2] a[3]); exit` prints uppercased either coreos or regular value
and exits, so only first line is always printed.

**Issue Fixed**:

Closes #1559

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

I didn't have time to test it yet, I only tested it with static `os-release` files.

Signed-off-by: Mateusz Gozdek <mateusz@kinvolk.io>

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -53,7 +53,7 @@ ERR_CIS_ASSIGN_FILE_PERMISSION=112 # Error assigning permission to a file in CIS
 ERR_CIS_COPY_FILE=113 # Error writing a file to disk for CIS enforcement
 ERR_CIS_APPLY_PASSWORD_CONFIG=115 # Error applying CIS-recommended passwd configuration
 
-OS=$(cat /etc/*-release | grep ^ID= | tr -d 'ID="' | awk '{print toupper($0)}')
+OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID_LIKE=(coreos)|ID=(.*))$/, a) { print toupper(a[2] a[3]); exit }')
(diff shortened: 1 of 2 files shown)


invidian:
Hm, I'm not sure why the tests are failing. @CecileRobertMichon could you please look into it and give me some hint?

invidian:
>Hm, I'm not sure why the tests are failing. @CecileRobertMichon could you please look into it and give me some hint?

Seems that everything is passing now, yay :)

PR #1579, opened 2019-07-05
test: improve crashing kube-system pod check

PatrickLang:
**Reason for Change**:
This test case was failing and didn't have enough logging. It wasn't clear which pods were crashing. Ideally none of them should, but what we've seen is that sometimes they restart a few times due to race conditions then stabilize later. It is failing when we're testing private vnets because some kube-system pods had 3 crashes before stabilizing. If 5 isn't enough, maybe this should move up to 10. Otherwise, we could rely on the test case "should have core kube-system componentry running" and cut this one.

It was also marked Windows only, which wasn't needed. There's nothing Windows specific about this test. I also moved it up to the top next to the test case that makes sure the right kube-system pods are running.

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -512,6 +512,18 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 		})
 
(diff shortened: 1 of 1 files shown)


PatrickLang:
related: #1550 - the new tests added are failing, this may fix them

Issue #1536, opened 2019-06-26
Generate network configuration file for Azure CNI on Azure Stack


t-chappl:
**Describe the request**
Azure CNI does not currently work on Azure Stack. The CNI queries wireserver for the network interface configuration which includes a list of available secondary IP addresses, and then assigns IP addresses to new pods based off that list. Wireserver is not available on Azure Stack at this point in time. In the meantime the Azure CNI will read the network configuration from a file. We are working with the Azure CNI team on this change and have a working proof of concept. We need AKS Engine to generate this network configuration file as it provisions each node. 

As soon as wireserver becomes available on Azure Stack AKS Engine will no longer need to generate the file. 

**Explain why AKS Engine needs it**
This will allow Kubernetes to be deployed on Azure Stack using the Azure CNI plugin. 

**Describe the solution you'd like**
We propose generating the network configuration file as part of the custom script extension that runs on each node. An extra function will be added that will generate a token for use with the Azure Resource Manager Rest API, query the API for the configuration of network interfaces in the resource group, select the correct network interface by matching MAC addresses, and then generate the config file that Azure CNI requires using the interface information. The address prefix of the subnet the interface is using will also be queried using the rest API and the subnet ID found in the interface configuration. 

The function will require the following parameters:
- tenant ID 
- service principal client ID
- service principal client secret
- subscription ID
- resource group
- network api version (eg: `2017-10-01`)

All but network api version are already present as environment variables to the setup script, so we will add the network api version as a parameter. The service management endpoint (eg: `https://management.azurestackci03.onmicrosoft.com/tenand_id`), active directory endpoint (eg: `https://login.microsoftonline.com/`), and resource manager endpoint (eg: `https://management.local.azurestack.external/`) are also required but can be pulled from the azure environment file that is already on each node. 

cURL will be used to make the requests for the token and network configuration and the json file that Azure CNI needs will be generated using jq. It will be placed at `/etc/kubernetes/interfaces.json` and will have read only permissions. 

On windows the Invoke-RestMethod command will be used to make requests for the token and network configuration and the json file will be generated using Powershell hashtables and arrays along with the ConvertTo-Json command. It will be placed at `/k/interfaces.json` and will have read only permisions. 

**Describe alternatives you've considered**

**Additional context**

t-chappl:
Also, here is an example of the json file that Azure CNI needs:
```
{
  "Interfaces": [
    {
      "MacAddress": "000D3A6E1825",
[...]

PR #1584, opened 2019-07-08
feat: generate network configuration file for Azure CNI on Azure Stack

t-chappl:
**Reason for Change**:

See https://github.com/Azure/aks-engine/issues/1536

**Issue Fixed**:

https://github.com/Azure/aks-engine/issues/1536

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/windowsazurecnifunc.ps1
```diff
@@ -81,3 +81,98 @@ Set-AzureCNIConfig
...
+        $tokenURL = "$($ActiveDirectoryEndpoint)adfs/oauth2/token"
+    } else {
+        $tokenURL = "$($ActiveDirectoryEndpoint)$TenantId/oauth2/token"
+    }
+
+    $encodedSecret = [System.Web.HttpUtility]::UrlEncode($AADClientSecret)
+
+    $body = "grant_type=client_credentials&client_id=$AADClientId&client_secret=$encodedSecret&resource=$ServiceManagementEndpoint"
+
+    $token = Invoke-RestMethod -Uri $tokenURL -Method Post -Body $body -ContentType 'application/x-www-form-urlencoded' | select -ExpandProperty access_token
```
rjaini: Good to add retry and token validation as we did in Linux case.

parts/k8s/windowsazurecnifunc.ps1
```diff
@@ -81,3 +81,98 @@ Set-AzureCNIConfig
...
+
+    Write-Log "Fetching network interface configuration for node"
+
+    $interfacesUri = "$($ResourceManagerEndpoint)subscriptions/$SubscriptionId/resourceGroups/$ResourceGroup/providers/Microsoft.Network/networkInterfaces/$($NetworkInterface)?api-version=$NetworkAPIVersion"
+    $headers = @{Authorization="Bearer $token"}
+
+    Invoke-RestMethod -Uri $interfacesUri -Method Get -ContentType 'application/json' -Headers $headers -OutFile $nicConfigFile
```
rjaini: Same here. Add config file content validation.

pkg/engine/armvariables.go
```diff
@@ -163,6 +168,12 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 		}
 		masterVars["environmentJSON"] = environmentJSON
 		masterVars["provisionConfigsCustomCloud"] = getBase64EncodedGzippedCustomScript(kubernetesCSECustomCloud)
+
+		if cs.Properties.OrchestratorProfile.IsAzureCNI() {
```
rjaini: Add cs.Properties.IsAzureStackCloud() condition as well.

pkg/engine/template_generator_test.go
```diff
@@ -131,21 +134,21 @@ func TestGetTemplateFuncMap(t *testing.T) {
 	}
 }
 
-func TestIsIdentitySystemADFS(t *testing.T) {
+func TestGetIdentitySystem(t *testing.T) {
 	for _, test := range []struct {
 		desc     string
 		apiModel string
-		isADFS   bool
+		correctResult   string
```
rjaini: Change the variable to expectedResult.

pkg/engine/template_generator_test.go
```diff
@@ -131,21 +134,21 @@ func TestGetTemplateFuncMap(t *testing.T) {
...
+			desc:     "should return adfs when identitySystem is set",
 			apiModel: `{"properties":{"customCloudProfile": {"identitySystem": "adfs"}}}`,
-			isADFS:   true,
+			correctResult:   "adfs",
 		},
 		{
-			desc:     "identitySystem=azure_ad should return false",
-			apiModel: `{"properties":{"customCloudProfile": {"identitySystem": "azure_ad"}}}`,
-			isADFS:   false,
+			desc:     "should return azure_ad when not azure stack",
```
rjaini: this is true for Azure Stack AAD scenario as well. Please reword the desc.

pkg/engine/template_generator_test.go
```diff
@@ -179,6 +182,67 @@ func TestGetBase64EncodedEnvironmentJSON(t *testing.T) {
 	}
 }
 
+func TestGetAzureStackEndpoints(t *testing.T) {
+	masApiModel := `{"properties":{"customCloudProfile":{"environment":{"serviceManagementEndpoint":"a","resourceManagerEndpoint":"b","activeDirectoryEndpoint":"c"}}}}`
```
rjaini: Can we use proper examples rather a,b or c.

pkg/engine/template_generator_test.go
```diff
@@ -179,6 +182,67 @@ func TestGetBase64EncodedEnvironmentJSON(t *testing.T) {
...
+		correctResult string
+	}{
+		{
+			desc: `GetServiceManagementEndpoint should return "" when not azure stack`,
+			apiModelString: getAPIModelString(),
+			method: "GetServiceManagementEndpoint",
+			correctResult: "",
+		},
+		{
+			desc: `GetServiceManagementEndpoint should return "a" when not azure stack`,
```
rjaini: Copy paste strings. Please fix it as appropriate.

pkg/engine/armvariables.go
t-chappl: It is already in an if block for IsAzureStackCloud() starting on line 159

pkg/engine/armvariables.go
```diff
@@ -110,8 +115,8 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
+			kubernetesVersion, isHostedMaster, isIPv6DualStackFeatureEnabled, azureStackCNIParams),
```
rjaini: Do we need comma as azureStackCNIParams has added comma. I assume you have tested the scenario where this value is empty.

pkg/engine/armvariables.go
rjaini: Code snippet didnt show as it was wraped.

pkg/engine/armvariables.go
t-chappl: The unit test in armvariables_test.go tests how this string is generated. I don't add a comma in the original string since we don't want a trailing comma if the insertion string is empty. The insertion string starts with a comma when it isn't empty.

parts/k8s/cloud-init/artifacts/cse_customcloud.sh
```diff
@@ -79,3 +79,78 @@ configureK8sCustomCloud() {
...
+    if [[ ! -s $NETWORK_INTERFACES_FILE ]]; then
+        echo "Error fetching network interface configuration for node"
+        exit $ERR_AZURE_STACK_GET_NETWORK_CONFIGURATION
+    fi
+
+    echo "Generating Azure CNI interface file"
+    echo "------------------------------------------------------------------------"
+    echo "Parameters"
+    echo "------------------------------------------------------------------------"
+    echo "SUBNET_PREFIX:              $SUBNET_PREFIX"
```
jadarsie: not aligned, in case you care

pkg/engine/armvariables.go
```diff
@@ -110,8 +115,8 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
...ig').cloudProviderBackoffJitter,' CLOUDPROVIDER_RATELIMIT=',toLower(string(parameters('cloudproviderConfig').cloudProviderRatelimit)),' CLOUDPROVIDER_RATELIMIT_QPS=',parameters('cloudproviderConfig').cloudProviderRatelimitQPS,' CLOUDPROVIDER_RATELIMIT_BUCKET=',parameters('cloudproviderConfig').cloudProviderRatelimitBucket,' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension'),' USE_INSTANCE_METADATA=',variables('useInstanceMetadata'),' LOAD_BALANCER_SKU=',variables('loadBalancerSku'),' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB'),' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount'),' CONTAINER_RUNTIME=',parameters('containerRuntime'),' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase'),' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec'),' KMS_PROVIDER_VAULT_NAME=',variables('clusterKeyVaultName'),' IS_HOSTED_MASTER=%t',' IS_IPV6_DUALSTACK_FEATURE_ENABLED=%t',' PRIVATE_AZURE_REGISTRY_SERVER=',parameters('privateAzureRegistryServer'),' AUTHENTICATION_METHOD=',variables('customCloudAuthenticationMethod'),' IDENTITY_SYSTEM=',variables('customCloudIdentifySystem')%s)]",
```
jadarsie: This is picky and just personal preference, but I would rather take the comma out of the new variable

`variables('customCloudIdentifySystem'),%s`

pkg/engine/template_generator_test.go
```diff
@@ -9,9 +9,8 @@ import (
 	"testing"
 	"text/template"
 
-	"github.com/pkg/errors"
-
 	"github.com/Azure/aks-engine/pkg/api"
+	"github.com/pkg/errors"
```
jadarsie: FYI, some linters will complain about the imports' order change. Not sure if AKSe's does.

pkg/engine/virtualmachinescalesets.go
```diff
@@ -300,7 +305,7 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
...
+				"commandToExecute": fmt.Sprintf("[concat('retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; "+outBoundCmd+" for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),%s,%svariables('provisionScriptParametersMaster'), ' /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]", generateUserAssignedIdentityClientIDParameter(userAssignedIDEnabled), azureStackCNIParams),
```
jadarsie: Same here, I would do this instead

`%s,variables('provisionScriptParametersMaster')`

pkg/engine/armvariables.go
t-chappl: The problem is that the variable is empty when azure cni is not being used on azure stack, so there would be a trailing comma that breaks the concat(). I put the comma inside the variable so that when the variable is empty the string is unchanged.

pkg/engine/template_generator_test.go
t-chappl: My IDE probably did that. It passes the linter though.

pkg/engine/armvariables.go
honcao: I would suggest: 
remove variable azureStackCNIParams 
Add NETWORK_API_VERSION into common parameters
Get the ACTIVE_DIRECTORY_ENDPOINT from /etc/kubernetes/azurestack.json

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -451,6 +451,8 @@ if (`$hnsNetwork)
 # Restart Kubeproxy, which would wait, until the network is created
 Restart-Service Kubeproxy
 
+`$env:AZURE_ENVIRONMENT_FILEPATH="c:\k\azurestackcloud.json"
```
honcao: Where do we set the values for c:\k\azurestackcloud.json?

pkg/engine/template_generator.go
```diff
@@ -631,8 +631,33 @@ func (t *TemplateGenerator) getTemplateFuncMap(cs *api.ContainerService) templat
...
+		"GetActiveDirectoryEndpoint": func() string {
+			if cs.Properties.IsAzureStackCloud() {
+				return cs.Properties.CustomCloudProfile.Environment.ActiveDirectoryEndpoint
+			}
+
+			return ""
+		},
+		"GetResourceManagerEndpoint": func() string {
+			if cs.Properties.IsAzureStackCloud() {
+				return cs.Properties.CustomCloudProfile.Environment.ResourceManagerEndpoint
```
honcao: Please check Environment is not nil

pkg/engine/virtualmachinescalesets.go
```diff
@@ -647,7 +652,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
...
+					"commandToExecute": fmt.Sprintf("[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('kubernetesAPIServerIP'),' -KubeDnsServiceIp ',parameters('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -TargetEnvironment ',parameters('targetEnvironment'),' -AgentKey ',parameters('clientPrivateKey'),' -AADClientId ',variables('servicePrincipalClientId'),' -AADClientSecret ',variables('singleQuote'),variables('singleQuote'),base64(variables('servicePrincipalClientSecret')),variables('singleQuote'),variables('singleQuote'),%s ' ',variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %%SYSTEMDRIVE%%\\AzureData\\CustomDataSetupScript.log 2>&1')]", azureStackCNIParams),
```
honcao: Is the double %% correct?

honcao: as I suggested, please remove azureStackCNIParams

parts/k8s/windowskubeletfunc.ps1
t-chappl: https://github.com/Azure/aks-engine/blob/master/parts/k8s/kuberneteswindowssetup.ps1#L206-L210

pkg/engine/virtualmachinescalesets.go
t-chappl: Yes, since I switched it to Sprintf now all % have to be %%

t-chappl: I removed the change now that I took out azureCNIParams

pkg/engine/template_generator.go
t-chappl: I just read the endpoints from the file and these functions aren't needed anymore

PatrickLang:
I checked with @t-chappl - this will also require an Azure-CNI update that's still in progress

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -52,6 +52,9 @@ ERR_CIS_ASSIGN_ROOT_PW=111 # Error assigning root password in CIS enforcement
 ERR_CIS_ASSIGN_FILE_PERMISSION=112 # Error assigning permission to a file in CIS enforcement
 ERR_CIS_COPY_FILE=113 # Error writing a file to disk for CIS enforcement
 ERR_CIS_APPLY_PASSWORD_CONFIG=115 # Error applying CIS-recommended passwd configuration
+ERR_AZURE_STACK_GET_ARM_TOKEN=116 # Error generating a token to use with Azure Resource Manager
```
CecileRobertMichon: let's allocate specific range for Azure Stack specific errors, eg. 120-130. Can you make these error codes 120, 121 and 122?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -647,7 +647,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
...
+					"commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command \"', '$arguments = ', variables('singleQuote'),'-MasterIP ',variables('kubernetesAPIServerIP'),' -KubeDnsServiceIp ',parameters('kubeDnsServiceIp'),' -MasterFQDNPrefix ',variables('masterFqdnPrefix'),' -Location ',variables('location'),' -TargetEnvironment ',parameters('targetEnvironment'),' -AgentKey ',parameters('clientPrivateKey'),' -AADClientId ',variables('servicePrincipalClientId'),' -AADClientSecret ',variables('singleQuote'),variables('singleQuote'),base64(variables('servicePrincipalClientSecret')),variables('singleQuote'),variables('singleQuote'),' -NetworkAPIVersion ',variables('apiVersionNetwork'),' ',variables('singleQuote'), ' ; ', variables('windowsCustomScriptSuffix'), '\" > %SYSTEMDRIVE%\\AzureData\\CustomDataSetupScript.log 2>&1')]",
```
CecileRobertMichon: does apiVersionNetwork need to be in single quotes? servicePrincipalClientSecret is a special case because it's a password that could contain special characters

t-chappl: It isn't in single quotes

t-chappl: The whole arguments string is in single quotes, and singe apiVersionNetwork is last the single quote after it is just ending the arguments string

t-chappl: The servicePrincipalClientSecret has two single quotes before and after because it has to escape any quotes inside the argument string. The single quote before apiVersionNetwork is actually part of servicePrincipalClientSecret.

CecileRobertMichon: yep sorry I misread the diff, thanks for clarifying!

Issue #1421, opened 2019-05-31
Upgrade from 1.12.9 is broken


CecileRobertMichon:
Repro:

{   "apiVersion": "vlabs",   "properties": {     "orchestratorProfile": {       "orchestratorType": "Kubernetes"     },     "masterProfile": {       "count": 1,       "dnsPrefix": "",       "vmSize": "Standard_D2_v2"     },     "agentPoolProfiles": [       {         "name": "agentpool1",         "count": 3,         "vmSize": "Standard_D2_v2",         "osType": "Linux"       }     ],     "linuxProfile": {       "adminUsername": "azureuser",       "ssh": {         "publicKeys": [           {             "keyData": ""           }         ]       }     },     "servicePrincipalProfile": {       "clientId": "",       "secret": ""     }   } }

from commit aacc7cfea5b37ff4c0e516d69f269b7d03a5ff5e consistently results in: 

```
Error: upgrading cluster: Code="NetworkingInternalOperationError" Message="Unknown network allocation error."
```

mo-saeed:
@CecileRobertMichon Just FYI i got this error sometimes when i was trying to create cluster with version 1.13.6, it happened one or two time then it worked ! I thought it's sth related network maintenance in azure but i am not sure as i didn't investigate.

CecileRobertMichon:
@mo-saeed that was my first intuition too but it's reproing consistently in our tests and doesn't repro with 1.12.8

CecileRobertMichon:
@andyzhangx any ideas?

CecileRobertMichon:
This is reproducible only if a scale happened prior to the upgrade.

andyzhangx:
what's the detailed error of this issue? e.g. node not ready or sth else?

CecileRobertMichon:
it's the error I pasted in the description, ie. `Error: upgrading cluster: Code="NetworkingInternalOperationError" Message="Unknown network allocation error."`

andyzhangx:
I could not repro by following these steps:
1. set up a 1.12.9 cluster with 3 vmss instances
2. scale down to 1 vmss instance
3. upgrade to 1.13.5

I used aks-engine version `678e02a52178d2b02702bf4372e7e66761572dca` on master branch, the upgrade operation completed successfully:
```
[...]

PR #1594, opened 2019-07-09
feat: add support for Kubernetes 1.12.10

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md#v11210

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

We need to verify that #1421 is fixed before we can merge this. 🤞

packer/install-dependencies.sh
```diff
@@ -369,6 +372,7 @@ K8S_VERSIONS="
 1.13.7-azs
 1.13.5
 1.13.5-azs
+1.12.10
 1.12.9
 1.12.9-azs
```
CecileRobertMichon: should we remove 1.12.9 s since it's disabled?

rjaini:
@mboersma : Can you please add Azure Stack support as well with this PR. As 1.12.10 image on Azure stack is also available now.

rjaini:
@mboersma : Thanks for adding the Azs changes..

PR #1595, opened 2019-07-09
feat: add support for Kubernetes 1.13.8

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md#v1138

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This also updates addon-manager to v8.9.1 (kubernetes/kubernetes#77623)


pkg/api/common/versions.go
```diff
@@ -111,9 +111,10 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.2":         false,
 	"1.13.3":         false,
 	"1.13.4":         false,
-	"1.13.5":         true,
+	"1.13.5":         false,
```
CecileRobertMichon: should we keep this until Azure Stack is added?

mboersma: Yes, good call. Maybe that should be our policy in general from now on: don't remove versions until Azure Stack supports two newer ones.

PR #1596, opened 2019-07-09
feat: add support for Kubernetes 1.14.4

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#v1144

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/versions.go
```diff
@@ -120,9 +120,10 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.14.0-beta.2":  false,
 	"1.14.0-rc.1":    false,
 	"1.14.0":         false,
-	"1.14.1":         true,
+	"1.14.1":         false,
```
CecileRobertMichon: same comment here about not disabling until Azure Stack supports 1.14.4

mboersma: Yep. 👍

Issue #1597, opened 2019-07-09
cluster-autoscaler restarts multiple times on new Linux cluster


mboersma:
**Describe the bug**
Our CI tests have been failing repeatedly since a new test was enabled that checks for too many component restarts in #1579. Specifically, we see cluster-autoscaler coming up as `Ready` but having restarted several times:

```go
Azure Container Cluster using the Kubernetes Orchestrator regardless of agent pool type 
  Should not have any unready or crashing pods right after deployment
  /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:515
STEP: Checking ready status of each pod in kube-system
2019/07/09 15:25:40 Checking azure-cni-networkmonitor-46mwk - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-cni-networkmonitor-9m9j5 - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-cni-networkmonitor-qzgwp - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-cni-networkmonitor-w5w6v - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-ip-masq-agent-4vjrs - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-ip-masq-agent-6p4mz - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-ip-masq-agent-njzqb - ready: true, restarts: 0
2019/07/09 15:25:40 Checking azure-ip-masq-agent-pmg2h - ready: true, restarts: 0
2019/07/09 15:25:40 Checking blobfuse-flexvol-installer-889q2 - ready: true, restarts: 0
2019/07/09 15:25:40 Checking cluster-autoscaler-7bd6bc964f-xgpj5 - ready: true, restarts: 7

• Failure [1.442 seconds]
Azure Container Cluster using the Kubernetes Orchestrator
/go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:94
  regardless of agent pool type
  /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:95
    Should not have any unready or crashing pods right after deployment [It]
    /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:515

    Expected
        <int>: 7
    to be <
        <int>: 5

[...]

PR #1598, opened 2019-07-09
test: temporarily allow 10 CA restarts

mboersma:
**Reason for Change**:

Provides a temporary workaround for #1597 so we don't miss other failures that might be occurring in CI. cluster-autoscaler restarting several times on Linux but coming up `Ready` is still something to investigate, but the overall test is useful, so this disables testing for just that specific and logs it instead.

**Issue Fixed**:

Refs #1597

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -520,7 +520,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			for _, currentPod := range pods.Pods {
 				log.Printf("Checking %s - ready: %t, restarts: %d", currentPod.Metadata.Name, currentPod.Status.ContainerStatuses[0].Ready, currentPod.Status.ContainerStatuses[0].RestartCount)
 				Expect(currentPod.Status.ContainerStatuses[0].Ready).To(BeTrue())
-				Expect(currentPod.Status.ContainerStatuses[0].RestartCount).To(BeNumerically("<", 5))
+				tooManyRestarts := 5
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -520,7 +520,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			for _, currentPod := range pods.Pods {
 				log.Printf("Checking %s - ready: %t, restarts: %d", currentPod.Metadata.Name, currentPod.Status.ContainerStatuses[0].Ready, currentPod.Status.ContainerStatuses[0].RestartCount)
 				Expect(currentPod.Status.ContainerStatuses[0].Ready).To(BeTrue())
-				Expect(currentPod.Status.ContainerStatuses[0].RestartCount).To(BeNumerically("<", 5))
+				tooManyRestarts := 5
+				if strings.Contains(currentPod.Metadata.Name, "cluster-autoscaler") && !eng.HasWindowsAgents() {
+					log.Print("need to investigate cluster-autoscaler restarts on Linux!")
```
PatrickLang: nothing in kube-system runs on Windows, so I don't see any need for if eng.HasWindowsAgents(). I think whatever tests we run against deployments in kube-system should be the same regardless of which node OS are used.

TL;DR - I'm ok with 10 restarts everywhere

CecileRobertMichon: +1

PR #1616, opened 2019-07-12
docs: proposal for unifying how node pool images are specified f…

marosset:
…or Linux and WIndows

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/design/unified-image-specification.md
+++ b/docs/design/unified-image-specification.md
@@ -3,12 +3,12 @@
 
 ## Motivation
 
-There are inconsistencies in where and how you can specify images in the aks-engine cluster definitions.
-This leads to unnecessary complexity in the aks-engine codebase and potential confusion for aks-engine users.
+There are inconsistencies in where and how you can specify images in the AKS Engine cluster definitions.
+This leads to unnecessary complexity in the aks-engine codebase and potential confusion for AKS Engine users.
 
 Additionally, it is not possible to specify different images for different pools running Windows which breaks some upgrade/test scenarios.
 
-Images for virtual machines in Azure are specified the same way regardless of guest OS. This proposal aims to unify how aks-engine cluster defintions specify images for Linux and Windows.
+Images for virtual machines in Azure are specified the same way regardless of guest OS. This proposal aims to unify how AKS Engine cluster defintions specify images for Linux and Windows.
 
 ## Current behavior
 
@@ -70,7 +70,8 @@ Images are specified as port of agentPoolProfile.
 ```json
 "Distro": "[ubuntu | rhel | coreos | AKSUbuntu1604 | AKSUbuntu1804 | AKSWindows1809 | AKSWindows1903 | AKSWindows2019]"
 ```
-* Image URL - path to a chd in azure blob storage
+
+* Image URL - path to a vhd in azure blob storage
 
 ```json
 "image" : {
@@ -83,21 +84,23 @@ Images are specified as port of agentPoolProfile.
   * If subscription, gallery, and version are also specified image will be pulled from a gallery
 
 ```json
-"imageReference": {
-    "name": "" ,
-    "resourceGroup": "",
+"image": {
+    "reference": {
+        "name": "" ,
+        "resourceGroup": "",
 
-    (optional)
-    "subscription": "",
-    "gallery": "",
-    "version": ""
+        (optional)
+        "subscription": "",
+        "gallery": "",
+        "version": ""
+    }
 }
 ```
 
 * Marketplace Image - Specifies an Azure marketplace image
 
 ```json
-"imageReference": {
+"image": {
     "marketplace": {
         "publisher": "",
         "offer": "",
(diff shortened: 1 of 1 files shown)


docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+### Linux and Windows
+
+Images are specified as port of agentPoolProfile.
+
+* Distro - Specifies a curated list of images for various 'distros' (pulled from marketplace or shared image gallery)
+
+```json
+"Distro": "[ubuntu | rhel | coreos | AKSUbuntu1604 | AKSUbuntu1804 | AKSWindows1809 | AKSWindows1903 | AKSWindows2019]"
+```
+* Image URL - path to a chd in azure blob storage
```
CecileRobertMichon: chd --> vhd

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+"imageVersion" = ""
+```
+
+## Proposed Behavior
+
+### Linux and Windows
+
+Images are specified as port of agentPoolProfile.
+
+* Distro - Specifies a curated list of images for various 'distros' (pulled from marketplace or shared image gallery)
```
CecileRobertMichon: I think by definition a "distro" is a Linux distribution so it might be confusing to have Windows in there as well

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+Images for virtual machines in Azure are specified the same way regardless of guest OS. This proposal aims to unify how aks-engine cluster defintions specify images for Linux and Windows.
+
+## Current behavior
+
+### Linux
+
+Images are specified in the agentPoolProfile one of the following ways. 
+Specifing the image in agentPoolProfile allows multiple pools running different images.
+
+* Distro - Specifies a curated list of images for various 'distros' (pulled from marketplace or shared image gallery)
```
CecileRobertMichon: the distros specify the Linux distribution to use and are mapped to an image from the marketplace  (not a SIG)

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+
+```json
+"Distro": "[ubuntu | rhel | coreos | AKSUbuntu1604 | AKSUbuntu1804 | AKSWindows1809 | AKSWindows1903 | AKSWindows2019]"
+```
+* Image URL - path to a chd in azure blob storage
+
+```json
+"image" : {
+    "url": ""
```
CecileRobertMichon: I've tried to implement this is the past because I wanted to be able to test VHDs before they are published as images but wasn't able to and had to abandon the PR. It might be worth retrying but just a heads up that this won't be trivial.

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+    "subscription": "",
+    "gallery": "",
+    "version": ""
+}
+```
+
+* Marketplace Image - Specifies an Azure marketplace image
+
+```json
+"imageReference": {
```
CecileRobertMichon: would this override the distro value as a "custom image"? In that case what is the value of "distro"?

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
+
+# Unified image specification for node pools
+
+## Motivation
+
+There are inconsistencies in where and how you can specify images in the aks-engine cluster definitions.
```
CecileRobertMichon: "AKS Engine" is the tool 
`aks-engine` the CLI command

docs/design/unified-image-specification.md
marosset: I was thinking that too... Need to think moer about that.
It would be good to re-use the same same terminology and api model syntax between Linux and Windows but only where it makes sense.

docs/design/unified-image-specification.md
marosset: I think that if image is specified it would override a distro.
Image/marketplace images would primarily be used for testing/validation of images before promoting them them as known-good images for the various distro/channel (at least on the windows side)

docs/design/unified-image-specification.md
marosset: It looks like this is coded up to work for windows images (although only for AvailabilitySets form what i can see in the engine code)
All the more reason to target images consistently for Linus and Windows :)

docs/design/unified-image-specification.md
```diff
@@ -0,0 +1,108 @@
...
+    "url": ""
+}
+```
+
+* Image Reference - Specifies an Azure manage image
+  * If only name and resourceGroup are specified, image will be pulled from current subscription
+  * If subscription, gallery, and version are also specified image will be pulled from a gallery
+
+```json
+"imageReference": {
```
marosset: This was supposed to be
"image": {
  "reference": {
    "name": ""...

PR #1618, opened 2019-07-12
[WIP] fix: don't set deprecated node labels in kubelet startup

### LINKED ISSUE #1546: kubernetes.io labels set via --node-label are unsupported for 1.16

mboersma:
**Reason for Change**:
Removes the use of deprecated node labels in the `--node-labels` argument of kubelet startup.

**Issue Fixed**:
Fixes #1546
Closes #1548 

**Requirements**:

- [ ] Azure Stack json test data needs updating
- [x] run label-nodes.sh script on a timer?
- [ ] need a label-nodes.ps1 for Windows too
- [ ] need an Azure cloud-provider change to tolerate this label as well as `kubernetes.io/role`
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:

Issue #1617, opened 2019-07-12
Some pods including metrics-server are not starting in 1.16 testing


PatrickLang:
**Describe the bug**

Some of the pods are not ready in the [aks-engine-azure-windows-master-staging](https://testgrid.k8s.io/sig-windows#aks-engine-azure-windows-master-staging)

https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-aks-engine-azure-master-staging-windows/1149707427701067782

```
test/e2e/e2e.go:72
Jul 12 17:07:17.584: Error waiting for all pods to be running and ready: 4 / 13 pods in namespace "kube-system" are NOT in RUNNING and READY state in 10m0s
POD                                   NODE                  PHASE   GRACE CONDITIONS
coredns-58c87db796-csl7l              k8s-master-37988155-0 Running       [{Type:Initialized Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-07-12 16:33:11 +0000 UTC Reason: Message:} {Type:Ready Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-07-12 16:33:11 +0000 UTC Reason:ContainersNotReady Message:containers with unready status: [coredns]} {Type:ContainersReady Status:False LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-07-12 16:33:11 +0000 UTC Reason:ContainersNotReady Message:containers with unready status: [coredns]} {Type:PodScheduled Status:True LastProbeTime:0001-01-01 00:00:00 +0000 UTC LastTransitionTime:2019-07-12 16:33:09 +0000 UTC Reason: Message:}]
[...]

PR #1621, opened 2019-07-15
fix: Remove pod-critical annotations in 1.16 manifests.

adelina-t:
Commit https://github.com/kubernetes/kubernetes/pull/79554 in
kubernetes/kubernetes removed the ExperimentalCriticalPodAnnotation feature
in kubernetes 1.16.

This PR aligns aks-engine with those changes, removing the feature-gate from
kube-proxy daemonset and the pod-critical annotations from any manifest that
uses it.

Fixes: https://github.com/Azure/aks-engine/issues/1617


PatrickLang:
Thanks! just added a `make generate` to unblock test pass

PR #1626, opened 2019-07-17
feat: add support for Kubernetes 1.16.0-alpha.1

### LINKED ISSUE #1546: kubernetes.io labels set via --node-label are unsupported for 1.16

PatrickLang:
**Reason for Change**:
Adds v1.16.0-alpha.1 support, pipelines to test it

~~Remaining work needed - push 1.16.0-alpha.1 binaries for Windows~~ Done.

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#v1160-alpha1

See also the [discussion on a previous PR](https://github.com/Azure/aks-engine/pull/1548#discussion_r300895806).

**Issue Fixed**:
Fixes #1546 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

- ~~[Should we](https://github.com/Azure/aks-engine/pull/1548#discussion_r300895806) use `node.kubernetes.io/role` instead of `kubernetes.azure.com/role`?~~ No.
- ~~Can the `label-nodes.service` be made to start only on Kubernetes 1.16?~~ Done.
- Does Windows need an equivalent label-nodes facility?
- There is [a Kubernetes operator PoC](https://github.com/mboersma/node-labeler) that could replace the systemd unit


lachie83:
Quick reminder that we should do a spot check and make sure all the addons are using non-deprecated APIs - https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

lachie83:
Seeing this in addon manager
```
error: unable to recognize "/etc/kubernetes/addons/audit-policy.yaml": no matches for kind "Policy" in version "audit.k8s.io/v1beta1"
```

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -330,6 +330,14 @@ ensureKubelet() {
     systemctlEnableAndStart kubelet || exit $ERR_KUBELET_START_FAIL
 }
 
+ensureLabelNodes() {
```
CecileRobertMichon: just to make sure I understand the intention here, this script runs for all kubernetes versions on master nodes _only_ to and adds labels to _both_ master and agent nodes?

pkg/api/types.go
```diff
@@ -1440,10 +1440,14 @@ func (a *AgentPoolProfile) IsUbuntuNonVHD() bool {
 }
 
 // GetKubernetesLabels returns a k8s API-compliant labels string for nodes in this profile
-func (a *AgentPoolProfile) GetKubernetesLabels(rg string) string {
+func (a *AgentPoolProfile) GetKubernetesLabels(rg string, deprecated bool) string {
```
CecileRobertMichon: unrelated to the PR but why is this in types and GetMasterKubernetesLabels in helper?

parts/k8s/cloud-init/artifacts/cse_config.sh
mboersma: Yes, that's the intent.

pkg/api/types.go
mboersma: No idea.

parts/k8s/cloud-init/artifacts/cse_config.sh
mboersma: @CecileRobertMichon can you think of a clean way only to start the `label-nodes.sh` service on Kubernetes 1.16 and later? It's safe to run on any cluster, but it's basically a no-op on earlier clusters.

mboersma: I'm going to try to do this using the `${KUBERNETES_VERSION}` env var.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -167,6 +167,9 @@ ensureKubelet
 ensureJournal
 
 if [[ -n "${MASTER_NODE}" ]]; then
+    if dpkg --compare-versions ${KUBERNETES_VERSION} ge 1.16; then
```
jackfrancis: `dpkg` only works on ubuntu (I think)

mboersma: 👍 I replaced it with a `version_lte` function that I tested on CoreOS as well as Ubuntu.

mboersma:
I added an e2e test spec `should have node labels specific to masters or agents`.

Since nearly all the changes in this PR are now isolated to 1.16, I think it's safe to merge. There may still be some outstanding issues but this will help us flush them out and fix them.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -236,3 +236,6 @@ sysctl_reload() {
         fi
     done
 }
+version_lte() {
```
jackfrancis: should the func name be `version_gte` ?

mboersma: No, because it's a `<=` comparator. When we use it, it's:
`version_lte 1.16 ${KUBERNETES_VERSION}` 
which *should* read as "version 1.16 is less than or equal to the current Kubernetes version," but I'm open to changing it if that's confusing.

jackfrancis: Got it, my thoughts are that from a code maintenance standpoint we are really using this function for >= type semantics, and so using a <= func (swapping the vals) is arguably less expressive.

From a human expressivity standpoint we are saying "do this thing for all k8s versions starting from k8s version _v_".

mboersma: I hear that. I had a `version_gt` func working and this renaming + swap args was the simplest path to get the behavior I wanted. I can refactor this.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -545,6 +545,31 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(len(nodeList)).To(Equal(nodes))
 		})
 
+		It("should have node labels specific to masters or agents", func() {
+			nodeList, err := node.Get()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(nodeList.Nodes)).To(Equal(eng.NodeCount()))
+			for _, node := range nodeList.Nodes {
+				labels := node.Metadata.Labels
+				Expect(labels).To(SatisfyAny(
```
jackfrancis: So the idea here is that *all* nodes in *all* versions on *all* OS flavors will have either `kubernetes.azure.com/role=master` or `kubernetes.azure.com/role=agent`. And then Linux nodes which are < 1.16 will additionally have either `kubernetes.io/role=master` or `kubernetes.io/role=agent` + either `node-role.kubernetes.io=master` or `node-role.kubernetes.io=agent` ?

jackfrancis: A nit would be: based on the first, universal test, do the follow up (if linux + < 1.16) tests in a way that ensures nodes don't have master + agent labels. So those follow-up tests would get "if master ensure add'l labels are master, if agent ensure add'l labels are agent" type thing.

mboersma: Just to clarify the state of the world in this PR:
- All nodes get `kubernetes.azure.com/role` (directly via `kubelet --node-labels`)
- All 1.15 nodes get `kubernetes.io/role` and `node-role.kubernetes.io` (directly via `kubelet --node-labels`)
- Linux 1.16 nodes get `kubernetes.io/role` and `node-role.kubernetes.io` (indirectly via the systemd service)
- Windows 1.16 nodes don't get `kubernetes.io/role` and `node-role.kubernetes.io` (need to fix this, see #1660)

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -241,7 +241,11 @@ write_files:
     KUBELET_CONFIG={{GetKubeletConfigKeyVals .KubernetesConfig }}
     KUBELET_IMAGE={{WrapAsParameter "kubernetesHyperkubeSpec"}}
     KUBELET_REGISTER_SCHEDULABLE=true
+{{if IsKubernetesVersionGe "1.16.0-alpha.1"}}
```
mboersma: This appears to return `true` in all cases on Windows, but work as expected on Linux. I'm not sure why, but I can reproduce it with an `aks-engine generate` command and I'll try to find a fix.

mboersma: Found, fixed.

mboersma:
acs-bot needs repair, so I'm merging this manually.

Issue #1625, opened 2019-07-17
Pre-pull docker distribution image


jadarsie:
Pre-pulling until ACR is ported to Azure Stack.

PR #1631, opened 2019-07-17
feat: pre-pull docker distribution for Azure Stack

jadarsie:
**Reason for Change**:
Stop gap solution to address registry reqs

**Issue Fixed**:
Fixes: #1625


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -302,6 +302,9 @@ cleanUpContainerImages() {
         docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep 'kube-svc-redirect') &
         docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep 'nginx') &
     fi
+
+    # TZL Financials: https://github.com/Azure/aks-engine/issues/1625
+    docker rmi registry:2 &
```
CecileRobertMichon: looks like you're removing this image for all clusters, wasn't there an Azure Stack specific cleanup at some point that we could reuse?

jadarsie:
@CecileRobertMichon it looks like codecov is stuck

PR #1632, opened 2019-07-17
chore: use kubernetes.azure.com/role label for nodes

mboersma:
**Reason for Change**:
Replaces uses of the `kubernetes.io/role` and `node-role.kubernetes.io` node labels with `kubernetes.azure.com/role`.

The first two labels are deprecated and can't be set via `kubelet --node-labels=` starting in Kubernetes 1.16. This continues to set the deprecated labels, however, for backward compatibility, and we will remove them just for 1.16 in that PR.

**Issue Fixed**:
This may be a prerequisite to #1626.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:

PR #1633, opened 2019-07-18
fix: Remove unnecessary transform on VMSS upgrade

### LINKED ISSUE #1638: VMSS agentpool upgrade doesn't update OS image version
**Describe the bug**
The code path on VMSS agentpool upgrade doesn't update OS image version. This is not a problem for AKS-Engine since the master template upgrade it, however this is a problem for AKS. The VMSS will continue have old VHD version in AKS after agent pool upgrade.

**Steps To Reproduce**

Upgrade an AKS VMSS agentpool.

**Expected behavior**
Update VMSS model always with desired OS image version.

**AKS Engine version**
v0.38.2

**Kubernetes version**
all

**Additional context**

yangl900:
**Reason for Change**:

VMSS upgrade path used a template normalization, which removes OS image reference. This doesn't seem right because we expect OS image version get updated from previous version. Also the scaling normalization is only called by upgrade path, this alone is very weird...

**Issue Fixed**:

Fixes #1638

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
`make test-style` is complaining about unused constants:

```go
==> Running Go linter <==
golangci-lint has version 1.17.1 built from 4ba2155 on 2019-06-10T09:06:49Z
Running inside container
pkg/engine/transform/transform.go:28:2: U1000: const `virtualMachineProfileFieldName` is unused (unused)
	virtualMachineProfileFieldName = "virtualMachineProfile"
	^
pkg/engine/transform/transform.go:40:2: U1000: const `vmssResourceType` is unused (unused)
	vmssResourceType = "Microsoft.Compute/virtualMachineScaleSets"
	^
```

pkg/engine/transform/transform.go
```diff
@@ -66,44 +66,6 @@ type Transformer struct {
...
-			continue
-		}
-
-		virtualMachineProfile, ok := resourceProperties[virtualMachineProfileFieldName].(map[string]interface{})
-		if !ok {
-			logger.Warnf("Template improperly formatted")
-			continue
-		}
-
-		if !t.removeCustomData(logger, virtualMachineProfile) || !t.removeImageReference(logger, virtualMachineProfile) {
```
CecileRobertMichon: the removeCustomData part might be needed for upgrade

yangl900:
`make test-style` passed:

==> Running Go linter <==
golangci-lint has version 1.17.1 built from 4ba2155 on 2019-06-10T09:06:49Z
==> Running shell linter <==
ShellCheck - shell script analysis tool
version: 0.6.0
license: GNU General Public License, version 3
website: https://www.shellcheck.net
==> Checking copyright headers <==

pkg/engine/transform/transform.go
yangl900: it should be allowed for VMSS, we will see the test result.

yangl900:
added #1638 for tracking.

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -451,8 +451,7 @@ func (ku *Upgrader) upgradeAgentScaleSets(ctx context.Context) error {
 			Translator: ku.Translator,
 		}
 
-		if err = transformer.NormalizeForVMSSScaling(ku.logger, templateMap); err != nil {
-			ku.logger.Errorf("unable to update template, error: %v.", err)
+		if err = transformer.NormalizeMasterResourcesForScaling(ku.logger, templateMap); err != nil {
```
CecileRobertMichon: This function is to normalize the master resource, why are we calling it in the upgradeAgent function? If it applies to both, we should rename it (also it says for scaling and this is an upgrade, I know it was already like that before, but maybe it's a good opportunity to make the name reflect what it's actually used for...)

yangl900: The template actually contains master node at this point. So we have to normalize it to make it work, I'm not sure if there is an option to generate agent only template. If there are, that seems the right approach. 

The function only normalize master node, nothing for agent. VMSS doesn't need anything to normalize. It's not for "Scaling" though, I can rename it NormalizeMasterResourcesForUpdate(), but that's a breaking change technically since it's exported function. If you are OK, I can make the change.

CecileRobertMichon: Got it. The template is the full template as far as I understand that's why it contains all the resources including master resources. What happens for a template that doesn't contain any masters (ie AKS)? Would this also work?

yangl900: yep, it will skip, the normalization only filter master node.

```
		// make sure this is only modifying the master vms
		if !strings.Contains(resourceName, "variables('masterVMNamePrefix')") {
			continue
		}
```

pkg/engine/transform/transformtestfiles/dcos_template.json
```diff
@@ -1,1140 +0,0 @@
-{
```
mboersma: Why did this file and `dcos_scale_template.json` have to be removed? Some of AKS Engine is still in use by the ACS service and we need to ensure any relevant unit tests are still passing.

yangl900: only the test TestNormalizeForVMSSScaling() uses this file, and the function has been deleted with my change. So I guess this file is not needed anymore?

pkg/operations/kubernetesupgrade/upgrader.go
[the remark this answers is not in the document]
mboersma: See #1645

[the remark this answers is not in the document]
yangl900: yes, I actually prepared a unit test yesterday, but didn't finish. :) Will send a PR today.

The caller at AKS side is actually ACS only, so it's on deprecation path.

PR #1636, opened 2019-07-18
chore: Update calico to v3.8.0

song-jiang:
**Reason for Change**:

Update calico to v3.8.0
The major change from current version v3.7.2 is `Calico v3.8.0 includes several improvements to its integration with Istio for application layer policy enforcement. Calico now supports Istio v1.1, and includes the necessary drivers and volumes in calico-node by default`

Please refer to https://docs.projectcalico.org/v3.8/release-notes/

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/containeraddons/kubernetesmasteraddons-calico-daemonset.yaml
+++ b/parts/k8s/containeraddons/kubernetesmasteraddons-calico-daemonset.yaml
@@ -16,7 +16,7 @@ data:
   cni_network_config: |-
     {
       "name": "k8s-pod-network",
-      "cniVersion": "0.3.0",
+      "cniVersion": "0.3.1",
       "plugins": [
         {
           "type": "calico",
@@ -190,7 +190,7 @@ spec:
 # Include a clusterrole for the calico-node DaemonSet,
 # and bind it to the calico-node serviceaccount.
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: calico-node
   labels:
@@ -291,12 +291,11 @@ rules:
   - create
   - update
 ---
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   name: calico-node
   labels:
-    kubernetes.io/cluster-service: "true"
     addonmanager.kubernetes.io/mode: "Reconcile"
 roleRef:
   apiGroup: rbac.authorization.k8s.io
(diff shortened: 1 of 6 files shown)


song-jiang:
FYI a new container image `calico/pod2daemon-flexvol:v3.8.0` has been added to calico manifest.  @andyzhangx

parts/k8s/containeraddons/kubernetesmasteraddons-calico-daemonset.yaml
```diff
@@ -626,7 +647,6 @@ metadata:
   name: calico-typha-horizontal-autoscaler
   namespace: kube-system
   labels:
-    kubernetes.io/cluster-service: "true"
```
mboersma: Is removing these lines intended to sync up with the upstream calico manifests?

parts/k8s/containeraddons/kubernetesmasteraddons-calico-daemonset.yaml
```diff
@@ -427,7 +426,7 @@ spec:
 # as the CNI plugins and network config on
 # each master and worker node in a Kubernetes cluster.
 kind: DaemonSet
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
```
mboersma: I had to reassure myself that `apps/v1` went GA in Kubernetes 1.9, because AKS Engine still has to support 1.10 and later. :+1:

parts/k8s/containeraddons/kubernetesmasteraddons-calico-daemonset.yaml
song-jiang: No. I thought this label is deprecated based on https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/addon-manager/README.md . Not sure if it is used by other components other than `addon manager`. WDYT?

mboersma: Based on that description (and the fact that we explicitly use `Reconcile` in these cases), it shouldn't change `addon-manager`'s behavior anyway. I just don't know if the label is used by anything else. I'm going with "probably safe." 😄

song-jiang: Cool. That is what I am thinking as well.

song-jiang:
@CecileRobertMichon @mboersma Is this PR pending on @andyzhangx 's review? Could we get it merged?

song-jiang:
Anyone please advise the status of this PR? Is it still pending on an internal validation? @CecileRobertMichon @mboersma @jackfrancis

mboersma:
@song-jiang I'm sorry for the slow response. We will do backward-compatibility testing this week and update here. cc: @jackfrancis.

song-jiang:
@mboersma No worries. Thanks for the update.

PR #1642, opened 2019-07-18
feat: add support for Kubernetes 1.15.1

lachie83:
Signed-off-by: Lachlan Evenson <lachlan.evenson@microsoft.com>

**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1150

**Issue Fixed**:
Closes #1641 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
#1641 is identical but has failed on 1.15 twice in CI. I'd like to investigate that before we merge this (or #1641).

mboersma:
They were both a (hopefully transient) 500 error from Azure Pipelines itself, so nothing to do with the PR contents. I'm running it again and assuming it's green then this lgtm.

PR #1661, opened 2019-07-23
fix: only include OrchestratorProfile validation during data marshalling for create

jackfrancis:
**Reason for Change**:

Based on pre-existing architectural debt, we are validating the `OrchestratorProfile` input data during the data marshalling flow, in particular in order to validate OrchestratorVersion for aks-engine deploy operations. See: #355

This PR modifies that vlabs hack a bit so that we only do this during create operations. The specific scenario that this fixes is scaling a pre-existing cluster with a pre-existing apimodel.json that was generated using an older version of aks-engine with a different (breaking) versions support matrix.

Also fixed up the E2E implementation a bit so when we marshall this data during an E2E run, we pass in the appropriate "update" context to skip validation, so that we can do back-compat style tests more easily (e.g., create from old branch, switch to new branch, scale in/out, re-run E2E).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -74,8 +74,12 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	csInput, err := engine.ParseInput(engCfg.ClusterDefinitionTemplate)
 	Expect(err).NotTo(HaveOccurred())
+	var isUpdate bool
+	if cfg.Name != "" {
+		isUpdate = true
+	}
 	validate := false
-	csGenerated, err := engine.ParseOutput(engCfg.GeneratedDefinitionPath+"/apimodel.json", validate)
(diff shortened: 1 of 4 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -74,7 +74,12 @@ var _ = BeforeSuite(func() {
 	Expect(err).NotTo(HaveOccurred())
 	csInput, err := engine.ParseInput(engCfg.ClusterDefinitionTemplate)
 	Expect(err).NotTo(HaveOccurred())
-	csGenerated, err := engine.ParseOutput(engCfg.GeneratedDefinitionPath + "/apimodel.json")
+	var isUpdate bool
```
mboersma: You could simplify this to:
```go

isUpdate := cfgName != ""
```

PR #1666, opened 2019-07-25
fix: distinct component update messages for upgrade/scale

jackfrancis:
**Reason for Change**:

Currently we erroneously state that we're upgrading certain components (e.g., moby) during a scale operation if we're using a newer version of the `aks-engine` binary, or if the user has manually updated the generated apimodel.json to a newer component version:

```
aks-engine scale --subscription-id <sub_id> ----api-model _output/kubernetes-uksouth-52692/apimodel.json --location uksouth --resource-group kubernetes-uksouth-52692 --master-FQDN kubernetes-uksouth-52692.uksouth.cloudapp.azure.com --node-pool agent1 --new-node-count 1 --auth-method client_secret --client-id **** --client-secret ****
INFO[0000] validating...                                
INFO[0006] Name suffix: 20965292                        
WARN[0006] Moby will be upgraded to version 3.0.6       
... 
```

This PR improves the clarity of the messaging:

```
aks-engine scale --subscription-id <sub_id> ----api-model _output/kubernetes-uksouth-52692/apimodel.json --location uksouth --resource-group kubernetes-uksouth-52692 --master-FQDN kubernetes-uksouth-52692.uksouth.cloudapp.azure.com --node-pool agent1 --new-node-count 1 --auth-method client_secret --client-id **** --client-secret ****
INFO[0000] validating...                                
INFO[0006] Name suffix: 20965292                        
WARN[0006] Any new scaled up nodes will have Moby version 3.0.6       
...
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -152,7 +152,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 					if isUpgrade {
 						log.Warnf("Moby will be upgraded to version %s\n", DefaultMobyVersion)
 					} else if isScale {
-						log.Warnf("Any new scaled up nodes will have Moby version %s\n", DefaultMobyVersion)
+						log.Warnf("Any new nodes will have Moby version %s\n", DefaultMobyVersion)
 					}
 				}
 				o.KubernetesConfig.MobyVersion = DefaultMobyVersion
@@ -163,7 +163,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 					if isUpgrade {
 						log.Warnf("containerd will be upgraded to version %s\n", DefaultContainerdVersion)
 					} else if isScale {
-						log.Warnf("Any new scaled up nodes will have containerd version %s\n", DefaultContainerdVersion)
+						log.Warnf("Any new nodes will have containerd version %s\n", DefaultContainerdVersion)
 					}
 				}
 				o.KubernetesConfig.ContainerdVersion = DefaultContainerdVersion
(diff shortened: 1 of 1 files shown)


jackfrancis:
We don't have a convenient way during the defaults enforcement flow to determine whether or not we're scaling up (in which case stuff will get upgraded) or down (in which case we just delete pre-existing nodes) so hopefully we can concisely express something that is useful info.

pkg/api/defaults.go
```diff
@@ -148,15 +148,23 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		switch o.KubernetesConfig.ContainerRuntime {
 		case Docker:
 			if o.KubernetesConfig.MobyVersion == "" || isUpdate {
-				if isUpdate && o.KubernetesConfig.MobyVersion != DefaultMobyVersion {
-					log.Warnf("Moby will be upgraded to version %s\n", DefaultMobyVersion)
+				if o.KubernetesConfig.MobyVersion != DefaultMobyVersion {
+					if isUpgrade {
+						log.Warnf("Moby will be upgraded to version %s\n", DefaultMobyVersion)
+					} else if isScale {
+						log.Warnf("Any new scaled up nodes will have Moby version %s\n", DefaultMobyVersion)
```
mboersma: Nitpick: I think `newly scaled-up` is more correct, or just say `Any new nodes...`.

jackfrancis: How about "new, scaled-up" ?

jackfrancis: Yeah, I like "new nodes". The user already knows it's a scale operation.

PR #1669, opened 2019-07-25
fix: add back-compat support for deprecated distro vals

### LINKED ISSUE #1652: revert distro const name change
to accommodate older api models, let's re-add back the const values for distro values

jackfrancis:
**Reason for Change**:

Changing a distro value, which deprecates the old value, should include a back-compat story for continued usages of that distro value. This PR adds an equivalent map value reference for the deprecated distro keys.

**Issue Fixed**:

Fixes #1652

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -170,13 +170,15 @@ var (
 		},
 
 		OSImageConfig: map[Distro]AzureOSImageConfig{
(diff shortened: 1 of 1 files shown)


jackfrancis:
@CecileRobertMichon FYI

PR #1671, opened 2019-07-25
test: use kill instead of interrupt signal to terminate kubectl port-forward

jackfrancis:
**Reason for Change**:

In practice it takes 5 minutes for the SIGINT signal to terminate the background kubectl port-forward process, making each test run 5 minutes longer.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@PatrickLang FYI

This may seem brute force on the surface, but I think in this case brute force is the appropriate intent once we're done validating the port-forward plumbing. The polite approach for some reason takes a long time.

PR #1675, opened 2019-07-29
chore: install ipvsadm

sylr:
Useful when kube-proxy is using ipvs mode.


sylr:
Generated package fixed.

PR #1676, opened 2019-07-29
chore: replace dep by go mod

sylr:


sylr:
golanci-lint fixed.

mboersma:
This looks good, but I'd like to test the branch locally. In particular, I need to be reassured that aks-engine can still be vendored successfully into other `dep` projects, since that has been a pain point in the past.

sylr:
They should migrate those projects to go mod too ;)

mboersma:
I changed `Gopkg.toml` in a project which vendors aks-engine to look like this:

```toml
[[constraint]]
  name = "github.com/Azure/aks-engine"
  branch = "go-mod"
  source = "github.com/sylr/aks-engine"
```

Then I did `GOOS=linux dep ensure -v` and it pulled in this branch and solved successfully for everything (as @devigned predicted). That's actually an improvement over current master which stumbles on `azure-sdk-for-go`.

I am reassured now.

sylr:
Rebased and implemented @devigned suggestion about go-bindata.

sylr:
Guys ?

devigned:
@jackfrancis and @mboersma, do we want to continue to vendor dependencies? If so, this PR will need to be updated to do so.

Let's talk about this item today at stand-up.

@sylr, thank you for your patience.

devigned:
Hi @sylr,

The team has discussed this change set. We are thankful for your contribution and perseverance to see this PR to this point. We appreciate your work and would like to pull this in. Unfortunately, behind the scenes, this is going to take a fair amount of effort, trust building and due diligence to integrate into projects that depend on this project. 

As I'm sure you know, a project does not approach changing the dependency management system nonchalantly.

We want to make sure your hard work is worth while and your commits land in the repo. To achieve this goal, we propose the PR is rebased on the newly created `gomod` branch, re-add module vendoring, and have someone from the internal team (me) help with the internal project integrations / due diligence. 

If this sounds good to you, please re-add module vendoring and I'll get this pulled into the `gomod` branch.

Thank you again for your contributions!!

sylr:
@devigned I understand the reluctance of changing the dependency manager.

However on the vendoring subject, since go 1.13 and the use by default of GOPROXY, the added value for vendoring dependencies has greatly decreased.

GOPROXY adds a layer of security that should prevent a dependency you pulled once through the proxy to become unavailable if it were to be removed from the VCS (github and such).

With this layer of security I believe vendoring deps only becomes an annoyance (big fat useless diff).

Regards.

devigned:
I agree with your point of view on GOPROXY and not vendoring dependencies. Let's go with it for now. If I get some static down the line about it, then I'll revisit.

devigned:
Closing in favor of #2165. I was going to have to rebase the branch @sylr shared and that didn't feel right. @sylr has taken the PR this far and I'll carry it the rest of the way. Thank you!

PR #1677, opened 2019-07-29
chore: use HAS_GOLANGCI value if defined

sylr:
Needed to build on windows.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ test-e2e:
 HAS_DEP := $(shell $(CHECK) dep)
 HAS_GOX := $(shell $(CHECK) gox)
 HAS_GIT := $(shell $(CHECK) git)
-HAS_GOLANGCI := $(shell $(CHECK) golangci-lint)
+HAS_GOLANGCI ?= $(shell $(CHECK) golangci-lint)
 HAS_GINKGO := $(shell $(CHECK) ginkgo)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -155,7 +155,7 @@ test-e2e:
 HAS_DEP := $(shell $(CHECK) dep)
 HAS_GOX := $(shell $(CHECK) gox)
 HAS_GIT := $(shell $(CHECK) git)
-HAS_GOLANGCI := $(shell $(CHECK) golangci-lint)
+HAS_GOLANGCI ?= $(shell $(CHECK) golangci-lint)
```
jackfrancis: Because `:=` is a slightly different way (simply expanded vs recursively expanded) method, do we need to do this in two steps?

```
GOLANGCI := $(shell $(CHECK) golangci-lint)
HAS_GOLANGCI ?= $GOLANGCI
```

See:

https://www.gnu.org/software/make/manual/html_node/Flavors.html#Flavors

sylr:
```shell
sylvain@khety:tmp$ cat Makefile
PWET ?= $(shell echo "pwet")

all:
	echo $(PWET)
sylvain@khety:tmp$ make
echo pwet
pwet
sylvain@khety:tmp$ PWET=plop make
echo plop
plop
```

PR #1681, opened 2019-07-29
fix: disable unattended upgrades during CSE execution

sylr:
Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -133,6 +133,17 @@ write_files:
     {{CloudInitData "aptPreferences"}}
 {{end}}
 
+- path: /etc/apt/apt.conf.d/99periodic
+  permissions: "0644"
+  owner: root
+  content: |
+    // Disable unattended upgrades at first boot
+    // This file must be removed by the provision script
+    APT::Periodic::Update-Package-Lists "0";
+    APT::Periodic::Download-Upgradeable-Packages "0";
+    APT::Periodic::AutocleanInterval "0";
+    APT::Periodic::Unattended-Upgrade "0";
+
 {{if IsIPv6Enabled}}
 - path: {{GetDHCPv6ServiceCSEScriptFilepath}}
   permissions: "0644"
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -137,8 +137,6 @@ write_files:
   permissions: "0644"
   owner: root
   content: |
-    // Disable unattended upgrades at first boot
-    // This file must be removed by the provision script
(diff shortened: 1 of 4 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -266,16 +266,24 @@ pullContainerImage() {
...
+        pids="$pids $!"
         docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep 'nginx') &
+        pids="$pids $!"
     fi
 
     # TODO: remove once ACR is available on Azure Stack
     docker rmi registry:2 &
+    pids="$pids $!"
+
+    wait $pids
```
jackfrancis: These changes don't specifically relate to disabling unattended upgrades, right? This change would block on returning control from `cleanUpContainerImages` until all the docker images are deleted, right?

sylr: Yeah I've just realized that.

It relates to the reboot of the node at the end of the cse if security upgrades have been installed.

I'll move it out and make a separate PR.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -118,6 +118,26 @@ write_files:
   content: !!binary |
     {{CloudInitData "aptPreferences"}}
 
+- path: /etc/apt/apt.conf.d/10periodic
```
jackfrancis: Do we need this `10periodic` configuration file? I would think that simply removing the `99periodic` configuration with all the `0` values would restore the default unattended upgrade configuration. Are you also suggesting we explicitly re-configure the default values for `Update-Package-Lists`, `Download-Upgradeable-Packages`, `AutocleanInterval`, and `Unattended-Upgrade`?

sylr: I'll come back to you about that cause it's been a while and I don't remember where i've taken the content of `10periodic`

parts/k8s/cloud-init/artifacts/cse_install.sh
jackfrancis: thanks!

parts/k8s/cloud-init/masternodecustomdata.yml
sylr: This `10periodic` seems to be part of the `update-notifier-common` package.

I guess we don't need to ship it ourselves.

sylr:
@jackfrancis can we move forward with this ?

sylr:
Guys ?

sylr:
ping

PR #1686, opened 2019-07-29
fix: improve container monitoring add-on

ganga1980:
**Reason for Change**:
This PR improves the on-boarding of the container monitoring add-on to the Azure Monitor. With this after successfully adding container monitoring add-on, customer can navigate to azure monitor and see health of the cluster and monitoring data without any additional steps or configurations.

This PR has the following changes 
1. creates default  log analytics workspace if the workspace not specified similar to the AKS Monitoring addon
2. customer can provider fully qualified azure resource id of the existing log analytics workspace
3. Adds the Container Insights solution on the default or existing log analytics work space
4. Adds the required tags on master node/vmss 

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


cmd/deploy.go
```diff
@@ -354,6 +356,68 @@ func autofillApimodel(dc *deployCmd) error {
...
+
+			resourceParts := strings.Split(workspaceResourceId, "/")
+			if len(resourceParts) != 9 {
+				return fmt.Errorf("%s is not a valid resourceID", workspaceResourceId)
+			}
+			workspaceSubscriptionId := resourceParts[2]
+			workspaceResourceGroup := resourceParts[4]
+			workspaceName := resourceParts[8]
+			log.Infoln("Retrieving log analytics workspace Guid, Key and location details")
+			wsId, wsKey, wsLocation, err := dc.client.GetLogAnalyticsWorkspaceInfo(ctx, workspaceSubscriptionId, workspaceResourceGroup, workspaceName)
```
CecileRobertMichon: deploy.go seems like the wrong place to do this as it's the CLI implementation for aks-engine. Some aks-engine users (such as AKS) don't use the CLI and instead call generate + az deployment create.
Also, this is creating Azure resources as far as I understand, it doesn't belong in the `autofillApimodel` function which is simply merging the apimodel with the command line flags provided by the user.

ganga1980: This change primarily for the AKS-Engine customers and not for AKS.  Followed the same pattern as Service Principal, role assignment and RG resource creation logic here. This piece of the code activated only if the user has the Container Monitoring add-on configuration in the API definition file.  For the monitoring add-on, we would need to fetch the Key of Log analytics Workspace  i.e. user provided or created by default.  Based on my understanding, this is not possible to make the  ARM APIs calls to during the generate experience.

Please let me know if you have any concerns or suggestions on this implementation.

CecileRobertMichon: Oh I see now. I guess this function is badly named (that's our fault) and is doing a lot more than it should. I'm still worried that this functionality relies entirely on using aks-engine deploy which not all customers do (not just AKS but also aks-engine customers). If that's the case, we should make it clear in the docs.

@mboersma thoughts?

ganga1980:
@mboersma @CecileRobertMichon,  do you have any insights why this error happening in CI build environment though there is no issue and that to this is error complaining on the old code .. 
is there any thing changed related to the go lint in the build environment? Appreciate any pointers on resolving this issue.

cmd/deploy.go:332:55: shadow: declaration of "err" shadows declaration at line 250 (govet)
            applicationResp, servicePrincipalObjectID, secret, err := dc.client.CreateApp(ctx, appName, appURL, replyURLs, requiredResourceAccess)

ganga1980:
> please add unit tests wherever possible for the newly added code
There are mock tests added already in the PR.  I am updating existing jenkins tests which we have for the container monitoring add-on to use this new functionality once the PR merged.

ganga1980:
> > please add unit tests wherever possible for the newly added code
> > There are mock tests added already in the PR.  I am updating existing jenkins tests which we have for the container monitoring add-on to use this new functionality once the PR merged.

Added the deployment tests to cover the new code.

pkg/api/types.go
```diff
@@ -1656,6 +1656,11 @@ func (k *KubernetesConfig) IsAADPodIdentityEnabled() bool {
 	return k.IsAddonEnabled(AADPodIdentityAddonName)
 }
 
+// IsContainerMonitoringAddonEnabled checks if the container monitoring addon is enabled
+func (k *KubernetesConfig) IsContainerMonitoringAddonEnabled() bool {
+	return k.IsAddonEnabled(ContainerMonitoringAddonName)
```
CecileRobertMichon: can you add a unit test for this? See TestIsClusterAutoscalerEnabled for an example

pkg/armhelpers/azurestack/loganalytics.go
```diff
@@ -0,0 +1,198 @@
...
+	"fmt"
+	"strings"
+
+	oi "github.com/Azure/azure-sdk-for-go/services/preview/operationalinsights/mgmt/2015-11-01-preview/operationalinsights"
+	om "github.com/Azure/azure-sdk-for-go/services/preview/operationsmanagement/mgmt/2015-11-01-preview/operationsmanagement"
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+// EnsureDefaultLogAnalyticsWorkspace ensures the default log analytics exists corresponding to specified location
+func (az *AzureClient) EnsureDefaultLogAnalyticsWorkspace(ctx context.Context, resourceGroup, location string) (workspaceResourceID string, err error) {
```
CecileRobertMichon: can this new file be unit tested as well? see disk.go and disk_test.go for an example of testing armhelpers functions. You should be able to use a mock client for this.

pkg/armhelpers/loganalytics.go
```diff
@@ -0,0 +1,240 @@
...
+	"fmt"
+	"strings"
+
+	oi "github.com/Azure/azure-sdk-for-go/services/preview/operationalinsights/mgmt/2015-11-01-preview/operationalinsights"
+	om "github.com/Azure/azure-sdk-for-go/services/preview/operationsmanagement/mgmt/2015-11-01-preview/operationsmanagement"
+	"github.com/Azure/go-autorest/autorest/to"
+)
+
+// EnsureDefaultLogAnalyticsWorkspace ensures the default log analytics exists corresponding to specified location in selected subscription
+func (az *AzureClient) EnsureDefaultLogAnalyticsWorkspace(ctx context.Context, resourceGroup, location string) (workspaceResourceID string, err error) {
```
CecileRobertMichon: same comment here about unit testing this new file

cmd/deploy.go
```diff
@@ -354,6 +355,67 @@ func autofillApimodel(dc *deployCmd) error {
 			}
 		}
 	}
+
+	if k8sConfig != nil && k8sConfig.Addons != nil && k8sConfig.IsContainerMonitoringAddonEnabled() {
+		log.Infoln("container monitoring addon enabled")
```
CecileRobertMichon: nit: can you please refactor the container monitoring addon setup to its own function since the `autofillApimodel` is already overloaded? Then you'd have something like:

```
if k8sConfig != nil && k8sConfig.Addons != nil && k8sConfig.IsContainerMonitoringAddonEnabled() {
      err = setupMonitoringAddon(k8sConfig)
      if err != nil {
           return errors.Wrap(err, "Failed to setup container monitoring")
      }
} 
```

This will help make the deploy cmd code hopefully little more readable and also allow you to isolate the monitoring addon stuff for unit testing.

CecileRobertMichon: (feel free to come up with a better name than setupMonitoringAddon, but you get the idea)

ganga1980: > (feel free to come up with a better name than setupMonitoringAddon, but you get the idea)

Thanks for the feedback. Refactored the code as suggested.

pkg/api/types.go
ganga1980: added tests for the container monitoring addon.

ganga1980: > can you add a unit test for this? See TestIsClusterAutoscalerEnabled for an example

Done.added tests for the container monitoring addon.

pkg/armhelpers/azurestack/loganalytics.go
ganga1980: @CecileRobertMichon, I look at the disk_test.go and I see these tests only for the azure stack. are you asking to do the same?

pkg/armhelpers/loganalytics.go
ganga1980: > same comment here about unit testing this new file

added the unit tests to cover these methods.

pkg/armhelpers/azurestack/loganalytics.go
ganga1980: > can this new file be unit tested as well? see disk.go and disk_test.go for an example of testing armhelpers functions. You should be able to use a mock client for this.

added the unit tests to cover these methods.

PR #1646, opened 2019-07-19
fix: omit NSG rules created by LoadBalancer during upgrade

marosset:
…ncer during upgrade

**Reason for Change**:

**Issue Fixed**:

Fixes #1624

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1687, opened 2019-07-29
Revert "fix: omit NSG rules created by LoadBalancer during upgrade"

jackfrancis:
Reverts Azure/aks-engine#1646


jackfrancis:
This change needs to be tested with upgrade before it's re-introduced.

PR #1693, opened 2019-07-30
feat: enable smart cloudprovider rate limiting

### LINKED ISSUE #420: kube-manager is being rate limited by Azure
No
---

ISSUE
---

0.28.0
---

**Kubernetes version**:
1.13.1

**What happened**:
The logs for kube-manager show frequent errors for being rate limited by Azure:

`1574:E0201 14:48:49.915225       1 attacher.go:138] azureDisk - Error checking if volumes (mycluster-dynamic-pvc-$uuid mycluster-dynamic-pvc-$uuid) are attached to current node ("k8s-agentpool1-1234567-vmss000002"). err=azure - cloud provider rate limited(read) for operation:VMSSGetInstanceView
1575:E0201 14:48:49.915258       1 operation_generator.go:186] VolumesAreAttached failed for checking on node "k8s-agentpool1-1234567-vmss000002" with: azure - cloud provider rate limited(read) for operation:VMSSGetInstanceView`

`714:E0201 09:17:48.805158       1 service_controller.go:219] error processing service instrumentation/elasticsearch-logging (will retry): error getting LB for service instrumentation/elasticsearch-logging: azure - cloud provider rate limited(read) for operation:LBList
715:I0201 09:17:48.805172       1 event.go:221] Event(v1.ObjectReference{Kind:"Service", Namespace:"instrumentation", Name:"myservice", UID:"919ec973-1358-11e9-a57f-000d3a29b594", APIVersion:"v1", ResourceVersion:"3891", FieldPath:""}): type: 'Warning' reason: 'CleanupLoadBalancerFailed' Error cleaning up load balancer (will retry): error getting LB for service mynamespace/myservice: azure - cloud provider rate limited(read) for operation:LBList
716:E0201 09:17:48.806361       1 azure_backoff.go:166] LoadBalancerClient.List(mycluster) - backoff: failure, will retry,err=azure - cloud provider rate limited(read) for operation:LBList
`

No errors when kube-manager makes requests to Azure APIs.

3 masters with availability set, agents in a VMSS. 

jackfrancis:
**Reason for Change**:

This PR changes the cloudprovider-enforced rate limiter configuration for VMSS scenarios to allow for more "burstability". Specifically, for Kubernetes LoadBalancer reconciliation, VMSS instance Get() calls are made during backend pool creation; one Get() call per VMSS instance. Because we can't anticipate in advance exactly how many VMSS instances will be served by an AKS Engine control plane (and because we don't currently have any functionality to update control plane cloudprovider configuration in response to scale up/down events), we need to configure the control plane so that it can handle the maximum number of supported VMSS instances. At present, we support up to 100 nodes per pool, and so we configure the "burstability" of the rate limit to be a factor of 100 per pool in VMSS cluster configurations.

In addition, this PR sets the actual QPS configuration to be at least 10% of the burstability tolerance, to prevent situations where the real world QPS (which is the allowed velocity of the rate limiter to release requests from the "burst queue") is never able to empty the "burst queue".

This PR does not change the node cloudprovider configuration for AKS (which would have no effect, but this skip is there for sanity).

**Issue Fixed**:

Fixes #420

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1827,6 +1827,26 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
 	return p.HasNSeriesSKU() && common.IsKubernetesVersionGe(p.OrchestratorProfile.OrchestratorVersion, "1.10.0")
 }
 
+// SetCloudProviderRateLimitDefaults sets default cloudprovider rate limiter config
+func (p *Properties) SetCloudProviderRateLimitDefaults() {
+	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS == 0 {
+		var rateLimitQPS float64
+		for _, profile := range p.AgentPoolProfiles {
+			if profile.AvailabilityProfile == VirtualMachineScaleSets {
+				rateLimitQPS += (3.0 * float64(profile.Count))
+			}
+		}
+		if rateLimitQPS > DefaultKubernetesCloudProviderRateLimitQPS {
+			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS = rateLimitQPS
+		} else {
+			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
+		}
+	}
+	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
+		p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
+	}
+}
+
 // IsReschedulerEnabled checks if the rescheduler addon is enabled
 func (k *KubernetesConfig) IsReschedulerEnabled() bool {
 	return k.IsAddonEnabled(ReschedulerAddonName)
@@ -1859,16 +1879,6 @@ func (k *KubernetesConfig) SetCloudProviderBackoffDefaults() {
 	}
 	if k.CloudProviderBackoffRetries == 0 {
 		k.CloudProviderBackoffRetries = DefaultKubernetesCloudProviderBackoffRetries
-	}
-}
-
-// SetCloudProviderRateLimitDefaults sets default cloudprovider rate limiter config
-func (k *KubernetesConfig) SetCloudProviderRateLimitDefaults() {
-	if k.CloudProviderRateLimitQPS == 0 {
-		k.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
-	}
-	if k.CloudProviderRateLimitBucket == 0 {
-		k.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
 	}
 }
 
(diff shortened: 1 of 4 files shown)


jackfrancis:
@feiskyer do you have any thoughts on an iterative improvement for programmatically generating a rate limit QPS value based on the cluster IaaS configuration? My initial prototype of multiplying `4` times the number of VMSS nodes in a cluster seems to improve reliability of LB creation for clusters that have >= 10 VMSS nodes at cluster creation time.

Again, this is a stopgap fix, as we don't currently have anything in place that will continually increase or decrease the rate limit enforcement of the control plane based on changes to the cluster's IaaS over time. The goal here is to improve real world LB functionality for folks who are using VMSS and who are unable to create LB resources immediately after the cluster comes online.

pkg/api/types.go
```diff
@@ -1827,6 +1827,38 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
 	return p.HasNSeriesSKU() && common.IsKubernetesVersionGe(p.OrchestratorProfile.OrchestratorVersion, "1.10.0")
 }
 
+// SetCloudProviderRateLimitDefaults sets default cloudprovider rate limiter config
+func (p *Properties) SetCloudProviderRateLimitDefaults() {
+	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
+		if p.HasVMSSAgentPool() {
+			var maxVMSSNodesSupported int
```
jackfrancis: @feiskyer the idea here is that for certain events (e.g., LB) VMSS read API calls are generated, so we need to provide burstability that accommodates large VMSS clusters

jackfrancis: also @yangl900 FYI

pkg/api/types.go
```diff
@@ -1827,6 +1827,38 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
...
+				}
+
+			}
+			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = rateLimitBucket
+		} else {
+			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
+		}
+	}
+	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS == 0 {
+		const minQPSToBucketFactor float64 = 0.1
```
jackfrancis: and @feiskyer @yangl900 the idea here is that we might not want to have the ratio of QPS to bucket size be too extreme; not sure if this is overthinking it...

feiskyer: makes sense

feiskyer:
@jackfrancis thanks of improving this. I think there's still no perfect way to determine those limits, because they're also depending on the number&update-frequency of services and nodes.

From Azure documentation, for each Azure subscription and tenant, Resource Manager allows up to 12,000 read requests per hour and 1,200 write requests per hour. So the safest read QPS is 3 and write QPS is 0.3 (though CloudProviderRateLimitQPSWrite should be an integer).

But those operations won't happen in every second. And usually, there would be a lot of requests during a short period when there're services or nodes changes. 

So I think the changes here make sense to reduce the rate limit issues from cloud provider side.

jackfrancis:
Validated this fix works to unblock LoadBalancer reconciliation for VMSS clusters up to 200 nodes. (2 pools of 100)

jackfrancis:
@yangl900 @palma21 FYI: these VMSS-influenced cloudprovider changes work for us in our testing of large VMSS node pool cluster configurations, I recommend investing in your own investigation to tune up the "burst queue" (i.e., bucket size) for large VMSS clusters.

palma21:
cc @chengliangli0918 @xizhamsft 
@jnoller (large cluster relation only)

PR #1694, opened 2019-07-31
feat: add support for Kubernetes 1.16.0-alpha.2

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#changelog-since-v1160-alpha1

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This also implied a (manual) revert of #1621. See kubernetes/kubernetes#80277. cc: @adelina-t.

### DIFF
--- a/parts/k8s/addons/1.16/kubernetesmasteraddons-cilium-daemonset.yaml
+++ b/parts/k8s/addons/1.16/kubernetesmasteraddons-cilium-daemonset.yaml
@@ -50,6 +50,7 @@ spec:
       annotations:
         prometheus.io/port: "9090"
         prometheus.io/scrape: "true"
+        scheduler.alpha.kubernetes.io/critical-pod: ""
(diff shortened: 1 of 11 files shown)


parts/k8s/addons/1.16/kubernetesmasteraddons-cilium-daemonset.yaml
```diff
@@ -50,6 +50,7 @@ spec:
       annotations:
         prometheus.io/port: "9090"
         prometheus.io/scrape: "true"
+        scheduler.alpha.kubernetes.io/critical-pod: ""
```
jackfrancis: So these annotations were removed from the initial 1.16.0-alpha.1 PR, because at that time in the 1.16 release cycle they had been deprecated? And we're adding them back because that deprecation was reverted?

mboersma: Exactly.

PR #1697, opened 2019-07-31
chore: use k8s test infra archive for nssm.exe

mboersma:
**Reason for Change**:
`make build-windows-k8s` began failing since the website https://nssm.cc is not accessible, so let's fetch it from a stable archive instead.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I used this change to create the Windows archive for #1694, so it has been tested. ✔️


mboersma:
Update: https://nssm.cc just came back from the dead, but since this is at least the second time it's gone down (and this time, for more than a week!), the k8s test infra archive will be more stable.

PR #1699, opened 2019-07-31
fix: document registry container image pull in VHD release notes

jackfrancis:
**Reason for Change**:

Fix VHD CI release notes omission for `registry` container image pre-pull.

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1700, opened 2019-07-31
chore: update VHDs to 2019.07.30

jackfrancis:
**Reason for Change**:

This VHD is built with `registry:2.7.1` pre-pulled. This VHD CI operation wasn't persisted to the release notes, filed #1699 to address that going forward.

Manually validated that `registry:2.7.1` was pre-pulled during these VHD image generations.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -137,7 +137,7 @@ var (
 		ImageOffer:     "aks",
 		ImageSku:       "aks-ubuntu-1604-201907",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "2019.07.29",
+		ImageVersion:   "2019.07.30",
 	}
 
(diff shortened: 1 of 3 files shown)


mboersma:
So `registry:2.7.1` isn't in these VHD release notes but #1699 fixes that...could we add it to these release notes manually?

jackfrancis:
I'm open to that, seems kind of an anti-pattern for relying on the release notes being a programmatically generated thing, also for historical tracking "when did we fix that registry thing in release notes?".

@CecileRobertMichon thoughts?

CecileRobertMichon:
I think adding it manually is fine, the fact that they are generated automatically is just a convenience in the end what matters is what's in the repo.

Issue #1689, opened 2019-07-30
Outbound rule for Standard Load balancer


sylr:
**Describe the request**

In order to be able to fix what would be the outbound public IP of pods it would be nice if we could tell aks-engine to generate a pool of X public IPs which would be used in a load balancer outbound rule.

**Explain why AKS Engine needs it**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

ritazh:
Are you referring to this? https://docs.microsoft.com/en-us/azure/virtual-network/public-ip-address-prefix

sylr:
After giving more thought about this I think it should be directly implemented in the kubernetes azure cloud provider.

See:
- https://github.com/kubernetes/cloud-provider-azure/issues/204
- https://github.com/kubernetes/cloud-provider-azure/issues/203

ritazh:
The default behavior when using SLB is that aks-engine will create the SLB, which includes a fontend PIP and the outbound rule for your agent nodes, unless you provide `agentProfile.LoadBalancerBackendAddressPoolIDs` , then the SLB and the outbound rule will need to be updated by the component that created that SLB outside of aks-engine. See this PR for more info: https://github.com/Azure/aks-engine/pull/1622

Once this SLB is created, then all services of loadbalancer type will be added to this SLB, unless its an ILB, in which case the cloud provider will create a standard ILB. So I think what you are asking for should be added to the SLB creation logic as part of aks-engine, not the cloud provider.

WDYT?

PR #1708, opened 2019-08-01
feat: add DisableOutboundSNAT feature

### LINKED ISSUE #1001: Add new options disableOutboundSNAT for standard load balancer
Refer https://github.com/kubernetes/kubernetes/pull/75282, cloud provider adds `disableOutboundSNAT` when standard load balancer is configured. 

> Azure outbound rules are not taking effect when used together with standard load balancer .
> This PR adds a new configuration for disabling SNAT for load balancer rules, so that outbound rules would be used as the egress IPs.
> In such case, disableOutboundSNAT should be set to true and loadBalancerSku should be set to standard.

Supported Kubernetes versions are v1.11.9+, v1.12.7+, v1.13.5+ and v1.14.0.

@jackfrancis

sylr:
Signed-off-by: Sylvain Rabot <s.rabot@lectra.com>

**Reason for Change**:

aks-engine currently does not allow to specify the cloud provider setting `disableOutboundSnat`.

**Issue Fixed**:

Related to #1689 
Fixes #1001

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@sylr What's the purpose of this change? Is it to override the default behavior and explicitly *disable* outbound SNAT for Standard LB configurations?

@feiskyer, does it make sense to expose this configuration to users?

jackfrancis:
Here is the current default implementation that I'm referring to:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/legacy-cloud-providers/azure/azure.go#L394

pkg/api/types.go
```diff
@@ -116,6 +116,7 @@ type FeatureFlags struct {
 	EnableCSERunInBackground bool `json:"enableCSERunInBackground,omitempty"`
 	BlockOutboundInternet    bool `json:"blockOutboundInternet,omitempty"`
 	EnableIPv6DualStack      bool `json:"enableIPv6DualStack,omitempty"`
+	DisableOutboundSNAT      bool `json:"disableOutboundSnat,omitempty"`
```
CecileRobertMichon: does this belong in `KubernetesConfig` instead? featureFlags is not meant to be used for k8s configuration but instead allow testing of experimental features without exposing them in docs. The `loadBalancerSku` and `excludeMasterFromStandardLB` knobs are in `KubernetesConfig` for example.

CecileRobertMichon: If you do move it to `KubernetesConfig`, could you also include it in docs/topics/clusterdefinitions.md?

feiskyer:
> @feiskyer, does it make sense to expose this configuration to users?

Yep, for SLB with outbound rules defined, `disableOutboundSnat` could be enabled to ensure external traffic is going through outbound rules. Refer https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview#outbound-rule.

pkg/engine/armvariables.go
```diff
@@ -80,6 +80,62 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 		kubernetesVersion = orchProfile.OrchestratorVersion + AzureStackSuffix
 	}
 
+	provisionScriptParametersCommonString := "[concat(" +
```
jackfrancis: Thank you @sylr for this code maintenance improvement :)

@CecileRobertMichon @devigned @mboersma  FYI

(Also, I manually validated that nothing was lost during translation.)

pkg/api/types.go
```diff
@@ -320,6 +320,7 @@ type CloudProviderConfig struct {
 	CloudProviderRateLimitQPSWrite    string `json:"cloudProviderRateLimitQPSWrite,omitempty"`
 	CloudProviderRateLimitBucket      int    `json:"cloudProviderRateLimitBucket,omitempty"`
 	CloudProviderRateLimitBucketWrite int    `json:"cloudProviderRateLimitBucketWrite,omitempty"`
+	CloudProviderDisableOutboundSNAT  bool   `json:"cloudProviderDisableOutboundSnat,omitempty"`
```
sylr: `Snat` instead of `SNAT` ?

marcel-dempers:
Hi all,

Just checking in to see updates on this item ?

Currently facing an issue of Standard LB not honoring outbound rules because of `disableOutboundSnat=false`. I've been comparing the LB between aks-engine and aks and support engineer from Microsoft told me disableOutboundSnat should be set to true for egress to work correctly. AKS currently sets it to true for all LB rules. AKS-engine seems to use default (false).

Let me know if there is anything I can contribute \ test out.

pkg/api/defaults.go
```diff
@@ -376,6 +376,15 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			a.OrchestratorProfile.KubernetesConfig.OutboundRuleIdleTimeoutInMinutes = DefaultOutboundRuleIdleTimeoutInMinutes
 		}
 
+		if o.KubernetesConfig.LoadBalancerSku == StandardLoadBalancerSku {
+			if o.KubernetesConfig.CloudProviderDisableOutboundSNAT == nil {
+				o.KubernetesConfig.CloudProviderDisableOutboundSNAT = to.BoolPtr(false)
+			}
+		} else {
+			// CloudProviderDisableOutboundSNAT is only valid in the context of Standard LB, statically set to false if not Standard LB
```
CecileRobertMichon: Can we drop a note in docs about this?

docs/topics/clusterdefinitions.md
```diff
@@ -73,6 +73,18 @@ $ aks-engine get-versions
...
 | outboundRuleIdleTimeoutInMinutes | no                        | Specifies a value for IdleTimeoutInMinutes to control the outbound flow idle timeout of the agent standard loadbalancer. This value is set greater than the default Linux idle timeout (15.4 min): https://pracucci.com/linux-tcp-rto-min-max-and-tcp-retries2.html                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
+| cloudProviderBackoff | no                        | Use the Azure cloudprovider exponential backoff implementation when encountering retry-able errors from the Azure API. Defaults to `true` for Kubernetes v1.14.0 and greater.                                                                                                                                                    |
```
jackfrancis: @CecileRobertMichon There was a rather large cloudprovider config documentation gap, so I used this PR as an excuse to backfill that info.

PR #1719, opened 2019-08-02
fix: address all cloud-init file waits

### LINKED ISSUE #1716: CSE race condition
Credit to @jasonzio:

> Your script has a race in it.
 
> `for i in $(seq 1 1200); do if [ -f /opt/azure/containers/provision.sh ]; then break; fi; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done;  <run the script> `
 
> The file is created (and its metadata made visible in the filesystem) before contents are written to the file; that’s just how Linux works. If your script executes the [ -f ] test between the time cloud-init creates the file and the time cloud-init puts data into that file, your attempt to run the script will fail (seeing EOF instead of useful contents).

jackfrancis:
**Reason for Change**:

Instead of checking for the existence of files (which includes empty files), we need to wait for evidence that cloud-init has actually fully paved the files.

This PR evolves the functionality of the `wait_for_file` CSE bash func so that it looks for a sentinel string to indicate that the file has been completely paved to disk via cloud-init. The func also removes the `#EOF` sentinel string after processing to accommodate future, more reliable concatenation of these files by later processes.

**Issue Fixed**:

Fixes #1716

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -243,6 +244,7 @@ write_files:
         user: client
       name: localclustercontext
     current-context: localclustercontext
+    #CLOUD_INIT_WAS_HERE
```
CecileRobertMichon: will this work? I don't know if it would be interpreted in a comment since this is a kubeconfig file

jackfrancis: `#` as the first char in a line is a valid yaml comment

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -293,7 +293,16 @@ ensureDocker() {
         wait_for_file 1200 1 $DOCKER_MOUNT_FLAGS_SYSTEMD_FILE || exit $ERR_FILE_WATCH_TIMEOUT
     fi
     DOCKER_JSON_FILE=/etc/docker/daemon.json
-    wait_for_file 1200 1 $DOCKER_JSON_FILE || exit $ERR_FILE_WATCH_TIMEOUT
+    for i in $(seq 1 1200); do
```
jackfrancis: Because this is JSON and doesn't support line comments, we have to manually validate this (instead of being able to use the `wait_for_file` convenience.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -343,7 +352,6 @@ ensureK8sControlPlane() {
     if $REBOOTREQUIRED || [ "$NO_OUTBOUND" = "true" ]; then
         return
     fi
-    wait_for_file 3600 1 $KUBECTL || exit $ERR_FILE_WATCH_TIMEOUT
```
jackfrancis: We don't need to wait for this file as `retrycmd_if_failure` will handle the error of the file not existing for us

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -208,3 +208,4 @@ else
       aptmarkWALinuxAgent unhold &
   fi
 fi
+#EOF
```
CecileRobertMichon: We aren't using the wait_for_file convenience for provision_source.sh (cse_helper) and it's still using a re-implementation of the old wait_for_file with `if [ -f $script_lib ]`:

https://github.com/Azure/aks-engine/pull/1719/files#diff-38350b45804f5298e3f609c9d4f77cacR9

(we can't use wait_for_file because the function gets defined in that file so we can't call it before having checked it's there).

We need to use a slightly different marker for that file though because `#EOF` is part of the function wait_for_file above so it being present in the file doesn't guarantee that the file has been completely written.

jackfrancis: Good catch

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -123,7 +124,7 @@ if [ -f $CUSTOM_SEARCH_DOMAIN_SCRIPT ]; then
 fi
 
 if [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
-    ensureDocker
+    ensureDocker || exit $ERR_DOCKER_START_FAIL
```
CecileRobertMichon: why this change?

CecileRobertMichon: This doesn't match the pattern in the rest CSE which is to have individual exit codes per line instead of returning non-zero and then having a broad exit code for the overall function.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -292,7 +292,16 @@ ensureDocker() {
         wait_for_file 1200 1 $DOCKER_MOUNT_FLAGS_SYSTEMD_FILE || exit $ERR_FILE_WATCH_TIMEOUT
     fi
     DOCKER_JSON_FILE=/etc/docker/daemon.json
-    wait_for_file 1200 1 $DOCKER_JSON_FILE || exit $ERR_FILE_WATCH_TIMEOUT
+    for i in $(seq 1 1200); do
+        if [ -s $DOCKER_JSON_FILE ]; then
+            jq '.' < $DOCKER_JSON_FILE && break
+        fi
+        if [ $i -eq 1200 ]; then
+            return 1
```
CecileRobertMichon: can we `exit $ERR_FILE_WATCH_TIMEOUT` here instead?

parts/k8s/cloud-init/artifacts/cse_main.sh
CecileRobertMichon: looking at the `ensureDocker` func this would happen if the json file isn't completely written in time so `ERR_FILE_WATCH_TIMEOUT` would be a more accurate exit code for that specific error. Let's replace the return by an exit in the func.

jackfrancis: I pushed a change that actually respects the return from "systemd start docker". I think pairing that with the (pre-existing and unused) `$ERR_DOCKER_START_FAIL` failure exit code makes sense.

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: done

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: Moved this inside the ensure func to be more consistent, but for the record I think pushing those error responses into the funcs themselves is less expressive/maintainable (possible improvements to consider for another day)

PR #1721, opened 2019-08-02
fix: cleanup cloud-init and waagent during VHD build

### LINKED ISSUE #1720: Clean up cloud-init artifacts on VHD-built images
@jasonzio said:

> Running waagent –deprovision cleans up the contents of /var/lib/waagent, removes user authentication credentials, (usually) removes user home directories, etc. When preparing a cloud-init based VM for clean reuse, you also need to tell cloud-init to reset itself. The command cloud-init clean -logs will remove cloud-init artifacts from /var/lib/cloud so cloud-init will re-run all stages as it did on first boot, and it cleans up the cloud-init log files so the next boot of the VM gets a clean log.
 
> When we build cloud-init-ready images, we perform both actions as the last commands in our packer scripts.

jackfrancis:
**Reason for Change**:

This PR adds waagent and cloud-init cleanup steps to the VHD CI build script.

**Issue Fixed**:

Fixes #1720 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -436,3 +436,5 @@ sudo mv /var/log /var/log.vhd
 sudo mkdir /var/log
 
 applyCIS
+cloud-init clean --logs
+waagent -force -deprovision
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -431,4 +431,8 @@ cat ${RELEASE_NOTES_FILEPATH}
 echo "END_OF_NOTES"
 set -x
 
+# Move logs from VHD creation out of /var/log
+sudo mv /var/log /var/log.vhd
+sudo mkdir /var/log
+
 applyCIS
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -431,8 +431,4 @@ cat ${RELEASE_NOTES_FILEPATH}
 echo "END_OF_NOTES"
 set -x
 
-# Move logs from VHD creation out of /var/log
-sudo mv /var/log /var/log.vhd
-sudo mkdir /var/log
-
 applyCIS
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -436,5 +436,3 @@ sudo mv /var/log /var/log.vhd
 sudo mkdir /var/log
 
 applyCIS
-cloud-init clean --logs
-waagent -force -deprovision
--- a/packer/vhd-image-builder.json
+++ b/packer/vhd-image-builder.json
@@ -46,11 +46,6 @@
         },
         {
             "type": "file",
-            "source": "packer/cleanup-vhd.sh",
-            "destination": "/home/packer/cleanup-vhd.sh"
-        },
-        {
-            "type": "file",
             "source": "packer/packer_source.sh",
             "destination": "/home/packer/packer_source.sh"
         },
@@ -193,10 +188,11 @@
             "type": "shell",
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
-                "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
-                "sudo rm -rf /home/packer",
-                "cloud-init clean --logs",
-                "waagent -force -deprovision"
+                "sudo rm -f /root/.ssh/authorized_keys",
+                "sudo rm -f /etc/machine-id",
+                "sudo touch /etc/machine-id",
+                "sudo cloud-init clean --logs",
+                "sudo -u packer /usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync"
             ]
         }
     ]
--- a/packer/vhd-image-builder.json
+++ b/packer/vhd-image-builder.json
@@ -188,13 +188,11 @@
             "type": "shell",
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
-                "sudo rm -f /root/.ssh/authorized_keys || exit 101",
-                "sudo rm -f /etc/machine-id || exit 102",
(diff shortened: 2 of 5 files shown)


packer/install-dependencies.sh
```diff
@@ -436,3 +436,5 @@ sudo mv /var/log /var/log.vhd
 sudo mkdir /var/log
 
 applyCIS
+cloud-init clean --logs
```
CecileRobertMichon: this can be done from the packer json

see https://github.com/hashicorp/packer/blob/master/examples/azure/linux_custom_image.json#L39

https://www.packer.io/docs/builders/azure.html#deprovision

jackfrancis: Is there any benefit to using the packer interface?

CecileRobertMichon: 1) by keeping the deprovision step separated from this script, the install-dependencies script can be reused as-is in other contexts (eg. managed image builder) as it only does what it's meant to, ie. install stuff on the VM
2) we also have a cleanup script which is currently running after install-dependencies. Deprovision should always be the very last step packer runs.
3) the Azure packer docs examples do it that way https://www.packer.io/docs/builders/azure.html#deprovision

packer/vhd-image-builder.json
```diff
@@ -194,7 +194,9 @@
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
                 "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
-                "sudo rm -rf /home/packer"
+                "sudo rm -rf /home/packer",
+                "cloud-init clean --logs",
```
jackfrancis: @CecileRobertMichon Is this what you're thinking?

CecileRobertMichon: yes. Is the diff with the Packer official examples intentional?  https://github.com/hashicorp/packer/blob/master/examples/azure/ubuntu_quickstart.json

compare `/usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync` with what you have below: `waagent -force -deprovision`

packer/vhd-image-builder.json
```diff
@@ -194,7 +194,9 @@
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
                 "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
-                "sudo rm -rf /home/packer"
+                "sudo rm -rf /home/packer",
```
CecileRobertMichon: this can be removed and replaced `-deprovision+user`

packer/vhd-image-builder.json
jackfrancis: Validated that the longer version works

```
azureuser@k8s-master-99506793-0:~$ sudo /usr/sbin/waagent -force -deprovision+user && export HISTSIZE=0 && sync
WARNING! The waagent service will be stopped.
WARNING! Cached DHCP leases will be deleted.
WARNING! root password will be disabled. You will not be able to login as root.
WARNING! /etc/resolvconf/resolv.conf.d/tail and /etc/resolvconf/resolv.conf.d/original will be deleted.
WARNING! azureuser account and entire home directory will be deleted.
azureuser@k8s-master-99506793-0:~$ echo $?
0
```

jackfrancis: Actually do we want to use the "+user" option to get rid of user info?

CecileRobertMichon: yes but in that case the user should be packer not azureuser

CecileRobertMichon: see my comment on the other line

jackfrancis: It's not clear how to tell waagent which users you want to "deprovision"

CecileRobertMichon: you don't, it's just the user that the process is running from - in that case it will be packer

packer/vhd-image-builder.json
```diff
@@ -193,8 +188,11 @@
             "type": "shell",
             "inline": [
                 "sudo FEATURE_FLAGS={{user `feature_flags`}} BUILD_NUMBER={{user `build_number`}} BUILD_ID={{user `build_id`}} COMMIT={{user `commit`}} /bin/bash -ux /home/packer/install-dependencies.sh",
-                "sudo /bin/bash -eux /home/packer/cleanup-vhd.sh",
-                "sudo rm -rf /home/packer"
+                "sudo rm -f /root/.ssh/authorized_keys",
+                "sudo rm -f /etc/machine-id",
+                "sudo touch /etc/machine-id",
+                "sudo cloud-init clean --logs",
```
CecileRobertMichon: thinking about this more this technically won't have any effect right now because we are moving the whole /var/log directory (which contains cloud-init logs) as part of the script

CecileRobertMichon: also we should probably move the var/log dir "clean-up" here

CecileRobertMichon: it doesn't really make sense for it to be in install-dependencies.sh

CecileRobertMichon: (or consider removing it all together if we don't want to keep /var/log.vhd for vhd cloud-init logs)

packer/install-dependencies.sh
```diff
@@ -431,8 +431,4 @@ cat ${RELEASE_NOTES_FILEPATH}
 echo "END_OF_NOTES"
 set -x
 
-# Move logs from VHD creation out of /var/log
-sudo mv /var/log /var/log.vhd
-sudo mkdir /var/log
```
CecileRobertMichon: so actually applyCIS changes the permissions on files in the /var/log dir so we might not actually want to run this after applyCIS (sorry I know I'm the one who told you we should in the first place...) I'm not sure how the `cloud-init clean --logs` affects applyCIS / if it reverts setting the permissions on cloud init logs. If that's the case we might want to consider moving all logs permission changes to CSE or cloud init instead of the VHD script. What do you think?

jackfrancis: Let's move the /var/log stuff back and then we'll see what cloud-init cleanup does

CecileRobertMichon:
implemented in #1865

Issue #1477, opened 2019-06-13
Add an e2e test that uses pod identity via keyvault flexvol


ritazh:
**Describe the request**

**Explain why AKS Engine needs it**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

CecileRobertMichon:
https://github.com/Azure/aks-engine/blob/master/examples/addons/aad-pod-identity/README.md

PR #1735, opened 2019-08-06
[WIP] chore: update AAD Pod Identity to v1.5.4

mboersma:
**Reason for Change**:
See https://github.com/Azure/aad-pod-identity/blob/master/CHANGELOG-1.5.md

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

Thanks @aramase, I'll go catch up with those changes.

jackfrancis:
Do we have comprehensive AAD pod identity E2E tests? Should we use this update as a forcing function to add them?

@ritazh thoughts?

CecileRobertMichon:
We don't: https://github.com/Azure/aks-engine/issues/1477

aramase:
We have e2e tests setup in jenkins which runs against an existing cluster. I've scheduled it to run every 6h.

jackfrancis:
Let's copy those tests over to the AKS Engine E2E tests, and include aad-pod-identity addon as part of the standard cluster config that PRs test

mboersma:
I updated the manifests with the upstream changes from AAD Pod Identity. I'll add E2E tests for the aad-pod-identity addon next.

zhiweiv:
AAD Pod Identity 1.15.1 has been released, it is better to use 1.15.1.

kkmsft:
@jackfrancis @mboersma - the e2e tests in aad-pod-identity Jenkins job (which runs periodically like @aramase mentioned), right now uses identityvalidator binary to verify the working. This can be simplified in ask-engine using az login or az login --identity and running az commands (I think @jackfrancis mentioned he preferences that  - right @jackfrancis ?). 

Additionally, there were issues with role assignments with the SP used in e2e pipeline, which is resulting in us running it on existing clusters and not recreating every time. Not sure whether that is going to go away with the aks-engine e2e infra.

PR #1746, opened 2019-08-08
chore: update kube-addon-manager to v9.0.2 and v8.9.1

mboersma:
**Reason for Change**:
As pointed out in the Kubernetes 1.16.0-alpha.3 release notes, there is a newer patch release of kube addon-manager that fixes a leader election bug.

See https://github.com/kubernetes/kubernetes#80575

Also, for all v8 versions, upgrade to v8.9.1 to ensure that all CVE-related updates are inherited for all versions of Kubernetes that use the v8 version.

See:

kubernetes/kubernetes#77623
kubernetes/kubernetes#76405

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/k8s_versions.go
```diff
@@ -88,7 +88,7 @@ var k8sComponentVersions = map[string]map[string]string{
 		"metrics-server":                   "metrics-server-amd64:v0.2.1",
 		"coredns":                          "coredns:1.5.0",
 		"kube-dns":                         "k8s-dns-kube-dns-amd64:1.15.4",
-		"addon-manager":                    "kube-addon-manager-amd64:v9.0.1",
+		"addon-manager":                    "kube-addon-manager-amd64:v9.0.2",
```
jackfrancis: Is there definitive documentation that says we should be running kube-addon-manager v8.9.x on 1.13 k8s, and v8.7.x on 1.12 k8s, and v8.6.x on 1.11/1.10 k8s ? Basically, why don't we rev to v9.0.2 everywhere?

mboersma: I think I revved these versions based on a previous compatibility matrix, but I'm not finding that right now. So I did the conservative thing and bumped 9.0.1 -> 9.0.2 wherever it showed up.

I'll try to find more details on version compatibility.

jackfrancis: Based on the fact that `v8.9.1` was released to address CVEs in the base image, we should probably rev all < v9-backed k8s versions to use that.

Ref:

https://github.com/kubernetes/kubernetes/pull/77623
https://github.com/kubernetes/kubernetes/pull/76405

PR #1747, opened 2019-08-08
test: Add stderr for kubectl port-forward

PatrickLang:
**Reason for Change**:

The test case I added for `kubectl port-forward` doesn't log stderr. If it fails, there's no clue why. This adds sufficient logging to troubleshoot it.


PatrickLang:
/hold

Relying on curl to retry the connection won't work. Since there's a race between `kubectl port-forward` setting up the connection and `curl` starting, then `curl --retry N --retry-connrefused` is needed. That's in [7.52.0](https://curl.haxx.se/changes.html#7_52_0), and the build hosts only have 7.58.0

PatrickLang:
/close

Fixed root cause and added better logs in #1775

PR #1761, opened 2019-08-12
ci: verify published VHD images

jackfrancis:
**Reason for Change**:

This isn't ideal, but due to last week's disappearance of published VHD images, maintaining a static int to validate that the number of published VHD images never changes will alert us in the place we're most likely to notice (PR CI) when something unexpected happens.

In practice, we'd move the static int forward in the "change VHD reference" PRs that land after the publication of new images.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/scripts/ensure-vhd-images.sh
+++ b/scripts/ensure-vhd-images.sh
@@ -1,0 +1,19 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+PUBLISHED_IMAGES=100
+
+exit_code=0
+
+echo "==> Using az CLI to validate published VHD images <=="
+
+IMAGES=$(az vm image list -p AKS -f aks --all -l eastus | jq '. | length')
+
+if [ "$IMAGES" != "$PUBLISHED_IMAGES" ]; then
+  echo "Expected to find $PUBLISHED_IMAGES published VHD images, instead found $IMAGES"
+  exit_code=1
+fi
+
+exit $exit_code
--- a/scripts/ensure-vhd-images.sh
+++ b/scripts/ensure-vhd-images.sh
@@ -9,6 +9,7 @@ exit_code=0
 
 echo "==> Using az CLI to validate published VHD images <=="
 
+az login --service-principal --username "${CLIENT_ID}" --password "${CLIENT_SECRET}" --tenant "${TENANT_ID}" &>/dev/null
 IMAGES=$(az vm image list -p AKS -f aks --all -l eastus | jq '. | length')
 
 if [ "$IMAGES" != "$PUBLISHED_IMAGES" ]; then
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -58,7 +58,10 @@ jobs:
   - script: make generate test-style
     displayName: Run linting rules
     workingDirectory: $(modulePath)
-  - script: make ensure-vhd
+  - script: |
+    export CLIENT_ID=$(SERVICE_PRINCIPAL_CLIENT_ID_E2E_KUBERNETES)
+    export CLIENT_SECRET=$(SERVICE_PRINCIPAL_CLIENT_SECRET_E2E_KUBERNETES)
+    make ensure-vhd
     displayName: Check if published VHD matches the expected count
     workingDirectory: $(modulePath)
   - script: make build-cross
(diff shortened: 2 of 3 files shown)


scripts/ensure-vhd-images.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+#
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT license.
+
+PUBLISHED_IMAGES=100
```
mboersma: We're capped at exactly 100 published images?

jackfrancis: That's how many images are published as of right now. This number will increment over time.

mboersma: Is this going to be a maintenance headache? (Are we going to be changing this value often?) Sorry for the naive questions.

jackfrancis: In terms of code maintenance, it shouldn't be a ton of extra overhead. Basically when we deliver a new VHD reference (e.g., a PR w/ new VHDs), we will update the count to reflect the new published VHD (at present, this is almost always 2 more: one for 16.04 and one for 18.04).

What might happen is that contributors would source their branches from commits that eventually become stale in terms of reflecting the correct number of published VHDs. Those PRs would then have the friction of failing unit tests.

Totally open to other ideas on maintaining this. I'd prefer managing this in code so we can track this over time. Perhaps we don't run this validation on PRs and instead run offline?

@CecileRobertMichon thoughts?

CecileRobertMichon: this will become a headache in my opinion.

1) When new images are published, there is a time window when the new images will appear in `az vm image list` before we merge a PR (+ rebase all other PRs) to update this count. This time window would be extended to when the images are in the Preview phase in this case because this test runs from our test sub which has preview access on our VHD images.  This means all PR E2E tests would start failing as soon as a new image is in Preview. I don't think that's acceptable. 

2) When we cut a patch release VHD just for AKS (eg. 0.38.7), we would need to update the count in master as well in order to make tests pass, which would require an additional PR (not as bad as 1) but still extra overhead that we might be able to avoid).

CecileRobertMichon: Trying to think of other ways we could do this... 

One thing that came to mind was using the release-notes dir file count but that's not very maintainable either for case 2). Will keep thinking.

CecileRobertMichon: illustration of my 1) point above: it's already 101 (so if this was checked in all PRs would be failing tests right now):

```
az vm image list -p AKS -f aks --all -l eastus | jq '. | length'                                      ✔  7899  12:00:48
101
```

jackfrancis: How about merging just the addition of the script file so we keep our validation of this (wherever that implementation goes) in one place. We can actually run it somewhere else.

CecileRobertMichon: And agreed it makes more sense to run this as a CI job not a PR job.

Maybe we could make PUBLISHED_IMAGES an environment variable so that we can update it as the expected count increments without having to merge a PR every time?

scripts/ensure-vhd-images.sh
```diff
@@ -0,0 +1,20 @@
...
+
+exit_code=0
+
+echo "==> Using az CLI to validate published VHD images <=="
+
+az login --service-principal --username "${CLIENT_ID}" --password "${CLIENT_SECRET}" --tenant "${TENANT_ID}" &>/dev/null
+IMAGES=$(az vm image list -p AKS -f aks --all -l eastus | jq '. | length')
```
CecileRobertMichon: for clarity I suggest renaming `IMAGES` to `ACTUAL_IMAGE_COUNT` and `PUBLISHED_IMAGES` to `EXPECTED_IMAGE_COUNT`

scripts/ensure-vhd-images.sh
jackfrancis: I just want to be able to track it. Like, if the test fails, I want to easily validate that it isn't because someone inadvertently messed up an env var assignment.

scripts/ensure-vhd-images.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [ "$IMAGES" != "$PUBLISHED_IMAGES" ]; then
+  echo "Expected to find $PUBLISHED_IMAGES published VHD images, instead found $IMAGES"
+  exit_code=1
```
CecileRobertMichon: why not just `exit 1` here?

seems like we don't really need `exit_code` since we're just exiting 0 at the end (which is default)

scripts/ensure-vhd-images.sh
mboersma: I think having it as a CI job makes sense, since it sounds like it has the potential to mess with the PR flow.

PR #1762, opened 2019-08-12
feat: Windows VHD build pipeline

marosset:
**Reason for Change**:

There is a desire to produce Windows VHDs curated by the AKS Engine team for use with AKS-Engine and AKS.
This packer build definition produces a Windows VHD that can be used be AKS-Engine without modification to any of the custom script extensions but also allows precise control over version of Windows Server 2019 (including patches) is used.

This work supports https://github.com/Azure/container-compute-upstream/issues/49

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -23,3 +23,6 @@ jobs:
       -e BUILD_NUMBER=$(Build.BuildNumber) \
       ${DEIS_GO_DEV_IMAGE} make run-packer-windows
     displayName: Building windows VHD
+  - task: PublishPipelineArtifact@0
+      artifactName: 'vhd-release-notes'
+      targetPath: 'release-notes.txt'
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -6,7 +6,7 @@ jobs:
   strategy:
     maxParallel: 0
   pool:
-    vmImage: Hosted Ubuntu 1604
+    name: Hosted Ubuntu 1604
   steps:
   - script: |
       docker run --rm \
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -24,5 +24,6 @@ jobs:
       ${DEIS_GO_DEV_IMAGE} make run-packer-windows
     displayName: Building windows VHD
   - task: PublishPipelineArtifact@0
+    inputs:
       artifactName: 'vhd-release-notes'
       targetPath: 'release-notes.txt'
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -17,6 +17,7 @@ jobs:
       -e TENANT_ID=${TENANT_ID} \
       -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
       -e AZURE_LOCATION=${AZURE_LOCATION} \
+      -e AZURE_VM_SIZE=${AZURE_VM_SIZE} \
       -e GIT_VERSION=$(Build.SourceVersion) \
       -e BUILD_ID=$(Build.BuildId) \
       -e BUILD_NUMBER=$(Build.BuildNumber) \
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -1,0 +1,24 @@
+trigger: none
+
+jobs:
+- job: build_vhd_windows
+  timeoutInMinutes: 120
+  strategy:
+    maxParallel: 0
+  pool:
+    vmImage: Hosted Ubuntu 1604
+  steps:
+  - script: |
+      docker run --rm \
+      -v ${PWD}:/go/src/github.com/Azure/aks-engine \
+      -w /go/src/github.com/Azure/aks-engine \
+      -e CLIENT_ID=${CLIENT_ID} \
+      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
+      -e TENANT_ID=${TENANT_ID} \
+      -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
+      -e AZURE_LOCATION=${AZURE_LOCATION} \
+      -e GIT_VERSION=$(Build.SourceVersion) \
+      -e BUILD_ID=$(Build.BuildId) \
+      -e BUILD_NUMBER=$(Build.BuildNumber) \
+      ${DEIS_GO_DEV_IMAGE} make run-packer-windows
+    displayName: Building windows VHD
--- a/packer/configure-windows-vhd.ps1
+++ b/packer/configure-windows-vhd.ps1
@@ -1,0 +1,161 @@
+<#
+    .SYNOPSIS
+        Used to produce Windows AKS images.
+
+    .DESCRIPTION
+        This script is used by packer to produce Windows AKS images.
+#>
+
+param()
+
+$ErrorActionPreference = "Stop"
+
+filter Timestamp {"$(Get-Date -Format o): $_"}
+
+function Write-Log($Message)
+{
+    $msg = $message | Timestamp
+    Write-Output $msg
+}
+
+function Disable-WindowsUpdates
+{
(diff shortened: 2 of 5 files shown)


packer/windows-vhd-builder.json
```diff
@@ -0,0 +1,67 @@
...
+            "elevated_password": "{{.WinRMPassword}}",
+            "environment_vars": "ProvisioningPhase=1",
+            "type": "powershell",
+            "script": "packer/configure-windows-vhd.ps1"
+        },
+        {
+            "restart_timeout": "10m",
+            "type": "windows-restart"
+        },
+        {
```
marosset: This second reboot is added for reliability because after the first reboot Hyper-V messes with the network adapters causing intermittent issues with the WinRM communicator.
After adding a second reboot here all the flakiness went away.
I'm looking for a good way to document that here/am open to suggestions but json doesn't all comments :(

.pipelines/vhd-builder-windows.yaml
```diff
@@ -0,0 +1,24 @@
+trigger: none
+
+jobs:
+- job: build_vhd_windows
```
CecileRobertMichon: is this pipeline running anywhere so we can see it succeeding before merging the PR?

packer/windows-vhd-builder.json
```diff
@@ -0,0 +1,67 @@
...
+            "client_secret": "{{user `client_secret`}}",
+            "tenant_id": "{{user `tenant_id`}}",
+            "subscription_id": "{{user `subscription_id`}}",
+            "location": "westus2",
+            "vm_size": "Standard_D4s_v3",
+            "os_type": "Windows",
+            "image_publisher": "MicrosoftWindowsServer",
+            "image_offer": "WindowsServer",
+            "image_sku": "2019-Datacenter-Core-smalldisk",
+            "image_version": "17763.615.1907121548",
```
CecileRobertMichon: should this be a variable so we can update it in the pipeline instead of having to merge a PR everytime it changes? Or do we want to repo check in so that we can have PRs + reviews for version bumps?

.pipelines/vhd-builder-windows.yaml
marosset: I have this running in my personal VSO instance. I can either try making the project public or finish getting a build set up in the CloudoNativeCompute project.

packer/windows-vhd-builder.json
marosset: I think we should have repo checkins to get reviews on version bumps. 
@PatrickLang for input as well.

marosset: I will move a few turn a few more fields like location and vs size into variables tho.

.pipelines/vhd-builder-windows.yaml
marosset: https://msazure.visualstudio.com/CloudNativeCompute/_build?definitionId=90028 is up and running :)

.pipelines/vhd-builder-windows.yaml
```diff
@@ -0,0 +1,29 @@
...
+      -e AZURE_VM_SIZE=${AZURE_VM_SIZE} \
+      -e GIT_VERSION=$(Build.SourceVersion) \
+      -e BUILD_ID=$(Build.BuildId) \
+      -e BUILD_NUMBER=$(Build.BuildNumber) \
+      ${DEIS_GO_DEV_IMAGE} make run-packer-windows
+    displayName: Building windows VHD
+  - task: PublishPipelineArtifact@0
+    inputs:
+      artifactName: 'vhd-release-notes'
+      targetPath: 'release-notes.txt'
```
PatrickLang: is there any reason we should add a timestamp or git hash to this filename? Not sure if we need to archive them into a single folder someplace

packer/configure-windows-vhd.ps1
```diff
@@ -0,0 +1,161 @@
...
+                    {
+                        0
+                        {
+                            Write-Log "Finished install of $fileName"
+                        }
+                        3010
+                        {
+                            WRite-Log "Finished install of $fileName. Reboot required"
+                        }
+                        default
```
PatrickLang: I think we should support .exe patches too so we can test pre-release Windows patches on a private branch

packer/windows-vhd-builder.json
```diff
@@ -0,0 +1,82 @@
...
+            "resource_group_name": "{{user `resource_group_name`}}",
+            "capture_container_name": "aksengine-vhds-windows-ws2019",
+            "capture_name_prefix": "aksengine-{{timestamp}}",
+            "storage_account": "{{user `storage_account_name`}}",
+            "communicator": "winrm",
+            "winrm_use_ssl": true,
+            "winrm_insecure": true,
+            "winrm_timeout": "10m",
+            "winrm_username": "packer"
```
PatrickLang: does the `sysprep /generalize` remove this user and disable winrm later?

packer/write-release-notes-windows.ps1
```diff
@@ -0,0 +1,93 @@
...
+$ErrorActionPreference = "Stop"
+
+function Log($Message) {
+    # Write-Output $Message
+    $Message | Tee-Object -FilePath "c:\release-notes.txt" -Append
+}
+
+Log "System Info"
+$systemInfo = Get-ItemProperty -Path 'HKLM:SOFTWARE\Microsoft\Windows NT\CurrentVersion'
+Log ("`t{0,-14} : {1}" -f "OS Name", $systemInfo.ProductName)
```
PatrickLang: learned a new PowerShell operator. need to remember  `-f`

packer/write-release-notes-windows.ps1
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+Log ""
+
+if (Test-Path 'C:\Program Files\Docker\') {
+    Log "Docker Info"
+    $dockerVersion = (docker --version) | Out-String
+    Log ("`tVersion: {0}" -f $dockerVersion)
+    Log "`tImages:"
+    $dockerImages = ((docker images) | Out-String).Split("`n") | Select-Object -Skip 1
```
PatrickLang: `docker images --format='{{json .}}'` will be easier to parse from PowerShell

Or if you want to format it directly within Docker, then check out https://docs.docker.com/config/formatting/

PatrickLang: honestly `docker images --format='{{ json . }}' | ConvertFrom-JSON | Format-Table Repository, Tag, ID` is probably enough to replace lines 78-91

.pipelines/vhd-builder-windows.yaml
marosset: The Linux pipeline uses a constant file name for the release notes as well. I'm not sure if those get archived or anything.
@CecileRobertMichon can you comment on this?

packer/configure-windows-vhd.ps1
marosset: As part of this PR or can we add that later?
Also do you have an exe to test with?

packer/write-release-notes-windows.ps1
marosset: Did not know about that - will update.
Thanks!

.pipelines/vhd-builder-windows.yaml
CecileRobertMichon: The devops job doesn't really need a timestamp because there's always just one artifact with that name per job (and we never rebuild a job, if anything we trigger a new build). When you download the artifact from devops it gets renames anyways so when we add the file to the repo we have to rename it no matter what. We archive the release notes of published images here: https://github.com/Azure/aks-engine/tree/master/releases/vhd-notes (the name of the file is sku_version of the AKS offer image).

packer/windows-vhd-builder.json
marosset: The user is removed.
Looks like WinRM is still enabled but is cleaned up some.

marosset: It looks like winrm is in the same state as a vanilla intall of ws2019-servercore-smalldisk in azure.

marosset: I found a winrm listener that was left behind.
I updated the packer job to purge the listeners right before shutdown and it it looks like that cleaned everything up.

packer/configure-windows-vhd.ps1
PatrickLang: later is fine

mboersma:
Codecov is AWOL on this one, so I'll merge manually.

PR #1429, opened 2019-06-03
refactor: run full validation for aks-engine deploy

jackfrancis:
**Reason for Change**:

This is a simpler method of running cluster configuration validation during `aks-engine deploy`

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1766, opened 2019-08-12
feat: autofill client Id and client secret for generate command

honcao:
**Reason for Change**:

feat: autofill client Id and client secret for generate command

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/generate.go
+++ b/cmd/generate.go
@@ -13,6 +13,7 @@ import (
 	"github.com/Azure/aks-engine/pkg/engine"
 	"github.com/Azure/aks-engine/pkg/engine/transform"
 	"github.com/Azure/aks-engine/pkg/i18n"
+	"github.com/gofrs/uuid"
 	"github.com/leonelquinteros/gotext"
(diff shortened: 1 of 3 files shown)


cmd/generate.go
```diff
@@ -132,7 +147,7 @@ func (gc *generateCmd) loadAPIModel() error {
 			Locale: gc.locale,
 		},
 	}
-	gc.containerService, gc.apiVersion, err = apiloader.LoadContainerServiceFromFile(gc.apimodelPath, true, false, nil)
```
jackfrancis: The existing implementation runs a full ContainerService validation (above). The new one seems only to run a vlabs ContainerService validation. Is that intentional?

CecileRobertMichon: this change is equivalent to one we made for deploy https://github.com/Azure/aks-engine/pull/1429/files

honcao: that is correct, I followed the pattern in deploy command.

PR #1769, opened 2019-08-12
chore: use numeric address to avoid DNS reverse lookup

haitch:
**Reason for Change**:

- avoid DNS reverse lookup when starting kubelet
  - in some case custom dns server used in vNet, quality of their DNS is not guaranteed, which could block kubelet service start.

I did compare impact of this change, only different is it displays **0.0.0.0/0** instead of **anywhere**, for a clean aks cluster.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
@haitch thanks!

It failed the generated files check, so you need to run `make generate` and commit the results to this PR.

jackfrancis:
@yangl900 should sign off on this

PR #1783, opened 2019-08-15
feat: enable configurable cloudprovider ratelimits for write

### LINKED ISSUE #1712: Enable configurable cloudprovider ratelimits for write
We currently support configurable cloudprovider read QPS and bucket size configuration, see `CloudProviderRateLimitQPS` and `CloudProviderRateLimitBucket`.

We need to do the same for the following two configurations:

- `CloudProviderRateLimitQPSWrite`
- `CloudProviderRateLimitBucketWrite`

These additions should mimic the existing cloudprovider implementations in terms of their type and the way they eventually reach `azure.json` on the node's VM disk.

We can inherit the existing `CloudProviderRateLimitBucket` default size for `CloudProviderRateLimitBucketWrite`, but we should set the `CloudProviderRateLimitQPSWrite` to `1.0` (instead of `3.0`).

The reason to enable this is to enable customers with large clusters to tune up or down their QPS in accordance with real world operational conditions. For example on a large cluster the current rate limits may slow down LB reconciliation time, which requres a VMSS write operation per instance:

```
  Warning  SyncLoadBalancerFailed  62s                  service-controller  Error syncing load balancer: failed to ensure load balancer: EnsureHostInPool(default/ingress-nginx-ilb): backendPoolID(/subscriptions/3014546b-7d1c-4f80-8523-f24a9976fe6a/resourceGroups/kubernetes-eastus-50990/providers/Microsoft.Network/loadBalancers/kubernetes-eastus-50990-internal/backendAddressPools/kubernetes-eastus-50990) - failed to ensure host in pool: "azure - cloud provider rate limited(write) for operation:VMSSUpdate"
```

devigned:
add write rate limits for cloudprovider

Fixes #1712 

**Requirements**:
- [x] adds unit tests


mboersma:
@devigned this needs a rebase already–sorry about that.

Could you `git commit --amend` the commit message so it starts with `feat: `? We use tooling to generate the CHANGELOG that expects to see commits with [this convention](https://www.conventionalcommits.org/en/v1.0.0-beta.2/).

devigned:
@mboersma, rebased and reworded the commit.

pkg/api/types.go
```diff
@@ -1850,6 +1854,27 @@ func (p *Properties) SetCloudProviderRateLimitDefaults() {
 			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitQPS = DefaultKubernetesCloudProviderRateLimitQPS
 		}
 	}
```
devigned: @feiskyer, I was hoping you could verify the default rate limit settings for bucket and qps write. Do the values seem reasonable or should they be decreased?

pkg/api/defaults-scheduler.go
```diff
@@ -18,16 +18,16 @@ var defaultSchedulerConfig = map[string]string{
 func (cs *ContainerService) setSchedulerConfig() {
 	o := cs.Properties.OrchestratorProfile
 
-	// If no user-configurable scheduler config values exists, use the defaults
```
jackfrancis: Let's move these optimizations to a separate PR.

Thanks!

pkg/api/types.go
feiskyer: the values look good to me.

pkg/api/defaults-scheduler.go
devigned: Resolved and applied here: https://github.com/Azure/aks-engine/pull/1791

PR #1785, opened 2019-08-15
fix: set kubelet config for containerd in templates

### LINKED ISSUE #1786: containerd scenarios CSE exit 6
For containerd scenarios, we are calling the `wait_for_file` func twice on `/etc/default/kubelet`.

This will always never succeed on the 2nd attemp, as `wait_for_file` is currently implemented to be a "run once to validate that cloud-init has successfully paved the entire file" func, and afterwards it cleans up the sentinel `#EOF` string (which means subsequent invocations of `wait_for_file` will fail, because they will never detect the presence of that string).

jackfrancis:
**Reason for Change**:

Simplify containerd kubelet config implementation: do this during cloud-init template generation instead of via sed at the last mile.

Also remove some deprecated "if less than k8s 1.8" logic to make the above more maintainable/expressive.

**Issue Fixed**:

Fixes #1786 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1733,6 +1733,11 @@ func (k *KubernetesConfig) GetOrderedKubeletConfigStringForPowershell() string {
 	return strings.TrimSuffix(buf.String(), ", ")
 }
 
+// HasContainerd returns whether or not we need the containerd runtime
+func (k *KubernetesConfig) HasContainerd() bool {
+	return k.ContainerRuntime == KataContainers || k.ContainerRuntime == Containerd
+}
+
 // IsNSeriesSKU returns true if the agent pool contains an N-series (NVIDIA GPU) VM
 func (a *AgentPoolProfile) IsNSeriesSKU() bool {
 	return common.IsNvidiaEnabledSKU(a.VMSize)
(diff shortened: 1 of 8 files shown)


pkg/api/types.go
```diff
@@ -1733,6 +1733,11 @@ func (k *KubernetesConfig) GetOrderedKubeletConfigStringForPowershell() string {
 	return strings.TrimSuffix(buf.String(), ", ")
 }
 
+// HasContainerd returns whether or not we need the containerd runtime
+func (k *KubernetesConfig) HasContainerd() bool {
+	return k.ContainerRuntime == KataContainers || k.ContainerRuntime == Containerd
```
CecileRobertMichon: this could be misleading in the future if we assume this function to mean `k.ContainerRuntime == Containerd`, now I understand your original naming `needsContainerd`...  Maybe we can make it explicit in the comment that Kata needs containerd as well so it shows in the function info when hovering from different files, wdyt? Also so we don't think later that this is an unintentional bug...

CecileRobertMichon: other than that lgtm

jackfrancis: I think `needsContainerd` actually is the best option here, with an appropriate comment

PR #1798, opened 2019-08-19
perf: Add images in aksrepos to VHD

gossion:
**Reason for Change**:

Currently AKS has already used all images in ACR (aksrepos.azurecr.io), those images were copied from external repositories. Even though the images are present in the VHD, but the ACR-specific URLs hasn't been pre-pulled into the VHD, that means AKS still need to connect to aksrepos to check the layers.

This PR is to add those images to VHD, so AKS can fully reap the benefits of pre-pulling.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
If I understand correctly, this shouldn't really affect the size of the VHD since the layers should match bitwise, it will just add the correct Docker image entries locally to avoid a round trip to `aksrepos.azurecr.io` to fetch the checksum. Correct?

We should verify the VHD produced here is roughly the same size (since we don't have much wiggle room). Is there any way to test this improves air-gapping in AKS without merging, releasing, vendoring, deploying, etc?

jackfrancis:
I confirmed that the CSE `cleanUpContainerImages` func should still work with these changes, but definitely another thing to manually validate.

gossion:
@mboersma , you are right. this shouldn't affect the size of the VHD. Those images in aksrepos.azurecr.io are copied from gcr.io/docker.io/quay.io, so they are same.

jackfrancis:
@gossion This approach looks good. What is the velocity for publishing mirroed images from k8s.gcr.io to ACR after they are published by k8.sgcr.io? AKS Engine normally kicks off a new VHD build within 24 hours of publication by k8s.gcr.io. We want to make sure that we have a similarly agile pipeline in place for ACR image publication, otherwise the entire VHD publication process will be blocked for new Kubernetes versions.

@CecileRobertMichon do you have any thoughts?

CecileRobertMichon:
This is okay with me if VHD size is unaffected. 

+1 on @jackfrancis' comment about velocity of publishing ACR images. This change means that all PRs for adding new k8s versions will fail the PR VHD pipeline check until the image is available on the ACR repo which effectively means we won't be able to merge new k8s versions until the ACR image is available so we want to make sure that velocity isn't too impacted.

packer/install-dependencies.sh
```diff
@@ -331,21 +361,27 @@ IP_MASQ_AGENT_VERSIONS="
 2.0.0
 "
 for IP_MASQ_AGENT_VERSION in ${IP_MASQ_AGENT_VERSIONS}; do
-    # TODO remove the gcr.io/google-containers image once AKS switches to use k8s.gcr.io
-    DEPRECATED_CONTAINER_IMAGE="gcr.io/google-containers/ip-masq-agent-amd64:v${IP_MASQ_AGENT_VERSION}"
-    pullContainerImage "docker" ${DEPRECATED_CONTAINER_IMAGE}
-    echo "  - ${DEPRECATED_CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
+  # TODO remove the gcr.io/google-containers image once AKS switches to use k8s.gcr.io
```
CecileRobertMichon: @gossion I think we can remove this if AKS isn't using the gcr.io/google-containers image anymore correct?

packer/install-dependencies.sh
```diff
@@ -331,21 +361,27 @@ IP_MASQ_AGENT_VERSIONS="
...
+  done
 done
 
 NGINX_VERSIONS="1.13.12-alpine"
 for NGINX_VERSION in ${NGINX_VERSIONS}; do
     CONTAINER_IMAGE="nginx:${NGINX_VERSION}"
     pullContainerImage "docker" ${CONTAINER_IMAGE}
     echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
+
+    CONTAINER_IMAGE="aksrepos.azurecr.io/mirror/nginx:${NGINX_VERSION}"
```
CecileRobertMichon: why not do a for loop on IMAGE_BASE here?

gossion: changed to for loop.

gossion:
@jackfrancis @CecileRobertMichon currently the [pipeline](https://msazure.visualstudio.com/CloudNativeCompute/_build?definitionId=89900) to copy images to ACR runs once a day. 
To unblock new k8s versions (built within 24h), maybe here is the solution:
1. run the pipeline twice a day (12h interval)
2. manually trigger the pipeline for any urgent case.

is that ok to you?

gossion:
We are moving to MCR, don't need to bake ACR images in the interim.
Closing the PR. Thanks.

PR #1801, opened 2019-08-19
fix: Increasing inotify limit to 1048576 from the default 8192

junaid-ali:
Links:
- https://linux.die.net/man/7/inotify
- https://github.com/guard/listen/wiki/Increasing-the-amount-of-inotify-watchers

Signed-off-by: Junaid Ali <junaidali.yahya@gmail.com>

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/sysctl-d-60-CIS.conf
+++ b/parts/k8s/cloud-init/artifacts/sysctl-d-60-CIS.conf
@@ -23,3 +23,5 @@ net.ipv6.conf.default.accept_redirects = 0
 vm.overcommit_memory = 1
 kernel.panic = 10
 kernel.panic_on_oops = 1
+# https://github.com/Azure/AKS/issues/772
+fs.inotify.max_user_watches = 1048576
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -15052,6 +15052,8 @@ net.ipv6.conf.default.accept_redirects = 0
 vm.overcommit_memory = 1
 kernel.panic = 10
 kernel.panic_on_oops = 1
+# https://github.com/Azure/AKS/issues/772
+fs.inotify.max_user_watches = 1048576
 `)
 
 func k8sCloudInitArtifactsSysctlD60CisConfBytes() ([]byte, error) {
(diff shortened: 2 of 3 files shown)


junaid-ali:
Backgroup for this PR: `kubectl logs <pod-name> -f` errors `No space left on device`. It appears the issue is due to inotify limit on worker nodes. Whenever more than ~15 pods are created on a node (Standard D16s v3 - 16 vcpus, 64 GiB memory), we start seeing this issue. Increasing the inotify limit from the default `8192` solves the issue.

I checked the limit on Ubuntu 18.04 and 16.04, where the limit is set to `1048576`.

junaid-ali:
@mboersma I have ran `make build` and `make ensure-generated`, pr-e2e should pass now. Also added a test for the change.

junaid-ali:
@mboersma  It looks like the `net-config-validate.sh` script is failing (please correct me if I'm wrong). I manually tested the test steps on my workstation and it seemed fine. Do you think I'm missing anything here for e2e tests?

junaid-ali:
It looks like the updated `/etc/sysctl.d/60-CIS.conf` should be baked into VHD else it will use the old `60-CIS.conf` which doesn't have my changes, so [net-config-validate.sh](https://github.com/Azure/aks-engine/pull/1801/files#diff-7814e26a85c7238422a5b1da421387a0R38) will always fail. Please correct me if I'm wrong.
CC: @mboersma @jackfrancis 

Ref: https://github.com/Azure/aks-engine/pull/999#issuecomment-486741298

CecileRobertMichon:
@junaid-ali you're correct. The fact that the `k8s_non_vhd_deployment` job passed confirms that E2E passes with your changes. In order to merge this change, we would need to comment out the net-config test for this change with a `//TODO` to re-enable when the next VHD is published (and open an issue on GitHub so we can track it on the backlog). cc @jackfrancis

junaid-ali:
@CecileRobertMichon thanks for the update. I have commented the test for this PR, and created [an issue](https://github.com/Azure/aks-engine/issues/1952) to re-enable it later.

jackfrancis:
@palma21 @yangl900 @wenwu449 FYI, we will merge this pending AKS approval

jackfrancis:
lgtm

jackfrancis:
@yangl900 @wenwu449 ping for your thoughts on this change

yangl900:
LGTM

PR #1802, opened 2019-08-19
fix: Remove public access from vhd storage container

marosset:
**Reason for Change**:

Fixes an access issue the storage container used to capture the image during VHD build pipeline is created with public access. The rest of the pipeline uses URLs with SAS tokens so public access is not needed.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
Azure DevOps build passed: https://msazure.visualstudio.com/One/_build/results?buildId=24348838

PR #1803, opened 2019-08-19
feat: add support for Kubernetes 1.13.10

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md#v11310

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -114,8 +114,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.5":         false,
 	"1.13.6":         false, // disabled because of https://github.com/kubernetes/kubernetes/issues/78308
 	"1.13.7":         false,
-	"1.13.8":         true,
+	"1.13.8":         false,
 	"1.13.9":         true,
+	"1.13.10":        true,
 	"1.14.0-alpha.1": false,
 	"1.14.0-alpha.2": false,
 	"1.14.0-beta.1":  false,


pkg/api/common/versions.go
```diff
@@ -114,8 +114,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.5":         false,
 	"1.13.6":         false, // disabled because of https://github.com/kubernetes/kubernetes/issues/78308
 	"1.13.7":         false,
-	"1.13.8":         true,
+	"1.13.8":         true, // TODO: disable when Azure Stack supports 1.13.9
```
jackfrancis: We mean 1.13.10, right?

mboersma: Yes, typo.

pkg/api/common/versions.go
```diff
@@ -114,8 +114,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.13.5":         false,
 	"1.13.6":         false, // disabled because of https://github.com/kubernetes/kubernetes/issues/78308
 	"1.13.7":         false,
-	"1.13.8":         true,
+	"1.13.8":         true, // TODO: disable when Azure Stack supports 1.13.10
```
jackfrancis: Let's remove 1.13.8 and keep 2 versions

PR #1804, opened 2019-08-19
feat: add support for Kubernetes 1.14.6

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.14.md#v1146

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -125,8 +125,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.14.1":         false,
 	"1.14.2":         false, // disabled because of https://github.com/kubernetes/kubernetes/issues/78308
 	"1.14.3":         false,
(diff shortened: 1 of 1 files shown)


pkg/api/common/versions.go
```diff
@@ -125,8 +125,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.14.1":         false,
 	"1.14.2":         false, // disabled because of https://github.com/kubernetes/kubernetes/issues/78308
 	"1.14.3":         false,
-	"1.14.4":         true,
+	"1.14.4":         true, // TODO: disable when Azure Stack supports 1.14.6
```
jackfrancis: Let's remove 1.14.4 and keep 2 versions

PR #1805, opened 2019-08-19
feat: add support for Kubernetes 1.15.3

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#v1153

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/versions.go
```diff
@@ -134,8 +134,9 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.15.0-beta.2":  false,
 	"1.15.0-rc.1":    false,
 	"1.15.0":         false,
-	"1.15.1":         true,
+	"1.15.1":         true, // TODO: disable when Azure Stack supports 1.15.3
```
jackfrancis: Let's remove 1.15.1 and keep 2 versions

PR #1807, opened 2019-08-19
chore: add k8s 1.13.10, 1.14.6, 1.15.3 to VHD script

mboersma:
**Reason for Change**:

Adds three new Kubernetes versions to the VHD.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -370,14 +370,17 @@ echo "  - busybox" >> ${RELEASE_NOTES_FILEPATH}
 
 # TODO: fetch supported k8s versions from an aks-engine command instead of hardcoding them here
 K8S_VERSIONS="
+1.15.3
```
jackfrancis: Let's get rid of the 1.15.1, 1.14.4, and 1.13.8 versions here

PR #1808, opened 2019-08-19
feat: update cloud provider rate limit for vmas.

### LINKED ISSUE #1806: enable smart cloudprovider rate limiting for VMAS agent node deployment.
Kubernetes version:
1.13
What happened:
The logs for kube-manager show frequent errors for being rate limited on Azure Stack kubernetes cluster deployment when we try to deploy an application with service type load balancer.

rjaini:
**Reason for Change**:
PR changes the cloudprovider-enforced rate limiter configuration for VMAS scenarios as well to allow for more "burstability". For large cluster on Azure Stack, app deployment which requires LoadBalancer reconciliation either NicGet call or NicUpdate call is failing. Updating the code to support "burstability" on VMAS as well.

**Issue Fixed**:
Fixes #1806 

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1834,10 +1834,12 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
 // SetCloudProviderRateLimitDefaults sets default cloudprovider rate limiter config
 func (p *Properties) SetCloudProviderRateLimitDefaults() {
 	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
-		if p.AnyAgentUsesAvailabilitySets() || p.AnyAgentUsesVirtualMachineScaleSets() {
-			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = len(p.AgentPoolProfiles) * common.MaxAgentCount
-		} else {
+		var agentPoolProfilesCount int
+		agentPoolProfilesCount = len(p.AgentPoolProfiles)
+		if agentPoolProfilesCount == 0 {
 			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
+		} else {
+			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = agentPoolProfilesCount * common.MaxAgentCount
 		}
 	}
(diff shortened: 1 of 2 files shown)


pkg/api/types.go
```diff
@@ -1834,15 +1834,8 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
...
-		if p.HasVMSSAgentPool() && !p.IsHostedMasterProfile() {
-			var rateLimitBucket int
-			for _, profile := range p.AgentPoolProfiles {
-				if profile.AvailabilityProfile == VirtualMachineScaleSets {
-					rateLimitBucket += common.MaxAgentCount
-				}
-
-			}
-			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = rateLimitBucket
+		if p.AnyAgentUsesAvailabilitySets() || p.AnyAgentUsesVirtualMachineScaleSets() {
```
jackfrancis: Thanks for simplifying the per-agent pool multiplier :)

Let's just get rid of the `if p.AnyAgentUsesAvailabilitySets() || p.AnyAgentUsesVirtualMachineScaleSets() {` altogether as there's no other kind of agent pool. We can also retire the `DefaultKubernetesCloudProviderRateLimitBucket` const if it's no longer being used.

rjaini: @jackfrancis : Sure we can do that. Only reason I didn't remove as was not sure if empty AgentPoolProfiles supported (load to default 3 Linux node deployment) ?.

jackfrancis: Perhaps we can accommodate that by keeping the default const in the event that we have zero pools (and thus try to set the rate to zero)

rjaini: I didn't get what you meant above.
Is below snippet what you are looking for where we set default value when agent poll is count is 0 ?.
		var agentPoolProfileCount int
		agentPoolProfileCount = len(p.AgentPoolProfiles)
		if agentPoolProfileCount == 0 {
			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
		} else {
			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = agentPoolProfileCount * common.MaxAgentCount

		}

jackfrancis: yes

rjaini: Updated the PR with the changes.

PR #1814, opened 2019-08-19
fix: kube-proxy addon not critical, doesn't reconcile

### LINKED ISSUE #1817: kube-proxy still running old version after upgrade
#679 appears to be a symptom of the fact that the kube-proxy daemonset is still running the older version after an `aks-engine upgrade` operation.

Investigate whether or not the daemonset spec is configured properly to reconcile via kube-addon-manager.

jackfrancis:
**Reason for Change**:

Ensure that upgrade operations always reconcile to the latest `kube-proxy` addons implementation. Also follow existing "critical addons" patterns.

Uses this as a reference:

https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/kube-proxy/kube-proxy-ds.yaml

**Issue Fixed**:

Fixes #1817

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/1.16/kubernetesmasteraddons-kube-proxy-daemonset.yaml
+++ b/parts/k8s/addons/1.16/kubernetesmasteraddons-kube-proxy-daemonset.yaml
@@ -18,6 +18,8 @@ spec:
       labels:
         component: kube-proxy
         tier: node
+      annotations:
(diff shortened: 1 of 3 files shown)


jackfrancis:
> /lgtm
> 
> I assume this fixed the upgrade issue in interactive testing?

Confirmed

PR #1816, opened 2019-08-19
fix: delete hyperkube images before starting kubelet

### LINKED ISSUE #679: kube-proxy stuck in ContainerCreating

jackfrancis:
**Reason for Change**:

Deleting hyperkube images in close proximity to bringing the kubelet online may trick the kubelet into thinking that it has a local image, which if it suddenly doesn't a few ticks later, puts that scheduling operation into an unrecoverable state.

This PR moves container image cleanup to the beginning of the CSE implementation, prior to the bootstrapping of the Kubernetes application layer, to prevent races between Kubernetes container scheduling and underlying container runtime image manipulation.

The time to do these image cleanup operations can be up to 30 seconds, and so we will continue to send the `docker rmi` processes to the background. By moving the commencement of execution to the beginning of CSE we will eliminate *most* of the flaky symptoms due to the race condition described above (which are already a low % of observed operations). The only way to entirely eliminate this race condition via CSE would be to block on image cleanup; that would add 20-30 seconds to each node bootstrap operation, which we do not want to do.

**Issue Fixed**:

Fixes #679

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -265,7 +265,14 @@ pullContainerImage() {
 
 cleanUpContainerImages() {
     # TODO remove all unused container images at runtime
-    docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'hyperkube') &
+    HYPERKUBE_IMAGES=$(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'hyperkube')
+    if [[ -n "${MASTER_NODE}" ]]; then
+      # a master node may be in the process of scheduling a kube-proxy daemonset pod with a prior version of k8s
+      # during upgrade first boot, so we wait a bit so we don't delete the image during pod bootsrapping
+      ( sleep 600; docker rmi $HYPERKUBE_IMAGES ) &
+    else
+      docker rmi $HYPERKUBE_IMAGES
+    fi
     docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'cloud-controller-manager') &
     if [ "$IS_HOSTED_MASTER" = "false" ]; then
         echo "Cleaning up AKS container images, not an AKS cluster"
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -265,14 +265,7 @@ pullContainerImage() {
 
 cleanUpContainerImages() {
     # TODO remove all unused container images at runtime
-    HYPERKUBE_IMAGES=$(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'hyperkube')
-    if [[ -n "${MASTER_NODE}" ]]; then
-      # a master node may be in the process of scheduling a kube-proxy daemonset pod with a prior version of k8s
-      # during upgrade first boot, so we wait a bit so we don't delete the image during pod bootsrapping
-      ( sleep 600; docker rmi $HYPERKUBE_IMAGES ) &
-    else
-      docker rmi $HYPERKUBE_IMAGES
-    fi
+    docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'hyperkube') &
     docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'cloud-controller-manager') &
     if [ "$IS_HOSTED_MASTER" = "false" ]; then
         echo "Cleaning up AKS container images, not an AKS cluster"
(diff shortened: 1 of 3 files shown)


jackfrancis:
The practical effect is an additional ~5-6 GB of data for the duration of the delay period until the older hyperkube images are deleted.

Before deletion:

```
azureuser@k8s-master-66023123-0:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            3.9G     0  3.9G   0% /dev
tmpfs           796M  1.4M  795M   1% /run
/dev/sda1        30G   14G   16G  47% /
tmpfs           3.9G     0  3.9G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sdb1        50G   52M   47G   1% /mnt
/dev/sdc1       252G  185M  239G   1% /var/lib/etcddisk
tmpfs           796M     0  796M   0% /run/user/1001
```

After deletion:

```
azureuser@k8s-master-66023123-0:~$ df -h
Filesystem      Size  Used Avail Use% Mounted on
udev            3.9G     0  3.9G   0% /dev
tmpfs           796M  1.4M  795M   1% /run
/dev/sda1        30G   14G   16G  47% /
tmpfs           3.9G     0  3.9G   0% /dev/shm
tmpfs           5.0M     0  5.0M   0% /run/lock
tmpfs           3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sdb1        50G   52M   47G   1% /mnt
/dev/sdc1       252G  185M  239G   1% /var/lib/etcddisk
tmpfs           796M     0  796M   0% /run/user/1001
```

Of course, there is also the side-effect of the vm being rebooted before file deletion, which would abandon the background process forever, and mean that these images would require manual cleanup.

jackfrancis:
Alternatively, we could just do the cleanup inline. Arguably that's the safest thing, although it does add (1) time and perhaps (2) some flakiness as we're reliant upon the `docker` CLI to find and delete images.

But the idea of deleting container images via the docker layer while kubelet is running (whether immediately after or 10 mins later) seems like perhaps the wrong approach.

Thoughts @mboersma @CecileRobertMichon ?

jackfrancis:
Verified that the time duration spent inside the `cleanUpContainerImages` function is negligible:

```
+ cleanUpContainerImages
+ date
Tue Sep 17 17:08:28 UTC 2019
+ '[' false = false ']'
+ echo 'Cleaning up AKS container images, not an AKS cluster'
Cleaning up AKS container images, not an AKS cluster
+ date
+ docker rmi registry:2.7.1
++ grep nginx
Tue Sep 17 17:08:28 UTC 2019
```

☝️ from a test run with the `date` command surrounding the executable code in the func, e.g.:

```
cleanUpContainerImages() {
    date
    # TODO remove all unused container images at runtime
    docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'hyperkube') &
    docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep -v "${KUBERNETES_VERSION}$" | grep 'cloud-controller-manager') &
    if [ "$IS_HOSTED_MASTER" = "false" ]; then
        echo "Cleaning up AKS container images, not an AKS cluster"
        docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep 'hcp-tunnel-front') &
        docker rmi $(docker images --format '{{.Repository}}:{{.Tag}}' | grep 'kube-svc-redirect') &
    fi

    # TODO: remove once ACR is available on Azure Stack
    docker rmi registry:2.7.1 &
    date
}
```

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -275,7 +275,7 @@ cleanUpContainerImages() {
     fi
 
     # TODO: remove once ACR is available on Azure Stack
-    docker rmi registry:2 &
+    docker rmi registry:2.7.1 &
```
jackfrancis: Cleaning this up while I'm here

PR #1819, opened 2019-08-20
feat: add support for Kubernetes 1.16.0-beta.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#changelog-since-v1160-alpha3

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
We may still need to follow up on this:

> scheduler.alpha.kubernetes.io/critical-pod annotation is removed. Pod priority (spec.priorityClassName) should be used instead to mark pods as critical.

### DIFF
--- a/parts/k8s/addons/1.16/kubernetesmasteraddons-cilium-daemonset.yaml
+++ b/parts/k8s/addons/1.16/kubernetesmasteraddons-cilium-daemonset.yaml
@@ -50,7 +50,6 @@ spec:
       annotations:
         prometheus.io/port: "9090"
         prometheus.io/scrape: "true"
-        scheduler.alpha.kubernetes.io/critical-pod: ""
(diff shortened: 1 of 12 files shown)


mboersma:
/hold

Need to audit the manifests to ensure we're not relying on the `scheduler.alpha.kubernetes.io/critical-pod` annotation.

mboersma:
```shell
$ k -n kube-system logs kube-proxy-6cv6z
W0821 16:41:49.577725       1 server.go:205] WARNING: all flags other than --config, --write-config-to, and --cleanup are deprecated. Please begin using a config file ASAP.
```
(Not an error per se, just taking notes here as I'm fixing the actual errors.)

PR #1821, opened 2019-08-20
fix: use azcopy-preview in VHD pipeline

CecileRobertMichon:
**Reason for Change**:

azcopy was removed from the latest go dev image. Move to use azcopy-preview which is the newest version of azcopy. Once the new go dev image is published, we will replace azcopy-preview with azcopy.

Also declare an in code variable for the container image instead of using a devops variable to allow us to move the image version forward with the rest of our pipelines.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder.yaml
+++ b/.pipelines/vhd-builder.yaml
@@ -7,7 +7,7 @@ trigger: none
 # - POST a new SKU to azure marketplace
 
 variables:
-  DEIS_GO_DEV_IMAGE:  'quay.io/deis/go-dev:v1.23.2' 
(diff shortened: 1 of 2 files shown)


.pipelines/vhd-builder.yaml
```diff
@@ -6,6 +6,9 @@ trigger: none
 # - generate SAS link from azure CLI
 # - POST a new SKU to azure marketplace
 
+variables:
+  CONTAINER_IMAGE:  'quay.io/deis/go-dev:v1.23.1'
```
CecileRobertMichon: @mboersma fyi moved this to a variable here

PR #1822, opened 2019-08-20
chore: add apache2-utils as an apt dependency

jadarsie:
**Reason for Change**:
This package includes _htpasswd_

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -17,6 +17,7 @@ echo "Components downloaded in this VHD build (some of the below components migh
 
 installDeps
 cat << EOF >> ${RELEASE_NOTES_FILEPATH}
+  - apache2-utils
   - apt-transport-https
   - auditd
   - blobfuse
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -47,6 +47,7 @@ ERR_APT_DAILY_TIMEOUT=98 # Timeout waiting for apt daily updates
 ERR_APT_UPDATE_TIMEOUT=99 # Timeout waiting for apt-get update to complete
 ERR_CSE_PROVISION_SCRIPT_NOT_READY_TIMEOUT=100 # Timeout waiting for cloud-init to place this (!) script on the vm
 ERR_APT_DIST_UPGRADE_TIMEOUT=101 # Timeout waiting for apt-get dist-upgrade to complete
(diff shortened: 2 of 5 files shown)


jackfrancis:
Can you add a backlog issue to add this package to the `test/e2e/kubernetes/scripts/ubuntu-installed-packages-validate.sh` E2E package validation script? If we add it before it's baked into a VHD then the test will always fail for VHD-backed cluster scenarios (all but one of the test scenarios for PRs).

@yangl900 FYI, this new package dependency is an Azure Stack requirement

jadarsie:
> Can you add a backlog issue to add this package to the `test/e2e/kubernetes/scripts/ubuntu-installed-packages-validate.sh` E2E package validation script? If we add it before it's baked into a VHD then the test will always fail for VHD-backed cluster scenarios (all but one of the test scenarios for PRs).
> 
> @yangl900 FYI, this new package dependency is an Azure Stack requirement

@jackfrancis I tried to help and pushed a few commits with my best guess. Is this going in the right direction?

jadarsie:
> Can you add a backlog issue to add this package to the `test/e2e/kubernetes/scripts/ubuntu-installed-packages-validate.sh` E2E package validation script? If we add it before it's baked into a VHD then the test will always fail for VHD-backed cluster scenarios (all but one of the test scenarios for PRs).
> 
> @yangl900 FYI, this new package dependency is an Azure Stack requirement

Here #1824

jackfrancis:
@palma21 @jluk FYI as this PR will pre-install this package into the VHD shared by AKS and AKS Engine Azure Stack.

palma21:
Assume it will be cleaned up from VHD as well as sources for AKS?

jadarsie:
> Assume it will be cleaned up from running node as well as sources list for AKS?

It can be removed, but at this point I do not think packages are removed anywhere in the code base (no new sources). What's your take @jackfrancis? I do not think that keeping these programs pose any harm.

jackfrancis:
@palma21 No, they are not cleaned up. Arguably creates more complexity during provisioning as more stuff could go wrong.

The idea here is we add things to the base image so long as they don't have any material negative effect if you aren't using it (i.e., they are just data on disk)

palma21:
@robbiezhang @wenwu449 for validation while @yangl900 is out

jackfrancis:
We'll want to look out for any CSE 102 error results once a VHD is produced with this package, and we start testing cluster buildouts using that VHD. @CecileRobertMichon FYI

PR #1823, opened 2019-08-20
fix: networkplugin conversion only if networkpolicy is empty

### LINKED ISSUE #1828: cilium NetworkPolicy api model no longer works
The example api model here no longer works:

https://github.com/Azure/aks-engine/blob/master/examples/networkpolicy/kubernetes-cilium.json

Error during `aks-engine generate`:

`Error: validating API model after populating values: networkPolicy 'cilium' is not supported with networkPlugin 'azure'`

jackfrancis:
**Reason for Change**:

This PR adds unit test coverage to generate + validate as vlabs (the new default flow with #1766), and fixes a faulty bit of logic in automatically setting the `networkPlugin` to the default value during "vlabs to unversioned" api model conversion.

This work reflects the fact that one of the examples api models we use to perform E2E tests was broken with the above commit, and we didn't know about it. In order to protect against that a full audit of the `examples/` directory was performed, and all vlabs JSON files were given basic `aks-engine generate` unit test coverage. Where appropriate, api model specs were updated when these new unit tests revealed that in fact they were non-working. Also, where appropriate, api model files (e.g., no-longer aks-engine-supported Kubernetes versions) were deleted from the `examples/` directory to improve the fidelity of stuff in there that actually works.

Additionally, `*.json.env` files which were used by an older, no longer used/maintained E2E script, were deleted.

Only api model examples under `examples/` that are `vlabs`-versioned were included in additional unit tests.

**Issue Fixed**:

Fixes #1828 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/convertertoapi.go
```diff
@@ -747,7 +747,7 @@ func setVlabsKubernetesDefaults(vp *vlabs.Properties, api *OrchestratorProfile)
 			api.KubernetesConfig.NetworkPolicy = vp.OrchestratorProfile.KubernetesConfig.NetworkPolicy
 		}
 	}
-	if api.KubernetesConfig.NetworkPlugin == "" {
+	if api.KubernetesConfig.NetworkPlugin == "" && api.KubernetesConfig.NetworkPolicy == "" {
```
jackfrancis: This is the fix to the prior logic during vlabs-->unversioned conversion. Because not every version of NetworkPolicy will work with the default NetworkPlugin values below, we don't want to brute force set the default unless NetworkPolicy is also empty.

examples/ipvs/kubernetes-msi.json
```diff
@@ -21,17 +21,16 @@
         "vmSize": "Standard_D2_v3",
         "availabilityProfile": "AvailabilitySet"
       }
```
jackfrancis: This wasn't valid JSON :|

examples/extensions/kubernetes.json
```diff
@@ -39,8 +39,8 @@
       }
     ],
     "servicePrincipalProfile": {
-      "servicePrincipalClientID": "",
-      "servicePrincipalClientSecret": ""
+      "clientId": "",
```
jackfrancis: These properties were not valid :|

examples/extensions/kubernetes.oms.json
```diff
@@ -40,8 +40,8 @@
       }
     ],
     "servicePrincipalProfile": {
-      "servicePrincipalClientID": "",
-      "servicePrincipalClientSecret": ""
+      "clientId": "",
```
jackfrancis: ditto

examples/extensions/kubernetes.preprovision.json
```diff
@@ -43,8 +43,8 @@
       }
     ],
     "servicePrincipalProfile": {
-      "servicePrincipalClientID": "",
-      "servicePrincipalClientSecret": ""
+      "clientId": "",
```
jackfrancis: ditto

examples/vnet/README.md
```diff
@@ -6,9 +6,6 @@ These examples show you how to build a customized Kubernetes cluster on Microsof
...
 
-1. first deploy a custom vnet.  An example of an arm template that does this is under directory vnetarmtemplate.
-2. next configure the example templates and deploy according to the examples:
- 1. **kubernetes.json** - deploying and using [Kubernetes](../../docs/topics/features.md#feat-custom-vnet)
- 2. **dcos.json** - deploying and using [DC/OS](../../docs/dcos.md)
- 3. **swarm.json** - deploying and using [Swarm](../../docs/swarm.md)
- 4. **swarmmodevnet.json** - deploying and using [Swarm Mode](../../docs/swarmmode.md)
+1. First, deploy a custom vnet.  An example of an arm template that does this is under directory vnetarmtemplate.
+2. Next configure the example templates, and deploy according to the examples:
+ 1. **kubernetesvnet.json** - deploying and using [Kubernetes](../../docs/topics/features.md#feat-custom-vnet)
```
jackfrancis: Cleaned up broken references and did a super quick version of making what remains more or less correct.

examples/e2e-tests/kubernetes/coreos/coreos.json
```diff
@@ -50,19 +50,11 @@
           },
           {
             "name": "keyvault-flexvolume",
-            "enabled": false,
```
jackfrancis: These properties were not valid :|

examples/e2e-tests/kubernetes/coreos/coreos.json
```diff
@@ -50,19 +50,11 @@
...
-            "cpuRequests": "20m",
-            "memoryRequests": "200Mi",
-            "cpuLimits": "20m",
-            "memoryLimits": "200Mi"
+            "enabled": false
           },
           {
             "name": "blobfuse-flexvolume",
-            "enabled": false,
```
jackfrancis: ditto

pkg/api/convertertoapi.go
CecileRobertMichon: so what happens now if NetworkPlugin is empty but not NetworkPolicy? Do we set the NetworkPlugin somewhere else?

pkg/api/defaults.go
```diff
@@ -99,7 +99,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		// we translate deprecated NetworkPolicy usage to the NetworkConfig equivalent
 		// and set a default network policy enforcement configuration
 		switch o.KubernetesConfig.NetworkPolicy {
-		case NetworkPluginAzure:
+		case NetworkPolicyAzure:
```
jackfrancis: @CecileRobertMichon I made this sort of superficial change because (1) it's more expressive to use the `NetworkPolicy` const here and (2) hopefully it answers your question below.

jackfrancis: Sorry, your question above ;)

PR #1825, opened 2019-08-21
test: retry all ssh-dependent E2E tests

jackfrancis:
**Reason for Change**:

This PR generalizes the "ssh to master and run commands on all nodes" pattern, and enables 3 retries. Ditto for "scp files to master and to nodes".

Specifically: we were establishing ssh connections for every test, and running `ssh-add` inline for every ssh operation. And little of that was being retried. So now we're establishing a re-usable ssh connection (to the first master node), and then every remote ssh execution on that connection is allowed 3 retries.

The idea is that the above will remove undesired flakiness due to network I/O flakiness that we're not in control over.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -61,6 +61,8 @@ var (
 	masterSSHPort                   string
 	masterSSHPrivateKeyFilepath     string
 	longRunningApacheDeploymentName string
+	sshConn                         *remote.Connection
+	kubeConfig                      *Config
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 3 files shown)


jackfrancis:
I'm actually gonna try to simplify this further by only establishing an ssh connection object to the primary master one time and then re-use it across tests.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -94,55 +96,44 @@ var _ = BeforeSuite(func() {
 	}
 	masterSSHPrivateKeyFilepath = cfg.GetSSHKeyPath()
 	longRunningApacheDeploymentName = "php-apache-long-running"
+	kubeConfig, err = GetConfig()
```
jackfrancis: We're now re-using a common `kubeConfig` instance across all tests (kubeconfig data representation is static from the perspective of a single test run)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -94,55 +96,44 @@ var _ = BeforeSuite(func() {
...
+	Expect(err).NotTo(HaveOccurred())
+	sshConn, err = remote.NewConnection(kubeConfig.GetServerName(), masterSSHPort, eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, masterSSHPrivateKeyFilepath)
```
jackfrancis: ditto for holding open a single ssh connection reference (to the first master in this case) for all test jobs

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -94,55 +96,44 @@ var _ = BeforeSuite(func() {
...
+	Expect(err).NotTo(HaveOccurred())
+	success := false
+	for i := 0; i < sshRetries; i++ {
+		err := util.AddToSSHKeyChain(masterSSHPrivateKeyFilepath)
```
jackfrancis: Also invoking `ssh-add` just once at the beginning of the test run as opposed to with each ssh execution.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -202,20 +176,13 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
 		It("should validate that every linux node has a root password", func() {
 			if eng.ExpandedDefinition.Properties.IsVHDDistroForAllNodes() {
-
-				kubeConfig, err := GetConfig()
-				Expect(err).NotTo(HaveOccurred())
-				master := fmt.Sprintf("%s@%s", eng.ExpandedDefinition.Properties.LinuxProfile.AdminUsername, kubeConfig.GetServerName())
 				nodeList, err := node.GetReady()
 				Expect(err).NotTo(HaveOccurred())
-				rootPasswdCmd := fmt.Sprintf("\"sudo grep '^root:[!*]:' /etc/shadow\"")
+				rootPasswdCmd := fmt.Sprintf("\"sudo grep '^root:[!*]:' /etc/shadow\" && exit 1 || exit 0")
```
jackfrancis: Changed this command so it actually succeeds on a zero exit code, instead of the prior implementation which expects an error response (which made the underlying ssh foo more complicated to implement)

PR #1831, opened 2019-08-21
chore: bump keyvault-flexvol to v0.0.12

aramase:
**Reason for Change**:

Update keyvault-flexvol to latest version v0.0.12

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Could you add this new version to the `KV_FLEXVOLUME_VERSIONS` bash array in `packer/install-dependencies.sh` so that we can pre-pull it during VHD creation?

aramase:
Thanks @jackfrancis. Done!

packer/install-dependencies.sh
```diff
@@ -312,7 +312,10 @@ for KUBE_SVC_REDIRECT_VERSION in ${KUBE_SVC_REDIRECT_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${RELEASE_NOTES_FILEPATH}
 done
 
-KV_FLEXVOLUME_VERSIONS="0.0.7"
+KV_FLEXVOLUME_VERSIONS="
+0.0.7
```
ritazh: Thanks for this PR @aramase! I don't think we need to keep the old version around. WDYT?

aramase: Yeah, I don't think we need to keep it.

PR #1832, opened 2019-08-21
chore: enable rg substring match in azure cleanup script

jackfrancis:
**Reason for Change**:

Evolves the cleanup script so that it can delete only certain rgs that match a substring.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/cleanup.sh
+++ b/test/e2e/cleanup.sh
@@ -40,7 +40,7 @@ az account set -s $SUBSCRIPTION_ID_TO_CLEANUP
 (( deadline=$(date +%s)-${expirationInSecs%.*} ))
 # find resource groups created before our deadline
 echo "Looking for resource groups created over ${EXPIRATION_IN_HOURS} hours ago..."
-for resourceGroup in $(az group list | jq --arg dl $deadline '.[] | select(.name | contains("acse-test") | not) | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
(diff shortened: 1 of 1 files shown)


test/e2e/cleanup.sh
```diff
@@ -40,12 +40,24 @@ az account set -s $SUBSCRIPTION_ID_TO_CLEANUP
 (( deadline=$(date +%s)-${expirationInSecs%.*} ))
 # find resource groups created before our deadline
 echo "Looking for resource groups created over ${EXPIRATION_IN_HOURS} hours ago..."
-for resourceGroup in $(az group list | jq --arg dl $deadline '.[] | select(.name | contains("acse-test") | not) | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
+if [ -z "$RESOURCE_GROUP_SUBSTRING" ]; then
+  for resourceGroup in $(az group list | jq --arg dl $deadline '.[] | select(.name | contains("acse-test") | not) | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
```
mboersma: `az --query` can probably handle this without piping to `jq` if you want to simplify.

Are we just trying not to match RGs with `acse-test` in the name? Here's the JMESPath way for example:

```
 az group list --query "[? !contains(name, 'acse-test')]"
```

CecileRobertMichon: + 1

this should be does _not_ contain though so ` az group list --query "[?!contains(name, 'acse-test')]"`

jackfrancis: I think the fact that we want to grab both the name *and* tags metatdata from the response suggests `jq` is the right tool, right?

PR #1835, opened 2019-08-22
fix: skip creating slb for hostedmaster

### LINKED ISSUE #1834: aks-engine shall not add standard LB with outboud publicIP in AKS scenario
**Describe the bug**
in AKS, when `networkprofile.loadbalancerSku: Standard`, it will trip this [code](https://github.com/Azure/aks-engine/blob/6e505a7750a71fc54bfac931f569d38d6fdb4b00/pkg/engine/armresources.go#L39) which undesirably added the SLB.

**Steps To Reproduce**

**Expected behavior**

**AKS Engine version**

**Kubernetes version**

**Additional context**

ritazh:
**Reason for Change**:

Skip creation of SLB when `isHostedMaster`

**Issue Fixed**:

fixes #1834 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/armresources.go
+++ b/pkg/engine/armresources.go
@@ -35,6 +35,7 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 	}
 
 	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
+		!isHostedMaster &&
 		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() &&
 		cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku {
 		isForMaster := false


pkg/engine/armresources.go
```diff
@@ -35,6 +35,7 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 	}
 
 	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
+		!isHostedMaster &&
 		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() &&
 		cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku {
```
feiskyer: could we also make `LoadBalancerSku` case insensitive?

jackfrancis: I assume we mean `strings.ToLower(cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku) == strings.ToLower(api.StandardLoadBalancerSku)` ? Not sure we have validation to prevent "standard" or "basic" strings from getting to this point, will investigate...

jackfrancis: Let's do this as a follow-up item, the critical change is to guard this altogether in AKS flows.

PR #1837, opened 2019-08-22
 [WIP] ci: add VHD rg cleanup script

CecileRobertMichon:
**Reason for Change**:

While auditing RGs I realized some resources were getting leaked by the VHD pipeline (even though there is already cleanup in place). This PR does a few things: 1) tag every resource group and storage account created by packer with a "now" date 2) add a condition `always()` to the cleanup step to ensure that it always run, even if the pipeline is interrupted or fails 3) add a script for cleaning up the resources in the sub older than 2 hours in a background scheduled job.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/vhd-rg-cleanup.sh
```diff
@@ -0,0 +1,83 @@
...
+# convert to seconds so we can compare it against the "tags.now" property in the resource group metadata
+(( expirationInSecs = ${EXPIRATION_IN_HOURS} * 60 * 60 ))
+# deadline = the "date +%s" representation of the oldest age we're willing to keep
+(( deadline=$(date +%s)-${expirationInSecs%.*} ))
+
+# clean up Packer resource groups
+
+# find packer resource groups created before our deadline
+echo "Looking for resource groups created over ${EXPIRATION_IN_HOURS} hours ago..."
+for resourceGroup in $( az group list --query "[?contains(name, 'packer-Resource-Group')]" | jq --arg dl $deadline '.[] | select(.tags.now < $dl).name' | tr -d '\"' || ""); do
```
CecileRobertMichon: @mboersma do you know any jmespath magic to check whether the value of the now tag is smaller than the deadline var? I tried for a while but couldn't get it to work...

```
 az storage account list -g aksimages --query "[?to_number(tags.now) < $dl]"
az storage account list: error: argument --query: invalid jmespath_type value: '[?to_number(tags.now) < 4]'
```

.pipelines/vhd-builder.yaml
```diff
@@ -69,8 +70,9 @@ phases:
       -e TENANT_ID=${TENANT_ID} \
       -e SA_NAME=${SA_NAME} \
       -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
-      ${CONTAINER_IMAGE} make delete-sa
-    displayName: Clean-up Storage Account
+      ${CONTAINER_IMAGE} /bin/bash -c "tag-packer-rg && tag-sa && make delete-sa"
+    displayName: Clean-up
+    condition: always()
```
CecileRobertMichon: cc @marosset, we might want to do this for the win pipeline as well

packer/vhd-image-builder.json
```diff
@@ -30,7 +30,9 @@
             "image_sku": "{{user `ubuntu_sku`}}-LTS",
             "image_version": "latest",
             "azure_tags": {
-                "dept": "Azure Container Service"
+                "os": "Linux",
+                "now": "{{user `create_time`}}",
```
marosset: can you use create_time on line 25 as well?
"capture_name_prefix": "aksengine-{{timestamp}}", -> "capture_name_prefix": "aksengine-{{user `create_time`}}",

marosset: thanks!

PR #1838, opened 2019-08-22
chore: tolerate lowercase LB SKU vals

jackfrancis:
**Reason for Change**:

Explicitly tolerate both "basic" and "Basic", and "standard" and "Standard" as values for `kubernetesConfig.loadBalancerSku`; and then do a conversion to the capitalized values before sending to Azure API.

(In fact we are tolerating all variants of "basiC", "STANDaRD", etc etc)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/defaults.go
```diff
@@ -306,6 +306,12 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			}
 		}
 
+		if strings.ToLower(a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku) == strings.ToLower(BasicLoadBalancerSku) {
```
jackfrancis: Here we are doing some "convenient" string rationalization so that by the time we compose our Azure API spec, we are always using LB SKU values of the "capitalized" form.

pkg/api/vlabs/validate.go
```diff
@@ -276,6 +276,12 @@ func (a *Properties) ValidateOrchestratorProfile(isUpdate bool) error {
 					}
 				}
 
+				if o.KubernetesConfig.LoadBalancerSku != "" {
```
jackfrancis: We were never actually doing basic SKU value validation for vlabs. Here we check against case-insensitive variants of "basic" and "standard"

pkg/api/defaults.go
```diff
@@ -306,6 +306,12 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
...
+			a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = BasicLoadBalancerSku
+		} else if strings.ToLower(a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku) == strings.ToLower(StandardLoadBalancerSku) {
```
CecileRobertMichon: let's unit test this

PR #1840, opened 2019-08-22
ci: Windows VHD winrm delay start

marosset:
**Reason for Change**:

I ran into some stability issues when trying to move the Windows VHD to start with the August images where winrm was not always reconnecting.
Hyper-V messes with networking on the reboot following it being enabled and delaying the winrm service start allows for windows to finish the reconfiguration before allowing packer to attempt connecting back to the machine.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/configure-windows-vhd.ps1
+++ b/packer/configure-windows-vhd.ps1
@@ -123,6 +123,21 @@ function Install-WindowsPatches
 
 }
 
+function Set-WinRmServiceAutoStart
+{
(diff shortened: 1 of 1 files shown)


marosset:
https://msazure.visualstudio.com/CloudNativeCompute/_build/results?buildId=24440366 successful build with this change

Issue #1541, opened 2019-06-27
Changing to reconcile mode for addons doesn't allow upgrade of tiller or customize coredns


fazdamoa:
**Describe the bug**
Upgrading tiller version now reverts to 2.11.0 after 60 seconds due to reconcile mode
Annoying for us as we have to downgrade helm version on all workstations and jenkins slaves to get helm working again (we use 2.13.0) - or reinstall tiller each upgrade?

Coredns configmap customization also doesn't work due to this. coredns-custom configmap doesn't work for us (does it work at all on non-managed aks...?)
As we rely on internal DNS for some apps we have to rework them to services/endpoints if this is staying.

**Steps To Reproduce**
helm init --upgrade

**Expected behavior**
Version of tiller should be updated to 2.13.0 and stay there
Proxy to custom dns servers should work in coredns

**AKS Engine version**
0.37.3
**Kubernetes version**
1.15

mboersma:
@fazdamoa in AKS Engine v0.37.3, the CoreDNS change was reverted, so you should be able to customize the ConfigMap again. For `helm init --upgrade` to stick, currently you'd have to edit the tiller deployment manifest to replace `Reconcile` with `EnsureExists`.

Without `Reconcile`, addons weren't being upgraded during `aks-engine upgrade`. You can read the discussion about the tradeoffs [here](https://github.com/Azure/aks-engine/pull/1401) and [here](https://github.com/Azure/aks-engine/pull/1493#issuecomment-503303279).

fazdamoa:
The CoreDNS change was reverted (to EnsureExists) in 0.37.2 - but was put back to reconcile in 0.37.3?
See commit
https://github.com/Azure/aks-engine/commit/5d48814d6995e50373c50fed6d4aad07fdef4271

Currently it is being wiped every 60 seconds when customized (on 0.37.3)

I can see the tradeoffs and fully understand :) as long as I can customize CoreDNS somehow...
Apologies for poor formatting/if I got something wrong, I'm pretty new to commenting on github.

mo-saeed:
@fazdamoa You can always change the config in the addons folder on any of the master machines (/etc/kubernetes/addons), from Reconcile to EnsureExists as @mboersma mentioned. then run helm upgrade

fazdamoa:
Thanks, updating the addons files fixed the issues. Although after the upgrade both our production cluster and dev cluster CoreDNS pods were failing the readiness probe.

[...]

PR #1841, opened 2019-08-22
feat: Allow customization of coredns (#1541)

UnwashedMeme:
This resolves one piece of #1541. The change to setting
`ConfigMap/coredns` to addon mode `reconcile` in #1493 made it so that
operators can't customize CoreDNS with stub domains anymore.

This patch adds a second volume mount from the configmap
`coredns-custom` (optional) that appends configuration to what is
shipped with aks-engine.

- [X] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/coredns.md
+++ b/docs/topics/coredns.md
@@ -1,0 +1,45 @@
+# CoreDNS customization
+
+The configuration provided by aks-engine handles most setups by forwarding to
+the dns server configured on the nodes.
+
+To customize CoreDNS ([kubernetes docs][Customizing DNS Service]) you can create
+a `configmap` that is appended to the builtin configuration. See the example below.
+
+The kubernetes docs on  also has some guidance.
+
+NB:
+
+* The custom configmap must be named `coredns-custom`
+* The configmap must contain the item `Corefile`. This is imported by the main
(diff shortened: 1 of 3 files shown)


jackfrancis:
Thanks @UnwashedMeme!

UnwashedMeme:
It looks like `pr-e2e` failed because `kubernetes-dashboard` is set to `runAsNonRoot` with a root image [I think this is the relevant bit in the logs](https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=2163&view=logs&jobId=0b1b32d8-89a7-5ec7-bd12-778f69c506c2&taskId=a5944003-1ba9-5bb4-b14f-61fd6a24c243&lineStart=847&lineEnd=848&colStart=1&colEnd=1)  I think the one with windows has the same problem

jackfrancis:
Thx for spelunking out that error @UnwashedMeme. I've observed that flake occasionally, doesn't make sense. Filed an issue here:

https://github.com/Azure/aks-engine/issues/1851

Re-kicked tests to in fact validate flakiness (obv. it doesn't correlate w/ the changeset in this PR in any event)

PR #1847, opened 2019-08-23
feat: enable deployment telemetry for generate command on Azure Stack

### LINKED ISSUE #1710: Add "deployment" resource to ARM templates targeting Azure Stack
**Describe the request**
If the target platform is Azure Stack, add the resource below to the ARM templates deployed as part of a deploy, upgrade or scale operation (a different GUID per command).

```
{
    "name": "pid-{GUID}",
    "type": "Microsoft.Resources/deployments",
    "apiVersion": "2015-11-01",
    "properties": {
        "mode": "Incremental",
        "template": {
            "$schema": "https://schema.managment.azure.com/schemas/2015-01-01/deploymentTemplage.json#",
            "contentVersion": "1.0.0.0",
            "resources": []
        }
    }
}
```

The operation GUID will generate extra telemetry data to track AKSe deployment issues.

https://docs.microsoft.com/en-us/azure/marketplace/azure-partner-customer-usage-attribution

**How**
Add a feature flag that controls whether to add or not the new resource.

deaborch:
**Reason for Change**:

Enable deployment with telemetry for generate command on Azure Stack

**Issue Fixed**:

Fixes #1710

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:


pkg/engine/armresources.go
```diff
@@ -14,6 +14,17 @@ import (
 func GenerateARMResources(cs *api.ContainerService) []interface{} {
 	var armResources []interface{}
 
+	var deploymentTelemetryEnabled bool
+	deploymentTelemetryEnabled = cs.Properties.FeatureFlags.IsFeatureEnabled("EnableTelemetry")
+	isAzureStack := cs.Properties.IsAzureStackCloud()
+
+	if deploymentTelemetryEnabled {
+		if isAzureStack {
```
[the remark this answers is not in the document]
jackfrancis: Let's add a `validateFeatureFlags()` method to the vlabs type definition to ensure that we don't enable telemetry in non-Azure Stack scenarios. See here for an example:

https://github.com/Azure/aks-engine/blob/master/pkg/api/vlabs/validate.go#L823

So, basically let's add a new method to validate feature flags generally, but for this PR the scope is just to validate the EnableTelemetry feature flag.

We can then wire in the call to this new method in this flow:

https://github.com/Azure/aks-engine/blob/master/pkg/api/vlabs/validate.go#L152

Basically, this will be a CLI convenience so that folks who enable telemetry in a non-Azure Stack configuration scenario get a fail fast error message like "telemetry is only supported in Azure Stack..."

Lemme know if that doesn't make sense!

jackfrancis:
@deaborch this looks great, thank you! Added a comment about including validation w/ this to ensure that it's clear that telemetry is a WIP and currently only functional in an Azure Stack context

examples/azure-stack/kubernetes-azurestack.json
```diff
@@ -15,6 +15,9 @@
         "customCloudProfile": {
             "portalURL": ""
         },
+        "featureFlags": {
+            "enableTelemetry": true
```
CecileRobertMichon: is this feature meant to be configurable by users? ie. users can opt out? The way I understood the intent of this change, this is to help the team track Azure stack aks-engine operations, not a feature for customers to get telemetry on their clusters. If that's the case, exposing it as an `exposeTelemetry` flag in the apimodel might be confusing for users (especially as a "featureFlag" since these are meant to be experimental knobs for feature development work). Thoughts?

jackfrancis: I think that introducing something like this via FeatureFlags is appropriate, and when it becomes more generally functional and defined (i.e., not just avail in Azure Stack) we can matriculate the feature elsewhere.

My 2¢

deaborch: I agree with Jack, this feature will at some point enable not only azurestrack to track deployments but also enable azure to track deployments. On azure however the user has to generate the guid used in the telemetry tracking.

CecileRobertMichon: sounds good

PR #1870, opened 2019-08-28
fix: set streaming-connection-idle-timeout as 4h

andyzhangx:
**Reason for Change**:

fix: set streaming-connection-idle-timeout as 4h

Per https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/#options, streaming-connection-idle-timeout means `Maximum time a streaming connection can be idle before the connection is automatically closed. 0 indicates no timeout. Example: '5m' (default 4h0m0s)`, we should set as `4h`, there is user complaining about the original `5m` value:
 `when user run kubectl exec , then session terminated after 5 idle minutes.`

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/assign @jackfrancis @CecileRobertMichon
/azp run pr-e2e

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -94,7 +94,7 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		"--pod-max-pids":                      strconv.Itoa(DefaultKubeletPodMaxPIDs),
 		"--image-pull-progress-deadline":      "30m",
 		"--enforce-node-allocatable":          "pods",
-		"--streaming-connection-idle-timeout": "5m",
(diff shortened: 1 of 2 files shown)


weinong:
lgtm. we really need to vet our k8s configurations... :-/

andyzhangx:
> @andyzhangx Does the connection interruption happen from the remote (i.e., node) side? If so, we want to get AKS sign-off on this change as it may effect their cluster behaviors if they use the default value.
> 
> @yangl900 @palma21 FYI

It's a minor fix, It happens between kubelet and user(e.g. use kubectl), user wants a bigger time since they  want to keep kubectl idle for a period

PR #1877, opened 2019-08-29
ci: Running e2e tests as part of windows VHD build

marosset:
**Reason for Change**:
Updating the windows VHD build pipeline definition and also adding an optional task to copy the VHD to classic storage for future publishing

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -103,5 +103,5 @@ jobs:
     condition: always()
   - script: |
       sudo chown -R $USER:$USER .
-    displayName: clueanup - chown all files in work directory 
+    displayName: cleanup - chown all files in work directory 
     condition: always()
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -43,10 +43,39 @@ jobs:
       -e BUILD_NUMBER=$(Build.BuildNumber) \
       ${DEIS_GO_DEV_IMAGE} make run-packer-windows
     displayName: Building windows VHD
-  - task: PublishPipelineArtifact@0
+  - task: PublishPipelineArtifact@1
     inputs:
-      artifactName: 'vhd-release-notes'
-      targetPath: 'release-notes.txt'
+      artifact: 'vhd-release-notes'
+      path: 'release-notes.txt'
+  - script: |
+      docker run --rm \
+      -v ${PWD}:/go/src/github.com/Azure/aks-engine \
+      -w /go/src/github.com/Azure/aks-engine \
+      ${DEIS_GO_DEV_IMAGE} make bootstrap
+    displayName: build aks-engine-test - bootstrap
+  - script: |
+      docker run --rm \
+      -v ${PWD}:/go/src/github.com/Azure/aks-engine \
+      -w /go/src/github.com/Azure/aks-engine \
+      ${DEIS_GO_DEV_IMAGE} make all
+    displayName: build aks-engine-test - make all
+  - script: |
+      OS_DISK_URI="$(cat packer-output | grep "OSDiskUri:" | cut -d " " -f 2)" && \
+      IMAGE_NAME="$(echo $OS_DISK_URI | cut -d "/" -f 3 | cut -d "." -f 1)" && \
+      docker run --rm \
+      -v ${PWD}:/go/src/github.com/Azure/aks-engine \
+      -w /go/src/github.com/Azure/aks-engine \
+      -e CLEANUP_ON_EXIT="true" \
+      -e CLIENT_ID=${CLIENT_ID} \
+      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
+      -e CLUSTER_DEFINITION="examples/e2e-tests/kubernetes/windows/hybrid/definition.json" \
+      -e LOCATION=${AZURE_LOCATION} \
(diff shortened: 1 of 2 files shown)


jackfrancis:
This is an interesting idea, should we do this for 16.04 and 18.04 VHDs as well @CecileRobertMichon ?

marosset:
I'll look into enabling this for the Linux VHDs as a seperate PR.

.pipelines/vhd-builder-windows.yaml
```diff
@@ -1,12 +1,32 @@
+# Required pipeline variables:
```
CecileRobertMichon: should we be running the pipeline on PRs?

marosset: It looks like both the this and the Linux VHD pipeline builds are getting triggered for all PRs.
I was attempting to restrict PR triggered builds to only PRs that changes these two folders.

Would you prefer these only get triggered manually? If so I can update this to be pr:none.

jchauncey: you can also just turn off triggers completely with `trigger: none`. What value is there with running this build for every pr?

marosset: I already have trigger:none in this build definition and I don't think that trigger: none prevents the pipeline from running for new PRs
If you look at the linux VHD pipeline the yaml also has trigger: none and I see a bunch of builds trigger by PR 
https://msazure.visualstudio.com/One/_build?definitionId=35580&_a=summary many show the trigger as 'PR automated' and only a few show 'Manually triggered' as the cause.

I took a look at https://docs.microsoft.com/en-us/azure/devops/pipelines/build/triggers?view=azure-devops&tabs=yaml and my understanding is that trigger: is used for pushing tags or commits to branches and pr: is used for whenever new pull requests created against branches.

My intention here was to reduce the amount of builds triggered by PRs from all PRs to just those changing files in specific directories.

If we want to turn off builds triggered by all PRs (which I'm ok with) i'll I can update to so both trigger:none and pr:none are set.

marosset: It looks like the default behavior if pr: is not specified is to trigger builds for all PRs but @CecileRobertMichon showed me that the PR behavior for the Linux builds is overridden in the Azure DevOps build settings.
I'll update the .yaml file to add pr:none so by default no builds get triggered and then set finer grained controls in DevOps

.pipelines/vhd-builder-windows.yaml
```diff
@@ -17,13 +32,76 @@ jobs:
...
+      -e CLIENT_SECRET="$(CLIENT_SECRET)" \
+      -e TENANT_ID=${TENANT_ID} \
+      -e SA_NAME=${SA_NAME} \
+      -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
+      ${DEIS_GO_DEV_IMAGE} make delete-sa
+    displayName: cleanup - deleate storage account
+    condition: always()
+  - script: |
+      sudo chown -R $USER:$USER .
+    displayName: clueanup - chown all files in work directory
```
CecileRobertMichon: typo here
```suggestion
    displayName: cleanup - chown all files in work directory 
```

jchauncey:
What subscription are these images being published to (the linux one too)? Should we think about moving them to the prod subscription we control to make publishing easier? I am trying to keep the amount of information we have to supply the release jobs down to a minimum so people dont have to worry about it breaking or how it works in case someone leaves. 

Obviously its not super important we move the artifacts right this moment but we should consider that transition shortly.

marosset:
@jchauncey i want to start publishing these to the prod sub right after this PR merges.

.pipelines/vhd-builder-windows.yaml
```diff
@@ -1,12 +1,27 @@
...
+pr: none
 
 jobs:
 - job: build_vhd_windows
   timeoutInMinutes: 120
   strategy:
     maxParallel: 0
   pool:
-    name: Hosted Ubuntu 1604
+    name: $(BUILD_POOL)
```
jchauncey: the only thing i notice is that this is a little weird but /shrug not a big deal

marosset: I needed to move away from using hosted agents because the jobs went past the 60 minute time limit after adding the E2E tests.
I made it a pipeline variable so the .yaml could be reusable across DevOps accounts.

.pipelines/vhd-builder-windows.yaml
```diff
@@ -17,13 +32,76 @@ jobs:
...
-      artifactName: 'vhd-release-notes'
-      targetPath: 'release-notes.txt'
+      artifact: 'vhd-release-notes'
+      path: 'release-notes.txt'
+  - script: |
+      docker run --rm \
+      -v ${PWD}:/go/src/github.com/Azure/aks-engine \
+      -w /go/src/github.com/Azure/aks-engine \
+      ${DEIS_GO_DEV_IMAGE} make bootstrap
+    displayName: build aks-engine-test - bootstrap
```
CecileRobertMichon: this shouldn't be required as a separate step, `make build` depends on `make generate` which itself depends on bootstrap

CecileRobertMichon: lgtm otherwise

PR #1884, opened 2019-08-30
chore: disable tiller addon by default

sozercan:
**Reason for Change**:

Disables Tiller addon by default. Users can choose to enable it via addon setting.
Tiller version in aks-engine cluster usually lags behind Helm version which users need to `helm init --upgrade`, which is broken due to #1541 

**Issue Fixed**:

**Requirements**:

- [x] updates documentation
- [x] updates unit tests

**Notes**:


jackfrancis:
If we did unit tests correctly, this should pass...

pkg/api/addons_test.go
```diff
@@ -243,7 +243,7 @@ func TestSetAddonsConfig(t *testing.T) {
 				},
 				{
 					Name:    TillerAddonName,
-					Enabled: to.BoolPtr(true),
+					Enabled: to.BoolPtr(DefaultTillerAddonEnabled),
```
jackfrancis: Thank you :)

PR #1891, opened 2019-09-02
chore: make the pki key size settable.

andyliuliming:
**Reason for Change**:

make the key size settable will help:
1. for some customer want longer or shorter lenght of this size.
2. this will help projects which reference this project save Unit Test time if we set it shorter in test code.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


andyliuliming:
Hi @jackfrancis  could you please help approve? thanks.

pkg/helpers/pki_test.go
```diff
@@ -235,3 +235,15 @@ func TestCreatePkiKeyCertPair(t *testing.T) {
 		t.Errorf("unexpected error thrown while executing CreatePkiKeyCertPair : %s", err.Error())
 	}
 }
+
+func TestSetSetPkiKeySize(t *testing.T) {
```
CecileRobertMichon: there's a typo here, should be `TestSetSetPkiKeySize`

pkg/helpers/pki_test.go
```diff
@@ -235,3 +235,15 @@ func TestCreatePkiKeyCertPair(t *testing.T) {
...
+	if GetPkiKeySize() != DefaultPkiKeySize {
+		t.Errorf("the default size is not correct.")
+	}
+
+	// set to the default size so it will have no impact on other testing.
+	SetPkiKeySize(DefaultPkiKeySize)
```
CecileRobertMichon: this test isn't really testing functionality since the expected value doesn't change before/after the set... Could we maybe set it to something different then set it back to 0?

andyliuliming: yes, I ever thought about it, but I think it may cause some problem if the testing is running in multi-thread mode.

pkg/helpers/pki_test.go
andyliuliming: good catch, will update it.

pkg/helpers/pki_test.go
CecileRobertMichon: in any case, we should undo all changes after the test runs and set back the override size to `0` so the default is used.

If we aren't able to set the value without impacting other tests running in parallel maybe it's a sign that a public var isn't the best way to expose this... Maybe it should be a parameter in SetPropertiesDefault instead?

andyliuliming: yes, add it into the SetPropertiesDefault method is better.
but that would be one API change.
would that be too big for this kind of purpose?

if you agree, I can move to that part :)

andyliuliming: change the API means a lot of the unit tests code change.
and even for the "projects" which reference aks-engine :)

CecileRobertMichon: Yes, on second thought maybe changing the SetPropertiesDefault function signature isn't the best idea since it's used in so many places. @devigned do you have any thoughts on this?

andyliuliming:
@devigned yes, give it one option struct and add key size into it is the best solution, but means a lot of change, but I will try to do that, thanks.

andyliuliming:
> Sorry to do this to you, but one last change. Would you change the struct type names to be Params rather than options? For example, `api.PropertiesDefaultsOptions` would be `api.PropertiesDefaultsParams`.
> 
> I was misleading in how I described this originally. Constructors should have options while Funcs have parameters. I think this strategy was described well in the go-cloud design doc here: https://github.com/google/go-cloud/blob/master/internal/docs/design.md#option-structs.
> 
> Thank you for undertaking this much larger change than you had originally signed up for. It is appreciated.

no problem, will update it.

devigned:
Just waiting on tests to pass, but the code looks great. Thank you, @andyliuliming.

PR #1906, opened 2019-09-04
feat: add support for Kubernetes 1.16.0-beta.2

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#v1160-beta2

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/service/service.go
+++ b/test/e2e/kubernetes/service/service.go
@@ -18,7 +18,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-const commandTimeout = 1 * time.Minute
+const commandTimeout = 3 * time.Minute
 
 // List holds a list of services returned from kubectl get svc
 type List struct {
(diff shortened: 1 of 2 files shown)


mboersma:
This is failing on an e2e test:

```
    with a linux agent pool
        [It] should be able to produce working LoadBalancers 
```

This obviously wasn't an issue with 1.16.0-beta.1. I'm looking into it.

test/e2e/kubernetes/service/service.go
```diff
@@ -18,7 +18,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-const commandTimeout = 1 * time.Minute
+const commandTimeout = 3 * time.Minute
```
jackfrancis: @feiskyer we're seeing something interesting w/ the introduction of 1.16.0-beta.2 (we haven't observed this in earlier 1.16 or other versions of k8s):

- kubectl delete svc <svcname> is taking > 1 minute, which is what our E2E tests have previously configured as the timeout threshold

When I've repro'd locally, the "service deleted" message returns immediately, e.g.:

```
$ kubectl delete svc ingress-nginx-elb
service "ingress-nginx-elb" deleted
```

... and then things stall for 1-2 minutes until control is finally returned to the shell.

We use the same version of kubectl as the version of the API we're testing, so is it possible something has regressed in the kubectl @ version 1.16.0-beta.2 ? Or is it possible that the API itself in -beta.2 is delaying its response to kubectl that would allow it to return control to the shell?

feiskyer: have you found something interesting from kube-controller-manager logs? The only changes may related I think is service load balancer finalizer goes to beta (beta means enabled by default) in this release (https://github.com/kubernetes/kubernetes/pull/81691).

jackfrancis: Will do if we keep seeing this.

PR #1908, opened 2019-09-04
chore: update ip-masq-agent to v2.5.0

aramase:
**Reason for Change**:

- Update ip-masq-agent to v2.5.0 which has support for ipv6

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/containeraddons/1.16/ip-masq-agent.yaml
```diff
@@ -24,32 +24,34 @@ spec:
...
+        - key: CriticalAddonsOnly
+          operator: Exists
+        - key: node-role.kubernetes.io/master
+          operator: Equal
+          value: "true"
+          effect: NoSchedule
+        - operator: "Exists"
+          effect: NoExecute
+        - operator: "Exists"
+          effect: NoSchedule
```
sylr: I would leave the list items aligned with the array name, i.e.:
```
azeaze:
- one
- two
```
That's how kubectl yaml is output

pkg/api/types.go
```diff
@@ -1196,13 +1196,25 @@ func (p *Properties) GetNonMasqueradeCIDR() string {
...
 			}
 		} else {
-			// kube-proxy still only understands single cidr and is not changed for ipv6 dual stack phase 1
-			// so only pass the ipv4 cidr which is the first one in the list as arg to kube proxy
 			if p.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack") {
 				nonMasqCidr = strings.Split(p.OrchestratorProfile.KubernetesConfig.ClusterSubnet, ",")[0]
-			} else {
-				nonMasqCidr = p.OrchestratorProfile.KubernetesConfig.ClusterSubnet
+				_, ipnet, _ := net.ParseCIDR(nonMasqCidr)
+				return ipnet.String()
```
jackfrancis: I have a slight preference for using the `nonMasqCidr` common reference to store the return value, and then having the single return statement at the bottom.

pkg/api/types.go
```diff
@@ -1196,13 +1196,25 @@ func (p *Properties) GetNonMasqueradeCIDR() string {
...
+}
+
+// GetSecondaryNonMasqueradeCIDR returns second cidr in case of dualstack clusters
+func (p *Properties) GetSecondaryNonMasqueradeCIDR() string {
+	var nonMasqCidr string
+	if !p.IsHostedMasterProfile() {
+		if p.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack") {
+			nonMasqCidr = strings.Split(p.OrchestratorProfile.KubernetesConfig.ClusterSubnet, ",")[1]
+			_, ipnet, _ := net.ParseCIDR(nonMasqCidr)
+			return ipnet.String()
```
jackfrancis: ditto above comment

pkg/api/types.go
aramase: Thanks @jackfrancis! Updated.

jackfrancis:
One last thing: let's add `2.5.0` to the `IP_MASQ_AGENT_VERSIONS` bash string array in `packer/install-dependencies.sh` so this new image version is pre-downloaded into the VHD.

Thanks @aramase!

PR #1913, opened 2019-09-05
chore: update accelerated networking-capable SKU list

### LINKED ISSUE #1894: Automate accelerated networking VM sizes whitelist
https://github.com/Azure/aks-engine/blob/master/pkg/helpers/helpers.go#L118

At least Standard_D16_v3 is missing.

mboersma:
**Reason for Change**:
Updates the list of accelerated networking-capable VM SKUs. Specifically it adds these:

```
 'Standard_B12ms',
 'Standard_B16ms',
 'Standard_B20ms',
 'Standard_D48_v3',
 'Standard_D48s_v3',
 'Standard_DS11-1_v2',
 'Standard_DS12-1_v2',
 'Standard_DS12-2_v2',
 'Standard_DS12_v2',
 'Standard_DS13-2_v2',
 'Standard_DS3_v2',
 'Standard_E16-4s_v3',
 'Standard_E16-8s_v3',
 'Standard_E20s_v3',
 'Standard_E4-2s_v3',
 'Standard_E48_v3',
 'Standard_E48s_v3',
 'Standard_E64i_v3',
 'Standard_E64is_v3',
 'Standard_E8-2s_v3',
 'Standard_E8-4s_v3',
 'Standard_F48s_v2'
```

**Issue Fixed**:
Fixes #1894

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I wrote a quick python script to do this update, but it's ugly. I can clean it up and include it in this PR or a later one.

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -143,6 +143,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.16.0-alpha.2": true,
 	"1.16.0-alpha.3": true,
 	"1.16.0-beta.1":  true,
+	"1.16.0-beta.2":  true,
 }
 
 // GetDefaultKubernetesVersion returns the default Kubernetes version, that is the latest patch of the default release
--- a/test/e2e/kubernetes/service/service.go
+++ b/test/e2e/kubernetes/service/service.go
@@ -18,7 +18,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-const commandTimeout = 1 * time.Minute
+const commandTimeout = 3 * time.Minute
 
 // List holds a list of services returned from kubectl get svc
 type List struct {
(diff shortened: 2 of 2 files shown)


jackfrancis:
Any way we could consume the whitelist in the `TestAcceleratedNetworkingSupported` test to validate our defaults flow that is responsible for turning accelerated networking on/off?

jackfrancis:
Also, it looks like there are test gaps for these assignments:

https://github.com/Azure/aks-engine/blob/master/pkg/engine/networkinterfaces.go#L331

(there may be indirect test coverage in `armresources_test.go`, but we should cover the property conversion from `AgentPoolProfile.AcceleratedNetworkingEnabled` (and the windows equivalent) to `compute.VirtualMachineScaleSetNetworkConfigurationProperties.EnableAcceleratedNetworking` in `networkinterfaces_test.go` for more maintainability.

Basically, as of now, all the interesting foo is in the api model composition flows (defaults enforcement + whitelist), and the ARM spec just literally tracks the property assignment in the profile, so if we do the following we should be good:

- UT for api model defaults flow against the whitelist
- validate api model property boolean assignments are respected during ARM spec composition

( hope that all makes sense :| )

PR #1918, opened 2019-09-06
chore: use hyperkube for ccm if >= 1.16.0-beta.1

### LINKED ISSUE #1915: Switch to  cloud-provider-azure for cloud-controller-manager image
Because of https://github.com/kubernetes/kubernetes/pull/81029, cloud-controller-manager image won't be released since v1.16.beta.x. 

~~We should switch to hyperkube image when configuring cloud-controller-manager.~~

Update: I didn't notice there's a change also dropping cloud-controller-manager support in hyperkube https://github.com/kubernetes/kubernetes/pull/81219. After this PR, we could only use the ccm image built from cloud-provider-azure repo.

Refer the example manifests here: https://github.com/kubernetes/cloud-provider-azure/tree/master/examples/out-of-tree.

jackfrancis:
**Reason for Change**:

Use hyperkube image for cloud-controller-manager for k8s 1.16.0.beta.1 and greater.

**Issue Fixed**:

Fixes #1915

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -401,9 +401,9 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
     else
       HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
       if (( $(echo ${KUBERNETES_VERSION} | cut -d"." -f2) < 16 )); then
-        CCM_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
-        pullContainerImage "docker" ${CCM_IMAGE}
-        echo "  - ${CCM_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+        CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
+        pullContainerImage "docker" ${CONTAINER_IMAGE}
+        echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
       fi
     fi
     extractHyperkube "docker"
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -400,9 +400,11 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
       HYPERKUBE_URL="mcr.microsoft.com/k8s/azurestack/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
       HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
-      CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
-      pullContainerImage "docker" ${CONTAINER_IMAGE}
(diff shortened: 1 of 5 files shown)


jackfrancis:
@feiskyer is this what we want?

@ritazh FYI

feiskyer:
@jackfrancis thanks of adding this.

jackfrancis:
@CecileRobertMichon Does `install-dependencies.sh` have k8s version context? Isn't it designed to accommodate pre-pulling artifacts used across a variety of versions?

CecileRobertMichon:
@jackfrancis it doesn't but right now ccm and hyperkube are using the same "for versions" loop which means the script will fail once we add 1.16.0 to the loop since `k8s.gcr.io/cloud-controller-manager-amd64:v1.16.0` won't be published.
 https://github.com/Azure/aks-engine/blob/master/packer/install-dependencies.sh#L401

feiskyer:
Sorry, I didn't notice there's a change also dropping cloud-controller-manager support in hyperkube https://github.com/kubernetes/kubernetes/pull/81219. After this one, we could only use the ccm image built from cloud-provider-azure repo.

/hold

jackfrancis:
@feiskyer is looks like kubernetes/cloud-provider-azure#231 is merged

Where should we be deriving CCM image for >= 1.16?

jackfrancis:
@CecileRobertMichon CCM image download in VHD CI for < 1.16 addressed

packer/install-dependencies.sh
```diff
@@ -400,9 +400,11 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
       HYPERKUBE_URL="mcr.microsoft.com/k8s/azurestack/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
       HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
-      CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
-      pullContainerImage "docker" ${CONTAINER_IMAGE}
-      echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+      if (( $(echo ${KUBERNETES_VERSION} | cut -d"." -f2) < 16 )); then
+        CCM_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
```
CecileRobertMichon: any reason for the variable rename? the rest of the file is using `CONTAINER_IMAGE` everywhere so this is breaking the pattern (and we can't search and replace anymore). Arguably it's better to have specific variables everywhere and not reuse the same var but then we should make that change for all the components otherwise it's inconsistent.

jackfrancis: I was having trouble reading the code, but sure, we can do that change independently of this PR and for all enumeration blocks distinctly.

feiskyer:
> Where should we be deriving CCM image for >= 1.16?

It's published on the MCR, but the release pipeline is not ready yet. Would update here when the images are ready.

jackfrancis:
@feiskyer Thank you. Just to be clear: we will introduce v1.16.0 without cloud-controller-manager support, correct?

@ritazh FYI

feiskyer:
> @feiskyer Thank you. Just to be clear: we will introduce v1.16.0 without cloud-controller-manager support, correct?

it's ok, cloud-controller-manager support could be added later since the images are still not ready yet.

PR #1921, opened 2019-09-09
fix: add hostname to CSE command so it shows in ARM deployment errors

### LINKED ISSUE #1496: Print hostname in VMSS CSE command
We are using vmss extension for provisioning vm with aks-engine!

At present we come across a scenario where provisioning failed with following error:

 

We have created a log analytics work space to track all logs but when I queried with following I couldn’t found more information.

 
I just see one output couldn’t get more information.

 

My question is :

Is there  a way to get more information or logs to investigate?

CecileRobertMichon:
**Reason for Change**:

Make troubleshooting easier when CSE fails for one node (especially when using scale set with many instances). Right now there is no easy/quick way to know which instance failed from a scale set.

**Issue Fixed**:

Fixes #1496 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
TODO:

- [ ] add to troubleshooting doc
- [ ] equivalent for windows


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -4,6 +4,9 @@ set -x
 echo $(date),$(hostname), startcustomscript>>/opt/m
 AZURE_STACK_ENV="azurestackcloud"
 
+
+exit 1
```
CecileRobertMichon: this is just for testing, will remove after

pkg/engine/virtualmachinescalesets.go
```diff
@@ -312,7 +312,7 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
...
+				"commandToExecute": fmt.Sprintf("[concat('echo $(date),$(hostname); retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; "+outBoundCmd+" for i in $(seq 1 1200); do grep -Fq \"EOF\" /opt/azure/containers/provision.sh && break; if [ $i -eq 1200 ]; then exit 100; else sleep 1; fi; done; ', variables('provisionScriptParametersCommon'),%s,variables('provisionScriptParametersMaster'), ' /usr/bin/nohup /bin/bash -c \"/bin/bash /opt/azure/containers/provision.sh >> /var/log/azure/cluster-provision.log 2>&1\"')]", generateUserAssignedIdentityClientIDParameter(userAssignedIDEnabled)),
```
CecileRobertMichon: @PatrickLang @marosset is this something that would be useful as well for Windows VMs? If so, what's the equivalent for powershell that I could add to https://github.com/Azure/aks-engine/pull/1921/files#diff-438c69b891b52277ec1344112d40e26cR729?

marosset: you could try replacing "commandToExecute": "[concat('powershell.exe -ExecutionPolicy Unrestricted -command with "commandToExecute": "[concat('**echo %TIME%,%COMPUTERNAME% &&** powershell.exe -ExecutionPolicy Unrestricted -command   below when the CSE gets run.
that should echo the current time and computer name before powershell.exe gets called to run the script.

marosset: If you want the date too (based on the doc comments) i think you'll need to echo echo %DATE%,%TIME%,%COMPUTERNAME% 
That will output something similar to the following 
marosset@MAROSSET-WKS C:\Users\marosset
$ echo %DATE%,%TIME%,%COMPUTERNAME%
Wed 09/11/2019, 9:33:53.08,MAROSSET-WKS

PR #1935, opened 2019-09-11
fix: not overwrite service principal profile for hosted master

norshtein:
**Reason for Change**:

For hosted master, we do not want to overwrite service principal with "msi" even if `useManagedIdentity` is `true`, that's because some addons running on agent nodes may not support authenticating using managed identity, so even managed identity is used, we still want to provide a service principal, which is dedicated to be used by addons.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/armvariables_test.go
+++ b/pkg/engine/armvariables_test.go
@@ -184,13 +184,34 @@ func TestK8sVars(t *testing.T) {
 		t.Errorf("unexpected diff while expecting equal structs: %s", diff)
 	}
 
-	// Test with ubuntu 16.04 distro
-	cs.Properties.AgentPoolProfiles[0].Distro = api.Ubuntu
+	// Test with MSI
+	cs.Properties.OrchestratorProfile.KubernetesConfig.UseManagedIdentity = true
 	varMap, err = GetKubernetesVariables(cs)
 	if err != nil {
 		t.Fatal(err)
 	}
 
+	expectedMap["servicePrincipalClientId"] = "msi"
+	expectedMap["servicePrincipalClientSecret"] = "msi"
+	expectedMap["useManagedIdentityExtension"] = "true"
+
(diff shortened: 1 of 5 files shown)


pkg/engine/armvariables_test.go
```diff
@@ -184,13 +184,34 @@ func TestK8sVars(t *testing.T) {
...
 	}
 
+	// Test with MSI
+	cs.Properties.OrchestratorProfile.KubernetesConfig.UseManagedIdentity = true
+	varMap, err = GetKubernetesVariables(cs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedMap["servicePrincipalClientId"] = "msi"
```
jackfrancis: These UT are for AKS Engine flows (i.e., not AKS "HostedMaster" flows)

pkg/engine/armvariables_test.go
```diff
@@ -390,6 +411,21 @@ func TestK8sVars(t *testing.T) {
...
 	}
 
+	// Test with HostedMaster + MSI
+	cs.Properties.OrchestratorProfile.KubernetesConfig.UseManagedIdentity = true
+	varMap, err = GetKubernetesVariables(cs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	expectedMap["useManagedIdentityExtension"] = "true"
```
jackfrancis: This AKS ("HostedMaster") UT reflects that when an AKS configuration includes the UseManagedIdentity = true setting we *only* set the `useManagedIdentityExtension` variable to `"true"`, but we *do not* change the `servicePrincipalClientId` and `servicePrincipalClientSecret` variables to `"msi"`. In other words, this section of the UT proves that @norshtein's change is doing what we want it to do, and that it will be protected by UT over time.

PR #1865, opened 2019-08-27
fix: add deprovision step to packer script

CecileRobertMichon:
**Reason for Change**:

Running waagent –deprovision cleans up the contents of /var/lib/waagent, removes user authentication credentials, (usually) removes user home directories, etc. When preparing a cloud-init based VM for clean reuse, you also need to tell cloud-init to reset itself. The command cloud-init clean -logs will remove cloud-init artifacts from /var/lib/cloud so cloud-init will re-run all stages as it did on first boot, and it cleans up the cloud-init log files so the next boot of the VM gets a clean log.
https://www.packer.io/docs/builders/azure.html#deprovision

**Issue Fixed**:

Fixes #1720
#1717 

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #1944, opened 2019-09-13
fix: update the location of the VHD footprint file on VMs

CecileRobertMichon:
**Reason for Change**:

https://github.com/Azure/aks-engine/pull/1865 added a deprovision step to the VHD script. It introduced a regression where "FULL_INSTALL_REQUIRED" was detected as false on all VMs deployed with the new VHD because the file `/var/log/azure/golden-image-install.complete` was missing (`/var/log.vhd` and `/var/log/` were deleted as part of the deprovision step). This PR does two things:
1) Update the location of the VHD log file to `/opt/azure/vhd-install.complete` so that VMs built from VHD still have the log file on disk.
2) Check the distro used and use that to determine if the VM built from a VHD in CSE. In addition, to avoid regressions like this one in the future, check for the existence of the footprint/log file if the distro is the AKS VHD distro and fail if the file is not found.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -50,12 +50,16 @@ else
...
+VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
+if [[ "${IS_VHD}" = true ]] then
+    # TODO (09/13): uncomment when new VHD is published
+    # if [ -f $VHD_LOGS_FILEPATH ]; then
+    #     echo "Using VHD distro but file $VHD_LOGS_FILEPATH not found"
+    #     exit $ERR_VHD_FILE_NOT_FOUND
+    # fi
     echo "detected golden image pre-install"
     FULL_INSTALL_REQUIRED=false
-    rm -rf /home/packer
```
CecileRobertMichon: this isn't needed in here anymore since the deprovision step is happening in the vhd build

packer/install-dependencies.sh
```diff
@@ -3,19 +3,19 @@ source /home/packer/provision_installs.sh
 source /home/packer/provision_source.sh
 source /home/packer/packer_source.sh
 
-RELEASE_NOTES_FILEPATH=/var/log/azure/golden-image-install.complete
+VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
```
jackfrancis: Just to be clear, the rename of this var is just to better describe it?

CecileRobertMichon: correct, and I used the same var in CSE so it's clear that they are the same thing and should both be updated if one of them changes.

jackfrancis:
Manually validated:

```
+ VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
+ [[ true = true ]]
+ echo 'detected golden image pre-install'
detected golden image pre-install
+ FULL_INSTALL_REQUIRED=false
+ [[ UBUNTU == UBUNTU ]]
+ '[' false = true ']'
+ echo 'Golden image; skipping dependencies installation'
Golden image; skipping dependencies installation
```

PR #1946, opened 2019-09-13
test: crash docker and validate against common node count

jackfrancis:
**Reason for Change**:

We are running docker crash simulations against a list of nodes whose length we determine at the time the test runs. We also need to do all validations against that common list.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -940,7 +940,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 
 						var expectedReadyNodes int
 						if !eng.ExpandedDefinition.Properties.HasLowPriorityScaleset() {
-							expectedReadyNodes = eng.NodeCount()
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -940,7 +940,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 
 						var expectedReadyNodes int
 						if !eng.ExpandedDefinition.Properties.HasLowPriorityScaleset() {
-							expectedReadyNodes = eng.NodeCount()
```
jackfrancis: By setting this expected count against the value in the api model, we don't take into account that the list we enumerate through in line 918 above may not have the exact same numbers due to <some> environmental behaviors. Because we are specifically testing docker crash simulation in this test, we want to make sure the node count before and after operate against a common number of nodes.

PR #1956, opened 2019-09-16
test: fix "run multiple commands in succession" implementation

jackfrancis:
**Reason for Change**:

Our current implementation was evaluating the duration time to execute a `kubectl` command, rather than actually introspecting a pod container's status to determine the duration of the container runtime execution lifecycle.

Also fixed some bugs around "Waiting for pods to complete".

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -48,7 +48,7 @@ const (
 	kubeSystemPodsReadinessChecks             = 6
 	sleepBetweenRetriesWhenWaitingForPodReady = 1 * time.Second
 	timeoutWhenWaitingForPodOutboundAccess    = 1 * time.Minute
-	stabilityCommandTimeout                   = 5 * time.Second
+	stabilityCommandTimeout                   = 10 * time.Second
 	windowsCommandTimeout                     = 1 * time.Minute
 	validateNetworkPolicyTimeout              = 3 * time.Minute
 	validateDNSTimeout                        = 2 * time.Minute
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -48,7 +48,7 @@ const (
 	kubeSystemPodsReadinessChecks             = 6
 	sleepBetweenRetriesWhenWaitingForPodReady = 1 * time.Second
 	timeoutWhenWaitingForPodOutboundAccess    = 1 * time.Minute
-	stabilityCommandTimeout                   = 10 * time.Second
+	stabilityCommandTimeout                   = 1 * time.Second
 	windowsCommandTimeout                     = 1 * time.Minute
 	validateNetworkPolicyTimeout              = 3 * time.Minute
 	validateDNSTimeout                        = 2 * time.Minute
--- a/test/e2e/kubernetes/pod/pod.go
+++ b/test/e2e/kubernetes/pod/pod.go
@@ -237,7 +237,7 @@ func RunWindowsPod(image, name, namespace, command string, printOutput bool, sle
 type podRunnerCmd func(string, string, string, string, bool, time.Duration, time.Duration, time.Duration) (*Pod, error)
 
 // RunCommandMultipleTimes runs the same command 'desiredAttempts' times
-func RunCommandMultipleTimes(podRunnerCmd podRunnerCmd, image, name, command string, desiredAttempts int, sleep, duration, timeout time.Duration) (int, error) {
+func RunCommandMultipleTimes(podRunnerCmd podRunnerCmd, image, name, command string, desiredAttempts int, sleep, commandTimeout, timeout time.Duration) (int, error) {
 	var successfulAttempts int
 	var actualAttempts int
 	logResults := func() {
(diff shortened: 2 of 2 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -48,7 +48,7 @@ const (
 	kubeSystemPodsReadinessChecks             = 6
 	sleepBetweenRetriesWhenWaitingForPodReady = 1 * time.Second
 	timeoutWhenWaitingForPodOutboundAccess    = 1 * time.Minute
-	stabilityCommandTimeout                   = 5 * time.Second
+	stabilityCommandTimeout                   = 1 * time.Second
```
jackfrancis: These are the commands that are re-using this const:

- `nc -vz 8.8.8.8 53 || nc -vz 8.8.4.4 53`
- `nc -vz kubernetes 443 && nc -vz kubernetes.default.svc 443 && nc -vz kubernetes.default.svc.cluster.local 443`
- `nc -vz apache-pod.default.svc.cluster.local 80`
- `nc -vz bbc.co.uk 80 || nc -vz google.com 443 || nc -vz microsoft.com 80`

I think it's fair to say that a 1 second execution duration of any of the above is "interesting" and we'd want our stability tests to fail as a result.

test/e2e/kubernetes/pod/pod.go
```diff
@@ -250,29 +250,25 @@ func RunCommandMultipleTimes(podRunnerCmd podRunnerCmd, image, name, command str
...
 		var p *Pod
 		var err error
-		p, err = podRunnerCmd(image, podName, "default", command, true, sleep, duration, timeout)
-
+		p, err = podRunnerCmd(image, podName, "default", command, true, sleep, 10*time.Second, timeout)
 		if err != nil {
 			return successfulAttempts, err
 		}
-		succeeded, err := p.WaitOnSucceeded(sleep, duration)
+		succeeded, err := p.WaitOnSucceeded(sleep, timeout)
```
jackfrancis: We don't want to apply the "command timeout" to this wait func. There are two things here:

- wait for the k8s api to schedule and run a pod to completion
- ensure that the pod container execution time is within expected timeouts

For the first one we want to be more tolerant of unexpected environmental conditions.

test/e2e/kubernetes/pod/pod.go
```diff
@@ -250,29 +250,25 @@ func RunCommandMultipleTimes(podRunnerCmd podRunnerCmd, image, name, command str
...
 		if err != nil {
-			log.Printf("pod did not succeed in time %s\n", err)
+			log.Printf("pod %s did not succeed in time\n", podName)
+			return successfulAttempts, err
 		}
-		cmd := exec.Command("k", "logs", podName, "-n", "default")
```
jackfrancis: pod logs will be printed out for us as part of `p.WaitOnSucceeded`

test/e2e/kubernetes/pod/pod.go
```diff
@@ -250,29 +250,25 @@ func RunCommandMultipleTimes(podRunnerCmd podRunnerCmd, image, name, command str
...
-		out, err := cmd.CombinedOutput()
+		terminated, err := p.WaitOnTerminated(podName, sleep, commandTimeout, timeout)
```
jackfrancis: There is some overlap here (anecdotally Terminated container state strongly correlates with Succeeded pod state), but we do this as a follow-up step to introspect the container execution time to validate that its duration is less than the passed in `commandTimeout`)

test/e2e/kubernetes/pod/pod.go
```diff
@@ -354,10 +350,7 @@ func Get(podName, namespace string, retries int) (*Pod, error) {
 	for i := 0; i < retries; i++ {
 		cmd := exec.Command("k", "get", "pods", podName, "-n", namespace, "-o", "json")
 		out, err = cmd.CombinedOutput()
-		if err != nil {
```
jackfrancis: These `kubectl get` errors are expected (thus the retries are built in), we don't need to print every one of them in-line in this function; rather we'll return the error to the consumer, who can decide to print out the equivalent error.

test/e2e/kubernetes/pod/pod.go
```diff
@@ -369,23 +362,6 @@ func Get(podName, namespace string, retries int) (*Pod, error) {
 	return &p, err
 }
 
-// GetTerminated will return a pod with a given name and namespace, including terminated pods
```
jackfrancis: Unused, old func

test/e2e/kubernetes/pod/pod.go
```diff
@@ -629,7 +605,9 @@ func WaitOnSucceeded(name, namespace string, sleep, timeout time.Duration) (bool
 			pod = result.pod
 			mostRecentWaitOnSucceededPodError = result.err
 			if mostRecentWaitOnSucceededPodError == nil {
-				return pod.IsSucceeded(), nil
+				if pod.IsSucceeded() {
```
jackfrancis: This actually fixes a bug :/

We don't want to return a `false, nil` result in such a func. A non-truthy succeeded result should always accompany an error.

PR #1957, opened 2019-09-16
chore: upgrade to metrics-server v0.3.1 for all k8s versions < 1.16

jackfrancis:
**Reason for Change**:

Follow-up from #1109: now that we have working signal of metrics-server v0.3.4 on 1.16, let's aim to backport this onto all prior versions of k8s.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -134,7 +134,7 @@ for HEAPSTER_VERSION in ${HEAPSTER_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-METRICS_SERVER_VERSIONS="0.2.1"
+METRICS_SERVER_VERSIONS="0.3.4 0.2.1"
```
PatrickLang: maybe 0.3.1 too if the other tests are good?

parts/k8s/containeraddons/1.11/kubernetesmasteraddons-metrics-server-deployment.yaml
```diff
@@ -126,7 +127,7 @@ spec:
         imagePullPolicy: IfNotPresent
         command:
         - /metrics-server
-        - --source=kubernetes.summary_api:''
+        - --kubelet-insecure-tls
```
juan-lee: Shouldn't we apply `--authorization-mode=Webhook` instead?

jackfrancis: We are doing that:

https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults-kubelet.go#L21

(Or do you mean there's a comparable arg context for metrics-server? Or are you suggesting that in a kubelet context w/ `--authorization-mode=Webhook` we don't need to be configuring metrics-server w/ `--kubelet-insecure-tls` ?)

pierluigilenoci:
Why "Closed"?

PR #1961, opened 2019-09-17
doc: Clarify defaults for enablePodSecurityPolicy

PatrickLang:
**Reason for Change**:
This parameter is overwritten to `true` for 1.15+, but the docs don't say so. This fixes the doc.


docs/topics/clusterdefinitions.md
```diff
@@ -51,7 +51,7 @@ $ aks-engine get-versions
...
-| enablePodSecurityPolicy         | no       | Enable [kubernetes pod security policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/).This is currently a beta feature. (boolean - default == false)                                                                                                                                                                                                                                       |
+| enablePodSecurityPolicy         | no       | Enable [kubernetes pod security policy](https://kubernetes.io/docs/concepts/policy/pod-security-policy/).This is currently a beta feature. (boolean - default == false for Kubernetes 1.14 and earlier, true for Kubernetes v1.15+)                                                                                                                                                                                                 |
```
CecileRobertMichon: this implies that the _default_  is true for 1.15+ when in fact (at least with how the code is written right now) this value gets set to true statically for all 1.15+ clusters, no matter what the apimodel `enablePodSecurityPolicy` value is. We should either make that clear or change the code to honor custom values.

PR #1976, opened 2019-09-18
chore: add new k8s releases to VHD CI spec

jackfrancis:
**Reason for Change**:

Add the following new Kubernetes versions to the VHD spec:

- 1.16.0
- 1.15.4
- 1.13.11
- 1.14.7

For AKS, AKS Engine, and Azure Stack.

Also deprecate VHD support for 1.11 and 1.12 versions on Azure Stack.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -401,9 +400,11 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
       HYPERKUBE_URL="mcr.microsoft.com/k8s/azurestack/core/hyperkube-amd64:v${KUBERNETES_VERSION}"
     else
       HYPERKUBE_URL="k8s.gcr.io/hyperkube-amd64:v${KUBERNETES_VERSION}"
-      CONTAINER_IMAGE="k8s.gcr.io/cloud-controller-manager-amd64:v${KUBERNETES_VERSION}"
-      pullContainerImage "docker" ${CONTAINER_IMAGE}
-      echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+      if (( $(echo ${KUBERNETES_VERSION} | cut -d"." -f2) < 16 )); then
```
CecileRobertMichon: the ident seems off here

jackfrancis: do you mean because 2 chars instead of 4?

CecileRobertMichon: yeah, just that the `fi` and the `if` aren't aligned

jackfrancis:
@CecileRobertMichon there were a mix of tabs and spaces inside that block, addressed here:

https://github.com/Azure/aks-engine/pull/1992

PR #2001, opened 2019-09-20
feat: enable azurestack telemetry for scale and upgrade scenarios

deaborch:
**Reason for Change**:

Configures a different Telemetry Tracking PID during scale, upgrade and deploy scenarios

**Issue Fixed**:

**Requirements**:

- [] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -160,7 +160,9 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		err = sc.containerService.Properties.SetAzureStackCloudSpec(false, true)
+		isUpgrade := false
+		isScale := true
+		err = sc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale)
 		if err != nil {
 			return errors.Wrap(err, "error parsing the api model")
 		}
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -161,7 +161,7 @@ func (sc *scaleCmd) load() error {
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
 
-		if err := sc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{IsUpgrade: false, IsScale: true}); err != nil {
+		if err = sc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{IsUpgrade: false, IsScale: true}); err != nil {
 			return errors.Wrap(err, "error parsing the api model")
 		}
 	}
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -160,7 +160,7 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		err = sc.containerService.Properties.SetAzureStackCloudSpec()
+		err = sc.containerService.Properties.SetAzureStackCloudSpec(false, true)
 		if err != nil {
 			return errors.Wrap(err, "error parsing the api model")
 		}
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -160,11 +160,8 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		
-		if err := sc.containerService.Properties.SetAzureStackCloudSpec(AzureStackCloudSpecDefaultsParams{
-			IsUpgrade: false,
-			IsScale:   true,
-		}){
+
+		if err := sc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{IsUpgrade: false, IsScale: true}); err != nil {
 			return errors.Wrap(err, "error parsing the api model")
 		}
 	}
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -160,10 +160,11 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		isUpgrade := false
-		isScale := true
-		err = sc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale)
-		if err != nil {
+		
+		if err := sc.containerService.Properties.SetAzureStackCloudSpec(AzureStackCloudSpecDefaultsParams{
+			IsUpgrade: false,
+			IsScale:   true,
+		}){
 			return errors.Wrap(err, "error parsing the api model")
 		}
 	}
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -158,7 +158,7 @@ func (uc *upgradeCmd) loadCluster() error {
 	if uc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(uc.containerService)
 
-		if err := uc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{
+		if err = uc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{
 			IsUpgrade: true,
 			IsScale:   false,
 		}); err != nil {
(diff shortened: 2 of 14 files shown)


cmd/scale.go
```diff
@@ -160,7 +160,7 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		err = sc.containerService.Properties.SetAzureStackCloudSpec()
+		err = sc.containerService.Properties.SetAzureStackCloudSpec(false, true)
```
jackfrancis: nit: rather than pass in boolean literals (e.g., `false, true`) create a named variable so it's more clear what context we're invoking the function in. E.g.:

```
isUpgrade := false
isScale := true
err = sc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale)
```

(same comment for all invocations of `SetAzureStackCloudSpec` below using the new params interface)

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -130,6 +143,8 @@ func (p *Properties) SetAzureStackCloudSpec() error {
 			//KubernetesSpecConfig
 			asccKubernetesSpecConfig := ascc.KubernetesSpecConfig
 			azsKubernetesSpecConfig := azureStackCloudSpec.KubernetesSpecConfig
+
+			azureStackCloudSpec.KubernetesSpecConfig.AzureTelemetryPID = helpers.EnsureString(asccKubernetesSpecConfig.AzureTelemetryPID, DefaultAzurestackDeployTelemetryPID)
```
jackfrancis: The reason we don't bother to set a scale- or upgrade-specific ID here is because the `AzureEnvironmentSpecConfig` will never be non-nil in a scale or upgrade scenario?

cmd/scale.go
```diff
@@ -160,7 +160,9 @@ func (sc *scaleCmd) load() error {
 
 	if sc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(sc.containerService)
-		err = sc.containerService.Properties.SetAzureStackCloudSpec()
+		isUpgrade := false
+		isScale := true
+		err = sc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale)
```
devigned: This is stylistic, but it's nice to shape these statements like:
```go
if err := sc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale); err != nil {
  // do stuff for error condition
}
```

cmd/upgrade.go
```diff
@@ -157,7 +157,9 @@ func (uc *upgradeCmd) loadCluster() error {
 
 	if uc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(uc.containerService)
-		err = uc.containerService.Properties.SetAzureStackCloudSpec()
+		isUpgrade := true
+		isScale := false
+		err = uc.containerService.Properties.SetAzureStackCloudSpec(isUpgrade, isScale)
```
devigned: Same guidance as scale.go

pkg/api/azenvtypes.go
```diff
@@ -64,6 +65,18 @@ type AzureOSImageConfig struct {
 	ImageVersion   string `json:"imageVersion,omitempty"`
 }
 
+// AzureTelemetryPID represents the current telemetry ID
+type AzureTelemetryPID string
```
devigned: If it represents a telemetry ID, what is a PID? I mean, I see `pid` as the beginning of the id below, but is it significant? If PID is significant, please describe it in the comment. If it is not, then please consider renaming to ID.

pkg/api/azenvtypes.go
```diff
@@ -64,6 +65,18 @@ type AzureOSImageConfig struct {
...
+
+const (
+	// DefaultAzurestackDeployTelemetryPID tracking ID for Deployment
+	DefaultAzurestackDeployTelemetryPID = "pid-1bda96ec-adf4-4eea-bb9a-8462de5475c0"
```
rjaini: Can we have AzureStack rather Azurestack in all the variable name.

pkg/api/defaults-custom-cloud-profile.go
deaborch: @jackfrancis that is exactly right. scale and upgrade scenarios are covered above.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/pkg/errors"
 )
 
-func (cs *ContainerService) setCustomCloudProfileDefaults() error {
+func (cs *ContainerService) setCustomCloudProfileDefaults(isUpgrade, isScale bool) error {
```
devigned: Multiple parameters of the same time is a code smell. Please think about adding a struct as input so the fields will be easily differentiated.

Also, I feel like this we clean up else where recently.

https://dave.cheney.net/2019/09/24/be-wary-of-functions-which-take-several-parameters-of-the-same-type

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -96,7 +96,7 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
 }
 
 // SetAzureStackCloudSpec sets the cloud spec for Azure Stack .
-func (p *Properties) SetAzureStackCloudSpec() error {
+func (p *Properties) SetAzureStackCloudSpec(isUpgrade, isScale bool) error {
```
devigned: Same guidance as above.

pkg/engine/deployment.go
```diff
@@ -8,12 +8,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
-const (
-	azurestackGenerateGUID = "pid-1bda96ec-adf4-4eea-bb9a-8462de5475c0"
-)
-
-//TODO create separate function to allow users to pass in deployment name for azure
-func createAzurestackTelemetry() DeploymentARM {
+func createAzurestackTelemetry(azureTelemetryPID string) DeploymentARM {
```
rjaini: Same here. Please change createAzurestackTelemetry to createAzureStackTelemetry

pkg/api/azenvtypes.go
deaborch: Done

pkg/engine/deployment.go
deaborch: done

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -114,8 +114,21 @@ func (p *Properties) SetAzureStackCloudSpec() error {
 		if p.CustomCloudProfile.Environment == nil || p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix == "" {
 			return errors.New("Failed to set Cloud Spec for Azure Stack due to invalid environment")
 		}
+
 		azureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
 		azureStackCloudSpec.CloudName = AzureStackCloud
+
+		//Sets default values for telemetry PID where none is set
+		if p.CustomCloudProfile.AzureEnvironmentSpecConfig == nil {
+			azureStackCloudSpec.KubernetesSpecConfig.AzureTelemetryPID = DefaultAzurestackDeployTelemetryPID
```
devigned: Please consider a case statement.

```go
	switch {
	  case isScale:
	    fmt.Println("scale")
	  case isUpgrade:
	    fmt.Println("upgrade")
	  default:
	    // deploy
	    fmt.Println("deploy")
	}
```

cmd/upgrade.go
```diff
@@ -157,8 +157,13 @@ func (uc *upgradeCmd) loadCluster() error {
 
 	if uc.containerService.Properties.IsAzureStackCloud() {
 		writeCustomCloudProfile(uc.containerService)
-		err = uc.containerService.Properties.SetAzureStackCloudSpec()
-		if err != nil {
+		isUpgrade := true
```
devigned: Looks like these vars are not needed any longer, or they should be used for the input to the construction of the struct below.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -95,8 +101,14 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
 	return nil
 }
 
+// AzureStackCloudSpecDefaultsParams is the parameters when we set the azure stack cloud spec defaults for ContainerService.
+type AzureStackCloudSpecDefaultsParams struct {
```
devigned: Probably should be `type AzureStackCloudSpecParams struct {`.

devigned:
lgtm

PR #2002, opened 2019-09-20
fix: don’t statically set EnablePodSecurityPolicy=true for AKS

jackfrancis:
**Reason for Change**:

AKS uses the `KubernetesConfig.EnablePodSecurityPolicy` property according to its own k8s version default requirements, so we should not statically set EnablePodSecurityPolicy to true in AKS scenarios.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -329,7 +329,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			a.OrchestratorProfile.KubernetesConfig.ExcludeMasterFromStandardLB = to.BoolPtr(DefaultExcludeMasterFromStandardLB)
 		}
 
-		if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.15.0-beta.1") {
+		if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.15.0") && !cs.Properties.IsHostedMasterProfile() {
 			a.OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy = to.BoolPtr(true)
 		}
 
--- a/pkg/api/defaults_test.go
+++ b/pkg/api/defaults_test.go
@@ -2927,14 +2927,14 @@ func getKubernetesConfigWithFeatureGates(featureGates string) *KubernetesConfig 
 	}
 }
 
-func TestDefaultEnablePodSecurityPolicy(t *testing.T) {
+func TestEnablePodSecurityPolicy(t *testing.T) {
 	cases := []struct {
 		name     string
 		cs       ContainerService
 		expected bool
 	}{
 		{
-			name: "default",
+			name: "default, < 1.15",
 			cs: ContainerService{
 				Properties: &Properties{
 					OrchestratorProfile: &OrchestratorProfile{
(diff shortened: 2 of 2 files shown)


jackfrancis:
@yangl900 @wenwu449 FYI

pkg/api/defaults_test.go
```diff
@@ -2947,44 +2947,104 @@ func TestDefaultEnablePodSecurityPolicy(t *testing.T) {
 			expected: false,
 		},
 		{
-			name: "default",
+			name: "default, >= 1.15",
 			cs: ContainerService{
 				Properties: &Properties{
 					OrchestratorProfile: &OrchestratorProfile{
 						OrchestratorType:    Kubernetes,
-						OrchestratorVersion: "1.15.0-alpha.1",
```
mboersma: Thanks for cleaning these -alpha and -beta references up. 

There are similar code droppings related to Kubernetes 1.16.0-(alpha,beta,rc). Should we make a PR to have those point to the release as well?

pkg/api/defaults.go
```diff
@@ -329,7 +329,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			a.OrchestratorProfile.KubernetesConfig.ExcludeMasterFromStandardLB = to.BoolPtr(DefaultExcludeMasterFromStandardLB)
 		}
 
-		if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.15.0-beta.1") {
+		if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.15.0") && !cs.Properties.IsHostedMasterProfile() {
```
yangl900: why we have this condition on IsHostedMasterProfile()?

the function is called setOrchestrator**Defaults**(), shouldn't we only set **default** value if it was not set?

e.g. EnablePodSecurityPolicy is a pointer, should we only set it to true if `OrchestratorProfile.KubernetesConfig.EnablePodSecurityPolicy == nil`?

yangl900: it seems we do that for other places, e.g. L325 we only set the SKU if it was not set. and L310 we only set `UseInstanceMetadata` if it is nil.

Why this is special?

jackfrancis: This is special in AKS Engine because we don't have a way to support privileged pods without the PodSecurityPolicy addon that we deliver. And we have pre-existing clusters that were built that expect that functionality.

We could change the logic for AKS Engine to only convert explicit false to true in an upgrade scenario (In fact in an upgrade scenario for an older cluster the `false` configuration is not necessarily user-influenced, as the prior default flow set this to `false` because prior to 1.15 a cluster didn't need the PodSecurityPolicy to support privileged pods).

@aramase @khenidak does the above description match your understanding for why this was merged?

https://github.com/Azure/aks-engine/pull/1454

yangl900: Right. In that case, shouldn't we validate it when generating template (or even earlier)? If input == 1.15 and EnablePodSecurityPolicy == false, return error.

the setdefaults() keep its simple job by setting appropriate default value **if not set**.

jackfrancis: Something like that would be preferable, yes. It would require some code refactor in rationalize the template generation code flows to be able to do this kind of "AKS Engine upgrade configuration rationalization (e.g., change false to true for 1.14 to 1.15 upgrade due to PSP requirement with >= 1.15).

In the meantime AKS Engine needs the current code as it fulfills those req's. Do we want to guard this requirement so that it doesn't mutate AKS scenarios during defaults enforcement? (That's what this PR does) Or close this PR and address the AKS scenario when the code is refactored?

yangl900: Could you elaborate more? what type of refactoring is needed? We already have a `validate.go`, which seems a good fit for this type of validation.

In upgrade case, user provide apiModel too. It seems perfectly fine to ask user explicitly set EnablePodSecurityPolicy=true when they upgrade to 1.15. It's also an acknowledge from user side that they are aware of the change and OK to proceed.

yangl900: @ritazh thoughts?

PR #2010, opened 2019-09-23
docs: Adding doc for new k8s version ingestion

marosset:
**Reason for Change**:
Adding a doc describing what actions need to happen when a new version of Kubernetes gets released

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/community/README.md
+++ b/docs/community/README.md
@@ -6,6 +6,7 @@ Here you'll find documentation geared towards learning about the development pro
 - [Developer Guide](developer-guide.md)
 - [Planning Process](planning-process.md)
 - [Release Checklist](release-checklist.md)
+- [New k8s releae](new-k8s-release.md)
 
 AKS Engine is a community effort. As it keeps growing, we always need more people to help others. As soon as you learn AKS Engine, you can contribute in many ways:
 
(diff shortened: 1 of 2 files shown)


docs/community/README.md
```diff
@@ -6,6 +6,7 @@ Here you'll find documentation geared towards learning about the development pro
 - [Developer Guide](developer-guide.md)
 - [Planning Process](planning-process.md)
 - [Release Checklist](release-checklist.md)
+- [New k8s releae](new-k8s-release.md)
```
CecileRobertMichon:
```suggestion
- [New k8s release](new-k8s-release.md)
```

mboersma:
Should we also add the Windows support file creation here? Hopefully this clunky process goes away, but for now you have to start with some environment vars set, and:

```shell
 make build-windows-k8s K8S_VERSION=1.17.1-alpha.1 PATCH_VERSION=1
```

Then create an IcM, get JIT, and move that artifact to the prod blob store.

marosset:
If that is something only maintainers/MS employees can perform it might be better to put in a devops wiki (similar to how the VHD build docs).
Thoughts?

PR #2011, opened 2019-09-23
fix: lowercase vm name before calling drain node

### LINKED ISSUE #1982: Should always change the node name to lower case before draining it
**Describe the bug**
For VMSS cluster, the virtual machine in Azure (instance name), and the machine name in OS and node name in Kubernetes is different, for example, for VMSS instance with instance id 11
The Instance name in Azure: aks-nodepool1-12345678-vmss_11
The machine name in OS:      aks-nodepool1-12345678-vmss00000B
The node name in Kubernetes: aks-nodepool1-12345678-vmss00000b

The upgrader gets the virtual machines to be upgraded using machine name in OS. So we should change the name to lower case before draining it. Otherwise the draining will fail with 'node not found' error.

https://github.com/Azure/aks-engine/blob/master/pkg/operations/kubernetesupgrade/upgrader.go#L692

**Steps To Reproduce**

**Expected behavior**

**AKS Engine version**

**Kubernetes version**

**Additional context**

mboersma:
**Reason for Change**:
Ensures that `SafelyDrainNodeWithClient` uses the lowercase version of a VM's name to avoid disagreement between node name and machine name.

**Issue Fixed**:

Fixes #1982 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/cordondrainvm.go
+++ b/pkg/operations/cordondrainvm.go
@@ -45,6 +45,7 @@ func SafelyDrainNode(az armhelpers.AKSEngineClient, logger *log.Entry, apiserver
 
 // SafelyDrainNodeWithClient safely drains a node so that it can be deleted from the cluster
 func SafelyDrainNodeWithClient(client armhelpers.KubernetesClient, logger *log.Entry, nodeName string, timeout time.Duration) error {
+	nodeName = strings.ToLower(nodeName)
 	//Mark the node unschedulable
 	var node *v1.Node
 	var err error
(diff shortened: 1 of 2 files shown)


pkg/operations/cordondrainvm.go
```diff
@@ -45,6 +45,7 @@ func SafelyDrainNode(az armhelpers.AKSEngineClient, logger *log.Entry, apiserver
 
 // SafelyDrainNodeWithClient safely drains a node so that it can be deleted from the cluster
 func SafelyDrainNodeWithClient(client armhelpers.KubernetesClient, logger *log.Entry, nodeName string, timeout time.Duration) error {
+	nodeName = strings.ToLower(nodeName)
```
jackfrancis: I asume this addresses other invocations of `SafelyDrainNodeWithClient` that weren't converting to lowercase in before calling?

mboersma: Yes. There was only one instance where we weren't coercing it to lowercase, but this should be slightly safer than fixing it before calling.

I looked at a comprehensive fix for VMSS name lowercasing as @CecileRobertMichon suggested, but there are several functions in different structs that should be treated this way and it wasn't a clear win. But I'm happy to revisit if we think a more general solution is warranted.

PR #2015, opened 2019-09-24
test: more resilient “wait for successful pod readiness checks”

jackfrancis:
**Reason for Change**:

This PR tweaks the way we validate that a pod or common set of pods is in a persistent, reliable "Ready" state. Instead of tracking and incrementing "successful" checks, and comparing to "unsuccessful" checks, we instead try to detect "flaps", which are in fact normal Kubernetes pod behaviors, and ensure that the transitions between Ready and non-Ready (i.e., a flap) are within a tolerable spectrum.

Specifically:

Once we get a positive Ready signal, we continue attempting to accumulate the desired success count until we detect a non-Ready signal. A ready --> non-ready transition counts as a single "flap". We then wait for the re-occurrence of a successive Ready signal, which then counts towards the targeted number of successes. Rinse, repeat.

We will return successful if the number of desired Ready counts is reached, and is at least 2 more than the number of flaps. In practice this means that a simple check of "just give me signal that one Ready state has occured" works fine (the "flap" detection will never be engaged in this scenario); and that 4 is probably the minimum number of successes to target for scenarios where you want more than 1 Ready count. 2 is super aggressive as basically any one flap will fail the test, and 3 is arguably not tolerant enough for "normal" reconciliation conditions.

As such all of the relevant "wait for more than 1 Ready signal" usages are set to 4.

Also renamed the func to `WaitOnSucesses` to more easily disambiguate from the single-validation-call `WaitOnReady`.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/pod/pod.go
+++ b/test/e2e/kubernetes/pod/pod.go
@@ -537,15 +537,16 @@ func GetPodAsync(name, namespace string, timeout time.Duration) GetPodResult {
 	}
 }
 
-// WaitOnReady returns true if all pods matching a prefix substring are in a succeeded state within a period of time
+// WaitOnSuccesses returns true if all pods matching a prefix substring are in a succeeded state within a period of time
 // successesNeeded is used to make sure we return the correct value even if the pod is in a CrashLoop
-func WaitOnReady(podPrefix, namespace string, successesNeeded int, sleep, timeout time.Duration) (bool, error) {
+func WaitOnSuccesses(podPrefix, namespace string, successesNeeded int, sleep, timeout time.Duration) (bool, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	defer cancel()
 	ch := make(chan AreAllPodsRunningResult)
-	var mostRecentWaitOnReadyErr error
+	var mostRecentWaitOnSuccessesErr error
 	successCount := 0
-	failureCount := 0
+	flapCount := 0
+	var lastResult bool
 	go func() {
 		for {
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/pod/pod.go
```diff
@@ -559,24 +560,28 @@ func WaitOnReady(podPrefix, namespace string, successesNeeded int, sleep, timeou
...
-					return true, nil
-				}
-			} else {
-				if successCount > 1 {
-					failureCount++
-					if failureCount >= successesNeeded {
-						PrintPodsLogs(podPrefix, namespace)
-						return false, errors.Errorf("Pods from deployment (%s) in namespace (%s) have been checked out as all Ready %d times, but NotReady %d times. This behavior may mean it is in a crashloop", podPrefix, namespace, successCount, failureCount)
+			mostRecentWaitOnSuccessesErr = result.err
+			if mostRecentWaitOnSuccessesErr == nil {
```
devigned: Total nit, but might be good to invert the if statement to reduce nesting. For example:

```go
if mostRecentWaitOnSuccessesErr != nil {
  continue
}
// do the rest of the things without the if nesting
```

jackfrancis: This is interesting. I propose we bikeshed this sort of abstractly (at later time), and then apply the team-approved pattern generally. Because this pattern is the one we currently use to handle channel responses in these goroutine validations.

PR #2027, opened 2019-09-25
[WIP] Enable IPv6 only kubernetes clusters

### LINKED ISSUE #1729: IPv6 only kubernetes cluster
**Describe the request**

Deploy IPv6 only clusters in Azure

Kubernetes has support for IPv6 only clusters since 1.9, Azure is implementing fully IPv6 support and AKS has already implemented dual-stack support.

The dual-stack PR has already implemented all the heavy lifting to enable IPv6 on the Azure networks and VMs #1424 and added a feature to deploy dual stack clusters.

We should be able to use a new feature in AKS to deploy an ipv6 only cluster (doesn't matter that the VMs are dual stack, the cluster will be IPv6 only), 

I'm not familiar with the code, but I can try to implement it if I have some guidance.

**Additional context**

This is a requirement to graduate Kubernetes IPv6 support to beta https://github.com/kubernetes/enhancements/pull/1139

aojea:
**Reason for Change**:

AKS just added support for dual-stack clusters, we should be able to deploy IPv6 only clusters too

**Issue Fixed**:

Fixes #1729 

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

Kubernetes IPv6 only clusters can work on dual stack hosts, so this
PR is based on the dual stack feature, just using IPv6 addresses to
configure the kubernetes components.

Kubernetes doesn't use flags to enable IPv6, it simply works in
IPv6 if the components use IPv6


aojea:
/close
duplicate of https://github.com/Azure/aks-engine/pull/2781

PR #2037, opened 2019-09-26
fix: verify LCOW in makedev.ps1

### LINKED ISSUE #2012: makedev.ps1 does not prevent from using windows containers
Currently, in the developer guide for aks-engine, https://github.com/Azure/aks-engine/blob/master/docs/community/developer-guide.md,  it has instructions to build the engine in a docker container. It does specify what kind of container it can be ran in though. The development process does not support a Windows container though so:

1. The documentation needs to be updated to reflect that it is only linux containers
2. Add a check in makedev.ps1 to prevent running the container if docker for desktop is running with Windows containers

This additional information is important to add as the container does run even if you are using Windows containers. It appears to work correctly but when attempting to 'make build' it will produce memory errors and the files are mounted to the wrong location.

Repro steps:

1. Follow https://github.com/Azure/aks-engine/blob/master/docs/community/developer-guide.md
2. When running makedev.ps1, change Docker for desktop to run Windows containers instead of linux containers

- notice that the directly it mounts does not contain any files

marosset:
**Reason for Change**:

Fixes #2012

**Issue Fixed**:

https://github.com/Azure/aks-engine/issues/2012

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/community/developer-guide.md
+++ b/docs/community/developer-guide.md
@@ -32,7 +32,7 @@ The easiest way to start hacking on AKS Engine is to use a Docker-based environm
 $ make dev
 ```
 
-Or on Windows:
+Or on Windows (ensure Docker is configured for Linux containers on Windows):
 
 ```console
 powershell ./makedev.ps1
--- a/makedev.ps1
+++ b/makedev.ps1
@@ -2,4 +2,12 @@ $REPO_PATH = "github.com/Azure/aks-engine"
 $DEV_ENV_IMAGE = "quay.io/deis/go-dev:v1.23.2"
 $DEV_ENV_WORK_DIR = "/go/src/$REPO_PATH"
 
+# Ensure docker is configured for linux containers
+$json = (docker version --format "{{json .}}" | ConvertFrom-Json)
+if ($json.Server.Os -ne "linux")
+{
+    Write-Error "Please switch Docker use to Linux containers on Windows"
+    exit 1
+}
+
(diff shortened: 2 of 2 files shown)


marosset:
@alexanderboswell as FYI (i can't seem to request review from you)

makedev.ps1
```diff
@@ -2,4 +2,12 @@ $REPO_PATH = "github.com/Azure/aks-engine"
 $DEV_ENV_IMAGE = "quay.io/deis/go-dev:v1.23.2"
 $DEV_ENV_WORK_DIR = "/go/src/$REPO_PATH"
 
+# Ensure docker is configured for linux containers
+$json = (docker version --format "{{json .}}" | ConvertFrom-Json)
```
PatrickLang: or `docker version --format "{{json .Server.Os }}"` either way works

docs/community/developer-guide.md
```diff
@@ -32,7 +32,7 @@ The easiest way to start hacking on AKS Engine is to use a Docker-based environm
 $ make dev
 ```
 
-Or on Windows:
+Or on Windows (ensure Docker is configured for Linux containers on Windows):
```
PatrickLang: or run `& 'C:\Program Files\Docker\Docker\DockerCli.exe' -SwitchLinuxEngine`

PatrickLang: but that shouldn't be done automatically - if someone has windows containers running they'll lose the ability to manage them

makedev.ps1
marosset: Thank, i'll keep that in mind for the future.

PR #2046, opened 2019-09-27
docs: propose dedicated AKS release process

jackfrancis:
**Reason for Change**:

This PR proposes a new, dedicated AKS release process.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/design/aks-release-process.md
+++ b/docs/design/aks-release-process.md
@@ -1,0 +1,36 @@
+# AKS Release Process Proposal
+
+## Problem Statement
+
+Today, AKS regularly updates its dependency upon AKS Engine by tracking the public release velocity of AKS Engine, in order to rapidly get the changes it needs that reside in the AKS Engine surface area. Because AKS Engine introduces non-AKS-required changes at a high velocity, in practice this approach brings more change, more frequently, than AKS actually desires, and thus increases the risk to AKS that some part of the changeset not strictly required by AKS may introduce undesired changes or regressions. So long as AKS strives to match AKS Engine release velocity, it will be required to regularly accept and validate that *all* AKS Engine changes work in the context of an AKS downstream library dependency.
+
+## Proposal
+
+Enable AKS to regularly update its downstream AKS Engine dependencies according to its own specific requirements.
+
+Rather than requiring AKS to pull down *all* changes as a result of matching AKS Engine release velocity, we are proposing that:
+
+1. AKS maintain its own release channel for rapid (aka weekly) iterations against its AKS Engine surface area, and
+2. AKS regularly (aka monthly) re-align with the "mainline" AKS Engine release channel.
+
+#1 above addresses the Problem Statement for most release iterations. #2 ensures that the AKS <-- AKS Engine relationship doesn't deviate too far from mainline, and that AKS isn't effectively maintaining a fork of AKS Engine.
+
+## How this will work
+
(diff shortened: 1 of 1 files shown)


docs/design/aks-release-process.md
```diff
@@ -0,0 +1,36 @@
...
+## How this will work
+
+Every month, we will cut a new AKS release @ n-1 compared to the "mainline", public AKS Engine release channel. For example:
+
+- For an "October" release: we derive a new AKS release branch from AKS Engine n-1:
+  - e.g., AKS Engine is at v0.42._n_
+  - e.g., we cut a new AKS release branch from AKS Engine the latest v0.41 release
+- Any additional commits in AKS Engine ahead of this new AKS release branch that are desired by AKS immediately will be cherry-picked by AKS into the new AKS release branch.
+- For the duration of the "October" release cycle, AKS continues to cherry-pick changes from `aks-engine@master` into its "October" release branch.
+- When "November" arrives, AKS cuts a new release branch from AKS Engine according to the above n-1 calculation.
```
xuto2: is this a typo when we say "AKS cuts new release branch from AKS Engine" here?If I remember correctly in our last discussion, AKS-Engine (but not AKS) will cut the new release in AKS Engine repo?

jackfrancis: In terms of requirements, AKS should drive those. So:

1. Create new branch from commit @ AKS-Engine n-1 release
2. Cherry-pick any desired changes after the above commit
3. Publish a release

I think AKS and AKS Engine will do the above in tandem. Especially #2 AKS will drive, and especially #3 AKS Engine will do.

Does that make sense? I think the details will probably look something like that, do we want to clarify that in the doc?

xuto2: yes makes sense. since we're here, it can help others to understand the process better, if we can just put what you added into the doc.

CecileRobertMichon: Do we need to do 3? From the doc it seems like we only cut a release branch without publishing an actual release which makes sense to me since these AKS specific changes don't always apply to aks-engine users who might be tracking releases to use the latest patch.

docs/design/aks-release-process.md
```diff
@@ -0,0 +1,36 @@
...
+
+At a glance, that's the process.
+
+## How this will work successfully
+
+- Additional testing/validation will be required for the monthly "catch-up" AKS releases. This additional validation overhead will be scheduled so as not to slow down feature and bugfix velocity during these intervals.
+  - I.e., these releases will include _all_ AKS Engine changes at the AKS Engine minor release that AKS derives its new release from, not just AKS-specific changes
+- All AKS-targeted changes to AKS Engine will be introduced via the normal PR-against-master workflow. In other words, the AKS release channel will not be a first class commit destination for novel additions/deletions to code; all changes will be cherry-picked from master.
```
CecileRobertMichon: +1 on this

xuto2:
> Can you add a note about always publishing a VHD from that release branch to ensure that component versions pre-installed match the ones in code?
> Also, let's add a new release process check to run the no-outbound test on every release/branch/tag before vendoring it to AKS.

+1 on this
VHD is also a part of the source code so it should come from the same branch.

With that, sometimes we may need to publish a VHD specifically for AKS(with or without "aks" in the vhd release name). I assume this is compatible with current AKS engine release policy?

jackfrancis:
Blocked by #2245

CecileRobertMichon:
keep open

PR #2048, opened 2019-09-27
feat: use aks windows vhds by default

marosset:
**Reason for Change**:
Use aks VHDs as default images for Windows nodes

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/defaults.go
```diff
@@ -701,11 +701,11 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName
 func (p *Properties) setWindowsProfileDefaults(isUpgrade, isScale bool) {
 	windowsProfile := p.WindowsProfile
 	if !isUpgrade && !isScale {
-		if windowsProfile.WindowsPublisher == "" {
-			windowsProfile.WindowsPublisher = DefaultWindowsPublisher
-		}
 
 		if p.IsAzureStackCloud() {
+			if windowsProfile.WindowsPublisher == "" {
+				windowsProfile.WindowsPublisher = DefaultWindowsPublisher
```
PatrickLang: @rjaini - do you want to change the AzureStack default in a separate PR or this one?

rjaini: yes. As the VHD issue is not resolve yet.

pkg/api/const.go
```diff
@@ -288,6 +288,14 @@ const (
 	DefaultImageVersion = "17763.737.1909062324"
 )
 
+// Aks windows VHD parts
+const (
```
jackfrancis: Is there any practical way to converge where we maintain Windows VHD default references and where we maintain the Linux defaults? See here for Linux VHD config:

https://github.com/Azure/aks-engine/blob/master/pkg/api/azenvtypes.go#L104

marosset: Hmm....  we could add now image config types there and then just grab the values we need installed of looking them up in a 'disto map'.
That wouldn't change the behavior in code but would keep all the image related changes to a single file.
Let me give that a try.

jackfrancis: Yeah I assume things are functionally ready here, just observing that we shouldn't be triaging Windows from Linux coding interfaces if we can help it. (A good justification for doing so would be that the existing Linux implementation has lots of tech debt and the Windows surface area is an excuse to improve maintenance. :))

marosset: Great sell!

marosset: @jackfrancis can you take a look at the lastest commit? I feel it is too big of a change to use how windows images are refenced internal to aks-engine but i did move the defaults align how defaults are specified into azenvtypes.go to align with how linux images are specified.

pkg/api/azenvtypes.go
```diff
@@ -148,6 +148,30 @@ var (
...
+	// AKSWindows2019OSImageConfig is the AKS image based on Windows Server 2019
+	AKSWindowsServer2019OSImageConfig = AzureOSImageConfig{
+		ImageOffer: "aks-windows",
+		ImageSku: "2019-datacenter-core-smalldisk",
+		ImagePublisher: "microsoft-aks",
+		ImageVersion: "17763.737.190923",
+	}
+
+	// AzureStackWindowsServer2019OSImageConfig is the 'vanilla' Windows Server 2019 image to be used to with Azure Stack deployments
+	AzureStackWindowsServer2019OSImageConfig = AzureOSImageConfig{
```
marosset: @rjaini as FYI - this change moves where the default imaging values for AzureStack gets defined (from consts.go to azenvtypes.go)

jackfrancis: Thanks! I think this fits in better with the existing implementation. (Hope you agree and aren't under the delusion that folks just tell you what to do :/)

marosset: I'm unhappy with how different image specification works for Linux vs Windows. 
I looked into unifying it but it looked like it was going to be very painful to do so and support existing API models. 
I would like to see converge for V2 tho.

PR #2050, opened 2019-09-27
feat: add linux support for kubenet + containerd

### LINKED ISSUE #1723: Missing network config with containerd + kubenet
> Still gathering details here, will remove this note once it's reproduced on master branch

**Describe the bug**

If you try to deploy a Linux-only cluster using containerd + kubenet, the nodes never reach the `Ready` state. `core-dns` never comes up.

**Steps To Reproduce**

Apimodel: https://github.com/PatrickLang/aks-engine/blob/containerd-kubenet/examples/kubernetes-containerd-kubenet.json

```
$ kubectl get pod -o wide --all-namespaces
NAMESPACE     NAME                                            READY   STATUS              RESTARTS   AGE     IP             NODE                        NOMINATED NODE   READINESS GATES
kube-system   azure-ip-masq-agent-6xszl                       1/1     Running             0          9m59s   10.240.0.5     k8s-linuxpool1-19153786-1   <none>           <none>
kube-system   azure-ip-masq-agent-n9zp4                       1/1     Running             0          9m59s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   azure-ip-masq-agent-xps9t                       1/1     Running             0          9m59s   10.240.0.6     k8s-linuxpool1-19153786-0   <none>           <none>
kube-system   coredns-7d6568df55-9t29j                        0/1     ContainerCreating   0          9m59s   <none>         k8s-linuxpool1-19153786-0   <none>           <none>
kube-system   kube-addon-manager-k8s-master-19153786-0        1/1     Running             0          9m58s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   kube-apiserver-k8s-master-19153786-0            1/1     Running             0          9m16s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   kube-controller-manager-k8s-master-19153786-0   1/1     Running             0          9m24s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   kube-proxy-4cc7p                                1/1     Running             0          9m58s   10.240.0.6     k8s-linuxpool1-19153786-0   <none>           <none>
kube-system   kube-proxy-cdwkq                                1/1     Running             0          9m58s   10.240.0.5     k8s-linuxpool1-19153786-1   <none>           <none>
kube-system   kube-proxy-fcfqw                                1/1     Running             0          9m58s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   kube-scheduler-k8s-master-19153786-0            1/1     Running             0          9m18s   10.240.255.5   k8s-master-19153786-0       <none>           <none>
kube-system   kubernetes-dashboard-54b795b69b-v6fm4           0/1     Pending             0          9m58s   <none>         <none>                      <none>           <none>
kube-system   metrics-server-6bf85bb69b-dzd6r                 0/1     Pending             0          9m59s   <none>         <none>                      <none>           <none>
kube-system   tiller-deploy-5d9ffb885b-m4dgc                  0/1     Pending             0          9m58s   <none>         <none>                      <none>           <none>

$ kubectl describe pod -n kube-system coredns-7d6568df55-9t29j
Name:               coredns-7d6568df55-9t29j
Namespace:          kube-system
Priority:           2000000000
PriorityClassName:  system-cluster-critical
Node:               k8s-linuxpool1-19153786-0/10.240.0.6
Start Time:         Fri, 02 Aug 2019 21:47:10 +0000
Labels:             k8s-app=kube-dns
                    pod-template-hash=7d6568df55
Annotations:        seccomp.security.alpha.kubernetes.io/pod: docker/default
Status:             Pending
IP:
Controlled By:      ReplicaSet/coredns-7d6568df55
Containers:
  coredns:
    Container ID:
    Image:         k8s.gcr.io/coredns:1.5.0
    Image ID:
    Ports:         53/UDP, 53/TCP, 9153/TCP
    Host Ports:    0/UDP, 0/TCP, 0/TCP
    Args:
      -conf
      /etc/coredns/Corefile
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Limits:
      memory:  170Mi
    Requests:
      cpu:        100m
      memory:     70Mi
    Liveness:     http-get http://:8080/health delay=60s timeout=5s period=10s #success=1 #failure=5
    Readiness:    http-get http://:8181/ready delay=0s timeout=1s period=10s #success=1 #failure=3
    Environment:  <none>
    Mounts:
      /etc/coredns from config-volume (ro)
      /var/run/secrets/kubernetes.io/serviceaccount from coredns-token-2q8t8 (ro)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  config-volume:
    Type:      ConfigMap (a volume populated by a ConfigMap)
    Name:      coredns
    Optional:  false
  coredns-token-2q8t8:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  coredns-token-2q8t8
    Optional:    false
QoS Class:       Burstable
Node-Selectors:  beta.kubernetes.io/os=linux
Tolerations:     :NoExecute
                 :NoSchedule
                 CriticalAddonsOnly
                 node-role.kubernetes.io/master:NoSchedule
Events:
  Type     Reason           Age                    From                                Message
  ----     ------           ----                   ----                                -------
  Normal   Scheduled        12m                    default-scheduler                   Successfully assigned kube-system/coredns-7d6568df55-9t29j to k8s-linuxpool1-19153786-0
  Warning  FailedMount      12m (x3 over 12m)      kubelet, k8s-linuxpool1-19153786-0  MountVolume.SetUp failed for volume "config-volume" : configmap "coredns" not found
  Warning  NetworkNotReady  2m16s (x301 over 12m)  kubelet, k8s-linuxpool1-19153786-0  network is not ready: runtime network not ready: NetworkReady=false reason:NetworkPluginNotR
eady message:Network plugin returns error: cni config load failed: no network config found in /etc/cni/net.d: cni plugin not initialized: failed to load cni config

```

**Expected behavior**

**AKS Engine version**
Version: v0.38.5
GitCommit: 3128e447e
GitTreeState: clean

**Kubernetes version**
1.15.1

**Additional context**

devigned:
**Reason for Change**:

Currently, Linux clusters will fail when specifying Kubenet as the network policy and containerd as the container runtime due to the fact that containerd requires a CNI configuration to run, even with Kubenet. This works with Docker runtime because the docker_shim in Kubelet creates a default CNI configuration if one does not exist.

This PR creates a CNI template (Golang template) in `/etc/containerd/kubenet_template.conf` and then references that template via `/etc/containerd/config.toml`.

**Issue Fixed**:

Fixes #1723

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -187,6 +187,54 @@ write_files:
...
+    [plugins.cri.containerd.untrusted_workload_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+{{if IsKataContainerRuntime }}
+    runtime_engine = "/usr/bin/kata-runtime"
+{{else}}
+    runtime_engine = "/usr/local/sbin/runc"
+{{end}}
+    [plugins.cri.containerd.default_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+    runtime_engine = "/usr/local/sbin/runc"
```
PatrickLang: I think this needs the same if/else to work with kata

PatrickLang: or we could cut kata support if it's not tested

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -174,6 +174,54 @@ write_files:
...
+    [plugins.cri.containerd.untrusted_workload_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+{{if IsKataContainerRuntime }}
+    runtime_engine = "/usr/bin/kata-runtime"
+{{else}}
+    runtime_engine = "/usr/local/sbin/runc"
+{{end}}
+    [plugins.cri.containerd.default_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+    runtime_engine = "/usr/local/sbin/runc"
```
PatrickLang: same as above - if IsKataRuntime needed here too

pkg/api/types.go
```diff
@@ -1660,7 +1660,12 @@ func (o *OrchestratorProfile) NeedsExecHealthz() bool {
 		!common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.9.0")
 }
 
-// HasAadProfile  returns true if the has aad profile
+// GetPodInfraContainerSpec returns the sandbox image as a string (ex: k8s.gcr.io/pause-amd64:3.1)
+func (o *OrchestratorProfile) GetPodInfraContainerSpec() string {
```
PatrickLang: This looks Linux specific, can you put Linux in the func name? We'll need a similar one for Windows later

devigned: Should we not just return the proper image based on the orchestrator profile? I would imagine this func could provide back a Windows based image with the information currently provided to it.

parts/k8s/cloud-init/masternodecustomdata.yml
devigned: I was hesitant to cut support for kata as I wasn't sure it it was in use. I've left it in for now. I propose we make that choice and remove it completely as a separate change.

parts/k8s/cloud-init/nodecustomdata.yml
devigned: Fixed

parts/k8s/cloud-init/masternodecustomdata.yml
PatrickLang: sounds good

pkg/api/types.go
PatrickLang: if we move to the multi-arch images (mcr.microsoft.com/kubernetes/pause:...) then it will work on all machines as-is. Otherwise we need to decide based off each node pool.

PatrickLang: it either needs to differ by node, or be a multi-arch multi-os image. `mcr.microsoft.com/k8s/pause` is such image. What's the default for `KubernetesConfig.KubernetesImageBase` ? If it matches then this is ok. the gcr images don't work with Windows

devigned: Great point about the node pools. I hadn't thought about that. I'll change the PR to use `mcr.microsoft.com/k8s/pause`.

Thanks for the great review!!

devigned: `mcr.microsoft.com/k8s/pause` was not correct. The correct image uri is `mcr.microsoft.com/k8s/core/pause`. I've standardized on tag `1.2.0`.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -187,6 +187,58 @@ write_files:
...
+  permissions: "0644"
+  owner: root
+  content: |
+    subreaper = false
+    oom_score = 0
+    [plugins.cri]
+    sandbox_image = "{{GetPodInfraContainerSpec}}"
+    [plugins.cri.containerd.untrusted_workload_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+{{if IsKataContainerRuntime }}
```
jadarsie: this looks like dead code, is it possible to specify both `containerd` and `kata-containers`?

devigned:
```go
// NeedsContainerd returns whether or not we need the containerd runtime configuration
// E.g., kata configuration requires containerd config
func (k *KubernetesConfig) NeedsContainerd() bool {
	return k.ContainerRuntime == KataContainers || k.ContainerRuntime == Containerd
}
```

Superfluous, perhaps, but I don't think it's dead.

jadarsie: I see, thx

jadarsie:
lgtm

examples/kubernetes-kubenet-containerd.json
```diff
@@ -0,0 +1,41 @@
...
+	"properties": {
+		"orchestratorProfile": {
+			"orchestratorType": "Kubernetes",
+			"orchestratorRelease": "1.13",
+			"kubernetesConfig": {
+				"networkPlugin": "kubenet",
+				"containerRuntime": "containerd",
+				"addons": [
+					{
+						"name": "kubernetes-dashboard",
```
jackfrancis: Is the dashboard literally non-functional in this cluster configuration?

pkg/engine/template_generator_test.go
```diff
@@ -139,6 +143,93 @@ func TestGetTemplateFuncMap(t *testing.T) {
 	}
 }
 
+func TestTemplateGenerator_FunctionMap(t *testing.T) {
```
jackfrancis: wow fun!

examples/kubernetes-kubenet-containerd.json
devigned: It's just copypasta. I'll change it to whatever you suggest.

jackfrancis: Cool, let's get rid of this addons config array altogether

packer/install-dependencies.sh
```diff
@@ -182,15 +182,12 @@ for KUBE_DNS_MASQ_VERSION in ${KUBE_DNS_MASQ_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-PAUSE_VERSIONS="3.1"
+PAUSE_VERSIONS="1.2.0"
```
jackfrancis: How confident are we that this new pause image is backwards compatible with all previous versions of k8s?

test/e2e/test_cluster_configs/network/kubenet_containerd.json
```diff
@@ -0,0 +1,43 @@
...
+	"env": {},
+	"apiModel": {
+		"apiVersion": "vlabs",
+		"properties": {
+			"orchestratorProfile": {
+				"orchestratorType": "Kubernetes",
+				"kubernetesConfig": {
+					"networkPlugin": "kubenet",
+					"containerRuntime": "containerd",
+					"addons": [
```
jackfrancis: same here btw

packer/install-dependencies.sh
devigned: uhm... I had thought about this a bit and reasoned that we only had one version difference used and that was in: https://github.com/Azure/aks-engine/pull/2050/files/096ad5d32ac7ba931f36876d1b7131a2823ae1aa#diff-bb9b6cb6f08a63800be959ba49eaf714L364. Everything else was pause:3.1. 

Based on the only difference being 3.0 and 3.1, I figured pause was following semantic versioning and thus would not have a breaking change between 3.0 and 3.1. Granted, could be more credit than due, but seemed reasonable.

Based on the above, I figured there is relatively low risk of using only pause:1.2.0.

What do you think?

test/e2e/test_cluster_configs/network/kubenet_containerd.json
devigned: good catch! fixing...

packer/install-dependencies.sh
jackfrancis: Are we saying that this particular `1.2.0` release has an equivalent "3 dot" release (currently we are using `3.1` from `k8s.gcr.io`.

In addition, the k8s.gcr.io pause image is the one referenced inside core components, so if we remove that image from being pre-pulled we break disconnected scenarios.

devigned: I have added to be pre-pulled in `install_dependencies.sh`.

AFAICT, the image works with the existing tests. I believe, @jchauncey is building this image to be a replacement for the `k8s.gcr.io` image.

jackfrancis: What I'm saying is: are we confident that all the changes in this PR mean that all cluster configurations will no longer refer to the k8s.gcr.io pause 3.1 image?

devigned: I'm not confident that all cluster configs will only use mcr pause.

Since that is the case, should we then continue to pull `k8s.gcr.io`? Should I run each config without outbound network access? What would you recommend?

jackfrancis: I recommend we add the new pause image to our VHD config spec, and keep the existing pause image as well.

(Generally we should be keeping existing pre-pulled images to protect the disconnected environments for back-compat scenarios; hyperkube images are an exception because we don't have the storage overhead to keep older than n-1 images around.)

devigned: Ok, so both are now getting pulled in `install_dependencies.sh`.

https://github.com/Azure/aks-engine/pull/2050/files#diff-62ed5765f6aaf1d428746b3d7b597100R191-R204

PatrickLang:
LGTM - my feedback was addressed but I'll leave it to @jackfrancis or @CecileRobertMichon for final since they had other feedback afterwards

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -192,6 +192,58 @@ write_files:
...
```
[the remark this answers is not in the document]
devigned: I was having issue with whitespace. I'll give it another try.

packer/install-dependencies.sh
```diff
@@ -188,8 +188,16 @@ for KUBE_DNS_MASQ_VERSION in ${KUBE_DNS_MASQ_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-PAUSE_VERSIONS="3.1"
-for PAUSE_VERSION in ${PAUSE_VERSIONS}; do
+MCR_PAUSE_VERSIONS="1.2.0"
+for PAUSE_VERSION in ${MCR_PAUSE_VERSIONS}; do
+    # Pull the arch independent MCR pause image which is built for Linux and Windows
+    CONTAINER_IMAGE="mcr.microsoft.com/k8s/core/pause:${PAUSE_VERSION}"
```
CecileRobertMichon: if I'm reading this right, we are already pulling the mcr image below

CecileRobertMichon:
```
# Image 'mcr.microsoft.com/k8s/azurestack/core/pause-amd64' is the same as 'k8s.gcr.io/pause-amd64'
    # At the time, re-tagging and pushing to mcr hub seemed simpler than changing how `defaults-kubelet.go` sets `--pod-infra-container-image`
    for IMAGE_BASE in k8s.gcr.io mcr.microsoft.com/k8s/azurestack/core; do
```

CecileRobertMichon: oh nevermind that one is in the Azure Stack MCR... the renaming to MCR/GCR is confusing though

CecileRobertMichon: maybe we can move the Azure Stack one to the MCR block, and also we should figure out if/why Azure Stack needs its own image, maybe one MCR image is good enough for everything

devigned: We are pulling the image from a different path. @jackfrancis requested that we keep the old images pulled for [backward compat](https://github.com/Azure/aks-engine/pull/2050#discussion_r333119287).

This work basically did what Azure Stack didn't want to do in the PR comments above.

devigned: In the PR discussion above, @jadarsie reviewed the change to move AzStack over to `mcr.microsoft.com/k8s/core/pause`. At this point, _all_ aks-engine clusters should depend on `mcr.microsoft.com/k8s/core/pause`.

parts/k8s/cloud-init/masternodecustomdata.yml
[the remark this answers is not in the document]
devigned: All flattened. Thanks for the nudge.

devigned:
@jackfrancis I missed a pause-amd64 for k8s 1.14. It was caught by the PR tests. This caused the lgtm to be removed. When you get a sec, please give it one more look.

jackfrancis:
> @jackfrancis I missed a pause-amd64 for k8s 1.14. It was caught by the PR tests. This caused the lgtm to be removed. When you get a sec, please give it one more look.

Yay PR tests ❤️

devigned:
codecov lies! I tested all of the functions it showed in red. I'm merging.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -192,6 +192,58 @@ write_files:
...
+            "type": "bridge",
+            "bridge": "cbr0",
+            "mtu": 1500,
+            "addIf": "eth0",
+            "isGateway": true,
+            "ipMasq": false,
+            "hairpinMode": false,
+            "ipam": {
+                "type": "host-local",
+                "subnet": "{{`{{.PodCIDR}}`}}",
```
PatrickLang: cc @marosset

devigned: Something wrong here?

marosset: Nope, we were wondering how the pod CIDR was determinted/discovered for kubenet on Linux

devigned: Awesome! I was totally scared I caused some catastrophic problem with this change.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -192,6 +192,58 @@ write_files:
...
+    runtime_type = "io.containerd.runtime.v1.linux"
+    {{if IsKataContainerRuntime }}
+    runtime_engine = "/usr/bin/kata-runtime"
+    {{else}}
+    runtime_engine = "/usr/local/sbin/runc"
+    {{end}}
+    [plugins.cri.containerd.default_runtime]
+    runtime_type = "io.containerd.runtime.v1.linux"
+    {{if IsKataContainerRuntime }}
+    runtime_engine = "/usr/bin/kata-runtime"
```
PatrickLang: @egernst is this something we should keep, or remove in favor of the v2-shim implementation?

egernst: V2-shim is definitely more idea, but this may introduce a dependency on containerd version.  Let me check into this.

The install method for Kata for AKS-engine will need to be updated, as well., if we agree this makes sense.

PR #2052, opened 2019-09-27
feat: Skip installing docker if already installed on windows vhd

### LINKED ISSUE #2051: Skip installing docker if already installed in windows vhd
Skip installing docker if already installed in windows vhd

rjaini:
**Reason for Change**:

For total disconnected case we don't want docker install flow to be trigger again when its already installed on windows vhd.

**Issue Fixed**:
Fixes #2051 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowsfunctions.ps1
+++ b/parts/k8s/kuberneteswindowsfunctions.ps1
@@ -29,7 +29,7 @@ function DownloadFileOverHttp
 
     if ($search.Count -ne 0)
     {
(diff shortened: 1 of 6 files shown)


marosset:
For this change we should introduce a new flag to indicate we want to run the CSE in a disconnected mode. If we take these changes as is we would skip installing the version of docker specified in the API model if any version is already installed.

Issue #2036, opened 2019-09-26
Don't depend on VHD distro in CSE


CecileRobertMichon:
https://github.com/Azure/aks-engine/pull/1944 introduced a dependency on using the AKS maintained VHD image for skipping outbound operations (in this case install apt packages).

We should consider reverting that change and instead skipping the apt install if the VHD footprint file is present (like before) or even better detect if the wanted apt packages are present. The issue with checking for distro value is that it impacts customers who might be using a custom image that is based on the AKS VHD and contain all the required packages but aren't using the AKS distro.

PR #2055, opened 2019-09-28
fix: don’t assign distro if image ref is provided

jackfrancis:
**Reason for Change**:

This PR addresses an unresolved collision between `imageReference` configuration and our curated OS images (which are delivered via the `distro` configuration property). Specifically, unknown outcomes will occur if we associate any specific `distro` value with a cluster spec that includes a custom `imageReference` configuration.

Includes additional validation to fail fast when `imageReference` + `distro` configurations are used.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -407,7 +407,7 @@ func (p *Properties) setExtensionDefaults() {
 }
 
 func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName string) {
-	if p.MasterProfile.Distro == "" {
+	if p.MasterProfile.Distro == "" && p.MasterProfile.ImageRef == nil {
 		if p.OrchestratorProfile.IsKubernetes() {
 			p.MasterProfile.Distro = AKSUbuntu1604
 		} else {
@@ -629,7 +629,7 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName 
 		}
 
 		if profile.OSType != Windows {
-			if profile.Distro == "" {
+			if profile.Distro == "" && profile.ImageRef == nil {
 				if p.OrchestratorProfile.IsKubernetes() {
 					if profile.OSDiskSizeGB != 0 && profile.OSDiskSizeGB < VHDDiskSizeAKS {
 						profile.Distro = Ubuntu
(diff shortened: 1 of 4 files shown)


pkg/api/defaults.go
```diff
@@ -404,7 +404,7 @@ func (p *Properties) setExtensionDefaults() {
 }
 
 func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName string) {
-	if p.MasterProfile.Distro == "" {
+	if p.MasterProfile.Distro == "" && p.MasterProfile.ImageRef == nil {
```
CecileRobertMichon: maybe we should assign a Distro value `custom` or something like that so it's explicit and doesn't run into unexpected usages of empty distro value.

jackfrancis: I have a preference for using empty string as a more expressive/semantic value of "this cluster is not using any distro configuration"

jackfrancis: @mboersma @devigned @PatrickLang @marosset any thoughts here?

mboersma: I think setting to `nil` is most clear if we want to say "no distro set." But that may have unfortunate ripple effects in the code, so I'm fine with empty string.

devigned: Hmm... might be taste, but I'd vote for either `*string` to indicate emptiness or just document what `""` means. I worry that adding a default that is not a golang idiomatic default value would further complicate. If we were to move down that path, then I think I'd not use a string, but rather a type (an enumeration).

jackfrancis: I don't understand this part

> adding a default that is not a golang idiomatic default value would further complicate

Is it non-idiomatic to assign significance to the zero value of a type? (in this case a string, so `""`)

devigned: I'm sorry for not being clear. `""` is the idiomatic default value for string. In contrast, a default of `"custom"` would not. If we were to use `"custom"`, I would vote we use a named type for `Distro` rather than a string.

marosset:
We should add a cluster config to use in a jenkins to validate this as well.
When i started looking at the way images were specified for windows pools there was documentation for targeting a VHD but the code behind that either never worked or had rotted :(

CecileRobertMichon:
@jackfrancis I took another look at this PR and I don't see how it fixes issue #2036. In order to fix it we need to change CSE to skip apt package installs for custom images that aren't using the AKS distro but have those packages pre-installed because they are based on the AKS VHD, does that make sense?

jackfrancis:
This tested out O.K. using the packer pipeline defined in #2118 🎉

jackfrancis:
@CecileRobertMichon I don't quite understand by I'm happy to remove the reference to #2036 if this doesn't address that. (i.e., I'd like to merge this as-is)

PR #2061, opened 2019-09-30
fix: updating windows release notes script to set a file encoding

marosset:
**Reason for Change**:

Powershell does not add a file encoding by default which is results in github treating the text file as a binary file which checking in release notes generated for windows VHD images.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
Thanks Matt!

PR #2071, opened 2019-09-30
chore: Windows update to docker 19.03.2

marosset:
**Reason for Change**:

New versions of docker have been released.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/windowsconfigfunc.ps1
+++ b/parts/k8s/windowsconfigfunc.ps1
@@ -63,11 +63,6 @@ function Install-Docker
             [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.37', [System.EnvironmentVariableTarget]::Machine)
         }
 
-        "19.03" {
-            Write-Log "Docker 19.03 found, setting DOCKER_API_VERSION to 1.40"
-            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.40', [System.EnvironmentVariableTarget]::Machine)
-        }
-
         default {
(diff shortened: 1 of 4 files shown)


parts/k8s/windowsconfigfunc.ps1
```diff
@@ -63,6 +63,11 @@ function Install-Docker
             [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.37', [System.EnvironmentVariableTarget]::Machine)
         }
 
+        "19.03" {
+            Write-Log "Docker 19.03 found, setting DOCKER_API_VERSION to 1.40"
+            [System.Environment]::SetEnvironmentVariable('DOCKER_API_VERSION', '1.40', [System.EnvironmentVariableTarget]::Machine)
```
PatrickLang: oh my forgot about this workaround for k8s. If you remove this case do the tests fail? I think this was supposed to be a point-in-time workaround for a specific version

marosset: checking - stay tuned

marosset: Is there a bug or something that i can reference here for future maintainers?

PatrickLang: thanks for adding the link. Those two specific versions needed special-case treatment. They could still be in use in some clusters running 1.14 so we shouldn't pull the code out. We just shouldn't set that env var for new deployments using newer Docker versions.

PR #2078, opened 2019-10-01
fix: auditd improvements

### LINKED ISSUE #2077: auditd systemd service is broken
When auditd is disabled via api model, we deliver a systemd service that doesn't actually work.

```
Could not open dir /var/log/audit (No such file or directory)
```

This is because we don't actually deliver the auditd implementation if `auditDEnabled` is set to `false` (which it is by default).

To allow disconnected AKS Engine scenarios to use auditd, we pre-install it via VHD. We should do a best-effort `apt-get purge` of auditd during CSE for all `"auditDEnabled": false` scenarios so that intrepid users don't notice a systemd service that doesn't actually work.

jackfrancis:
**Reason for Change**:

This PR addresses auditd cruft left on vms that are built without the auditd flag enabled:

If `auditDEnabled` is not true in the cluster configuration, we should clean up the auditd apt package so that folks don't try and use it (we only deliver a working configuration *if* `auditDEnabled` is set to true.

This PR adds best-effort (i.e., it runs in the background and failures are ignored) `apt-get purge auditd`.

This PR also enables auditd for non-VHD distros.

Finally, this PR adds `"auditDEnabled": true` configuration for the "everything" test config to validate that we protect this feature with regular tests.

**Issue Fixed**:

Fixes #2077 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -22,16 +22,6 @@ systemctlEnableAndStart() {
         return 1
     fi
 }
-systemctlDisableAndStop() {
-    if ! systemctl_stop 100 5 30 $1; then
-        echo "$1 could not be stopped"
-        return 1
-    fi
-    if ! retrycmd_if_failure 120 5 25 systemctl disable $1; then
-        echo "$1 could not be disabled by systemctl"
-        return 1
-    fi
-}
 
 configureEtcdUser(){
     useradd -U "etcd"
@@ -132,7 +122,7 @@ ensureAuditD() {
     systemctlEnableAndStart auditd || exit $ERR_SYSTEMCTL_START_FAIL
   else
     if apt list --installed | grep 'auditd'; then
-      systemctlDisableAndStop auditd || exit $ERR_SYSTEMCTL_START_FAIL
+      apt_get_purge 20 30 120 auditd &
     fi
   fi
 }
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1299,6 +1299,11 @@ func (m *MasterProfile) IsVHDDistro() bool {
 	return m.Distro == AKSUbuntu1604 || m.Distro == AKSUbuntu1804
 }
 
+// IsAuditDEnabled returns true if the master profile is configured for auditd
+func (m *MasterProfile) IsAuditDEnabled() bool {
+	return m.AuditDEnabled != nil && to.Bool(m.AuditDEnabled)
+}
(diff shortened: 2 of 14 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -132,7 +122,7 @@ ensureAuditD() {
     systemctlEnableAndStart auditd || exit $ERR_SYSTEMCTL_START_FAIL
   else
     if apt list --installed | grep 'auditd'; then
-      systemctlDisableAndStop auditd || exit $ERR_SYSTEMCTL_START_FAIL
+      apt_get_purge 20 30 120 auditd &
```
mboersma: Removing the package I assume does `systemctl stop` as well?

pkg/api/types.go
```diff
@@ -1299,6 +1299,11 @@ func (m *MasterProfile) IsVHDDistro() bool {
 	return m.Distro == AKSUbuntu1604 || m.Distro == AKSUbuntu1804
 }
 
+// IsAuditDEnabled returns true if the master profile is configured for auditd
+func (m *MasterProfile) IsAuditDEnabled() bool {
+	return m.AuditDEnabled != nil && to.Bool(m.AuditDEnabled)
```
mboersma: A `nil` check isn't necessary since [`to.Bool`](https://github.com/Azure/go-autorest/blob/master/autorest/to/convert.go#L71) returns `false` in that case.

pkg/api/types.go
```diff
@@ -1428,6 +1433,11 @@ func (a *AgentPoolProfile) IsVHDDistro() bool {
 	return a.Distro == AKSUbuntu1604 || a.Distro == AKSUbuntu1804
 }
 
+// IsAuditDEnabled returns true if the master profile is configured for auditd
+func (a *AgentPoolProfile) IsAuditDEnabled() bool {
+	return a.AuditDEnabled != nil && to.Bool(a.AuditDEnabled)
```
mboersma: Same comment as above: `nil` check not necessary.

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: It removes the systemd service definition altogether, so yeah explicitly stopping/disabling on startup is no longer a thing. (Also this was the only use of that func hence it goes altogether)

jackfrancis:
@mboersma this is ready for review

PR #2092, opened 2019-10-02
feat: add support for Kubernetes v1.16.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#changelog-since-v1160

**Issue Fixed**:

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Still waiting on publishing of k8s.gcr.io/hyperkube:v1.16.1. *Update*: published now.

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -142,13 +142,14 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.15.2":         false,
 	"1.15.3":         true,
 	"1.15.4":         true,
-	"1.16.0-alpha.1": true,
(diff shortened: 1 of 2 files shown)


pkg/api/common/versions.go
```diff
@@ -142,13 +142,14 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.15.2":         false,
 	"1.15.3":         true,
 	"1.15.4":         true,
-	"1.16.0-alpha.1": true,
```
mboersma: The prerelease versions should have been set to `false` in #1972. Better late than never.

PR #2093, opened 2019-10-02
chore: add legal notice to vhd

### LINKED ISSUE #2084: discussion: vhd legal notice
**What**

Because Azure Stack users own the hardware, CELA asked us to take a few things into account. One of them is to create a NOTICE file that lists all the OSS components inside the pre-pulled docker images.

**How**

- `tern` is used to inspect container images (https://github.com/vmware/tern)
- For now, the process that extracts components and composes the notice file will be executed offline
- Eventually, it will be moved to a pipeline (Az or AzS)
- There is no need to execute tern on each image more than once.
- We will keep the output file in the AKSe repo and only run the process against new images

jadarsie:
**Reason for Change**:
Fixes #2084

**Notes**:
A few files are empty, that's ok.


packer/vhd-image-builder.json
```diff
@@ -191,6 +191,11 @@
             "source": "parts/k8s/cloud-init/artifacts/docker_clear_mount_propagation_flags.conf",
             "destination": "/home/packer/docker_clear_mount_propagation_flags.conf"
         },
+        {
+            "type": "file",
+            "source": "releases/notice/notice.txt",
+            "destination": "/NOTICE.txt"
```
jackfrancis: Do we really want to put this in the root volume?

jadarsie: That seems to be the case, I can check if alternative locations are accepted.

jadarsie: BTW, pushed a commit to validate if creating the file in the root vol is causing the pipeline to fail

releases/notice/sha256:23df717980b4aa08d2da6c4cfa327f1b730d92ec9cf740959d2d5911830d82fb.json.txt
```diff
@@ -0,0 +1,55 @@
+---
```
jackfrancis: How will these `sha256:<sha>.json.txt` files be maintained over time? Can we throw them away and simply keep the `notice.txt` file?

jadarsie: those files avoid the need to execute tern more than once for each image, producing them took a few hours, if this was a quicker process I do not see a need to keep them around
I can store them in a different place

jackfrancis: Cool, yeah, I think I prefer keeping them out of the codebase, but if it makes your pipeline more difficult to maintain we can revisit that

PR #2095, opened 2019-10-02
chore: update default IPv6 subnet for dual stack, update docs

aramase:
**Reason for Change**:

- Updates docs with changes in dual-stack phase2
- Update default ipv6 cluster cidr to `fc00::/8`

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

/cc @lachie83

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -439,7 +439,7 @@ const (
 	// DefaultKubernetesClusterSubnet specifies the default subnet for pods.
 	DefaultKubernetesClusterSubnet = "10.244.0.0/16"
 	// DefaultKubernetesClusterSubnetIPv6 specifies the IPv6 default subnet for pods.
-	DefaultKubernetesClusterSubnetIPv6 = "fd00:101::/8"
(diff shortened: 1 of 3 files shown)


pkg/api/const.go
```diff
@@ -439,7 +439,7 @@ const (
 	// DefaultKubernetesClusterSubnet specifies the default subnet for pods.
 	DefaultKubernetesClusterSubnet = "10.244.0.0/16"
 	// DefaultKubernetesClusterSubnetIPv6 specifies the IPv6 default subnet for pods.
-	DefaultKubernetesClusterSubnetIPv6 = "fd00:101::/8"
+	DefaultKubernetesClusterSubnetIPv6 = "fc00::/8"
```
jackfrancis: This change isn't a doc change (although of course it's good to document it :))

Just to make sure we want to deliver this change?

aramase: yeah, we want this to be changed.

aramase: got it! will separate such changes next time and create separate PRs. Thanks @jackfrancis!

PR #2103, opened 2019-10-03
chore: document some bug workarounds

PatrickLang:
There's a bunch of bug workarounds in the Windows setup scripts. Document them.

/hold


PatrickLang:
/hold cancel
no reason not to merge as-is. we can add more comments later if needed

PR #2127, opened 2019-10-08
feat: expose PlatformFaultDomainCount in the apimodel

### LINKED ISSUE #2123: Expose Fault Domain count in the apimodel
**Describe the request**

The FD count is currently automatically set based on the region of deployment.
This default behavior is fine. However, some SKUs do not support FD, even in regions that support FD.
To avoid tracking every new SKUs and their support for FD, we can expose a property on the agentpool to override the FD count and leave it to the user to set it for these specific SKUs.

Alternative, maintain a map of SKUs -> maxDefaultFDcount or SKUs -> singleFD boolean.
this can could be an addition to the solution above to make the api mode user friendly: automate the choice, but leave an override for the user for SKUs that we did not list yet.

serbrech:
**Reason for Change**:
see #2123

**Issue Fixed**:
fixes #2123

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests


pkg/api/vlabs/types.go
```diff
@@ -390,6 +390,7 @@ type MasterProfile struct {
 	AgentSubnet              string            `json:"agentSubnet,omitempty"`
 	AvailabilityZones        []string          `json:"availabilityZones,omitempty"`
 	SinglePlacementGroup     *bool             `json:"singlePlacementGroup,omitempty"`
+	PlatformFaultDomainCount *int              `json:"platformFaultDomainCount,omitEmpty"`
```
jackfrancis: In order to compose both vlabs and "unversioned" data types we need these property additions to `MasterProfile` and `AgentPoolProfile` to be in `pkg/api/convertertoapi.go` and `pkg/api/converterfromapi.go`

pkg/engine/availabilitysets.go
```diff
@@ -28,10 +28,6 @@ func CreateAvailabilitySet(cs *api.ContainerService, isManagedDisks bool) Availa
 			avSet.AvailabilitySetProperties = &compute.AvailabilitySetProperties{
 				PlatformUpdateDomainCount: to.Int32Ptr(3),
 			}
-			if cs.Properties.MasterProfile.PlatformFaultDomainCount != nil {
```
jackfrancis: So it looks like in fact we were basing the FD on the api model already, we just weren't exposing that as a user-viewable property. Does it make sense to explicitly pass in a `PlatformFaultDomainCount` value for Availability Zone-configured AvailabilitySet VMs? (I know that's a mouthful)

@ritazh @sozercan fyi

serbrech: Apparently yes, you can fix the FaultDomainCount of a AZ enabled VMSS.
default is 1, and means MaxSpread and is recommended. but it can also be fixed to _n_
So if the property is there, I figured its best to apply it. if you want the default, don't set it.
If we apply the property for non AZ but don't apply if for AZ enabled, then we differ in behavior from VMSS.
https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-manage-fault-domains

pkg/api/vlabs/types.go
serbrech: done in a22edcb

pkg/engine/availabilitysets.go
```diff
@@ -40,6 +36,11 @@ func CreateAvailabilitySet(cs *api.ContainerService, isManagedDisks bool) Availa
 		}
 	}
 
+	if cs.Properties.MasterProfile.PlatformFaultDomainCount != nil {
+		p := int32(*cs.Properties.MasterProfile.PlatformFaultDomainCount)
```
CecileRobertMichon: For agents the equivalent value assignment is be inside the `profile.IsManagedDisks() {` if statement, what's the different between master and agent?

serbrech: managed disks integrate with the VM faultdomains. they align on the same FD.
UnManaged disks can support less FD and don't guarantee alignment.
So you should not use FD with unmanaged disks really.
But I don't think that silently swallowing the setting is the right approach. if you set FD and you use UnManaged Disks, you should know what you're doing.
Finally, who still uses unmanaged disks? :)

serbrech: I didn't move the agent one because I didn't want to change its behavior, but I would say that it would be better to move it out of the if else.
It's pretty much 0 impact anyway, because who still uses unmanaged disks?

palma21:
See all is looking good, could we get one more lgtm? Any concerns?

pkg/engine/availabilitysets.go
serbrech: @jackfrancis It seems that I didn't digest your mouthful properly. You did point out AvailabilitySet, my answer was documentation about VMSS. And you were right I think. 
The template does not include any AvailabilitySet when we set Zones, and so, there is no point setting the FDCount as it won't even show up anywhere in the arm template. 
see https://github.com/Azure/aks-engine/blob/master/pkg/engine/masterarmresources.go#L20-L23
So I'll revert this part and send a new PR, we can run BackCompat tests on the branch again and take it from there

PR #2130, opened 2019-10-09
feat: enable re-use of VHD

### LINKED ISSUE #2036: Don't depend on VHD distro in CSE

jackfrancis:
**Reason for Change**:

This PR re-orders the "do I have the 'created by AKS VHD CI' OS signature?" and the "does my cluster configuration explicitly declare a VHD distro?" checks in order to enable AKS VHD re-use scenarios. E.g.:

- create a new managed image / shared gallery image / VHD _from_ the AKS VHD

Specifically, the boolean variable `FULL_INSTALL_REQUIRED` in our CSE implementation is functionally coupled with the VHD CI implementation that pairs with that CSE implementation at that commit of aks-engine. In practice, this means that for folks who want to re-use the AKS VHD as a base for creating their own Kubernetes node OS images can (1) take advantage of the optimized CSE code paths (e.g., don't execute unnecessary `apt` operations) and (2) implement the proper cleanup functionality so that the resultant Kubernetes node isn't unnecessarily burdened with the storage overhead of dozens of unused, pre-pulled Kubernetes hyperkube images (we use a common VHD to support a large set of Kubernetes versions).

**Issue Fixed**:

Fixes #2036 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -58,15 +58,15 @@ if [[ "${GPU_NODE}" != "true" ]]; then
 fi
 
 VHD_LOGS_FILEPATH=/opt/azure/vhd-install.complete
-if [[ "${IS_VHD}" = true ]]; then
-    if [ ! -f $VHD_LOGS_FILEPATH ]; then
-        echo "Using VHD distro but file $VHD_LOGS_FILEPATH not found"
-        exit $ERR_VHD_FILE_NOT_FOUND
-    fi
+if [ -f $VHD_LOGS_FILEPATH ]; then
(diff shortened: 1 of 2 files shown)


jackfrancis:
@CecileRobertMichon I think this might fix #2036

Issue #2072, opened 2019-10-01
E2E: only run Windows custom image test variations on a single k8s version


jackfrancis:
As a practical way of reducing our test overhead, let's run the various Windows custom image / VHD / etc tests against a single version of k8s; perhaps the default that ships for cluster w/ Windows node pools?

That would be modifying the JSON files here:

`test/e2e/test_cluster_configs/windows`

That would be adding a root `"options"` object that has the `"allowedOrchestratorVersions"` array in it, e.g.:

```
"options": {
	"allowedOrchestratorVersions": ["1.15"]
},
```

jackfrancis:
@devigned @marosset FYI

devigned:
I think this would be relatively trivial to implement.

Check out: https://github.com/Azure/aks-engine/blob/7a6981d8367d3bf135e6e043916bfa8cbe92b2f1/Jenkinsfile#L154-L159

I believe all one would need to do is the following:
```groovy
// define some job level or server parameter that is "CurrentVersion", 
// thus `params.CurrentVersion`

def isAllowedVersion = jobCfg.options?.allowedOrchestratorVersions == null ? true : version in jobCfg.options.allowedOrchestratorVersions

isAllowedVersion |= jobCfg.options?.useCurrentVersion && params.CurrentVersion == version

// ....
```

marosset:
Cool, we can start using this with the windows tests but i imaging this would be useful for more test cluster configs

marosset:
This work was completed but the issue was not closed

PR #2133, opened 2019-10-09
test: Add 'latestReleasedVersion' target to Jenkins jobs

marosset:
**Reason for Change**:
This change 
- Fixes some of the test configs by moving allowedOrchestratorRelease to options instead of env
- Adds the abililty to target a 'latest released' k8s version for jobs that do not benefits from running against multiple k8s versions
- Updates some additional test configs to use the 'latestReleasedVersion' allowedOrchestratorRelease

**Issue Fixed**:

Addresses issue #2072 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -158,9 +158,9 @@ stage ("discover tests") {
 				// (version being process equals latestReleasedVersion AND allowedOrchestratorVersions contains "latestReleasedVersion")
 				def allowedVersions = jobCfg.options?.allowedOrchestratorVersions
 				def isVersionAllowed = allowedVersions == null ? true  : version allowedVersions
-				isAllowedVersion |= version == latestReleasedVersion && allowedVersions && "latestReleasedVersion" in allowedVersions
+				isVersionAllowed |= version == latestReleasedVersion && allowedVersions && "latestReleasedVersion" in allowedVersions
 
-				if(!isAllowedVersion) {
+				if(!isVersionAllowed) {
 					// the job config has limited this job to not run for this verion of the orchestrator
 					echo("${jobName} is limited to ${jobCfg.options?.allowedOrchestratorVersions}; not running for ${version}")
 					return // this is a continue and will not exit the entire iteration
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -153,11 +153,13 @@ stage ("discover tests") {
 				}
 
 				def isAllowedVersion = jobCfg.options?.allowedOrchestratorVersions == null ? true : version in jobCfg.options.allowedOrchestratorVersions
-				isAllowedVersion |= (version == stableVersion) && stableVersion in jobCfg.options.allowedOrchestratorVersions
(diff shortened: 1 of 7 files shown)


marosset:
I tested this out using the following jenkins job
https://upstream.jenkins.azure-containers.io/view/aks-engine/job/curentversion-test/
Please let me know what additional testing I can do here.

Jenkinsfile
```diff
@@ -6,6 +6,7 @@ defaultEnv = [
 	] + params
 
 def k8sVersions = ["1.12", "1.13", "1.14", "1.15", "1.16"]
+def latestReleasedVersion = "1.16"
```
devigned: why not just take the last index of k8sVersions?

Jenkinsfile
```diff
@@ -152,10 +153,13 @@ stage ("discover tests") {
 				}
 
 				def isAllowedVersion = jobCfg.options?.allowedOrchestratorVersions == null ? true : version in jobCfg.options.allowedOrchestratorVersions
+				isAllowedVersion |= (version.equals(latestReleasedVersion)) && jobCfg.options?.allowedOrchestratorVersions != null && "latestReleasedVersion" in jobCfg.options.allowedOrchestratorVersions
```
devigned: I think we can clean this up a bit.
```groovy
nullOps = null
fooOps = ["foo"]
blahOps = ["blah"]
foo = "foo"

println(nullOps && "blah" in nullOpts)   // false
println(fooOps && "blah" in fooOps)      // false
println(blahOps && "blah" in blahOps)    // true
println(nullOps ?: "foo")                // foo
println(fooOps[0] == foo)                // true

isTrue = false
isTrue |= foo == "foo" && fooOps && foo in fooOps
println(isTrue)                          // true

// check out https://groovy-playground.appspot.com
```
So...
```groovy
def allowedVersions = jobCfg.options?.allowedOrchestratorVersions
def isAllowedVersion = allowedVersions ?: version in allowedVersions
isAllowedVersion |= version == latestReleasedVersion && allowedVersions && "latestReleasedVersion" in allowedVersions
```

Jenkinsfile
marosset: My reasoning was that we'll probably add versions like 1.17  to this list which they under active development and I thought it would help reduce noise to run these jobs for the latest stable/released k8s version.
Thoughts?

devigned: sgtm -- ty

Jenkinsfile
marosset: i'll try that out. I'm a total noob w.r.t. js/java/groovy/whatever this is :P

marosset: updated.
I needed to change the second line to isAllowedVersions = allowedVersions == null ? true : version in allowedVersions because ?: assigned the value of allowedVersions to isAllowedVersions if it is not null.

PR #2053, opened 2019-09-27
fix: fixes cluster preprovisioning script download on azurestack.

deaborch:
**Reason for Change**:

Cluster pre-provisioning was failing on Azurestack due to an untrusted certificate on the host preventing the pre-provisioning scripts from being downloaded

**Issue Fixed**:

**Requirements**:

- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [X] includes documentation
- [X] adds unit tests
- [X] tested upgrade from previous version

**Notes**:

PR #2135, opened 2019-10-09
fix: revert 'trust Azure Stack's CA cert before pre-provision extensions'

jadarsie:
**Reason for Change**:
We have a bizarre DNS issue that I think is caused by this #2053. 
Reverting until we figure out what's going on.

- [ ] includes documentation
- [ ] adds unit tests


jadarsie:
@jackfrancis, the checks are not reflecting the status pipeline status, it looks like it is done
https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=2758

PR #2147, opened 2019-10-11
fix: pass 'cacert' option to curl pre-provision scripts on Azure Stack

jadarsie:
**Reason for Change**:
Same reason as #2053 (then reverted here #2135). Downloading pre-provision extensions fails if the scripts are hosted on the local Azure Stack instance. Pass the 'cacert' option to curl fixed the problem and does not cause the unintended side-effects.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests


rjaini:
lgtm

PR #2152, opened 2019-10-14
fix: VM and VMSS do not have user assigned identity in its dependsOn

### LINKED ISSUE #2082: Deployment will fail for cluster using user assigned identity when identity service is throttled
**Describe the bug**
If the k8s cluster uses user assigned managed identity, in AKS-Engine generated template, the VM will not add the identity to its `dependsOn`:
```
    "resources": [
        {
            "apiVersion": "[variables('apiVersionManagedIdentity')]",
            "name": "[variables('userAssignedID')]",
            "location": "[variables('location')]"
        },
        {
            "type": "Microsoft.Compute/virtualMachines",
            "apiVersion": "[variables('apiVersionCompute')]",
            "name": "[concat(variables('nodepool1VMNamePrefix'), copyIndex(variables('nodepool1Offset')))]",
            "location": "[variables('location')]",
            "dependsOn": [
                "[concat('Microsoft.Network/networkInterfaces/', variables('nodepool1VMNamePrefix'), 'nic-', copyIndex(variables('nodepool1Offset')))]",
                "[concat('Microsoft.Compute/availabilitySets/', variables('nodepool1AvailabilitySet'))]"
            ],
            "identity": {
                "type": "UserAssigned",
                "userAssignedIdentities": {
                    "[variables('userAssignedIDReference')]": {}
                }
            },
            ...
```
Normally it's OK because the creation of the identity is really fast, but when identity service is throttled(limitation is 8 requests per second per tenant according to identity team), the creation of identity will be slow and the deployment will fail because it tries to associate an non-existing identity to VM. The identity will be created finally but the deployment is early failed when trying to create VM.

**Steps To Reproduce**

Not always reproduce. Can only be reproduced for cluster using user assigned identity when identity service is throttled in tenant.

**Expected behavior**
The deployment should success.
**AKS Engine version**
All version.
**Kubernetes version**
All version.
**Additional context**

norshtein:
**Reason for Change**:

Background is described in #2082 .

**Issue Fixed**:

Fixes #2082 

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -35,6 +35,9 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 	if isStorageAccount {
 		dependencies = append(dependencies, "[variables('masterStorageAccountName')]")
 	}
+	if userAssignedIDEnabled {
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
 
 	armResource := ARMResource{
 		APIVersion: "[variables('apiVersionCompute')]",
@@ -344,6 +347,10 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 	dependencies = append(dependencies, fmt.Sprintf("[concat('Microsoft.Network/networkInterfaces/', variables('%[1]sVMNamePrefix'), 'nic-', copyIndex(variables('%[1]sOffset')))]", profile.Name))
 
 	dependencies = append(dependencies, fmt.Sprintf("[concat('Microsoft.Compute/availabilitySets/', variables('%[1]sAvailabilitySet'))]", profile.Name))
+
+	if userAssignedIDEnabled {
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
 
 	if profile.IsWindows() {
 		windowsProfile := cs.Properties.WindowsProfile
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -53,6 +53,10 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 		dependencies = append(dependencies, "[variables('masterLbID')]")
 	}
 
+	if userAssignedIDEnabled {
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
+
 	armResource := ARMResource{
 		APIVersion: "[variables('apiVersionCompute')]",
 		DependsOn:  dependencies,
@@ -369,6 +373,10 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	orchProfile := cs.Properties.OrchestratorProfile
 	k8sConfig := orchProfile.KubernetesConfig
 	linuxProfile := cs.Properties.LinuxProfile
+
+	if k8sConfig != nil && k8sConfig.UseManagedIdentity && k8sConfig.UserAssignedID != "" {
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
 
 	armResource.DependsOn = dependencies
 
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -374,7 +374,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	k8sConfig := orchProfile.KubernetesConfig
 	linuxProfile := cs.Properties.LinuxProfile
 
-	if k8sConfig != nil && k8sConfig.UseManagedIdentity && k8sConfig.UserAssignedID != "" {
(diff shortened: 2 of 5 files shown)


jackfrancis:
@yangl900 @serbrech FYI

jackfrancis:
@xuto2 FYI

pkg/engine/virtualmachinescalesets.go
```diff
@@ -385,6 +389,10 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	k8sConfig := orchProfile.KubernetesConfig
 	linuxProfile := cs.Properties.LinuxProfile
 
+	if k8sConfig != nil && k8sConfig.UseManagedIdentity && k8sConfig.UserAssignedID != "" {
```
xuto2: there is a util method UserAssignedIDEnabled() in pkg/api/types.go, should we reuse that?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -385,6 +389,10 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
...
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
```
xuto2: I don't see where we deploy the identity in the ARM template. and if that's true, the ARM deployment will fail.

xuto2: found it. nvm.

pkg/engine/virtualmachines.go
```diff
@@ -34,6 +34,9 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 	if isStorageAccount {
 		dependencies = append(dependencies, "[variables('masterStorageAccountName')]")
 	}
+	if userAssignedIDEnabled {
+		dependencies = append(dependencies, "[concat('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]")
+	}
```
xuto2: Shall we filter the user assigned identity resource and dependency for scalingup/upgrading template? There is some normalization method in transform.go that filter those infra resources/dependencies like nsg/vnet/rt, etc.

pkg/engine/virtualmachinescalesets.go
norshtein: yeah, it makes sense, updated.

pkg/engine/virtualmachines.go
norshtein: Sorry, I'm not familiar with that. What's the purpose of filtering resources and dependencies? To prevent the redeployment swiping existing configuration? If so, for user assigned identity, that will be ok. Even after redeployment, the identity itself, all role assignment on it and all association relationship will keep unchanged; and keep identity in VM's dependency is also fine.

xuto2: Will the identity have the same principal ID after being redeployed?

norshtein: Yes.

xuto2: Ideally when we scale/upgrade a cluster, we should only deploy new VM/VMSS without redeploying the cluster infrastructures, such as VNET/RouteTable/NSG and this UserAssignedIdentity. The consequence of touching these infrastructure is nondeterministic and is tied to how these infra are built/maintained.

 If I were you I would filter them out(Look for these below methods in transform.go)

NormalizeMasterResourcesForScaling
NormalizeForK8sVMASScalingUp
NormalizeResourcesForK8sMasterUpgrade
NormalizeResourcesForK8sAgentUpgrade

Having said that, In this case, the identity is pretty static and if the principal ID doesn't change on redeployment, it seems it's ok if we redeploy it.

norshtein: I think user assigned identity is different from VNET/RouteTable/NSG:
As you said "the consequence of touching these infrastructure is nondeterministic and is tied to how these infra are built/maintained", the biggest difference is it is guaranteed by identity team that a redeployment on a user assigned identity will not change current user assigned identity -- it's a deterministic behavior. And in current AKS-Engine's implementation, the definition of user assigned identity and role assignment on it(at the scope of cluster resource group, that is MC_ resource group in AKS) are not removed from the template when upgrading and scaling, as far as I can see, it brings following benefit:
- When the identity or the role assignment on it is deleted by mistake, a redeployment can bring them back, that is similar as a "reconciliation". And we don't need to worry about a redeployment on user assigned identity will wipe out anything so a redeployment will never harm.

The main purpose of this PR is to add the user assigned identity to VM's `dependsOn`, I think we may want to keep current AKS-Engine's transformer implementation.

norshtein: ping @xuto2

jackfrancis:
Is this stale?

norshtein:
@jackfrancis not stale. Any thing I need to do to get this PR merged into master branch?

jackfrancis:
@norshtein the reason it wasn't merged is the E2E test failures. I've rebased against master, let's run E2E tests again.

timja:
> @norshtein the reason it wasn't merged is the E2E test failures. I've rebased against master, let's run E2E tests again.

just glancing through the PRs on this repo, it's green now, is there anything else blocking it?

jackfrancis:
I'm running a back-compat test to ensure this change works w/ `aks-engine upgrade` flows of pre-existing clusters (i.e., clusters created before this change)

mboersma:
/azp run pr-e2e

> I'm running a back-compat test...

@jackfrancis were you able to run this and get any good signal?

jackfrancis:
We've done some work in this surface area recently, so I'd prefer we start any net new changes from scratch at this point.

PR #2153, opened 2019-10-14
feat: enable containermonitoring addon support for Azurestack and also update the agent to latest version

ganga1980:
**Reason for Change**:

 This PR  has following updates
  1.  update the container monitoring add-on to latest version 
   2.  enable container monitoring addon support for azure stack aks-engine clusters
   3. doc updates on how to use container monitoring for azure stack aks-engine clusters

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


docs/topics/azure-stack.md
```diff
@@ -135,6 +136,110 @@ The `AKS Base Image` marketplace item has to be available in your Azure Stack's
 
 Each AKS Engine release is validated and tied to a specific version of the AKS Base Image. Therefore, you need to take note of the base image version required by the AKS Engine release that you plan to use, and then download exactly that base image version. New builds of the `AKS Base Image` are frequently released to ensure that your disconnected cluster can be upgraded to the latest supported version of each component.
 
+## Azure Monitor for containers
```
jadarsie: This section seems to apply to AKSe in general.
Should we update this section https://github.com/Azure/aks-engine/blob/504e1039bc357d949ec4c62dd0fa9375ec182e85/docs/topics/monitoring.md#azure-monitor-for-containers instead? 
Then from the azure stack page we can refer to the monitoring page and highlight the differences

cmd/deploy.go
```diff
@@ -530,3 +518,30 @@ func (dc *deployCmd) configureContainerMonitoringAddon(ctx context.Context, k8sC
...
+
+//get domain of azure log analytics workspace based on the cloud or azure stack dependenciesLocation
+func (dc *deployCmd) getLogAnalyticsWorkspaceDomain() string {
+	var workspaceDomain string
+	var cloudOrDependenciesLocation string
+	cloudOrDependenciesLocation = dc.containerService.GetCloudSpecConfig().CloudName
+	if dc.containerService.Properties.IsAzureStackCloud() {
+		cloudOrDependenciesLocation = string(dc.containerService.Properties.CustomCloudProfile.DependenciesLocation)
+	}
+	switch cloudOrDependenciesLocation {
```
devigned: I'm not sure this switch statement does what you think it does. For example, `workspaceDomain` will be `""` if `"AzureGermanCloud"` is the value of `cloudOrDependenciesLocation`.

See: https://play.golang.org/p/WkA59EajxGP

ganga1980: > I'm not sure this switch statement does what you think it does. For example, `workspaceDomain` will be `""` if `"AzureGermanCloud"` is the value of `cloudOrDependenciesLocation`.
> 
> See: https://play.golang.org/p/WkA59EajxGP
Thanks @devigned  for catching this issue. Yes, your right.  interesting bug in my code. Fixed.

docs/topics/azure-stack.md
ganga1980: Hi, @jadarsie , As we discussed, there are differences between Azure and Azure stack AKS-Engine with regards to Azure Monitor for containers on-boarding and viewing etc.  I think, we can keep them separate and at some point, we can combine them and just cover the differences.

jadarsie: This document was initially conceived as a short reference for people targeting AzS. We mainly want to list what's possible, what's not possible and highlight procedural differences between Az and AzS. We add links to the main source of information for people that want to dig deeper on each topic.

I would like people to know that Az Monitor for Containers works on AzS, but this new section is breaking the premise by doubling the page size with mostly not-AzS specific information. I think this new section is out of place. Would you rather create a new `azure-stack-monitoring.md` (or similar) document instead of updating `monitoring.md`?

ganga1980: @jadarsie ,  I have updated common monitoring  doc and just referenced that here.

jadarsie:
lgtm

ganga1980:
Hi, @devigned , can you please help having quick look at the PR and  if everything looks good, help on approving and merging this PR?

devigned:
@ganga1980 looks like the patch is going to lower the test coverage. Would you please increase the code coverage.

ganga1980:
> @ganga1980 looks like the patch is going to lower the test coverage. Would you please increase the code coverage.

Sure, @devigned , will add the some tests to cover these updates. Hopefully, that will bring the coverage back.

ganga1980:
> > @ganga1980 looks like the patch is going to lower the test coverage. Would you please increase the code coverage.
> 
> Sure, @devigned , will add the some tests to cover these updates. Hopefully, that will bring the coverage back.

Hi, @devigned , I have added these tests and hopefully, it will get the coverage back. can you please help on running PR e2e tests on this PR via "/azp run pr-e2e" ?

cmd/deploy_test.go
```diff
@@ -903,3 +903,201 @@ func TestAPIModelWithContainerMonitoringAddonWithWorkspaceGuidAndKeyConfigInCmd(
 		t.Fatalf("expected workspaceKey : %s but got : %s", expectedWorkspaceKeyInBase64, addon.Config["workspaceKey"])
 	}
 }
+
+func TestAPIModelWithContainerMonitoringAddonWithWorkspaceGuidAndKeyConfigInCmdInMooncake(t *testing.T) {
```
devigned: I love the test coverage, but I must ask you to dry up this code a bit. Please introduce these tests as a single test with cases. A good example of this is: https://github.com/Azure/aks-engine/blob/bdd4cf5622f0b31a16be60ce48d8a6fb7499c31a/cmd/deploy_test.go#L161-L251.

devigned: I notice you added cases to the other test, so you don't need an example :). Sorry if that was a little pedantic.

ganga1980: Hi, @devigned, I refactored the tests to have single test with different cases. can you please look at the changes and also help running pr e2e tests on this?

PR #2157, opened 2019-10-14
fix: fix cni conf json for single exception address

marosset:
**Reason for Change**:

Fixes an issue where invalid cni conf json is generated when route exception lists contain a single value by forcing powershell function to return an array even for single values in GetBroadestRangesForEachAddress function 

**Issue Fixed**:

This should fix failing windows/custom_vnet jenkins test jobs

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


PatrickLang:
sometimes a bug just needs some coercion to go away

/lgtm

PR #2161, opened 2019-10-15
chore: use Azure CCM for k8s 1.16 and later

### LINKED ISSUE #1915: Switch to  cloud-provider-azure for cloud-controller-manager image

mboersma:
**Reason for Change**:
Switches to the [out-of-tree cloud provider components](https://github.com/kubernetes/cloud-provider-azure/releases) if `useCloudControllerManager` is enabled and the Kubernetes version is 1.16.x or later. These components are considered alpha status currently but intended to become the default by 1.17 final release (early December!).

**Issue Fixed**:
Fixes #1915

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
The cloud-controller-manager is deployed as a [static pod manifest](https://kubernetes.io/docs/tasks/configure-pod-container/static-pod/) as it already was, and the cloud-node-manager is deployed as an addon. This is odd but I would argue the best we can do, but I'm open to feedback.

### DIFF
--- a/parts/k8s/manifests/kubernetesmaster-cloud-controller-manager.yaml
+++ b/parts/k8s/manifests/kubernetesmaster-cloud-controller-manager.yaml
@@ -15,17 +15,24 @@ spec:
       imagePullPolicy: IfNotPresent
       command: ["cloud-controller-manager"]
       args: [<config>]
+      resources:
+        requests:
+          cpu: 100m
+          memory: 128Mi
+        limits:
+          cpu: 4
+          memory: 2Gi
       volumeMounts:
-        - name: etc-kubernetes
-          mountPath: /etc/kubernetes
-        - name: etc-ssl
-          mountPath: /etc/ssl
-          readOnly: true
-        - name: var-lib-kubelet
-          mountPath: /var/lib/kubelet
-        - name: msi
-          mountPath: /var/lib/waagent/ManagedIdentity-Settings
-          readOnly: true
+      - name: etc-kubernetes
+        mountPath: /etc/kubernetes
+      - name: etc-ssl
+        mountPath: /etc/ssl
+        readOnly: true
+      - name: var-lib-kubelet
+        mountPath: /var/lib/kubelet
+      - name: msi
+        mountPath: /var/lib/waagent/ManagedIdentity-Settings
+        readOnly: true
   volumes:
     - name: etc-kubernetes
       hostPath:
--- a/pkg/api/addons_test.go
+++ b/pkg/api/addons_test.go
@@ -378,6 +378,10 @@ func TestSetAddonsConfig(t *testing.T) {
 					Name:    AzureDiskCSIDriverAddonName,
 					Enabled: to.BoolPtr(false),
 				},
+				{
+					Name:    CloudNodeManagerAddonName,
+					Enabled: to.BoolPtr(false),
+				},
 			},
 		},
 		{
@@ -570,6 +574,10 @@ func TestSetAddonsConfig(t *testing.T) {
 				},
 				{
 					Name:    AzureDiskCSIDriverAddonName,
+					Enabled: to.BoolPtr(false),
+				},
+				{
+					Name:    CloudNodeManagerAddonName,
 					Enabled: to.BoolPtr(false),
 				},
 			},
@@ -767,6 +775,10 @@ func TestSetAddonsConfig(t *testing.T) {
 				},
 				{
 					Name:    AzureDiskCSIDriverAddonName,
+					Enabled: to.BoolPtr(false),
+				},
+				{
+					Name:    CloudNodeManagerAddonName,
 					Enabled: to.BoolPtr(false),
 				},
 			},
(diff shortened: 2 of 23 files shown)


mboersma:
Let's revisit this after #2191 merges because I anticipate some (small) refactorings needed.

parts/k8s/manifests/kubernetesmaster-cloud-controller-manager.yaml
```diff
@@ -15,17 +15,24 @@ spec:
       imagePullPolicy: IfNotPresent
       command: ["cloud-controller-manager"]
       args: [<config>]
```
feiskyer: could you also update args for ccm according to https://github.com/kubernetes/cloud-provider-azure/blob/master/examples/out-of-tree/cloud-controller-manager.yaml#L16-L28? especially, `"--controllers=*,-cloud-node"` is a new configuration.

feiskyer:
@mboersma any updates on this?

mboersma:
@feiskyer I think it's ready to merge, just trying to get CI to agree and looking for reviews from the AKS Engine team.

pkg/api/defaults-cloud-controller-manager.go
```diff
@@ -10,14 +10,16 @@ import (
...
-		"--configure-cloud-routes": strconv.FormatBool(o.RequireRouteTable()),
-		"--cloud-provider":         "azure",
-		"--cloud-config":           "/etc/kubernetes/azure.json",
-		"--cluster-cidr":           o.KubernetesConfig.ClusterSubnet,
-		"--kubeconfig":             "/var/lib/kubelet/kubeconfig",
-		"--leader-elect":           "true",
-		"--v":                      "2",
+		"--allocate-node-cidrs":         strconv.FormatBool(!o.IsAzureCNI()),
+		"--configure-cloud-routes":      strconv.FormatBool(o.RequireRouteTable()),
+		"--controllers":                 "*,-cloud-node",
```
jackfrancis: Does this `--controllers` change work for pre-1.16 clusters using CCM?

mboersma: I tested a few days ago with a 1.15 cluster and didn't see a problem, but let me test again to see if it should be made conditional.

mboersma: The 1.15 cloud-controller-manager container starts up ok with this flag according to Kubernetes but logs seem to halt after an unrelated (not!) error, although the container stays healthy. I'll make this flag conditional for k8s >= 1.16.

mboersma:
CI reports that we may be exceeding custom data max length for 1.17 clusters on this PR. :-(

pkg/api/defaults-cloud-controller-manager.go
mboersma: I made the addition of this argument conditional and added a unit test for that.

mboersma:
Rebased off of #2290 temporarily so I can see what CI failures are left to solve after the `customData` beast is slain.

mboersma:
The "should be able to kubectl port-forward to a running pod" spec fails on Windows because:
`Warning  FailedScheduling  <unknown>  default-scheduler  0/4 nodes are available: 2 node(s) didn't match node selector, 2 node(s) had taints that the pod didn't tolerate.`

The problem here is that the Windows nodes retain this taint:
`node.cloudprovider.kubernetes.io/uninitialized=true:NoSchedule`
which in the cloud-provider-azure code is the `schedulerapi.TaintExternalCloudProvider` constant. (Linux nodes don't have this taint and thus pass the `kubectl port-forward` test.)

mboersma:
Reading the cloud-provider-azure code, it's only in `nodemanager.go` that code exists to remove the taint. But the `"--controllers=*,-cloud-node"` change that was requested seems to have disabled the node-manager components:
`W1112 17:31:40.091363       1 controllermanager.go:229] "cloud-node" is disabled`

I'm going to revert that change locally to see if it removes the taint from the Windows nodes.

pkg/api/addons_test.go
```diff
@@ -3056,11 +3100,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Enabled: to.BoolPtr(false),
 				},
 				{
-					Name:    AzureFileCSIDriverAddonName,
```
mboersma: This was a duplicated pair of entries introduced in an earlier PR.

pkg/api/addons_test.go
```diff
@@ -3323,11 +3363,7 @@ func TestSetAddonsConfig(t *testing.T) {
...
```
mboersma: Same as above: duplicate entries removed.

PR #2162, opened 2019-10-15
chore: add Kubernetes patch releases to VHD

mboersma:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -401,10 +401,12 @@ K8S_VERSIONS="
 1.15.4-azs
 1.15.3
 1.15.3-azs
+1.14.8
 1.14.7
 1.14.7-azs
 1.14.6
```
jackfrancis: We need to remove both 1.14.6 versions. Until Azure Stack has a hyperkube image they are effectively unable to accept the resultant VHD.

packer/install-dependencies.sh
```diff
@@ -401,10 +401,12 @@ K8S_VERSIONS="
...
 1.14.6-azs
+1.13.12
 1.13.11
 1.13.11-azs
 1.13.10
```
jackfrancis: Ditto for 1.13.10

PR #2165, opened 2019-10-15
chore: replace dep by go mod

devigned:
**Reason for Change**:

Replace `dep` with go modules to provide better tooling support for the future.

I've created this PR to build off of what was done in #1676. A change of this magnitude will take a maintainer to work with upstream MSFT parties to ensure seamless integration.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
this seems more like a refactor than a chore

devigned:
Perhaps, you are right, @marosset. Though, I don't want to rewrite the previous commits for attribution sake.

marosset:
You can just update the PR title (which should get reflected in the merge commit)
Chores don't show up on https://github.com/Azure/aks-engine/releases and we probably want to surface this change there.

Makefile
```diff
@@ -155,21 +154,17 @@ ensure-generated:
...
 .PHONY: bootstrap
 bootstrap:
-ifndef HAS_DEP
-	go get -u github.com/golang/dep/cmd/dep
-endif
 ifndef HAS_GOX
 	go get -u github.com/mitchellh/gox
 endif
-	go install ./vendor/github.com/go-bindata/go-bindata/...
+	go get github.com/go-bindata/go-bindata
```
mboersma: We installed `go-bindata` from our vendored directory to ensure we used a specific version, IIRC. Is there a way we can still do that? Ideally v3.1.2 since we have that vendored in `go.mod`.

devigned: We actually specify the version in the go.mod file.

mboersma: Oh, **right**–`go get` is mod-aware now. Excellent!

CecileRobertMichon:
@marosset chores do show in releases, it's the "maintenance" category

devigned:
@marosset, @jackfrancis or @mboersma  what do you think about a lgtm?

devigned:
Thank you, @mboersma!

This is going to take an admin merge due to codecov and azdevops not rerunning checks, so I'm going to wait til stand-up to push the button.

CecileRobertMichon:
chore maps to maintenance but the PR title still has to be `chore: `

PR #2177, opened 2019-10-18
feat: add version file

egernst:
Maintaining a simple version file would allow end-users a quick way to
determine which version of AKS-engine to install on their system.
Example usage would be:

AKS_ENGINE_VER=$(curl
https://raw.githubusercontent.com/Azure/aks-engine/master/version)

Users could then automatically pull the latest released version's
tar.gz. Example usage could be:

```bash
curl -LO https://github.com/Azure/aks-engine/releases/download/${AKS_ENGINE_VER}/aks-engine-${AKS_ENGINE_VER}-linux-${ARCH}.tar.gz \
  && tar xvf aks-engine-${AKS_ENGINE_VER}-linux-${ARCH}.tar.gz \
  && mv aks-engine-${AKS_ENGINE_VER}-linux-${ARCH}/aks-engine /usr/local/bin/aks-engine \
  && rm aks-engine-${AKS_ENGINE_VER}-linux-${ARCH}.tar.gz
```

**Reason for Change**:

This improves aks-engine usability for automated jobs. At least it does for me!

**Issue Fixed**:
n/a

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

I fully appreciate this may be the lamest PR ever. The real work will be how this is tied into the aks-engine project's release process. Is this feasible and/or desired?  

Also, feel free to shout at me if there's already a better way to achieve same.


CecileRobertMichon:
@egernst what about something like this?

https://gist.github.com/lukechilds/a83e1d7127b78fef38c2914c4ececc3c

egernst:
Well, I did say *easy* way to grab version.  Will do some curl-redirect-fu.

mboersma:
I can see how this is useful, but I think something `git`-based like Cecile suggested would be simpler and more maintainable than updating a static file. (It's our fault the the release process is still manual, but as it is currently this would just be another detail for a human to miss.)

egernst:
I understand, regarding release processes. That's more or less what I'm trying to work out / automate on my end which started this.

Pulling from the GitHub release page itself is cumbersome and error prone AFAICT (if page format changes, which I believe has since that gist was put together).

Assuming the tags stay 'nice' and sorted going forward, I can just awk the remote:

```
git ls-remote --tags https://github.com/azure/aks-engine | awk -F ' ' '{print $2}' | cut -d/ -f3- | tail -n1
```

egernst:
worked around by pulling the tags associated with the aks-engine GitHub repository.

PR #2178, opened 2019-10-19
[WIP] test: add statefulset e2e test

jackfrancis:
**Reason for Change**:

Ref: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/workloads/statefulset.yaml
+++ b/test/e2e/kubernetes/workloads/statefulset.yaml
@@ -1,0 +1,35 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: web
+spec:
+  selector:
+    matchLabels:
+      app: nginx
+  serviceName: "nginx"
+  podManagementPolicy: "Parallel"
+  replicas: 3 # by default is 1
+  template:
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/workloads/statefulset.yaml
```diff
@@ -0,0 +1,35 @@
...
+kind: StatefulSet
+metadata:
+  name: web
+spec:
+  selector:
+    matchLabels:
+      app: nginx
+  serviceName: "nginx"
+  podManagementPolicy: "Parallel"
+  replicas: 3 # by default is 1
```
jackfrancis: note to self get rid of the inline comment

kkmsft: Is there a way in the aks-engine tests to scale up and down the replicas for this specific tests ?

devigned: Should be able to add that to the tests. @jackfrancis what do you think?

PR #2185, opened 2019-10-21
chore: update cluster-autoscaler to latest patch releases

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/autoscaler/releases. Some minor fixes are included in each patch.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -219,11 +219,14 @@ for TILLER_VERSION in ${TILLER_VERSIONS}; do
 done
 
 CLUSTER_AUTOSCALER_VERSIONS="
+1.16.2
 1.16.1
```
mboersma: Can we remove 1.16.1 (or others)?

PR #2188, opened 2019-10-21
chore: default to ubuntu distro for customHyperkubeImage

jackfrancis:
**Reason for Change**:

Set the distro to "ubuntu" by default so `customHyperkubeImage` scenarios don't have to do that manually.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -413,7 +413,7 @@ func (p *Properties) setExtensionDefaults() {
 
 func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName string) {
 	if p.MasterProfile.Distro == "" && p.MasterProfile.ImageRef == nil {
-		if p.OrchestratorProfile.IsKubernetes() && p.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage == "" {
+		if p.OrchestratorProfile.IsKubernetes() && p.OrchestratorProfile.KubernetesConfig != nil && p.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage == "" {
 			p.MasterProfile.Distro = AKSUbuntu1604
 		} else {
 			p.MasterProfile.Distro = Ubuntu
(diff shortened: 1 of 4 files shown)


pkg/api/defaults.go
```diff
@@ -653,6 +655,8 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool, cloudName
 				} else if profile.Distro == AKS1804Deprecated {
 					profile.Distro = AKSUbuntu1804
 				}
+			} else if p.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage != "" {
```
CecileRobertMichon: we don't want to override a non-empty distro otherwise the instructions in the doc above to set it to 18.04 won't work... this needs to be inside the `if profile.Distro == "" && profile.ImageRef == nil` if statement

pkg/api/defaults_test.go
```diff
@@ -3564,3 +3564,84 @@ func TestImageReference(t *testing.T) {
...
+					ImageRef: nil,
+				},
+			},
+		},
+		{
+			name: "custom hyperkube",
+			cs: ContainerService{
+				Properties: &Properties{
+					OrchestratorProfile: &OrchestratorProfile{
+						OrchestratorType: Kubernetes,
```
CecileRobertMichon: let's add a test case for custom hyperkube + ubuntu 18.04 distro to prove out my point above

pkg/api/defaults.go
jackfrancis: yep

pkg/api/defaults.go
```diff
@@ -424,6 +424,8 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade, isScale bool, cloudName
 		} else if p.MasterProfile.Distro == AKS1804Deprecated {
 			p.MasterProfile.Distro = AKSUbuntu1804
 		}
+	} else if p.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage != "" && p.MasterProfile.Distro == "" {
```
CecileRobertMichon: why not just put this in the if statement? I think all conditions above apply... If imageRef is set we don't want to set the distro here either

jackfrancis: not sure what you mean

CecileRobertMichon: I mean:

```
if p.MasterProfile.Distro == "" && p.MasterProfile.ImageRef == nil {
		if p.OrchestratorProfile.IsKubernetes() && p.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage != "" {
			p.MasterProfile.Distro = AKSUbuntu1604
		} else {
			p.MasterProfile.Distro = Ubuntu
		}
	} else if p.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) {
		if p.MasterProfile.Distro == AKSDockerEngine || p.MasterProfile.Distro == AKS1604Deprecated {
			p.MasterProfile.Distro = AKSUbuntu1604
		} else if p.MasterProfile.Distro == AKS1804Deprecated {
			p.MasterProfile.Distro = AKSUbuntu1804
		}
	}
```

CecileRobertMichon: All the unit tests you wrote should pass with this and it's more readable (the 3rd if is redundant otherwise) and it should cover all the logic we want (I think). If not, can you add a unit test case that passes in the current code that doesn't pass with the code above?

PR #2191, opened 2019-10-22
refactor: install kube- components without hyperkube for k8s 1.17+

### LINKED ISSUE #1842: Drop requirement for hyperkube
**Describe the request**
Hyperkube is being deprecated & removed in 1.17. This will break AKS-Engine deployments.

References:

- Thread asking for owners, people still using it https://groups.google.com/forum/#!msg/kubernetes-sig-architecture/4uTHr74ucv0/KImea_0RFAAJ
- https://github.com/kubernetes/kubernetes/pull/81268 - removing from GCE setup scripts

Use the tarballs published in Kubernetes releases and builds from PR tests

**Additional context**

mboersma:
**Reason for Change**:
Hyperkube is no more as of Kubernetes 1.17.0-alpha.2, so AKS Engine needs to provision its core `kube-` components and binaries from other official builds.

**Issue Fixed**:
Fixes #1842
Unblocks #2158 (yes, I tested)
Related to #2161 (`UseCloudControllerManager=true` is broken for k8s 1.16+)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Still TODO:

- [x] Update Linux VHD script
- [ ] ~~Update Windows VHD script~~
- [ ] ~~Other Windows changes still needed?~~

TODO in another PR, let's keep this one simple:
- [ ] Switch to (temporary) `upstream.azurecr.io/oss/kubernetes/` container repo when builds are working
- [ ] Switch to final MCR container repo when it's available and populated
- [ ] Remove 1.17 conditional clauses when back catalog is populated

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -6,6 +6,7 @@ CNI_CONFIG_DIR="/etc/cni/net.d"
 CNI_BIN_DIR="/opt/cni/bin"
 CNI_DOWNLOADS_DIR="/opt/cni/downloads"
 CONTAINERD_DOWNLOADS_DIR="/opt/containerd/downloads"
+K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
 
 removeEtcd() {
@@ -253,11 +254,22 @@ extractHyperkube() {
 
 installKubeletAndKubectl() {
     if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]]; then
-        if [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
-            extractHyperkube "docker"
+        if version_gte ${KUBERNETES_VERSION} 1.17; then  # don't use hyperkube
+            # TODO: move this URL var to install-dependencies.sh?
+            K8S_DOWNLOAD_URL="https://kubernetesreleases.blob.core.windows.net/k8s-staging/v${KUBERNETES_VERSION}/release-tars/kubernetes-server-linux-amd64.tar.gz"
+            K8S_TGZ_TMP=$(echo ${K8S_DOWNLOAD_URL} | cut -d "/" -f 7)
+            mkdir -p "${K8S_DOWNLOADS_DIR}"
+            retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${K8S_DOWNLOAD_URL} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
+            tar --transform="s|.*|&-${KUBERNETES_VERSION}|" --show-transformed-names -xzvf "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" \
+                --strip-components=3 -C /usr/local/bin kubernetes/server/bin/kubelet kubernetes/server/bin/kubectl
+            rm -f "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}"
         else
-            installImg
-            extractHyperkube "img"
+            if [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
+                extractHyperkube "docker"
+            else
+                installImg
+                extractHyperkube "img"
+            fi
         fi
     fi
     mv "/usr/local/bin/kubelet-${KUBERNETES_VERSION}" "/usr/local/bin/kubelet"
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -252,16 +252,19 @@ extractHyperkube() {
     fi
 }
 
+extractKubeBinaries() {
+    K8S_TGZ_TMP=$(echo ${KUBE_BINARY_URL} | cut -d "/" -f 5)
+    mkdir -p "${K8S_DOWNLOADS_DIR}"
+    retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${KUBE_BINARY_URL} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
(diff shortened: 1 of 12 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -395,11 +395,17 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     # Redirect ILB (4443) traffic to port 443 (ELB) in the prerouting chain
     iptables -t nat -A PREROUTING -p tcp --dport 4443 -j REDIRECT --to-port 443
 {{end}}
-
     sed -i "s|<img>|{{WrapAsParameter "kubernetesAddonManagerSpec"}}|g" /etc/kubernetes/manifests/kube-addon-manager.yaml
+{{if IsKubernetesVersionGe "1.17.0-alpha.1"}}
+    sed -i "s|<img>|{{WrapAsParameter "kubeAPIServerSpec"}}|g" /etc/kubernetes/manifests/kube-apiserver.yaml
+    sed -i "s|<img>|{{WrapAsParameter "kubeControllerManagerSpec"}}|g" /etc/kubernetes/manifests/kube-controller-manager.yaml
+    sed -i "s|<img>|{{WrapAsParameter "kubeSchedulerSpec"}}|g" /etc/kubernetes/manifests/kube-scheduler.yaml
+    sed -i "s|\"/hyperkube\", ||g" /etc/kubernetes/manifests/kube-*.yaml
```
jackfrancis: can we just use 1.17-specific yamls with no hyperkube reference so we can omit this sed replacement?

mboersma: Sure, that's another approach, although we don't follow the version -> subdirectory convention in `parts/k8s/manifests` currently, just in addons. 

I was thinking of this `sed` as a temporary hack because the end goal is to replace hyperkube usage in all supported versions, at which point we can just fix the command in the manifests and drop the `sed`.

jackfrancis: Yeah, all good with iteratation.

mboersma: I went with 1.17-specific yamls as you suggested, because I'm not sure we can make a wholesale change off of hyperkube. We expose `customHyperkubeImage` in KubernetesConfig, so there's a backward-compatibility issue there.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -247,11 +248,21 @@ extractHyperkube() {
 
 installKubeletAndKubectl() {
     if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]]; then
-        if [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
-            extractHyperkube "docker"
+        if version_gte ${KUBERNETES_VERSION} 1.17; then  # don't use hyperkube
+            K8S_DOWNLOAD_URL="https://dl.k8s.io/v${KUBERNETES_VERSION}/kubernetes-node-linux-amd64.tar.gz"
+            K8S_TGZ_TMP=$(echo ${K8S_DOWNLOAD_URL} | cut -d "/" -f 5)
+            mkdir -p "${K8S_DOWNLOADS_DIR}"
+            retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${K8S_DOWNLOAD_URL} || exit $ERR_K8S_DOWNLOAD_TIMEOUT
```
CecileRobertMichon: is the pre-download on VHD a todo?

mboersma: /hold

Yes it is a `// TODO:`, thanks for reviewing and keeping me honest.

mboersma: Fixed now, PTAL.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -245,13 +246,26 @@ extractHyperkube() {
     fi
 }
 
+extractKubeBinaries() {
+    K8S_TGZ_TMP=$(echo ${KUBE_BINARY_URL} | cut -d "/" -f 5)
```
CecileRobertMichon: this is called both by CSE at runtime and by the VHD to pre-install versions. We need a way to pre-install several versions of 1.17 (eg. 1.17.1 and 1.17.2) in the VHD and, at CSE runtime: 1) download the desired version if not present 2) if it's already present, use it without going out to the internet 3) remove all unused version binaries

mboersma: But when it's called in the `installKubeletAndKubectl` func below, it's guarded by a `[[ ! -f ]]` test that checks for the presence of the `kubectl-1.17.0` binary (for example). Only if the binary isn't already present do we go out to the internet to fetch it. The logic is parallel to how we use `extractHyperkube` so I think it's doing what we want.

Could be entirely mistaken, so apologies if I'm still off-base. Long day.

CecileRobertMichon: My bad, you're completely right 🙏  I read the code too fast and forgot that `[[ ! -f ]]` was in `installKubeletAndKubectl` and not `extractHyperkube`...
So just to confirm, this function will create `/usr/local/bin/kubelet-${KUBERNETES_VERSION}` and `/usr/local/bin/kubectl-${KUBERNETES_VERSION}`?

mboersma: Yes, it does, so the binaries in the same place with the same names that `extractHyperkube` would have put them and we can reuse that bit of logic.

(The `tar --transform=` argument is what renames the binaries while extracting them. It's a neat trick but it's hard to grok the resulting command line.)

CecileRobertMichon: Got it, in that case it's exactly what we want. 

VHD integration for kube binaries lgtm, let me know if you'd like another review of the whole PR.

mboersma:
The code coverage report here is incorrect: all the new lines of code are actually invoked by unit tests (I've checked in the debugger). It's penalizing us for `templates_generated.go` despite our best efforts to ignore that one.

parts/k8s/cloud-init/artifacts/cse_install.sh
mboersma: /hold cancel

PR #2196, opened 2019-10-22
fix: run all CSE apt-get purge operations in background

jackfrancis:
**Reason for Change**:

In order to improve the average CSE run, we should not block apt-get purge operations (which include waiting for apt locks that can be held by other concurrent processes)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@palmerabollo No, those errors are an inability to establish a TCP 443 connection to `aksrepos.azurecr.io` before bootstrap begins. That's an AKS requirement, and we're seeing a non-trivial number of errors for AKS Engine, which is driving this PR: https://github.com/Azure/aks-engine/pull/2210

PR #2210 will remove the outbound connectivity check to ACR entirely (which means CSE error 50 will no longer be a thing)

PR #2201, opened 2019-10-23
fix: Resolve incorrect Windows extension params

zachomedia:
**Reason for Change**:

This resolves an issue where the Windows extensions parameters were not being passed to the VM correctly.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

We were manually making this change in our ARM templates, but that didn't persist when running a cluster upgrade. This fix worked when upgrading that cluster from 1.15.4 -> 1.15.5.

Credit to @sylus for doing the research as to what the fix was.


zachomedia:
I've just pushed another commit which updates the failing test.

marosset:
I manually trigged the e2e pipeline
https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=2930&view=results

mboersma:
Thanks again @zachomedia!

PR #2204, opened 2019-10-23
feat: blang/semver and gofrs/uuid does not match module compatibility

### LINKED ISSUE #2203: blang/semver and gofrs/uuid does not match module compatibility
**Describe the request**

For packages that use go modules, importing aks-engine is failing because of two dependencies: [blang/semver](https://github.com/blang/semver) and [gofrs/uuid](https://github.com/gofrs/uuid) that don't satisfy [module compatibility](https://golang.org/cmd/go/#hdr-Module_compatibility_and_semantic_versioning) (version is 2.0 and above, and they don't have `go.mod`).

As more and more repos start using go modules, it would be great if aks-engine can be imported by those repos

1. **blang/semver**: starting version 3.6 moved to go modules, so all that is needed is an update of dependency to the latest version
2. **gofrs/uuid**: this package could easily be exchanged with [google/uuid package](https://github.com/google/uuid) which supports go mod 

**Additional context**

hichemaichour:
**Reason for Change**:

I just wanted to include the changes that would resolve the issue I just created, so here is how resolving that issue would be.

**Issue Fixed**:

Fixes #2203 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


hichemaichour:
It looks like the error I am getting is the same with other builds: [#20191024.2](https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=2921), [#20191022.18](https://dev.azure.com/aks-engine/AKS-Engine/_build/results?buildId=2893)
Should I assume it is not related to my changes? Is this a known issue?

mboersma:
@hichemaichour I think that's an unrelated flake, yes. I can't re-run those tests currently, but I consider this a pass.

I'm ok with these changes, but a major use case for AKS Engine is as a library called by the AKS service. So ensuring that the project can be vendored into that codebase is a high priority, and we'll need to test that first.

Also @devigned has taken on go modules support and lots of vendoring tasks, so his opinion on this should be more authoritative.

devigned:
/lgtm

Thank you! I'll pull this in.

PR #2207, opened 2019-10-23
feat: expose PlatformFaultDomainCount in the apimodel 

### LINKED ISSUE #2123: Expose Fault Domain count in the apimodel

serbrech:
same as https://github.com/Azure/aks-engine/pull/2127 minus the master VM Availability set change (https://github.com/Azure/aks-engine/pull/2127/commits/37a0f6ba4daaf90d4e0995b5d519c9fe0e872426)

- adds the possibility to define FDCount/agentpool	
- remove arm calls to fetch fdcount during upgrade/scale
- remove fdcount and AS from arm template completely during upgrade/scale

**Reason for Change**:
see #2123

**Issue Fixed**:
fixes #2123

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**: Need to pass back-compat


serbrech:
back compat take 1 : https://upstream.jenkins.azure-containers.io/view/aks-engine/job/k8s-back-compat-latest-release/186/console

pkg/api/types.go
```diff
@@ -554,7 +554,7 @@ type AgentPoolProfile struct {
 	MinCount                            *int                 `json:"minCount,omitempty"`
 	EnableAutoScaling                   *bool                `json:"enableAutoScaling,omitempty"`
 	AvailabilityZones                   []string             `json:"availabilityZones,omitempty"`
-	PlatformFaultDomainCount            *int                 `json:"platformFaultDomainCount,omitempty"`
+	PlatformFaultDomainCount            *int                 `json:"platformFaultDomainCount"`
```
serbrech: changed this to match MasterProfile : 
https://github.com/Azure/aks-engine/pull/2207/commits/7802a7af9118f48cfb4a460dd35e19b3d166c592#diff-9ff06ad723720f9428a65da3710cf436R481

serbrech:
back compat passed : https://upstream.jenkins.azure-containers.io/view/aks-engine/job/k8s-back-compat-latest-release/210/console

pkg/engine/virtualmachinescalesets.go
```diff
@@ -106,6 +106,9 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 
 	vmProperties := &compute.VirtualMachineScaleSetProperties{}
 
+	if masterProfile.PlatformFaultDomainCount != nil {
+		vmProperties.PlatformFaultDomainCount = to.Int32Ptr(int32(*masterProfile.PlatformFaultDomainCount))
```
mboersma: What does dereferencing and then casting back to an int32 pointer actually do here (as opposed to just passing through `*masterProfile.PlatformFaultDomainCount`?

serbrech: oh... int on one side int32 on the other side :/
so it takes the *int dereferences it to cast it to int32 and takes the pointer of that... pretty ugly

serbrech:
rebased on master and fixed the failing test, it was using a now unsupported version of k8s (1.10)

jackfrancis:
It looks like this PR as-is will reveal the bug behavior described here: https://github.com/Azure/aks-engine/issues/2240

How do we ensure that we don't change the FD on upgrade/scale?

It sounds like the current code does something wrong and "re-sets" the FD during scale and upgrade, which is an anti-pattern because the FD value is immutable.

What part of this PR will expose this underlying anti-pattern? Can we fix this in this PR?

jackfrancis:
Also, @serbrech, I ran upgrade/scale tests across various versions (from 1.12 to 1.13, from 1.15 to 1.16, etc) using this branch commit and didn't observe that this PR introduces/reveals any pre-existing bugs. Do we know how we can produce a scale/upgrade workflow that will show us how bad the bug is?

serbrech:
yes.
create a cluster with 2 vmss agent pool give them different FDCount, and run upgrade or scale.
that will fail

jackfrancis:
That makes sense, thank you. Your fix makes sense as well, thanks!

I also think this clarifies that it's not necessarily a "bug" in aks-engine, but (without your fixes) a known limitation in aks-engine after this change: per-pool FD count is not supported. (it's not even possible today)

So, I definitely prefer your fix, thank you! But an alternate approach could be to add validation error or aks-engine vlabs that ensures the same FD count is used across all pools.

Is all that correct?

pkg/engine/transform/transform.go
```diff
@@ -117,10 +117,6 @@ func (t *Transformer) RemoveImmutableResourceProperties(logger *logrus.Entry, te
 			resource.RemoveProperty(logger, platformFaultDomainCountFieldName)
 			resource.RemoveProperty(logger, singlePlacementGroupFieldName)
 		}
-
-		if resource.Type() == vmasResourceType {
```
serbrech: we remove the whole resource, no need to remove a property from vmas

pkg/engine/transform/transform.go
```diff
@@ -380,6 +376,9 @@ func (t *Transformer) removeImageReference(logger *logrus.Entry, resourcePropert
 // NormalizeResourcesForK8sMasterUpgrade takes a template and removes elements that are unwanted in any scale up/down case
 func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry, templateMap map[string]interface{}, isMasterManagedDisk bool, agentPoolsToPreserve map[string]bool) error {
 	resources := templateMap[resourcesFieldName].([]interface{})
+	resourceTypeToProcess := map[string]bool{vmResourceType: true, vmExtensionType: true,
```
serbrech: extracting the if or list into an appropriately named map for readability

pkg/engine/transform/transform.go
```diff
@@ -398,7 +397,14 @@ func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry
...
 		}
 
-		if !(resourceType == vmResourceType || resourceType == vmExtensionType || resourceType == nicResourceType || resourceType == vnetResourceType || resourceType == nsgResourceType || resourceType == lbResourceType || resourceType == vmssResourceType) {
+		_, process := resourceTypeToProcess[resourceType]
+		if !process {
+			continue
+		}
+
+		// never include vmas resource in upgrade arm template
+		if strings.EqualFold(resourceType, vmasResourceType) {
```
serbrech: removing vmas resource types

PR #2209, opened 2019-10-23
[WIP] feat: command to add a single agent pool to an existing cluster 

DavidLangworthy:
This is a work in progress. 

This adds an "addpool" command to aks-engine that takes a single agent pool spec as a JSON file and deploys it to an existing cluster.  It is basically copies from scale on Jack's excellent suggestion.


damoodamoo:
Out of interest - would this also work for an arbitrary Kubernetes cluster deployed to Azure (not deployed using AKS-Engine but another tool, such as KubeSpray) ? Or does this only apply to clusters initially deployed using AKS-Engine?

DavidLangworthy:
The short answer is no.  It uses the apimodel that is input to aks-engine and does stuff with that.  

You could use the same approach with a different tool.  At the bottom all this does is deploy an azure vm scale set and run the setup scripts on them.  There is nothing magic.

damoodamoo:
Ok cheers - and it wouldn’t be possible to generate an apimodel for a non AKS-Engine deployed cluster? If not would you be able to point me to where the setup scripts are in the repo so I could investigate reusing them?

DavidLangworthy:
There is a bunch of stuff under here: https://github.com/Azure/aks-engine/tree/master/parts/k8s.  It gets pulled together by packer, then aks-engine picks out the right stuff, fills in parameters, and injects it in the right places.  I've kind of figured out what it's doing by reading the generated ARM template.

My profile should show that I'm at Microsoft, but I'm not on the AKS team.  So that's about the limit of my knowledge.

mtparet:
@jackfrancis @DavidLangworthy any idea if we can get this one day ?

We are stuck with a cluster with too small disks and we cannot afford recreate (a another cluster...  (almost the same cost than moving to GCP)

DavidLangworthy:
@mtparet You can add a pool manually by adding it to your apimodel.json, regenerating, and redeploying.  If you don't change anything else, it wont change your existing deployment, it will just add the new pool.  You can do this with something like 12 pools.

pierluigilenoci:
Why closed?

DavidLangworthy:
@pierluigilenoci   Because I didn't have time to finish before stale bot.

pierluigilenoci:
@jackfrancis could you please reopen the PR?

DavidLangworthy:
I opened a new one.  #2557

pierluigilenoci:
Cool!

PR #2213, opened 2019-10-24
test: skip tests when block ssh port enabled

aramase:
**Reason for Change**:

- Set `longRunningApacheDeploymentName` in BeforeSuite even if `blockSSHPort` is true as it's used by tests that don't use ssh.
- Skip `should be able to access the dashboard` as it requires ssh connection. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -89,6 +89,8 @@ var _ = BeforeSuite(func() {
 		ClusterDefinition:  csInput,
 		ExpandedDefinition: csGenerated,
 	}
+	longRunningApacheDeploymentName = "php-apache-long-running"
+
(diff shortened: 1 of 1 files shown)


CecileRobertMichon:
@aramase can you also add https://github.com/Azure/aks-engine/commit/26f1c7a3b6603001fee9532525d9dbed3cd8df50?
 otherwise lgtm

aramase:
@CecileRobertMichon I don't think we need to skip that test anymore. It was failing because `longRunningApacheDeploymentName` was not set before if block ssh port was true. But that should be fixed with this PR.

```
$ k get deploy  -n default -o json
2019/10/24 00:25:35 Error getting deployment: exit status 1
```

PR #2231, opened 2019-10-28
fix: Ensured identity system passed from cli is propagated to api model

### LINKED ISSUE #2054: Identitiy system not propagated from cli parameters
**Describe the bug**
When the identity system is specified from the cli, It does not get included in the final template generated where there is no identity system specified in the in the api model. It often results in the use of the default identity system (azure_ad) regardless of what is passed in the cli. 

**Steps To Reproduce**

- In the API model, do not specify the identity system
- Within when deploying the cluster using deploy command, specify the identity system as "adfs". 
- Check the generated arm template and notice the identity system set to "azure_ad"

**Expected behavior**
Where there is an Identity system specified in the cli, it should be used as the identity system when deploying the Kuberentes cluster. 

**AKS Engine version**
0.41.0
**Kubernetes version**
1.11
**Additional context**

deaborch:
**Reason for Change**:

Ensures that in cases where the cloud profile in the api model has not specified an identity system, the identity system passed in the cli will be used before using defaults. Issue only happens in deploy command. 

**Issue Fixed**:

Fixes #2054

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -220,6 +220,10 @@ func (dc *deployCmd) loadAPIModel() error {
 		return errors.New("--location does not match api model location")
 	}
 
+	if dc.containerService.Properties.CustomCloudProfile.IdentitySystem == "" {
+		dc.containerService.Properties.CustomCloudProfile.IdentitySystem = dc.authProvider.getAuthArgs().IdentitySystem
+	}
+
 	if dc.containerService.Properties.IsAzureStackCloud() {
 		err = dc.containerService.SetCustomCloudProfileEnvironment()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


devigned:
@deaborch you got a test for that?

deaborch:
@devigned just taken care off

devigned:
@deaborch maybe I'm missing it, but where in the test does it check the `IdentitySystem`? Is it verified indirectly.

deaborch:
@devigned I see what you mean, the previous test is just to ensure that the code does not break the `loadAPIModel()` method. will update the PR to check to see if the identity system is altered correctly

PR #185, opened 2018-12-19
fix: update MSI

ritazh:
**What this PR does / why we need it**:
- Remove MSI Extension since it will be deprecated in Jan 2019 and it is not available in sovereign clouds
- Create separate apiVersionAuthorization to support both user assigned and system assigned identity for clouds that do not have the latest apis: `apiVersionAuthorizationUser` and `apiVersionAuthorizationSystem`
- Add VMs to dependsOn for each role assignment if it's system assigned identity
- Bump version of kms plugin to support instance metadata service identity endpoint 

**Which issue this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*: fixes #175

**Special notes for your reviewer**:

**If applicable**:
- [ ] documentation
- [ ] unit tests
- [ ] tested backward compatibility (ie. deploy with previous version, upgrade with this branch)

**Release note**:

```release-note
Update MSI to remove MSI extension and add support for older api versions of role assignment
```

PR #2238, opened 2019-10-30
feat: support using existing identity for addons

### LINKED ISSUE #2233: Support associating existing user assigned identity to nodes
**Describe the request**
Support associating existing user assigned identity to nodes. Currently AKS-Engine only supports creating a new one when creating the cluster.
We're working on switching all addons in AKS to use user assigned identity to authenticate, and AKS RP will create one user assigned identity for each addon. After this, we need to associate the identity to every agent node in the cluster, it would be very convenient if AKS-Engine supports associating existing identity to agent nodes.
A new struct containing the resource id of existing user assigned identities will be added to “ContainerService” and specified identities will be associated to every (agent) node in generated template. 
Associating identities can also be done by AKS RP, but this way has a big shortcoming: when upgrading the cluster, for a newly created VM, identities will not be associated with it. So if addon pod is scheduled to a newly created VM, it can't work correctly for a period of time.
**Additional context**

norshtein:
**Reason for Change**:

Detailed background can be found in #2233 

**Issue Fixed**:

Fixes #2233 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
 - Though AKS will only generate agent nodes' ARM template, I associate addons' identity to master VM(SS) to keep consistent with agent nodes' behaviror.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -109,6 +109,16 @@ type ClusterMetadata struct {
 type AddonProfile struct {
 	Enabled bool              `json:"enabled"`
 	Config  map[string]string `json:"config"`
+	// Identity contains information of the identity associated with this addon.
+	// This property will only appear in an MSI-enabled cluster.
+	Identity *UserAssignedIdentity `json:"identity,omitempty"`
+}
+
+// UserAssignedIdentity contains information that uniquely identifies an identity
+type UserAssignedIdentity struct {
+	ResourceID string `json:"resourceId,omitempty"`
+	ClientID   string `json:"clientId,omitempty"`
+	ObjectID   string `json:"objectId,omitempty"`
 }
 
 // FeatureFlags defines feature-flag restricted functionality
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -18,6 +18,7 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 	hasAvailabilityZones := cs.Properties.MasterProfile.HasAvailabilityZones()
 	isStorageAccount := cs.Properties.MasterProfile.IsStorageAccount()
 	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
+	addonProfiles := cs.Properties.AddonProfiles
 
 	var useManagedIdentity, userAssignedIDEnabled bool
 	if kubernetesConfig != nil {
@@ -88,6 +89,21 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 			identity.Type = compute.ResourceIdentityTypeSystemAssigned
 		}
 		virtualMachine.Identity = identity
+	}
+
+	for _, addonProfile := range addonProfiles {
+		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
+			if virtualMachine.Identity == nil {
+				virtualMachine.Identity = &compute.VirtualMachineIdentity{}
+				virtualMachine.Identity.Type = compute.ResourceIdentityTypeUserAssigned
+				virtualMachine.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+			} else if virtualMachine.Identity.Type == compute.ResourceIdentityTypeSystemAssigned {
+				virtualMachine.Identity.Type = compute.ResourceIdentityTypeSystemAssignedUserAssigned
+				virtualMachine.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+			}
+
+			virtualMachine.Identity.UserAssignedIdentities[addonProfile.Identity.ResourceID] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
+		}
 	}
 
 	vmProperties := &compute.VirtualMachineProperties{}
@@ -321,6 +337,7 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 	isStorageAccount := profile.IsStorageAccount()
 	hasDisks := profile.HasDisks()
 	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
+	addonProfiles := cs.Properties.AddonProfiles
 
 	var useManagedIdentity, userAssignedIDEnabled bool
 
@@ -402,6 +419,21 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 			virtualMachine.Identity = &compute.VirtualMachineIdentity{
 				Type: compute.ResourceIdentityTypeSystemAssigned,
 			}
+		}
+	}
+
+	for _, addonProfile := range addonProfiles {
+		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
+			if virtualMachine.Identity == nil {
+				virtualMachine.Identity = &compute.VirtualMachineIdentity{}
+				virtualMachine.Identity.Type = compute.ResourceIdentityTypeUserAssigned
+				virtualMachine.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+			} else if virtualMachine.Identity.Type == compute.ResourceIdentityTypeSystemAssigned {
+				virtualMachine.Identity.Type = compute.ResourceIdentityTypeSystemAssignedUserAssigned
+				virtualMachine.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+			}
+
+			virtualMachine.Identity.UserAssignedIdentities[addonProfile.Identity.ResourceID] = &compute.VirtualMachineIdentityUserAssignedIdentitiesValue{}
 		}
 	}
 
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -607,6 +607,9 @@ func addCustomTagsToVM(tags map[string]string, vm *compute.VirtualMachine) {
 }
 
 func associateAddonIdentitiesToVM(addonProfiles map[string]api.AddonProfile, virtualMachine *compute.VirtualMachine) {
+	if virtualMachine == nil {
+		return
+	}
 	for _, addonProfile := range addonProfiles {
 		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
 			// We need to associate addon's identity to VM, there're 3 cases:
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -18,7 +18,6 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
 	hasAvailabilityZones := cs.Properties.MasterProfile.HasAvailabilityZones()
 	isStorageAccount := cs.Properties.MasterProfile.IsStorageAccount()
 	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
-	addonProfiles := cs.Properties.AddonProfiles
 
 	var useManagedIdentity, userAssignedIDEnabled bool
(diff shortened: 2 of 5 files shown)


norshtein:
I noticed when managed identity is enabled, one extension will be installed on master VMSS(there is no such extension in agent VMSS): https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachinescalesets.go#L274-L289 . What's the purpose of the extension? @jackfrancis

pkg/api/types.go
```diff
@@ -109,6 +109,16 @@ type ClusterMetadata struct {
 type AddonProfile struct {
 	Enabled bool              `json:"enabled"`
 	Config  map[string]string `json:"config"`
+	// Identity contains information of the identity associated with this addon.
+	// This property will only appear in an MSI-enabled cluster.
+	Identity *UserAssignedIdentity `json:"identity,omitempty"`
```
kkmsft: The add-on facility used in aks-engine is different from the add-on facility used in AKS. Hence I am not sure how this can be used in AKS. I was expecting the facility to be more generic and be VMSS/VMA template specific. What is the missing piece piece connecting this ?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -440,6 +457,21 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		}
 	}
 
+	for _, addonProfile := range addonProfiles {
+		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
+			if virtualMachineScaleSet.Identity == nil {
+				virtualMachineScaleSet.Identity = &compute.VirtualMachineScaleSetIdentity{}
+				virtualMachineScaleSet.Identity.Type = compute.ResourceIdentityTypeUserAssigned
+				virtualMachineScaleSet.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue)
```
kkmsft: Can we make it a single step by using an initialization which is already filled with Type and UserAssignedIdentities ?

pkg/engine/virtualmachines.go
```diff
@@ -405,6 +422,21 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 		}
 	}
 
+	for _, addonProfile := range addonProfiles {
+		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
+			if virtualMachine.Identity == nil {
+				virtualMachine.Identity = &compute.VirtualMachineIdentity{}
+				virtualMachine.Identity.Type = compute.ResourceIdentityTypeUserAssigned
+				virtualMachine.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineIdentityUserAssignedIdentitiesValue)
+			} else if virtualMachine.Identity.Type == compute.ResourceIdentityTypeSystemAssigned {
```
kkmsft: Aren't there other possibilities such as:
1. Identity.Type is None ?
2. Identity.Type is already ResourceIdentityTypeUserAssigned? 

Can we please handle/add comments on how those scenarios works ?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -98,6 +99,21 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 		virtualMachine.Identity = identity
 	}
 
+	for _, addonProfile := range addonProfiles {
+		if addonProfile.Enabled && addonProfile.Identity != nil && addonProfile.Identity.ResourceID != "" {
+			if virtualMachine.Identity == nil {
+				virtualMachine.Identity = &compute.VirtualMachineScaleSetIdentity{}
```
kkmsft: Is there a possibility to move the Identity handling to a utility function and reuse it in VMA and VMSS case ?

ritazh:
> one extension will be installed on master VMSS(there is no such extension in agent VMSS)

The `ManagedIdentityExtensionForLinux` extension was deprecated in Jan 2019 and was removed in this PR: https://github.com/Azure/aks-engine/pull/185. But it was added back in a subsequent PR: https://github.com/Azure/aks-engine/blame/master/pkg/engine/virtualmachinescalesets.go#L274-L289

pkg/api/types.go
norshtein: When generating ARM template, AKS RP will also pass `aksEngineAPI.AddonProfile` to AKS-Engine. After adding this in AKS-Engine, AKS RP will add the resourceID of created identity in `AddonProfile.Identity` so that AKS-Engine can associate the identity to every agent node.

pkg/engine/virtualmachines.go
norshtein: Thanks for reminding, will add comment.

pkg/api/types.go
kkmsft: Was wondering won't the deployment of the template add the add-on enabled to add-on manager and then add-on manager pull the corresponding images when you deploy the template ? Wouldn't that interfere with the helm installation of add-on in case of AKS ?

pkg/engine/virtualmachinescalesets.go
norshtein: Thanks for reminding, updated.

pkg/engine/virtualmachinescalesets.go
norshtein: Thanks for reminding, it makes sense, updated.

norshtein:
> > one extension will be installed on master VMSS(there is no such extension in agent VMSS)
> 
> The `ManagedIdentityExtensionForLinux` extension was deprecated in Jan 2019 and was removed in this PR: #185. But it was added back in a subsequent PR: https://github.com/Azure/aks-engine/blame/master/pkg/engine/virtualmachinescalesets.go#L274-L289
> 
> This should probably be removed.

Thanks for the info. Just removed the extension.

pkg/api/types.go
norshtein: > wouldn't the deployment of the template add the enabled aks-engine add-ons to add-on manager and then pull the corresponding images when deployed?

I guess not? I haven't seen such code in AKS-Engine. The deployment of the template will do nothing related to addon. And seems in current AKS-Engine, it just ignores all `AddonProfile`. @jackfrancis @ritazh  could you please help confirm? If I understand this correctly, the management of all addons is fully controlled by AKS RP.

norshtein: Seems AKS's addon is recorded in `ContainerService.Properties.AddonProfiles`, AKS-Engine's addon is recorded in `ContainerService.Properties.OrchestratorProfile.KubernetesConfig.Addons`. Currently AKS-Engine will do nothing about addons in `ContainerService.Properties.AddonProfiles`.

norshtein:
Hey @jackfrancis , @yangl900 , could you please have a review?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -271,23 +273,6 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 
 	var extensions []compute.VirtualMachineScaleSetExtension
 
-	if useManagedIdentity {
```
jackfrancis: Why are we getting rid of this? This breaks the following cluster configuration scenario:

```
cs.Properties.OrchestratorProfile.KubernetesConfig.UseManagedIdentity = true
cs.Properties.OrchestratorProfile.KubernetesConfig.UserAssignedID = "fooAssignedID"
```

Am I missing something?

@ritazh @sozercan Am I correct in recalling that one of you introduced MSI into aks-engine?

sozercan: Wasn't me, but I believe the extension is deprecated in favor of IMDS.
https://docs.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/howto-migrate-vm-extension

jackfrancis: Thanks @sozercan, that suggests this deletion actually fixes a bug? Should we do that as a separate PR so we can track any regressions distinctly?

sozercan: missed this, looks like @ritazh commented earlier: https://github.com/Azure/aks-engine/pull/2238#issuecomment-548163703

jackfrancis: Ah, thank you @ritazh and @sozercan, I should read the comment history in addition to the changeset. :)

I think ideally this would have been fixed in a distinct PR, but we don't have to be strict here. Looks like a good change that helped to clarify the scope of work here.

pkg/engine/virtualmachines.go
```diff
@@ -601,3 +605,33 @@ func addCustomTagsToVM(tags map[string]string, vm *compute.VirtualMachine) {
 		}
 	}
 }
+
+func associateAddonIdentitiesToVM(addonProfiles map[string]api.AddonProfile, virtualMachine *compute.VirtualMachine) {
```
jackfrancis: We need UT for this new func. Please add a test case for a nil `addonProfiles` and `virtualMachine`. Which means that we'll need to exit the func right away if a nil `virtualMachine` input is passed in.

pkg/engine/virtualmachinescalesets.go
```diff
@@ -828,3 +815,33 @@ func addCustomTagsToVMScaleSets(tags map[string]string, vm *compute.VirtualMachi
 		}
 	}
 }
+
+func associateAddonIdentitiesToVMSS(addonProfiles map[string]api.AddonProfile, virtualMachineScaleSet *compute.VirtualMachineScaleSet) {
```
jackfrancis: ditto above comment about UT

pkg/api/types.go
ritazh: I have a similar question as @kkmsft. The changes you are making updates the aks-engine addons, which will be deployed by the add-on manager. How does this work with AKS addons? Can you pls verify this?

jackfrancis: The `AddonProfile` struct is actually not used by aks-engine. aks-engine uses the `KubernetesAddon` struct which is composed as a child property of `KubernetesConfig`.

pkg/engine/virtualmachines.go
norshtein: Thanks for reviewing! added unit test.

pkg/engine/virtualmachinescalesets.go
norshtein: added :D

norshtein:
Hi @jackfrancis , is there any ETA of a new release containing this PR?

norshtein:
Rebased upon latest master.

pkg/engine/virtualmachinescalesets.go
```diff
@@ -828,3 +815,36 @@ func addCustomTagsToVMScaleSets(tags map[string]string, vm *compute.VirtualMachi
...
+			if virtualMachineScaleSet.Identity == nil {
+				virtualMachineScaleSet.Identity = &compute.VirtualMachineScaleSetIdentity{
+					Type:                   compute.ResourceIdentityTypeUserAssigned,
+					UserAssignedIdentities: make(map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue),
+				}
+			} else if virtualMachineScaleSet.Identity.Type == compute.ResourceIdentityTypeSystemAssigned {
+				virtualMachineScaleSet.Identity.Type = compute.ResourceIdentityTypeSystemAssignedUserAssigned
+				virtualMachineScaleSet.Identity.UserAssignedIdentities = make(map[string]*compute.VirtualMachineScaleSetIdentityUserAssignedIdentitiesValue)
+			} else if virtualMachineScaleSet.Identity.Type == compute.ResourceIdentityTypeNone {
+				// Note: in current AKS-Engine's implementation, we will never enter into this branch. Just handle it here in case implementation
```
xuto2: late comment: why not return error in this case?

xuto2: at least we should log some message?

norshtein: Yeah, logging a warning here is better, but there is no logger available in current context. As for return an error, as the function name `associateAddonIdentitiesToVMSS` indicates, this function should associate identities to VMSS, if `virtualMachineScaleSet.Identity.Type` is `None`(which will never happen in current AKS-Engine's implementation but may happen later), instead of returning error, just change `virtualMachineScaleSet.Identity.Type` to `UserAssigned` and associate addons' identities to the VMSS makes more sense.

PR #2262, opened 2019-11-05
fix: move e2e runner into a module

devigned:
**Reason for Change**:
The e2e runner is causing instability with go modules when running `go run ./tests/e2e/runner.go`. Really, the code under test should be relegated to only be used in testing scenarios, so it should be broken into its own module.

This PR marks all of the test runner code with build tag "test" and creates an `github.com/Azure/aks-engine/tests/e2e` module which has a dependency on the main aks-engine module. 

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


devigned:
for e2e test validation, see Jenkins: k8s-create-cluster-matrix-1.13-gh2262

PR #2269, opened 2019-11-05
fix: set DefaultOutboundRuleIdleTimeoutInMinutes in VMSS defaults

chewong:
**Reason for Change**:

OutboundRuleIdleTimeoutInMinutes is not set to DefaultOutboundRuleIdleTimeoutInMinutes when availability zone is used and loadBalancerSku & outboundRuleIdleTimeoutInMinutes were not specified by the API model, causing the following ARM error: 
```
LBOutboundRule has invalid Idle Timeout. The value must be between 4 and 120
```
The following condition is false because loadBalancerSku is an empty string under the conditions described above, causing OutboundRuleIdleTimeoutInMinutes to be 0.
https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults.go#L374-L377

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/defaults.go
```diff
@@ -563,6 +563,9 @@ func (p *Properties) setVMSSDefaultsForMasters() {
 	if p.MasterProfile.HasAvailabilityZones() && (p.OrchestratorProfile.KubernetesConfig != nil && p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "") {
 		p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = StandardLoadBalancerSku
 		p.OrchestratorProfile.KubernetesConfig.ExcludeMasterFromStandardLB = to.BoolPtr(DefaultExcludeMasterFromStandardLB)
+		if p.OrchestratorProfile.KubernetesConfig.OutboundRuleIdleTimeoutInMinutes == 0 {
```
jackfrancis: This defaults code flow is only entered in when the master node(s) is/are VMSS. Is that what we want?

jackfrancis: I suppose because we're doing it in the `setVMSSDefaultsForAgents` flow below, we're covering both vectors. Maybe my question is: is this a VMSS-specific default?

jackfrancis:
So reading through this it does seem like the `OutboundRuleIdleTimeoutInMinutes` only comes into play via this func:

`pkg/engine/loadbalancers.go-->CreateAgentLoadBalancer()`

Which is only called for Standard LB. So I think the code as-is makes sense. What's the api model input that's producing this ARM error?

chewong:
> So reading through this it does seem like the `OutboundRuleIdleTimeoutInMinutes` only comes into play via this func:
> 
> `pkg/engine/loadbalancers.go-->CreateAgentLoadBalancer()`
> 
> Which is only called for Standard LB. So I think the code as-is makes sense. What's the api model input that's producing this ARM error?

https://github.com/kubernetes-sigs/cloud-provider-azure/blob/master/tests/k8s-azure/manifest/linux-vmss-multi-zones.json

pkg/api/defaults.go
chewong: It is the default value when a standard load balancer is used. Since VMSS with availability zone uses a standard load balancer, I just reuse the same value.

PR #2275, opened 2019-11-06
refactor: recreate missing masters before upgrading

sylr:
**Reason for Change**:

Assume that during the upgrade of a master the recreation fails after the delete then if you run the upgrade again it will upgrade the next master before trying to recreate the one which is missing.

If your 2nd upgrade fails you will end up with 1 master left (in case you have 3 masters).

Starting by recreating missing masters before continuing upgrading the next ones ensures you are not left with too few masters, if not none.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
I manually tested this change in lots of ways, thank you so much, this re-ordering makes sense.

PR #2276, opened 2019-11-06
chore: remove support for Kubernetes 1.11 and 1.12

mboersma:
**Reason for Change**:
These old versions are now out of Kubernetes' support policy (so is 1.13), and therefore out of AKS Engine's support matrix.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I removed both versions in one PR because I'm told that's how AKS will do it. But if we need to deprecate each minor separately, the commits are distinct.


mboersma:
/hold

This shouldn't be merged until AKS has officially removed support for 1.11 and 1.12, to be safe.

PR #2282, opened 2019-11-07
fix: addons "update config" is upgrade-only

jackfrancis:
**Reason for Change**:

We are unnecessarily re-configuring the api model addons spec on both scale _and_ upgrade operations. We should only be doing this for upgrade operations, as scale operations do not affect the master vms, and thus do not change the control plane implementation.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -377,7 +377,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		// First, Configure addons
-		cs.setAddonsConfig(isUpdate)
+		cs.setAddonsConfig(isUpgrade)
 		// Defaults enforcement flows below inherit from addons configuration,
 		// so it's critical to enforce default addons configuration first
 
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -13,7 +13,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
-func (cs *ContainerService) setAddonsConfig(isUpdate bool) {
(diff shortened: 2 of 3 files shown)


pkg/api/defaults.go
```diff
@@ -377,7 +377,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		// First, Configure addons
-		cs.setAddonsConfig(isUpdate)
+		cs.setAddonsConfig(isUpgrade)
```
CecileRobertMichon: was this maybe done to ensure back compat?

jackfrancis: Passing in upgrade context is indeed to ensure back-compat, and to signal forward-compat transformations. The thing is: back-compat requirements are not a factor for scale operations w/ addons spec configs, because scale is node-only, and addons configurations are control plane-only. See L93 above which sets the `isUpdate` var based on this evaluation:

`isUpdate := isUpgrade || isScale`

What I'm proposing in this PR is that passing `isUpdate` into `setAddonsConfig` is a bug, we should be passing `isUpgrade`.

CecileRobertMichon:
do we also need to update the unit tests https://github.com/Azure/acs-engine/pull/4224/files#diff-daf624fb49e7cf74e670db52d41e949a ?

can you run a backcompat scale test on this (with a release before PR 4224 merged) just to make sure we're not missing anything?

jackfrancis:
@CecileRobertMichon in fact the proof that this is the right thing is in *not* changing the UT and validating everything is still green.

Issue #2146, opened 2019-10-11
CoreDNS custom domains not working


SuperMarioo:
**Describe the bug**

I have followed this documentation to add custom domain

https://github.com/MicrosoftDocs/azure-docs/blob/master/articles/aks/coredns-custom.md

```apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system
data:
 mycompany.com.server: |
    mycompany.com:53 {
        errors
        cache 30
        proxy . 192.11.0.1  
    }
```
Pods are unable to resolve custom domain .
curl: (6) Could not resolve host:

**Steps To Reproduce**

kubectl apply -f corednsms.yaml
kubectl delete pod --namespace kube-system --label k8s-app=kube-dns

Exec  to pod and curl http URL with custom domain

**Expected behavior**

custom domain should be resolved 

**AKS Engine version**

v0.41.0

**Kubernetes version**

1.15.4

**Additional context**

CoreDNS-1.6.2 version 

Logs from CoreDNS:

[WARNING] No files matching import glob pattern: conf.d/Corefile*
[WARNING] No files matching import glob pattern: conf.d/Corefile*
[WARNING] No files matching import glob pattern: conf.d/Corefile*

devigned:
Hi @SuperMarioo, thx for reaching out. Let's see if we can get this resolved.

Per coredns, I don't think there is anything special going on here. That is to say, we don't have any special Azure configuration happening here, so the main docs for coredns might be more helpful to you: https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/.

Is you DNS server reachable at `192.11.0.1`?
```
$ nslookup microsoft.com 192.11.0.1
;; connection timed out; no servers could be reached
```
This also might help: https://stackoverflow.com/questions/55612141/resolve-custom-dns-in-kubernetes-cluster-aks

SuperMarioo:
Hi @devigned , thx for reply . 

DNS Server is reachable 

I have tried your links but I'm still unable to resolve custom DNS

jjburr:
if you look at your CoreDNS config does it get overwritten? 

If so you will have to switch reconcile with ensureexists on /etc/kubernetes/addons/coredns.yml otherwise it will keep getting reconciled with the default coredns settings.

jackfrancis:
Fixed in #2288

PR #2288, opened 2019-11-08
fix: Modified coredns.yaml to include import parameters

evillgenius75:
**Reason for Change**:

Currently coreDNS is set to auto-reconcile and any customization such as override parameters or stub zone proxy customization will not work in AKS-Engine as the Corefile is reset to default config map settings. Adding the import additional parameters can be added by using a coredns-custom config map with the following parameters:
i.e. Rewrite DNS
```
 apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system
data:
  test.server: |
    <domain to be rewritten>.com:53 {
        errors
        cache 30
        rewrite name substring <domain to be rewritten>.com default.svc.cluster.local
        forward .  /etc/resolv.conf # you can redirect this to a specific DNS server such as 10.0.0.10
    }
```
i.e. CoreDNS Hosts plugin using overrride
```apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom # this is the name of the configmap you can overwrite with your changes
  namespace: kube-system
data:
    test.override: |
          hosts example.hosts example.org { # example.hosts must be a file
              10.0.0.1 example.org
              fallthrough
          }
```

**Issue Fixed**:

#2146 and #1541 

**Requirements**:

- [ x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -74,7 +74,9 @@ data:
         loop
         reload
         loadbalance
+        import custom/*.override
     }
+    import custom/*.server
 ---
 apiVersion: v1
 kind: ConfigMap


jackfrancis:
Thank you @evillgenius75 !

CecileRobertMichon:
@jackfrancis @evillgenius75 squash commit message needs to amended to follow release note format.

jackfrancis:
Re-running E2E because the one Windows fail signal was a DNS error.

PR #2293, opened 2019-11-09
chore: improve build through parallel tests and stable tools

devigned:
**Reason for Change**:
- Builds take too long and thought it would be nice to start making them run in parallel. Started down the path of doing this with `ginkgo`, but noticed our regular old go tests were actually being run `N` times where `N` is the num_cpus (https://github.com/onsi/ginkgo/issues/618). So, I decided to remove some of the ginkgo stuff we are using and start moving more to regular old golang tests.
- Also took a moment to add `make reload` and `make clean` to the tools directory so that folks can more easily refresh the version of our build tools they are using. 

Well, things have gotten a bit out of hand... I ended up finding the source for over 3/4 of the build time, generating PKI. We so often generate 4096 bit keys that most of our build time is taken up by that activity.

I had started off down a path replacing them through out the tests, but found a more elegant solution, introducing a build tag to use a smaller default key size. The link below shows where the helper will respond to the `-tags=fast` golang build tag which replaces 4096 keys with 512 keys. This change on its own would probably take the unit tests from 35+m to less than 12m.

```go
//+build fast
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.

 package helpers

 const (
 	// SSHKeySize is the size (in bytes) of SSH key to create (only used when `-tags fast` is used)
 	SSHKeySize = 512

 	// DefaultPkiKeySize is the default size in bytes of the PKI key (only used when `-tags fast` is used)
 	DefaultPkiKeySize = 512
 )
```
https://github.com/Azure/aks-engine/pull/2293/files#diff-db3d36d8c86a42b9418b82d5b0b933cc

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


devigned:
Unit tests passed in 11m 49s

PR #2297, opened 2019-11-11
ci: enforce shell script formatting

mboersma:
**Reason for Change**:
Enforces a minimal-whitespace style for shell scripts using the `shfmt` tool that's already in our toolbox. Run `make validate-shell` to check. (To format instead of just checking, change the `-d` flag to `-w` in `validate-shell.sh`.)

If we're happy with this tool and this approach, it can be used to minify shell scripts as well should that be required.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
My 2¢ are that we should defer this work until after we fully convert the CSE surface area to cluster configuration-context-sensitive go templating. And then UT against customData payloads for all supported cluster configurations, to asses the need for this type of surgical approach.

PR #2298, opened 2019-11-11
fix: extract kubelet and kubectl binaries from hyperkube's binary folder when possible

chewong:
**Reason for Change**:

With https://github.com/kubernetes/kubernetes/pull/84662, we can no longer obtain kubelet and kubectl binaries by renaming the hyperkube binary. Instead, we should extract them through hyperkube's binary folder when possible.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
See also #2191, which is necessary for 1.17.0-alpha.2 and later, but is currently blocked until #2161 merges.

chewong:
@mboersma thanks for the info. Are we still planning to support `customHyperkubeImage` after when provided?

mboersma:
> Are we still planning to support `customHyperkubeImage`...?

It's a good question, since there is no official hyperkube image for 1.17 onward. I guess the most backward-compatible decision would be to continue to honor `customHyperkubeImage` regardless, so if the user actually builds a custom 1.17.x hyperkube image that will still work.

jackfrancis:
Just to insert myself where I wasn't invited... we should continue to support hyperkube images (including "bring your own hyperkube") indefinitely.

mboersma:
@chewong since there isn't a hyperkube for 1.17 (except 1.17.0-alpha.1), is this a useful change?

chewong:
> @chewong since there isn't a hyperkube for 1.17 (except 1.17.0-alpha.1), is this a useful change?

We could still build a hyperkube image in 1.17 using this [Dockerfile](https://github.com/kubernetes/kubernetes/blob/master/cluster/images/hyperkube/Dockerfile) after building component binaries in k8s/k8s. This PR accounts for the new hyperkube  introduced by https://github.com/kubernetes/kubernetes/pull/84662. However, the long term solution is to move away from building hyperkube in k8s/test-infra and build individual components instead.

mboersma:
Thanks for the explanation. If this helps with upstream testing even temporarily, that's important.

PR #2313, opened 2019-11-13
refactor: render manifest specs at template generation time

jackfrancis:
**Reason for Change**:

Rather than delivering manifest specs w/ incomplete references and then sed replacing them via kubelet systemd, we will simply render the appropriate data based on the cluster configuration context.

This PR leverages go template interpretation for all cloud-init-delivered `/etc/kubernetes/manifests/`-bound files.

A related, targeted refactor of sed usage is also included in this PR:

- custom search domain script no gets its data injection via go template instead of at runtime via kubelet systemd shell script sed replacement

In addition to making the various cloud-init file sources more maintainable, this PR allows us to reduce the number of ARM parameters.

This PR includes a new UT for `engine.getContainerServiceFuncMap`, w/ coverage for the effected/new funcs. This is additive coverage, as there is currently no coverage at present for `engine.getContainerServiceFuncMap`.

A follow-up PR will do the equivalent refactor work for cloud-init-delivered addons specs.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -396,15 +396,12 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     iptables -t nat -A PREROUTING -p tcp --dport 4443 -j REDIRECT --to-port 443
 {{end}}
     a=/etc/kubernetes/manifests/kube-apiserver.yaml
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.APIServerConfig}}|g" $a
 {{ if HasCosmosEtcd  }}
     sed -i "s|<etcdEndPointUri>|{{ GetCosmosEndPointUri }}|g" $a
 {{ else }}
     sed -i "s|<etcdEndPointUri>|127.0.0.1|g" $a
 {{ end }}
     sed -i "s|<advertiseAddr>|{{WrapAsVariable "kubernetesAPIServerIP"}}|g" $a
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.ControllerManagerConfig}}|g" /etc/kubernetes/manifests/kube-controller-manager.yaml
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.SchedulerConfig}}|g" /etc/kubernetes/manifests/kube-scheduler.yaml
 {{if IsKubernetesVersionGe "1.17.0-alpha.1"}}
     {{ if IsIPv6DualStackFeatureEnabled }}
     sed -i "s|<img>|{{WrapAsParameter "kubeProxySpec"}}|g; s|<CIDR>|{{WrapAsParameter "kubeClusterCidr"}}|g; s|<kubeProxyMode>|{{ .OrchestratorProfile.KubernetesConfig.ProxyMode}}|g; s|<IPv6DualStackFeature>|IPv6DualStack: true|g" /etc/kubernetes/addons/kube-proxy-daemonset.yaml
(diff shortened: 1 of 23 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -395,27 +395,7 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.APIServerConfig}}|g" $a
-{{ if HasCosmosEtcd  }}
-    sed -i "s|<etcdEndPointUri>|{{ GetCosmosEndPointUri }}|g" $a
-{{ else }}
-    sed -i "s|<etcdEndPointUri>|127.0.0.1|g" $a
-{{ end }}
-    sed -i "s|<advertiseAddr>|{{WrapAsVariable "kubernetesAPIServerIP"}}|g" $a
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.ControllerManagerConfig}}|g" /etc/kubernetes/manifests/kube-controller-manager.yaml
-    sed -i "s|<args>|{{GetK8sRuntimeConfigKeyVals .OrchestratorProfile.KubernetesConfig.SchedulerConfig}}|g" /etc/kubernetes/manifests/kube-scheduler.yaml
+    sed -i "s|<advertiseAddr>|{{WrapAsVariable "kubernetesAPIServerIP"}}|g" /etc/kubernetes/manifests/kube-apiserver.yaml
```
jackfrancis: We need to keep the sed replacement for API Server IP because that value is determined at ARM runtime. (Not 100% sure if this is a strict requirement but unwinding that implementation will be a distinct effort.)

pkg/engine/engine_test.go
```diff
@@ -1728,8 +1728,7 @@ func TestVerifyGetBase64EncodedGzippedCustomScriptIsTransparent(t *testing.T) {
 		c := c
 		t.Run(c.name, func(t *testing.T) {
 			t.Parallel()
-			for _, file := range []string{kubernetesCustomSearchDomainsScript,
-				kubernetesMasterGenerateProxyCertsScript,
+			for _, file := range []string{kubernetesMasterGenerateProxyCertsScript,
```
jackfrancis: This UT is for cloud-init files that are being processed via go template, but do not have any (purposeful) go template layer expressions in them; in other words, the UT validates that the go template processing is a data no-op.

The deletion of the files in here reflects that we are in fact now doing purposeful go template processing of those files.

pkg/api/defaults-apiserver.go
```diff
@@ -38,9 +38,16 @@ func (cs *ContainerService) setAPIServerConfig() {
...
-	if !(cs.Properties.MasterProfile != nil && cs.Properties.MasterProfile.HasCosmosEtcd()) {
-		staticAPIServerConfig["--etcd-cafile"] = "/etc/kubernetes/certs/ca.crt"
+
+	if cs.Properties.MasterProfile != nil {
+		if cs.Properties.MasterProfile.HasCosmosEtcd() {
+			// Configuration for cosmos etcd
+			staticAPIServerConfig["--etcd-servers"] = fmt.Sprintf("https://%s:%s", cs.Properties.MasterProfile.GetCosmosEndPointURI(), strconv.Itoa(DefaultMasterEtcdClientPort))
+		} else {
+			// Configuration for local etcd
+			staticAPIServerConfig["--etcd-cafile"] = "/etc/kubernetes/certs/ca.crt"
```
CecileRobertMichon: the logic here is slightly different: before we would set the `--etcd-cafile` when the `MasterProfile` was nil OR if `HasCosmosEtcd`. Now we don't set it for nil `MasterProfile`. The etcd setting is probably not needed for AKS (is it?) but we should be careful about changing AKS logic in a refactor PR.

Also cc @serbrech this is one of those situations where AKS won't get these changes anymore, we need to establish a process for back-porting fixes to the default files.

jackfrancis: Right. I think this new logic is what we want (UT should tell the tale.)

Definitely worth pointing out to @serbrech although in this case (apiserver config) this doesn't affect AKS at all. But comparable changes to kubelet default config (for example) *will* affect the AKS node kubelet runtime configuration, if AKS doesn't otherwise set all kubelet runtime configuration values explicitly.

PR #2321, opened 2019-11-15
(docs/sgx): Add SGX Device plugin installation details

lachie83:
Signed-off-by: Lachlan Evenson <lachlan.evenson@microsoft.com>

**Reason for Change**:

Adds SGX device plugin installation details so that the Kubernetes schedule may be used to schedule workloads to hardware that supports the creation of Trusted Execution Environment (enclaves)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/sgx.md
+++ b/docs/topics/sgx.md
@@ -61,3 +61,192 @@ spec:
       path: /dev/sgx
       type: CharDevice
 ```
+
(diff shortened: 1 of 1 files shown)


devigned:
@lilyfang can you expand upon why this needs to be changed?

PR #2323, opened 2019-11-15
chore: Updating Azure CNI and Azure NPM to v1.0.29

jaer-tsun:
**Reason for Change**:

Updates to the latest release of Azure CNI and Azure NPM:

https://github.com/Azure/azure-container-networking/releases/tag/v1.0.29

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


packer/install-dependencies.sh
```diff
@@ -64,8 +64,8 @@ installGPUDrivers
 echo "  - nvidia-docker2 nvidia-container-runtime" >> ${VHD_LOGS_FILEPATH}
 
 VNET_CNI_VERSIONS="
+1.0.29
```
mboersma: It looks like we need to update `AZURE_NPM_VERSIONS` in this same file–it still points to 1.0.18. And I don't see us fetching `azure-vnet-telemetry` anywhere either, so that should probably be added here as well.

jaer-tsun: updated

PR #2327, opened 2019-11-16
fix: userAssignedIdentityId in windows azure.json missing quotes

xizhamsft:
**Reason for Change**:

The userAssignedIdentity property in azure.json for windows does not have quotes around the values. So when the value is empty string, you will get the following line in the json:

"userAssignedIdentityID": ,

That is a syntax error. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Azure DevOps seems not to have updated the overall "pr-e2e" job status, although all the actual test jobs passed. Merging...

PR #2333, opened 2019-11-18
feat: Support custom k8s components in v1.17

chewong:
**Reason for Change**:

Follow-up PR for https://github.com/Azure/aks-engine/pull/2191. With hyperkube removed, this PR adds support for custom kube-apiserver, kube-controller-manager, kube-proxy and kube-scheduler. customKubeBinary field is also added for custom kubelet and kubectl.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -284,9 +285,6 @@ installKubeletAndKubectl() {
 pullContainerImage() {
     CLI_TOOL=$1
     DOCKER_IMAGE_URL=$2
-    if [[ -n "${PRIVATE_AZURE_REGISTRY_SERVER:-}" ]]; then
```
jackfrancis: Why are we getting rid of this?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -109,6 +109,9 @@ if [[ "${GPU_NODE}" = true ]]; then
     ensureGPUDrivers
 fi
 {{end}}
+if [[ -n "${PRIVATE_AZURE_REGISTRY_SERVER:-}" ]] && [[ "$CONTAINER_RUNTIME" == "docker" ]]; then
```
jackfrancis: Oh I see. Does moving it out of the install script make the private login more general?

parts/k8s/cloud-init/artifacts/cse_install.sh
chewong: Docker login is only executed when we were extracting hyperkube. Since it is removed for 1.17 and we will be pushing custom kube components to a private azure registry, we will have to move the login function to somewhere else.

pkg/api/types.go
```diff
@@ -392,7 +392,7 @@ type KubernetesConfig struct {
 	UserAssignedID                    string            `json:"userAssignedID,omitempty"`
 	UserAssignedClientID              string            `json:"userAssignedClientID,omitempty"` //Note: cannot be provided in config. Used *only* for transferring this to azure.json.
 	CustomHyperkubeImage              string            `json:"customHyperkubeImage,omitempty"`
-	CustomKubeAPIServerImage          string            `json:"CustomKubeAPIServerImage,omitempty"`
```
jackfrancis: Thanks for spotting this despite my lgtm :)

Issue #2076, opened 2019-10-01
proposal: fail fast if OS image is not available


jadarsie:
**What**

Add a new command that returns the OS image metadata for a given AKSe distro. Something like this maybe.

```
$> aks-engine distro --name aks-ubuntu-16.04 [--cloud AzurePublicCloud]
Offer: aaa
Publisher: bbb
SKU: ccc
Version: xxxx.yy.zz
```

**Why**

The Azure Stack doc page documents which VHD is required for a specific AKSe version. However, this approach is not sufficient, users do not read the docs or are not aware of them. 

Users (and troubleshooters) of disconnected AzS stamps  will enjoy a better experience if they can quickly figure out  which VHD is expected.

**Next Steps**

In addition to the proposed command, it would be nice to short-circuit the execution of the `deploy` command (and print a nicer error message) if the expected VHD is not available in the local marketplace.

jadarsie:
The original proposal is not properly addressing the Windows VHD case.

A next iteration of this proposal would be to short-circuit the execution of the `deploy` command if any of the VHDs specified in the generated apimodel is not available in the target cloud's marketplace.

In that case, AKSe should NOT start the deployment and print a clean message (currently a stacktrace), that roughly says: "Hey, you are trying to use image XYZ, but it is not available, go get it before you re-try."

The `deploy` command output could potentially include the list of available images, but this may be confusing to users (at least for Linux nodes) as it won't give them the entire picture. The corresponding AKSe version that goes with the image will be missing.

jadarsie:
FYI @marosset

marosset:
Right now all windows agent pools must use the same image and the image details are stored in the windowsProfile.
https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#marketplace-images has more details

marosset:
(in case you were looking for info on how windows images are referenced to perform the check :))

jadarsie:
I was mostly interested in feedback to make sure this works for everyone. Thx for the details!

marosset:
This seems like a useful check to perform as part of  `deploy` operations

CecileRobertMichon:
sgtm

PR #2342, opened 2019-11-20
feat: validate VHD availability on azurestack

deaborch:
**Reason for Change**:

Check Azurestack for availability of VHDs 

**Issue Fixed**:

#2076

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
*TODO*:
- [ ] Validation fixes
- [ ] Include unit tests
- [ ] Make feature only available on azure stack 
- [ ] Support for feature in scale 
- [ ] Support for feature in upgrade


pkg/armhelpers/interfaces.go
```diff
@@ -95,6 +95,12 @@ type AKSEngineClient interface {
 	// DeployTemplate can deploy a template into Azure ARM
 	DeployTemplate(ctx context.Context, resourceGroup, name string, template, parameters map[string]interface{}) (resources.DeploymentExtended, error)
 
+	// ListVirtualMachineImages return a list of images
```
jackfrancis: What do you think @devigned? Should we create a `pkg/armhelpers/v2/interfaces.go` to create the additive method changes here? Making these changes in place will indeed break contracts of pre-existing `AKSEngineClient` type usages.

pkg/armhelpers/azurestack/vmImage.go
```diff
@@ -0,0 +1,39 @@
...
+
+//ListVirtualMachineImages returns the list of images available in the current environment
+func (az *AzureClient) ListVirtualMachineImages(ctx context.Context, top int32, location, publisherName, offer, skus, filter, orderBy string) (result azcompute.ListVirtualMachineImageResource, err error) {
+	vmImages, err := az.virtualMachineImageClient.List(ctx, location, publisherName, offer, skus, filter, &top, orderBy)
+	azVMImages := azcompute.ListVirtualMachineImageResource{}
+
+	if err != nil {
+		return azVMImages, fmt.Errorf("fail to get virtual machine images, %s", err)
+	}
+	err = DeepCopy(&azVMImages, vmImages)
```
devigned: please consider using the following:
```go
if err := DeepCopy(...); err != nil {
  // ...
}
```

pkg/armhelpers/azurestack/vmImage.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		return azVMImages, fmt.Errorf("fail to convert virtual machine images, %s", err)
+	}
+	return azVMImages, err
+}
+
+//GetVirtualMachineImage returns an image or an error where there is no image
+func (az *AzureClient) GetVirtualMachineImage(ctx context.Context, location, publisherName, offer, skus, version string) (result azcompute.VirtualMachineImage, err error) {
```
devigned: Why use named returns if you are not going to use them in the function? Either remove the names on the returns or use them in the function.

pkg/armhelpers/azurestack/vmImage.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	azcompute "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2018-10-01/compute"
+)
+
+//ListVirtualMachineImages returns the list of images available in the current environment
+func (az *AzureClient) ListVirtualMachineImages(ctx context.Context, top int32, location, publisherName, offer, skus, filter, orderBy string) (result azcompute.ListVirtualMachineImageResource, err error) {
```
devigned: Either remove the names on the returns or use them in the function.

pkg/armhelpers/azurestack/vmImage.go
```diff
@@ -0,0 +1,39 @@
...
+//GetVirtualMachineImage returns an image or an error where there is no image
+func (az *AzureClient) GetVirtualMachineImage(ctx context.Context, location, publisherName, offer, skus, version string) (result azcompute.VirtualMachineImage, err error) {
+	vmImage, err := az.virtualMachineImageClient.Get(ctx, location, publisherName, offer, skus, version)
+
+	azVMImage := azcompute.VirtualMachineImage{}
+
+	if err != nil {
+		return azVMImage, fmt.Errorf("fail to get virtual machine image, %s", err)
+	}
+	err = DeepCopy(&azVMImage, vmImage)
```
devigned: Again, please consider making this an error set and if err != nil block.

pkg/armhelpers/interfaces.go
devigned: I agree with you. This should be moved into a new interface. Perhaps, there should be an interface like below:

```go
type VMImageFetcher interface {
	ListVirtualMachineImages(ctx context.Context, top int32, location, publisherName, offer, skus, filter, orderBy string) (compute.ListVirtualMachineImageResource, error)
 	GetVirtualMachineImage(ctx context.Context, location, publisherName, offer, skus, version string) (compute.VirtualMachineImage, error)
}
```

If the client implements that type, then the image checking functionality could be used. For example:

```go
if imageFetcher, ok := akseClient.(VMImageFetcher); ok {
  // do the image checking stuff...
}
```

That doesn't limit the ability of AKS Engine to provide additional functionality and it allows that new functionality to be non-breaking to existing consumers. I favor this method since it is a "nice to have" behavior, not something that is required to function. That is to say, the client will still build a cluster without the ability to check if an image exists, but there may be an error raised by ARM at a later time stating the image does not exist. It is not optimal behavior since with the ability to check if an image exists prior to the deployment, we could respond quicker. In either case, the behavior will reach the same end state.

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type requiredImage struct {
+	image     api.AzureOSImageConfig
+	errorData error
+}
+
+//ValidateRequiredImages checks both master ang agent profiles and ensures these images are available in the environment
+func ValidateRequiredImages(ctx context.Context, location string, p *api.Properties, client AKSEngineClient) error {
+
```
devigned: less vert spacing please

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,113 @@
...
+		return api.RHELOSImageConfig
+	case api.CoreOS:
+		return api.CoreOSImageConfig
+	case api.AKSUbuntu1604:
+		return api.AKSUbuntu1604OSImageConfig
+	case api.AKSUbuntu1804:
+		return api.AKSUbuntu1804OSImageConfig
+	case api.ACC1604:
+		return api.ACC1604OSImageConfig
+	default:
```
devigned: I would expect default to return an error rather than just guess what the correct distro is based on exhaustion of options.

In constrained lists, I'd generally opt toward being precise.

deaborch: That is true and i do agree that we should be percise whenever possible. However, seeing as when dealing with the distro values, it is possible for the value to be empty and when it is empty, aksEngine defaults. with this in mind i felt the Image checker should behave in the same way

devigned: hmm... I guess I can buy that.

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,113 @@
...
+	image     api.AzureOSImageConfig
+	errorData error
+}
+
+//ValidateRequiredImages checks both master ang agent profiles and ensures these images are available in the environment
+func ValidateRequiredImages(ctx context.Context, location string, p *api.Properties, client AKSEngineClient) error {
+
+	var missingVMImages map[string]requiredImage
+
+	imageInformaition := getDistroVMImageInformation(p.MasterProfile.Distro)
```
jadarsie: typo

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,113 @@
...
+
+		} else {
+			missingVMImages[string(p.MasterProfile.Distro)] = requiredImage{
+				image:     imageInformaition,
+				errorData: err,
+			}
+		}
+	}
+
+	log.Infof("Image %s", *vmImage.Name)
```
jadarsie: this log message offers little information unless you have a lot of context

cmd/deploy.go
```diff
@@ -404,6 +404,16 @@ func (dc *deployCmd) run() error {
 		return errors.Wrapf(err, "in SetPropertiesDefaults template %s", dc.apimodelPath)
 	}
 
+	//Validate image
+	cx, cancel := context.WithTimeout(context.Background(), armhelpers.DefaultARMOperationTimeout)
```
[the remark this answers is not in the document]
devigned: 1+ to ctx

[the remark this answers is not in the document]
deaborch: So in the deploy code, there is actually already a variable that is `ctx` which is a different context object. It is in light of this that I believe that the original author called it `cx` all I did was move the declaration of `cx` higher up. See DeployTemplate method which also uses the `cx`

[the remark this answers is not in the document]
devigned: Ahh... interesting. I missed this.

You may have some unexpected side effects from this change. By moving the context higher up in the function and using it for **both** your calls for fetching images as well as the deployment of the template, you may introduce a a bug from the context timing out.

I would provide a new context for `ValidateDependencies()` which is the max amount of time to wait for that function to complete, and a second context for `DeployTemplate()`. This will separate the expiry of the two contexts.

cmd/deploy.go
```diff
@@ -404,6 +404,14 @@ func (dc *deployCmd) run() error {
 		return errors.Wrapf(err, "in SetPropertiesDefaults template %s", dc.apimodelPath)
 	}
 
+	// Validate image
+
```
devigned: nit: less vert spacing. Again, spaces after braces.

examples/azure-stack/kubernetes-azurestack.json
```diff
@@ -44,9 +44,9 @@
                 ]
             }
         },
-        "servicePrincipalProfile": {
-            "clientId": "",
-            "secret": ""
+        "servicePrincipalProfile": {	
+            "clientId": "",	
+            "secret": ""
```
devigned: Please clean up the spacing here. Probably could just `git checkout` the file. Doesn't look like you made any changes.

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,118 @@
...
+)
+
+type requiredImage struct {
+	image     api.AzureOSImageConfig
+	errorData error
+}
+
+//ValidateRequiredImages checks both master ang agent profiles and ensures these images are available in the environment
+func ValidateRequiredImages(ctx context.Context, location string, p *api.Properties, client AKSEngineClient) error {
+
```
devigned: nit: delete empty line

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,118 @@
...
+	missingVMImages := make(map[string]requiredImage)
+
```
devigned: nit: delete empty line

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,118 @@
...
+	imageInformation := getDistroVMImageInformation(p.MasterProfile.Distro)
+
```
devigned: Please clean up the vert spacing. Spaces after braces.

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,118 @@
...
+	if imageFetcher, ok := client.(VMImageFetcher); ok {
+		vmImage, err := imageFetcher.GetVirtualMachineImage(ctx, location, imageInformation.ImagePublisher, imageInformation.ImageOffer, imageInformation.ImageSku, imageInformation.ImageVersion)
+
+		if err != nil {
+			if p.MasterProfile.Distro == "" {
```
devigned: Please refactor as described below.

``` go
var distro = string(api.Ubuntu)
if p.MasterProfie.Distro != "" {
  distro = p.MasterProfile.Distro
}

missingVMImages[distro] = requiredImage{
  image:     imageInformation,
  errorData: err,
}
```

pkg/armhelpers/support-validator.go
```diff
@@ -0,0 +1,118 @@
...
+				if err != nil {
+					missingVMImages[string(api.Windows)] = requiredImage{
+						image:     imageInformaition,
+						errorData: err,
+					}
+				} else {
+					log.Infof("Image %s-%s-%s found in your environment", imageInformation.ImageOffer, imageInformation.ImageSku, *vmImage.Name)
+				}
+			} else {
+				imageInformaition := getDistroVMImageInformation(profile.Distro)
```
devigned: Please refactor to below. Also, please consider removing all non-essential logging statements.

``` go
imageInformaition := getDistroVMImageInformation(profile.Distro)
if _, err := imageFetcher.GetVirtualMachineImage(ctx, location, imageInformaition.ImagePublisher, imageInformaition.ImageOffer, imageInformaition.ImageSku, imageInformaition.ImageVersion)
  var distro = string(api.Ubuntu)
  if p.MasterProfie.Distro != "" {
    distro = profile.Distro
  }
  missingVMImages[distro] = requiredImage{
    image:     imageInformaition,
    errorData: err,
}
```

devigned: Better yet, change the function to do the following:

```go
for _, profile := range p.AgentPoolProfiles {
  var imgInfo api.AzureOSImageConfig
  var distro string
  if profile.OSType == api.Windows {
    imgInfo := api.WindowsServer2019OSImageConfig
    distro = string(api.Windows)
  } else {
    distro = string(api.Ubuntu)
    if profile.Distro != "" {
      distro = string(profile.Distro)
    }

    imgInfo := getDistroVMImageInformation(distro)
  }

  if _, err := imageFetcher.GetVirtualMachineImage(ctx, location, imgInfo.ImagePublisher, imgInfo.ImageOffer, imgInfo.ImageSku, imgInfo.ImageVersion)
    missingVMImages[distro] = requiredImage{
      image:     imageInformaition,
      errorData: err,
  }
}
```

cmd/deploy.go
```diff
@@ -404,6 +404,12 @@ func (dc *deployCmd) run() error {
 		return errors.Wrapf(err, "in SetPropertiesDefaults template %s", dc.apimodelPath)
 	}
 
+	// Validate image
+	err = dc.validateDependencies(context.Background())
```
devigned: `context.Background()` is not appropriate to use since it means that this command could wait indefinitely for validateDependencies to return.

You should create a context with the maximal amount of time you expect this call to take.

devigned:
@jadarsie thank you for fixing up the context timeout issue. Looks like the only thing left is raising coverage a bit.

jadarsie:
@deaborch please review the changes I pushed yesterday

PR #2345, opened 2019-11-21
fix: enable cloud-node-manager addon on upgrade when appropriate

### LINKED ISSUE #2337: Upgrade to 1.17 leaves cloud-node-manager addon disabled
**Describe the bug**
When upgrading from an earlier Kubernetes version with `useCloudControllerManager` equal to false, the `cloud-node-manager` addon is not enabled as it should be since it's required on 1.17.

**Steps To Reproduce**

**Expected behavior**

**AKS Engine version**

**Kubernetes version**

**Additional context**

mboersma:
**Reason for Change**:
Enables the cloud-node-manager addon if appropriate during `aks-engine upgrade`.

**Issue Fixed**:
Fixes #2337

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -540,6 +540,13 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		}
 	}
 
+	// Ensure cloud-node-manager is enabled on appropriate upgrades
+	if isUpgrade {
+		if i := getAddonsIndexByName(o.KubernetesConfig.Addons, CloudNodeManagerAddonName); i > -1 {
+			o.KubernetesConfig.Addons[i] = defaultCloudNodeManagerAddonsConfig
+		}
+	}
+
 	// Back-compat for older addon specs of cluster-autoscaler
 	if isUpgrade {
 		i := getAddonsIndexByName(o.KubernetesConfig.Addons, ClusterAutoscalerAddonName)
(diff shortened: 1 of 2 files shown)


mboersma:
/hold

I think a similar fix is needed for the CSI components.

pkg/api/addons.go
```diff
@@ -540,6 +540,13 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		}
 	}
 
+	// Ensure cloud-node-manager is enabled on appropriate upgrades
```
jackfrancis: I think this will work as-is, but it is also general and may introduce weird stuff in the future. For example, if we introduce additional configuration vectors to this addon w/ defaults, me may (or may not) want to use the default values in an upgrade scenario. In other words, I think the back-compat implementation should be a targeted transformation: "if you were using this specific version of the addon based and are upgrading from this particular version of k8s to this one, do this specific transformation".

What are the specific conditions that this back-compat foo is trying to fix?

mboersma: Yes, using the default for cloud-node-controller should always be safe since it has no config attached...at least currently. But we could run afoul of config changes for the CSI addons.

The specific case I need to test for is `UseCloudControllerManager == true` and upgrade version is 1.16 or greater. If those conditions are met and the addon isn't already enabled, we use the default config for the addon (which will also enable it).

jackfrancis: Got it, I think this should be fine until we evolve this addon, in which case the "isUpgrade" foo will have to change with it (and UT should be used to support the evolution of all these scenarios)

PR #2349, opened 2019-11-22
test: remove windows-specific tests, use windows node pools

jackfrancis:
**Reason for Change**:

Rather than test Windows by creating entirely new clusters, let's just add a Windows node pool to every cluster configuration we test across Kubernetes versions.

This PR adds a Windows node pool to the "standard" cluster configuration we use for our current "Linux" Kubernetes version tests, effectively making them OS-agnostic (or more appropriately: OS-generic, as they will cover both OS flavors).

This PR also removes 1.11 and 1.12 from our PR test matrix to reflect that 1.13-->1.17 are the versions of Kubernetes that are actively under development upstream.

Fewer tests for the same amount of coverage!

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -76,7 +76,6 @@ jobs:
     k8sRelease: '1.13'
     apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
     createVNET: true
-    stabilityIterations: '3'
 
 - template: e2e-job-template.yaml
   parameters:
@@ -84,7 +83,6 @@ jobs:
     k8sRelease: '1.14'
     apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
     createVNET: true
-    stabilityIterations: '3'
 
 - template: e2e-job-template.yaml
   parameters:
@@ -92,7 +90,6 @@ jobs:
     k8sRelease: '1.15'
     apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
     createVNET: true
-    stabilityIterations: '3'
 
 - template: e2e-job-template.yaml
   parameters:
@@ -100,7 +97,6 @@ jobs:
     k8sRelease: '1.16'
     apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
     createVNET: true
-    stabilityIterations: '3'
 
 - template: e2e-job-template.yaml
   parameters:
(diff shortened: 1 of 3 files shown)


mboersma:
> This PR also removes 1.11 and 1.12 from our PR test matrix

Could we merge #2303 instead? It's just better hygiene to have 1 PR (commit) == 1 concern, plus it was here first.

.pipelines/pr-e2e.yaml
```diff
@@ -70,94 +70,37 @@ jobs:
...
-    createVNET: true
-    stabilityIterations: '3'
-
 - template: e2e-job-template.yaml
   parameters:
     name: 'k8s_1_13_release_e2e'
     k8sRelease: '1.13'
     apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
     createVNET: true
-    stabilityIterations: '3'
```
mboersma: 3 is the default, I assume?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -1891,11 +1891,11 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				if clusterAutoscalerEngaged {
 					cpuTarget = 50
 					for _, profile := range eng.ExpandedDefinition.Properties.AgentPoolProfiles {
-						maxPods, _ := strconv.Atoi(profile.KubernetesConfig.KubeletConfig["--max-pods"])
-						var n []node.Node
-						n, err = node.GetByRegexWithRetry(fmt.Sprintf("^k8s-%s", profile.Name), 3*time.Minute, cfg.Timeout)
```
mboersma: So now we're just using `profile.Count` instead of actually fetching nodes to find the `maxTotalPods`. Is this because of Windows, and was there value in making the API call to get the nodes?

jackfrancis:
> > This PR also removes 1.11 and 1.12 from our PR test matrix
> 
> Could we merge #2303 instead? It's just better hygiene to have 1 PR (commit) == 1 concern, plus it was here first.

Yep

.pipelines/pr-e2e.yaml
jackfrancis: So we've been using 3 for the "linux" tests, which are occasionally flaky :(

We were skipping for "windows" tests.

The most cautious way to converge the two was to stop running these "stability" tests on PR tests for now.

(That's my thinking, at least.)

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Right, it's because the prior regex would fail (Windows nodes have a different name).

The purpose of this enumeration is to set an appropriate number of pod replicas that exceeds the pods-per nodes setting across all nodes (so that we induce quick scale up by creating pods that are unable to be scheduled). Doing the node count by calling the k8s api at runtime is more strictly accurate, but with Windows nodes in the mix, got more complicated: basically correlating the nodes in the cluster with a particular pool is trickier. All that being said, the theoretical benefits of doing this over simply taking the `Count` value in the api model seemed to be not worth wrangling this complexity.

PR #2353, opened 2019-11-22
docs: update kubernetes-developers.md

chewong:
**Reason for Change**:

Follow-up PR for https://github.com/Azure/aks-engine/pull/2333.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/kubernetes-developers.md
+++ b/docs/topics/kubernetes-developers.md
@@ -1,57 +1,57 @@
 # For Kubernetes Developers
 
-If you're working on Kubernetes upstream, you can use AKS Engine to test your build of Kubernetes in the Azure environment.  The option that allows you to do this is `orchestratorProfile.kubernetesConfig.customHyperkubeImage`, which you should set to point to a Docker image containing your build of hyperkube.
+If you are working on Kubernetes upstream, you can use AKS Engine to test your build of Kubernetes in the Azure environment.
+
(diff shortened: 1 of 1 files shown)


docs/topics/kubernetes-developers.md
```diff
@@ -1,76 +1,124 @@
...
+```
+
+After that, you will have to upload a tarball (`_output/release-tars/kubernetes-node-linux-amd64.tar.gz`) built from the previous step to online storage such as GitHub or Azure Storage. AKS Engine uses the tarball to extract kubelet and kubectl.
+
+### AKS Engine API Model
+
+* Open the AKS Engine API Model (e.g. a file from the examples directory) and add the following to the `orchestratorProfile` section:
+
+```
+"kubernetesConfig": {
```
jackfrancis: Do we want to mention that these workflows should also include specifying `"orchestratorRelease": "1.17"` in the api model? I think that will be required to properly instruct the template generation flow that this cluster configuration input should be interpreted in a 1.17 context.

Also, we might just mention this generally; when using custom images (hyperkube or not) you want to tell aks-engine which minor version (via `"orchestratorRelease"`) context you're in, so various defaults and configuration are properly applied to the ARM template + artifacts.

PR #2358, opened 2019-11-22
fix: correct defaults order for setting IPAddressCount

jackfrancis:
**Reason for Change**:

This fixes a regression that landed w/ #2274 

The IPAddressCount values in MasterProfile and each AgentPoolProfile depend (in part) upon the kublet configuration `--max-pods`; essentially these values are related and need to be configured in concert according to a deterministic ordinality.

This PR moves the IPAddressCount default enforcement for both masters and pools to occur after the kubelet defaults codeflow.

The primary fix here is missing UT. :(

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Manually confirmed that had the UT added in this PR been in place prior to #2274 this bug would not have landed. Tears.

PR #2371, opened 2019-11-27
feat: node-problem-detector addon

### LINKED ISSUE #2364: Support for node-problem-detector
**Describe the request**
node-problem-detector is a useful daemon for detecting common problems a k8s node might encounter. It can reliably detect frequent docker, containerd, or kubelet restarts, hung tasks, unavailable network, just to name a few. node-problem-detector is also pluggable. Writing a new plugin can be as easy as authoring a regex for a log line that's indicative of a problem.

It would useful to have node-problem-detector installed by default as a systemd unit on the aks-engine vm image to surface common node stability issues as events and node conditions.

Users of AKS Engine haven't had an easy time diagnosing common problems with nodes. node-problem-detector would make some of these common issues more observable by surfacing them as events and node conditions. Not to mention, additional detections can be added as new problems arise.

Install node-problem-detector as a systemd unit by default or at least allow users to opt in.

Hoping for the best.

**Additional context**
https://github.com/kubernetes/node-problem-detector

mboersma:
**Reason for Change**:
Adds the Kubernetes [node problem detector](https://github.com/kubernetes/node-problem-detector) as an addon.

**Issue Fixed**:
Fixes #2364

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

Should this be enabled by default as suggested in #2364?


parts/k8s/containeraddons/node-problem-detector.yaml
```diff
@@ -0,0 +1,88 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: node-problem-detector
+  namespace: kube-system
+  labels:
+    kubernetes.io/cluster-service: "true"
+    addonmanager.kubernetes.io/mode: Reconcile
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
juan-lee: Consider adding a [Service](https://gist.github.com/juan-lee/b4a98b9aa7384afc74e311134ee98cac) and the --prometheus-address arg to expose npd's metrics endpoint. Also, I think we also need the /dev/kmsg mount.

devigned:
All looks good, but would be nice to get the metrics endpoint exposed.

parts/k8s/containeraddons/node-problem-detector.yaml
mboersma: I added all the above.

devigned:
/lgtm

I'll create an issue to make this add-on to default to enabled after we get more signal.

PR #2375, opened 2019-12-02
fix: change kernel settings based on number of cores

xizhamsft:
**Reason for Change**:

For nodes with more CPU/Memory, the default kernel settings for  ubuntu is not big enough for high-performance server scenarios.

**Issue Fixed**:

We changed the values for ARP table, network connections. These optimizations are applied widely in the industry.

**Requirements**:

- [X ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/kubelet.service
+++ b/parts/k8s/cloud-init/artifacts/kubelet.service
@@ -17,6 +17,15 @@ ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
 {{/* This is a partial workaround to this upstream Kubernetes issue: */}}
 {{/* https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731 */}}
 ExecStartPre=/sbin/sysctl -w net.ipv4.tcp_retries2=8
+ExecStartPre=/sbin/sysctl -w net.core.somaxconn=16384
+ExecStartPre=/sbin/sysctl -w net.ipv4.tcp_max_syn_backlog=16384
+ExecStartPre=/sbin/sysctl -w net.core.message_cost=40
+ExecStartPre=/sbin/sysctl -w net.core.message_burst=80
+
+ExecStartPre=/bin/bash -c "if [ $(nproc) -gt 8 ]; then /sbin/sysctl -w net.ipv4.neigh.default.gc_thresh1=4096; fi"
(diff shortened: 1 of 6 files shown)


jackfrancis:
@xizhamsft run `make build` to generate new code based on these changes, and then commit that change

jackfrancis:
If you have trouble you may need to run `cd ./hack/tools && make reload` locally to get the version of code generator tools that github CI uses. Then, run `make build`.

devigned:
If you rebase on the latest master, you no longer need to run `cd ./hack/tools && make reload`. The correct version of `go-bindata` will be installed in `./hack/tools/bin` and will be used in the build.

xizhamsft:
@jackfrancis , @CecileRobertMichon , please take a look. The only e2e failure is exit code 14, unable to start etcd runtime. Should not be related. 
@devigned . Thanks for the tip. Works perfectly.

jackfrancis:
@xizhamsft Can you run some manual tests using > 8 core VM SKUs to validate the desired OS configuration changes?

xizhamsft:
> @xizhamsft Can you run some manual tests using > 8 core VM SKUs to validate the desired OS configuration changes?

Sure. Doing that now.

palma21:
Hey, any eta for this? Pending just a few manual tests right?

jackfrancis:
@xizhamsft could you add a new E2E configuration file under this directory?

- `test/e2e/test_cluster_configs`

You can use `base.json` as a template for a new file, we can call it `large-vm-sku.json`.

Let's replace `Standard_D2_v3` with `Standard_D16_v3`, that way the additional "if greater than 8 cores" configuration will kick in.

Also, please replace the empty `"env": {},` line with:

```
"env": {
		"REGION_OPTIONS": "eastus,westeurope,southeastasia,westus2"
	},
```

This is because we have add'l quota granted to our test subs for Dv3 in those regions.

Thanks!

xizhamsft:
@jackfrancis, @palma21 , a few changes in the new iteration:
1. Included the new test file as suggested by Jack Francis
2. removed the conntrac_max settings since turns out kube-proxy is already setting that value based on number of cores: --conntrack-max-per-core="32768" and the value is large enough for us.
3. tested by copying the kubelet.service file to the node and restart kublet and  verified the settings are applied

Issue #2251, opened 2019-11-01
feat: enable configurable image source


craiglpeters:
**Describe the request**
The repository for Kubernetes and related images is being moved from ACR to mcr.microsoft.com. AKS Engine needs the ability to be configured to pull from the old or the new location to enable testing of the new images as an optional configuration, and then later make the new location the default. The configuration needs to apply to both the images packaged into VHDs and to the CSE.

**Explain why AKS Engine needs it**
AKS and other customers of AKS Engine need to be able to consume the new images from MCR

**Describe the solution you'd like**
- work w/ aks-engine team to validate that image changes are applied to VHD CI spec
- do not merge changes to AKS until above changes have been applied to a new VHD publication iteration, and the resultant VHD is being referenced in the RP @ HEAD

**Describe alternatives you've considered**

**Additional context**

thomas1206:
Here is the image list currently used by AKS:

rp: https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp?path=%2Fpkg%2Fmicrosoft.com%2Fcommon%2Fk8sversions%2Fk8sversions.go&version=GBmaster&line=120&lineStyle=plain&lineEnd=491&lineStartColumn=1&lineEndColumn=1

overlay: https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-overlaymgr?path=%2Fpkg%2Fhelmctlr%2Fhelmvalues%2Fconst.go&version=GBguwe%2Fexpose-diagnostic-settings&line=33&lineStyle=plain&lineEnd=95&lineStartColumn=1&lineEndColumn=1

jackfrancis:
Status update: a stream of addons refactor PRs is landing in master right now, which will contribute to cleaning up the underlying surface area for custom images.

jackfrancis:
See #2408 for the design doc that will guide this work

PR #2381, opened 2019-12-02
refactor: remove unnecessary enabled addon config state representation

jackfrancis:
**Reason for Change**:

The definitive configuration state representation for addons is in the `KubernetesConfig` `Addons` array. This PR removes the unnecessary setting of an `isEnabled` property during addons data marshalling, and instead consumes the authoritative `Enabled` boolean in the `Addons` config array.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/artifacts.go
```diff
@@ -36,133 +36,111 @@ func kubernetesContainerAddonSettingsInit(p *api.Properties) map[string]kubernet
 			sourceFile:      "kubernetesmasteraddons-heapster-deployment.yaml",
 			base64Data:      k.GetAddonScript(HeapsterAddonName),
 			destinationFile: "kube-heapster-deployment.yaml",
-			isEnabled:       k.IsAddonEnabled(HeapsterAddonName),
```
jackfrancis: Upside: we remove duplicate code!

pkg/engine/artifacts_test.go
```diff
@@ -12,339 +12,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
-func TestKubernetesContainerAddonSettingsInit(t *testing.T) {
```
jackfrancis: This UT essentially tests the functionality that we're removing, so it goes.

pkg/engine/artifacts.go
CecileRobertMichon: do we still need `isEnabled` to be part of `kubernetesComponentFileSpec` or is it removed everywhere?

CecileRobertMichon: nevermind it's still used in `kubernetesAddonSettingsInit`. What's the difference with `kubernetesContainerAddonSettingsInit` and why remove it in one but not both?

jackfrancis: It's still being used by the non-user-configurable addons, and the manifests.

I'm scaffolding out a refactor plan to move all addons into the "user-configurable addons" interface, and to do something similar for the `/etc/k8s/manifests` files as well, after which time we can possibly remove the property altogether.

jackfrancis: The difference is that `kubernetesContainerAddonSettingsInit` is able to correlate with an `KubernetesAddon` item in the `Addons` array.

`kubernetesAddonSettingsInit`, on the other hand, *is* where the authoritative state of whether or not a particular "non-user-configurable addon" is enabled or not. (This is tech debt.)

jackfrancis:
Related to #2251

PR #2383, opened 2019-12-02
refactor: move addons consts to pkg/api/common

jackfrancis:
**Reason for Change**:

Addon name consts are being used by both template generation (`pkg/engine`)  and cluster data model composition (`pkg/api`). This PR moves all addon consts to `pkg/api/common` so that we no longer have duplicate const definitions of the same values for local usage in `pkg/engine` and `pkg/api`, respectively.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Related to #2251

PR #2388, opened 2019-12-03
feat: Adding a unique id to Windows VHDs for telemtry correlation

marosset:
**Reason for Change**:

Adding a unique ID file to Windows VHDs so we can correlate telemetry from Windows node setup extensions to a specific VHD image

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/configure-windows-vhd.ps1
+++ b/packer/configure-windows-vhd.ps1
@@ -241,6 +241,7 @@ switch ($env:ProvisioningPhase)
         Install-Docker
         Get-ContainerImages
         Get-FilesToCacheOnVHD
+        (New-Guid).Guid | Out-File -FilePath 'c:\vhd-id.txt'
     }
     default
     {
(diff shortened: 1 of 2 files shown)


packer/configure-windows-vhd.ps1
```diff
@@ -241,6 +241,7 @@ switch ($env:ProvisioningPhase)
         Install-Docker
         Get-ContainerImages
         Get-FilesToCacheOnVHD
+        (New-Guid).Guid | Out-File -FilePath 'c:\vhd-id.txt'
```
PatrickLang: would putting this in a machine env var help with AppInsights integration?

PatrickLang: n/m, the CSE will be able to read this just fine

PR #2396, opened 2019-12-05
feat: upgrade uses keyvault to create kubeconfig

artburkart:
This commit adds a `GetKeyVaultSecret` function to the `AKSEngineClient`
interface. This is used in [cmd/upgrade.go](./cmd/upgrade.go) to fetch
the caCrt, kubectlCrt, and kubectlKey from Azure KeyVault and populate a
kubeconfig, which it then uses to roll a kubernetes cluster.

As this is a WIP PR, it is probably unnecessary to point out that I've
made some not-ideal design decisions in getting this to solve my use
case. The first that comes to mind is how I opted to log a warning
whenever a KeyVault request is attempted but fails. I wasn't sure what
the desired behavior was there. Perhaps it should just fail out.

I can't pretend I am a golang developer. What I present here is a hack
job, and I'm sure I broke several golang conventions along the way. One
part of the code that comes to mind is my instantiation of an
`api.CertificateProfile` struct to use as a temporary placeholder while
I manipulate the properties of the uc `UpgradeCommand` object. I do this
because the `GenerateKubeConfig` function takes the entire `Properties`
structure, and I only required 3 properties from it.

I will follow up with a second, much smaller PR, that shows how I worked
around the same limitation by simply reading a kubeconfig file from a
path that I pass in from the command line.

**Reason for Change**:
From Slack: https://kubernetes.slack.com/archives/C5HJXTT9Q/p1575468425031100?thread_ts=1574685642.479700&cid=C5HJXTT9Q

> I tried performing a cluster upgrade, and I noticed the aks-engine upgrade command is incompatible with usage of KeyVault. The reason is because this line of code here doesn't take KeyVault into consideration:
> https://github.com/Azure/aks-engine/blob/ac9a074b4cff58f6e4db89dc190d7b1c85f09ee8/pkg/engine/engine.go#L59
> 
> Since I have a ready use case, I'm happy to write the code, but I'm curious how you recommend accessing/instantiating the keyvault client.
> 
> I think this is the client import statement I should use: https://github.com/Azure/aks-engine/blob/ac9a074b4cff58f6e4db89dc190d7b1c85f09ee8/pkg/engine/armtype.go#L12

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


artburkart:
@jackfrancis – Here's that WIP you requested.

PR #2397, opened 2019-12-05
feat: Adds kubeconfig option to upgrade cmd

artburkart:
This PR accounts for my second strategy to solve the limitation
described in #2396. Rather than introduce the KeyVault client, so we can
run `GetSecret` and fetch our certs and keys, I read a local config file
based on what's passed into the command line.

This is not a complete PR, because I think the same stuff would need to
be done anywhere else `GenerateKubeConfig` is called.

**Reason for Change**:

See #2396 

**Issue Fixed**:

See #2396  

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"context"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path/filepath"
 	"time"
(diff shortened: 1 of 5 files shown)


artburkart:
@jackfrancis – this is the second strategy I suggested

artburkart:
I just realized I didn't open this as a draft PR 😞

Issue #2392, opened 2019-12-04
Specifying multiple VM extensions in apimodel causes occasional deployment failures


marosset:
**Describe the bug**
Specifying multiple VM extensions in an apimodel can cause random deployment failures.

We are seeing frequently on the sig-windows k8s test passes with the following apimodel/deployment template:
https://github.com/kubernetes-sigs/windows-testing/blob/master/job-templates/kubernetes_release_1_17.json

The deployments fail with the following error:
W1203 14:18:21.338] 2019/12/03 14:18:21 main.go:319: Something went wrong: starting e2e cluster: error creating cluster: cannot deploy: cannot get the create deployment future response: Code="DeploymentFailed" Message="At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details." Details=[{"code":"Conflict","message":"{\r\n  \"status\": \"Failed\",\r\n  \"error\": {\r\n    \"code\": \"ResourceDeploymentFailure\",\r\n    \"message\": \"The resource operation completed with terminal provisioning state 'Failed'.\",\r\n    \"details\": [\r\n      {\r\n        \"code\": \"DeploymentFailed\",\r\n        \"message\": \"At least one resource deployment operation failed. Please list deployment operations for details. Please see https://aka.ms/DeployOperations for usage details.\",\r\n        \"details\": [\r\n          {\r\n            \"code\": \"Conflict\",\r\n            \"message\": \"{\\r\\n  \\\"error\\\": {\\r\\n    \\\"code\\\": \\\"Conflict\\\",\\r\\n    \\\"message\\\": \\\"The request failed due to conflict with a concurrent request. \\\"\\r\\n  }\\r\\n}\"\r\n          }\r\n        ]\r\n      }\r\n    ]\r\n  }\r\n}"}]

Example:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-kubernetes-e2e-aks-engine-azure-1-17-windows/1201856259427405824

**Expected behavior**
Deployments do not fail

**AKS Engine version**
latest

**Kubernetes version**
any 

**Additional context**

marosset:
It looks like ARM is not happy having 3 different extensions (billing extension, master_extension, and gmsa-coredns extension) having deponsOn relationships with the cse-master-* extension and is throwing conflicts when processing the ARM template.

[...]

PR #2398, opened 2019-12-05
fix: fix ARM dependency issues with vm user-specified extensions on node pools

marosset:
**Reason for Change**:

**Issue Fixed**:

Fixes issue #2392

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/vmextensions.go
+++ b/pkg/engine/vmextensions.go
@@ -235,9 +235,9 @@ func CreateCustomExtensions(properties *api.Properties) []DeploymentARM {
 	var extensionsARM []DeploymentARM
 
 	if properties.MasterProfile != nil {
-		// The first extension needs to depend on the billing extension created for all pools
+		// The first extension needs to depend on the master cse created for all nodes
 		// Each proceeding extension needs to depend on the previous one to avoid ARM conflicts in the Compute RP
-		nextDependsOn := "[concat('Microsoft.Compute/virtualMachines/', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')), '/extensions/computeAksLinuxBilling')]"
+		nextDependsOn := "[concat('Microsoft.Compute/virtualMachines/', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')), '/extensions/cse-master-', copyIndex(variables('masterOffset')))]"
 
 		for _, extensionProfile := range properties.ExtensionProfiles {
 			masterOptedForExtension, singleOrAll := validateProfileOptedForExtension(extensionProfile.Name, properties.MasterProfile.Extensions)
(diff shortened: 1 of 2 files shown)


marosset:
Code changes are functional but I want to improve the unit tests and add possibly add a new config to the jenkins matrix.

marosset:
I'll need to follow up with a Jenkin's job in a separate PR.
I'm having issues finding extensions that are both function and don't conflict with test content in the e2e tests (most extensions muck with dns which causes the dns livenessprobe test case to fail).
Also, specifying SKIP_TEST=true in the test cluster config does not cause the tests to skip.

pkg/engine/vmextensions.go
```diff
@@ -234,39 +234,53 @@ func CreateAgentVMASAKSBillingExtension(cs *api.ContainerService, profile *api.A
 func CreateCustomExtensions(properties *api.Properties) []DeploymentARM {
 	var extensionsARM []DeploymentARM
 
-	for _, extensionProfile := range properties.ExtensionProfiles {
-		if properties.MasterProfile != nil {
+	if properties.MasterProfile != nil {
+		// The first extension needs to depend on the billing extension created for all pools
+		// Each proceeding extension needs to depend on the previous one to avoid ARM conflicts in the Compute RP
+		nextDependsOn := "[concat('Microsoft.Compute/virtualMachines/', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')), '/extensions/computeAksLinuxBilling')]"
```
CecileRobertMichon: doesn't the billing extension change names depending on AKS/AKS Engine? Also it can be disabled. I don't think we can hardcode a dependency on `computeAksLinuxBilling` here.

marosset: the name of the extension doesn't change (which is weird)

marosset: I didn't know that the billing extension can be disabled, is there a way to detect that in the properties @CecileRobertMichon.

CecileRobertMichon: You could disable the extension though. See `IsAKSBillingEnabled`.

marosset: Thanks, I'll take a look

CecileRobertMichon: Yeah looks like it's the type that changes, not the name. AKS did change the vmss one at some point https://github.com/Azure/aks-engine/pull/1561/files but that was for VMSS so shouldn't affect this.

marosset: It looks like we only need to serialize the deployments of the 'CustomScriptExtension' extensions. The billing extensions are their own type of extension so we don't need to worry about those.

It also looks like the way aks-engine is using custom script extensions for additional extensions is a bit fragile. According to https://azure.microsoft.com/en-us/blog/automate-linux-vm-customization-tasks-using-customscript-extension/ you are supposed to refernce all of the scripts you want to run in a single ARM resource since there can only be on 'CustomScriptExtension' active on a machine/scale set at any given time. 

A larger refactor should probably be considered when we have time and/or if we need to support additional extensions for scale sets.

This is working pretty reliably as is in my local testing.

marosset:
I'm going to do a little more testing before i remove the hold.

PR #2399, opened 2019-12-05
test: add UT for kubernetesContainerAddonSettingsInit

jackfrancis:
**Reason for Change**:

In preparation for converging all addons into the user-configurable "addons" interface (and hey, also, we're actually testing some more code!) this adds some basic UT coverage to the `pkg/engine/kubernetesContainerAddonSettingsInit()` func that statically composes the set of addons to be prepared for reading from the filesystem and concatenating into cloud-init-appropriate output.

There's not much there, but this function was literally not UT'ed.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/artifacts.go
```diff
@@ -33,120 +33,119 @@ func kubernetesContainerAddonSettingsInit(p *api.Properties) map[string]kubernet
...
 			base64Data:      "",
-			destinationFile: "cloud-node-manager.yaml",
+			destinationFile: cloudNodeManagerAddonDestinationFilename,
 		},
 		common.NodeProblemDetectorAddonName: {
-			sourceFile:      "node-problem-detector.yaml",
+			sourceFile:      nodeProblemDetectorAddonSourceFilename,
 			base64Data:      k.GetAddonScript(common.NodeProblemDetectorAddonName),
-			destinationFile: "node-problem-detector.yaml",
-			isEnabled:       k.IsAddonEnabled(common.NodeProblemDetectorAddonName),
```
jackfrancis: Cleaning this up. This addon was introduced while this PR was in development: https://github.com/Azure/aks-engine/pull/2381

pkg/engine/artifacts.go
```diff
@@ -33,120 +33,119 @@ func kubernetesContainerAddonSettingsInit(p *api.Properties) map[string]kubernet
...
-			sourceFile:      "azure-policy-deployment.yaml",
+			sourceFile:      azurePolicyAddonSourceFilename,
 			base64Data:      k.GetAddonScript(common.AzurePolicyAddonName),
-			destinationFile: "azure-policy-deployment.yaml",
+			destinationFile: azurePolicyAddonDestinationFilename,
 		},
 		common.CloudNodeManagerAddonName: {
-			sourceFile:      "kubernetesmasteraddons-cloud-node-manager.yaml",
+			sourceFile:      cloudNodeManagerAddonSourceFilename,
 			base64Data:      "",
```
jackfrancis: @mboersma do you have any idea why the cloud-node-manager addon doesn't currently support arbitrary override data like all the other user-configurable addons?

mboersma: There aren't any documented command-line arguments to the `cloud-node-manager` binary, so I left this empty when I implemented it.

jackfrancis: So, the `.Data` property of an addon is actually just meant to be base64-encoded data that represents the entire yaml manifest. It's a way of brute-force overriding what's in the blessed aks-engine implementation, either for testing, or just for the fun of living on the edge.

Because cloud-node-manager is just a yaml spec that we drop into /etc/kubernetes/addons/ I think there's no reason to specifically call it out as not able to use this brute-force override mechanism. Right?

mboersma: Gotcha. No, no reason that I can think of–it's just an oversight due to my misunderstanding.

jackfrancis: 10-4, thanks for confirming! I'll go ahead and clean that up in this PR since it'll have the new UT coverage to prove the change out.

jackfrancis:
Related to #2251

PR #2400, opened 2019-12-05
feat: Cse tracing for windows nodes

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowsfunctions.ps1
+++ b/parts/k8s/kuberneteswindowsfunctions.ps1
@@ -47,7 +47,19 @@ function DownloadFileOverHttp
     
         $oldProgressPreference = $ProgressPreference
         $ProgressPreference = 'SilentlyContinue'
+
+        $downloadTimer = [System.Diagnostics.Stopwatch]::StartNew()
         Invoke-WebRequest $Url -UseBasicParsing -OutFile $DestinationPath -Verbose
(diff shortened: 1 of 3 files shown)


marosset:
@devigned here is my WIP PR for cse-tracing for windows. I've plumped an application insights key through the API model and also used the already existing EnableTelemetry feature flag to collecting telemetry in the CSEs.
Let's make sure to use the same API model constructs for linux vs windows settings :)

pkg/api/types.go
```diff
@@ -768,6 +769,12 @@ type CustomCloudProfile struct {
 	PortalURL                  string                      `json:"portalURL,omitempty"`
 }
 
+// TelemetryProfile contains settings for collecting telemtry.
+// Note telemtry is currently enabled/disabled with the 'EnableTelemetry' feature flag.
```
CecileRobertMichon:
```suggestion
// Note telemetry is currently enabled/disabled with the 'EnableTelemetry' feature flag.
```

marosset:
This is part 1 where I'm only reporting on some east to collect metrics.
I will follow this PR up with another set of changes to query the instance metadata service to get metrics such as vm size and other data points that aren't already available in the CSE powershell.

PR #2401, opened 2019-12-05
chore: don't calculate code coverage on const-only files

jackfrancis:
**Reason for Change**:

We are currently counting various const libs against test coverage statistics, which we shouldn't be doing.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.codecov/codecov.yml
+++ b/.codecov/codecov.yml
@@ -40,3 +40,6 @@ ignore:
   - "pkg/engine/engine_dcos.go"
   - "pkg/engine/templates_generated.go"
   - "pkg/i18n/translations_generated.go"
+  - "pkg/engine/const.go"
+  - "pkg/api/common/const.go"
+  - "pkg/api/const.go"


mboersma:
I guess the const files could be considered a no-op, since they *should* always have 100% coverage. (Although we don't have any mechanism other than reviewers' eyeballs to prevent helper functions or other code from creeping into a const.go.)

Shouldn't codecov have reported a lower number for coverage on this PR, since we're now ignoring several files with complete coverage? Or are declarations not counted anyway?

CecileRobertMichon:
I think codecov already ignores const lines otherwise coverage would have increased in this PR.

jackfrancis:
It's based on the codecov results from this PR, which *only* add UT coverage:

https://github.com/Azure/aks-engine/pull/2399

It's likely that codcov wouldn't give any add'l credit for the changes in PR 2339 because the additional UT (for the func `kubernetesContainerAddonSettingsInit`) was probably being credited as already covered due to indirect traversal through existing upstream UT.

So why did we lose coverage?

CecileRobertMichon:
Not sure but https://codecov.io/gh/Azure/aks-engine/commit/f3effd97cc94abc3d4eed584bf4e26c6b2c2cb1c show that the const file is ignored 

Screen Shot 2019-12-05 at 3 16 19 PM

jackfrancis:
@CecileRobertMichon cool, thanks for chiming in. I think we should close this PR.

PR #2412, opened 2019-12-08
feat: BYOK support on os disk

andyzhangx:
**Reason for Change**:

[BYOK support](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disk-encryption) on os disk on agent node pools, this PR adds a new field(`diskEncryptionSetID`)
```
    "agentPoolProfiles": [
      {
        "name": "agentpool",
        ...
        "diskEncryptionSetID": "/subscriptions/{subs-id}/resourceGroups/{rg-name}/providers/Microsoft.Compute/diskEncryptionSets/{diskEncryptionSet-name}" # ResourceId of the disk encryption set to use for enabling encryption at rest
      }
```

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
[PR#84605](https://github.com/kubernetes/kubernetes/pull/84605) already added BYOK support on data disk in azure disk in-tree driver from k8s v1.17.0


devigned:
The code looks good, but it would be great to see some unit tests as well as an E2E test. 

If I remember correctly from past experiences with disk encryption, the tests we wrote made clear several error cases which we didn't expect simply by setting the encryption key id.

andyzhangx:
> The code looks good, but it would be great to see some unit tests as well as an E2E test.
> 
> If I remember correctly from past experiences with disk encryption, the tests we wrote made clear several error cases which we didn't expect simply by setting the encryption key id.

@devigned thanks for review, I have already added the tests. If user inputs an incorrect DiskEncryptionSetID format, there would be error in validation step:
```
Error: validating API model after populating values: DiskEncryptionSetID(diskEncryptionSets/andydes) is of incorrect format, correct format: ^/subscriptions/\S+/resourceGroups/\S+/providers/Microsoft.Compute/diskEncryptionSets/[^/\s]+$
```

/assign @jackfrancis @CecileRobertMichon @mboersma

This PR is the final PR to make BYOK feature work in aks-engine.
After this PR merged into master branch, I would like to make this feature into `aks-release-v0.x.x` branch, how can I achieve that?
BTW, this work is in my highest priority, pls help code review, thanks!

devigned:
@andyzhangx, the unit tests are solid. Though, I wasn't referring only to unit tests, but to end to end tests as I previously stated (see: https://github.com/Azure/aks-engine/tree/master/test/e2e/test_cluster_configs).

I'm sure if a resource id is provided for disk encryption sets and passes validation, it should "just work", but it would feel much safer with an end to end test to ensure all parts of the system are working and will not regress in the future.

I'm happy to work with you to add this test.

If anyone else on @Azure/aks-engine feels like we don't need to validate e2e, then please overrule. I've just been burnt by disk encryption in the past through almost every phase of the lifecycle (creating encrypted volumes, maintaining, backing up, etc...).

marosset:
@devigned more coverage rarely hurts

jackfrancis:
@devigned I assume we'd need to manage a highly available `diskEncryptionSet` resource in order to make a test, right?

devigned:
@jackfrancis either that, or add creation to the test the test suite. The latter would provide guidance on how to use the feature from beginning to end, which is always nice for cryptic features (pun intended).

andyzhangx:
> @andyzhangx, the unit tests are solid. Though, I wasn't referring only to unit tests, but to end to end tests as I previously stated (see: https://github.com/Azure/aks-engine/tree/master/test/e2e/test_cluster_configs).
> 
> I'm sure if a resource id is provided for disk encryption sets and passes validation, it should "just work", but it would feel much safer with an end to end test to ensure all parts of the system are working and will not regress in the future.
> 
> I'm happy to work with you to add this test.
> 
> If anyone else on @Azure/aks-engine feels like we don't need to validate e2e, then please overrule. I've just been burnt by disk encryption in the past through almost every phase of the lifecycle (creating encrypted volumes, maintaining, backing up, etc...).

thanks @devigned it would be great if you could help, I have created one issue to track this: https://github.com/Azure/aks-engine/issues/2420

 this feature is in preview and can only run in below two regions now:
 - `west central us`
 - `centraluseuap` (not suggested, it always run out of cpu core quota)

Will send you more details about this BYOK feature.

docs/topics/clusterdefinitions.md
```diff
@@ -578,6 +578,7 @@ A cluster can have 0 to 12 agent pool profiles. Agent Pool Profiles are used for
...
+| diskEncryptionSetID                                            | no                                                             | Specifies ResourceId of the disk encryption set to use for enabling encryption at rest (ie: `"/subscriptions/{subs-id}/resourceGroups/{rg-name}/providers/Microsoft.Compute/diskEncryptionSets/{diskEncryptionSet-name}"`).                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       |
```
CecileRobertMichon: if this feature is not available in all regions can we make sure to point that out here for aks-engine users

andyzhangx: added an official link

andyzhangx:
Thanks for the help @devigned
I have added the e2e test, while this test may not work now since our testing subs is not whitelisted by BYOK team. Anyway, pls take a review.
cc @raeitan

test/e2e/test_cluster_configs/byok.json
```diff
@@ -0,0 +1,42 @@
+{
+	"env": {
```
andyzhangx: @devigned is this config all for such an e2e test? do I need to configure sth else?

devigned: Yes, this is all that’s needed for an Jenkins e2e test. Jenkins won’t run this until a job is triggered or created to run this branch / fork. Once in master, it’ll run periodically. Perhaps work with @jackfrancis tomorrow to create that job to verify this config.

jackfrancis: FYI, I've added a `INFRA_RESOURCE_GROUP` env var to the E2E template generation processor, and we're injecting that and the `SUBSCRIPTION_ID` env var into the `diskEncryptionSetID` property via string replacement, so we can formulate the URI per-test via env vars.

andyzhangx: @jackfrancis thanks, I have configured a `diskEncryptionSet` and added into e2e test config:
```
"/subscriptions/SUB_ID/resourceGroups/RESOURCE_GROUP/providers/Microsoft.Compute/diskEncryptionSets/des"
```

jackfrancis: I've put together a one-off that successfully deploys a cluster using a VMSS pool w/ `diskEncryptionSetID`. How do we actually test this functionality from the Kubernetes layer? How can I prove that the OS disk on these VMSS instances have the desired configuration?

andyzhangx: @jackfrancis
good question, here is the full E2e test, already verified on an aks-engine cluster:
```
### E2E test
1. Create a cluster with BYOK feature enabled
2. Make sure all agent nodes are up
3. Revoke DiskEncryptionSet access to azure key vault
4. After a few minutes, make sure all agent nodes are down
5. Grant DiskEncryptionSet access to azure key vault
6. After a few minutes, make sure all agent nodes are up
```

Not sure whether aks-engine test CI supports such testing scenario, and also need to "translate" below DiskEncryptionSet related script to golang code if we want to write a fully automatic e2e test:
https://github.com/andyzhangx/demo/blob/master/aks/byok/create-diskencryptionset.sh

Shall we go with this PR first? Thanks.

jackfrancis: > 3. Revoke DiskEncryptionSet access to azure key vault

How do you do this?

> 5. Grant DiskEncryptionSet access to azure key vault

How do you do this?

andyzhangx: @jackfrancis by be

> > 1. Revoke DiskEncryptionSet access to azure key vault
> 
> How do you do this?
> 
> > 1. Grant DiskEncryptionSet access to azure key vault
> 
> How do you do this?

@jackfrancis by below command in https://github.com/andyzhangx/demo/blob/master/aks/byok/create-diskencryptionset.sh
```
az role assignment create --assignee $desIdentity --role Reader --scope $keyVaultId
az role assignment delete --assignee $desIdentity --role Reader
```
How can we support such a test case? Is there already an example?

andyzhangx:
just took a look at the e2e test:
https://github.com/Azure/aks-engine/blob/master/docs/community/developer-guide.md#end-to-end-tests
I already added such a test config as `test/e2e/test_cluster_configs/byok.json`.

while I was also asked to prove that test is running on BYOK enabled cluster, I would say that by adding `diskEncryptionSet` field in VM template, it's already BYOK enabled, it's ensured by BYOK feature by azure storage team.
```
        "storageProfile": {
          "imageReference": "[variables('imageReference')]",
          "osDisk": {
            "name": "[variables('diskName')]",
            "createOption": "FromImage",
            "managedDisk": {
              "storageAccountType": "[parameters('diskSkuName')]",
              "diskEncryptionSet": {
                "id": "/subscriptions/{subs-id}/resourceGroups/{rg-name}/providers/Microsoft.Compute/diskEncryptionSets/{diskEncryptionSet-name}"
              }
            }
          },
```

To further ensure it's really a BYOK enabled cluster, here is the testing scenario I am doing on AKS(already working now, wait for this PR merge):
```
### E2E test
1. Create a cluster with BYOK feature enabled
2. Make sure all agent nodes are up
3. Revoke DiskEncryptionSet access to azure key vault
4. After a few minutes, make sure all agent nodes are down
5. Grant DiskEncryptionSet access to azure key vault
6. After a few minutes, make sure all agent nodes are up
```

I doubt whether aks-engine e2e test framework would support this scenario, since current e2e test framework is based on a cluster definition file, and then all tests are based on that to make sure those e2e tests all pass.

test/e2e/test_cluster_configs/byok.json
devigned: I think what @jackfrancis was really trying to get at is whether or not DES will be observable from Kubernetes. Would some process running on the node be able to tell that the underlying infrastructure has DES enabled without asking Azure Resource Manager? Is this feature completely transparent to the node? Could someone check `blkid` or something to see type of block device type, or is it at rest encryption and only known by Azure?

andyzhangx: > I think what @jackfrancis was really trying to get at is whether or not DES will be observable from Kubernetes. Would some process running on the node be able to tell that the underlying infrastructure has DES enabled without asking Azure Resource Manager? Is this feature completely transparent to the node? Could someone check `blkid` or something to see type of block device type, or is it at rest encryption and only known by Azure?

@devigned it's a [server side encryption](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disk-encryption), not Azure Disk Encryption, `Azure Disk Encryption` is another feature:
"
Azure Disk Encryption leverages the BitLocker feature of Windows and the DM-Crypt feature of Linux to encrypt managed disks with customer-managed keys within the guest VM. Server-side encryption with customer-managed keys improves on ADE by enabling you to use any OS types and images for your VMs by encrypting data in the Storage service.
"
details: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/disk-encryption#server-side-encryption-versus-azure-disk-encryption

devigned: Yep. That is what I had figured. The VM should have no idea that this feature is enabled. 

The extent that we can test this seems to be Azure accepts the deployment template, provisions the infra with configured data encryption, and everything runs as expected.

andyzhangx: > Yep. That is what I had figured. The VM should have no idea that this feature is enabled.
> 
> The extent that we can test this seems to be Azure accepts the deployment template, provisions the infra with configured data encryption, and everything runs as expected.

so shall we merge this PR now? Thanks.

jackfrancis: O.K., we have basic, working E2E coverage of the ARM deployment + BYOK scenario. That will do for now.

PR #2413, opened 2019-12-08
feat: add support for Kubernetes 1.17.0

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.17.md

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This will fail until Kubernetes 1.17.0 is actually released (presumably on or after 12/9), and until the Windows support .zip is mirrored.


mboersma:
/hold

We need to create and mirror the Windows support .zip after the release is final.

PR #2415, opened 2019-12-09
feat: add linux CSE tracing and metrics  

### LINKED ISSUE #2248: Instrument CSE scripts for Linux
// TODO @devigned

devigned:
**Reason for Change**:

It would help development if we were able to have more insight into the behavior of aks-engine.

This PR introduces a new command line tool for sending telemetry to Application Insights, [apmz](https://github.com/devigned/apmz) (docs in progress). `apmz` outputs each trace and metric to an intermediate file, then flushes them at the end. This behavior is only activated when the feature is turned on in the API model. When the feature is not active, there are still stub no-op bash functions available.

### To query this data, you can run the following kql
```
customMetrics
| where customDimensions.os_type == "linux" and customDimensions.correlation_id == "41b61b03-ae2d-495d-a0f7-869afb024ed3"
| order by timestamp asc 
```

**Issue Fixed**:

Fixes #2248 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
@devigned i opened a new PR with only the apimodel changes for enabling telemetry. Let's try and get that merged and we can work on enabling telemetry for linux and windows CSEs independently
 https://github.com/Azure/aks-engine/pull/2417

marosset:
this PR is much cleaner now :)

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -55,6 +55,7 @@ ERR_CIS_APPLY_PASSWORD_CONFIG=115 {{/* Error applying CIS-recommended passwd con
 
 ERR_VHD_FILE_NOT_FOUND=124 {{/* VHD log file not found on VM built from VHD distro */}}
 ERR_VHD_BUILD_ERROR=125 {{/* Reserved for VHD CI exit conditions */}}
+ERR_APMZ_INSTALL_ERROR=126 {{/* Unable to install apmz telemetry client */}}
```
marosset: Are you planning on having this be a fatal error?
I've been debating what we should do if we can't get the app insights bins on the Windows side.

devigned: Probably, not.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -177,6 +178,17 @@ downloadContainerd() {
     retrycmd_get_tarball 120 5 "$CONTAINERD_DOWNLOADS_DIR/${CONTAINERD_TGZ_TMP}" ${CONTAINERD_DOWNLOAD_URL} || exit $ERR_CONTAINERD_DOWNLOAD_TIMEOUT
 }
 
+installAPMZ() {
```
CecileRobertMichon: are we going to pre-install this on the vhd? Also, does uploading metrics require outbound internet? if so, should be disable it/skip it in "locked down environments"?

devigned: I'd like to pre-install. What do you think about that?

pkg/engine/templates_generated.go
```diff
@@ -13357,6 +13401,21 @@ source {{GetCSEHelpersScriptFilepath}}
...
+function exitAndFlush {
+  if [[ $? != "0" ]]; then
+    trace_err "cse-exit" code=$?
+  else
+    trace_info "cse-exit" code=$?
+  fi
+  apmz batch -f ${TMP_APMZ_BATCH_FILE} --api-key {{GetApplicationInsightsTelemetryKey}}
+}
+
+trap exitAndFlush EXIT
```
CecileRobertMichon: this is 💯 

I meant to do something like this a while ago and actually didn't know you could do it like this, just learned something

pkg/engine/templates_generated.go
```diff
@@ -13357,6 +13401,21 @@ source {{GetCSEHelpersScriptFilepath}}
...
+{{- if HasTelemetryEnabled }}
+installAPMZ
+
+function exitAndFlush {
+  if [[ $? != "0" ]]; then
+    trace_err "cse-exit" code=$?
+  else
+    trace_info "cse-exit" code=$?
+  fi
+  apmz batch -f ${TMP_APMZ_BATCH_FILE} --api-key {{GetApplicationInsightsTelemetryKey}}
```
CecileRobertMichon: does this ever fail intermittently? Worth retrying?

pkg/engine/templates_generated.go
```diff
@@ -13176,6 +13208,17 @@ downloadContainerd() {
     retrycmd_get_tarball 120 5 "$CONTAINERD_DOWNLOADS_DIR/${CONTAINERD_TGZ_TMP}" ${CONTAINERD_DOWNLOAD_URL} || exit $ERR_CONTAINERD_DOWNLOAD_TIMEOUT
 }
 
+installAPMZ() {
+  apmz_filepath=/usr/local/bin/apmz; apmz_url="https://github.com/devigned/apmz/releases/download/v0.0.1/apmz_v0.0.1_linux_amd64.tar.gz"
```
CecileRobertMichon: why is this download path in your fork?

parts/k8s/cloud-init/artifacts/cse_helpers.sh
CecileRobertMichon: +1 on not making it fatal, right now looks like it is

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -55,6 +55,7 @@ ERR_CIS_APPLY_PASSWORD_CONFIG=115 {{/* Error applying CIS-recommended passwd con
 
 ERR_VHD_FILE_NOT_FOUND=124 {{/* VHD log file not found on VM built from VHD distro */}}
 ERR_VHD_BUILD_ERROR=125 {{/* Reserved for VHD CI exit conditions */}}
```
CecileRobertMichon: 120s are reserved for VHD-related errors, let's make the new code 130+

parts/k8s/cloud-init/artifacts/cse_install.sh
CecileRobertMichon: actually we can probably add `{{- if HasTelemetryEnabled }}` here

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -177,6 +178,17 @@ downloadContainerd() {
...
+  apmz_filepath=/usr/local/bin/apmz; apmz_url="https://github.com/devigned/apmz/releases/download/v0.0.1/apmz_v0.0.1_linux_amd64.tar.gz"
```
CecileRobertMichon: why is this download path in your fork?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -19,6 +20,21 @@ source {{GetCSEHelpersScriptFilepath}}
...
+{{- if HasTelemetryEnabled }}
+installAPMZ
+
+function exitAndFlush {
+  if [[ $? != "0" ]]; then
+    trace_err "cse-exit" code=$?
+  else
+    trace_info "cse-exit" code=$?
+  fi
+  apmz batch -f ${TMP_APMZ_BATCH_FILE} --api-key {{GetApplicationInsightsTelemetryKey}}
```
CecileRobertMichon: does this ever fail intermittently? Worth retrying?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -19,6 +20,21 @@ source {{GetCSEHelpersScriptFilepath}}
...
+}
+
+trap exitAndFlush EXIT
```
CecileRobertMichon: this is 💯

I meant to do something like this a while ago and actually didn't know you could do it like this, just learned something

parts/k8s/cloud-init/artifacts/cse_main.sh
devigned: retry is handled within apmz

parts/k8s/cloud-init/artifacts/cse_install.sh
devigned: Well... I wrote `apmz`. I'm happy to move it to Azure if we like what's going on with it :).

CecileRobertMichon: aaah nvm :)

parts/k8s/cloud-init/artifacts/cse_install.sh
marosset: +1 for pre-installing

marosset: But also keep the install path in there if customers are using other distros

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -265,4 +265,34 @@ sysctl_reload() {
...
+}
+{{else}}
+trace_err() {
+    return
+}
+trace_info() {
+    return
+}
+time_metric() {
+    return
```
devigned: This should be a call to `"$@"` rather than return.

parts/k8s/cloud-init/artifacts/cse_install.sh
devigned: `apmz` is now pre-installed and will check to verify the version installed matches the version symlinked.

examples/kubernetes.json
```diff
@@ -29,6 +29,9 @@
     "servicePrincipalProfile": {
       "clientId": "",
       "secret": ""
+    },
+    "featureFlags": {
+      "enableTelemetry": true
```
CecileRobertMichon: I think we want to revert this change to the main example file before the PR is merged

devigned: I left it there intentionally, but in hindsight, I think you are right. I'll remove it.

devigned:
/hold

Going to move `apmz` into acs-mirror and then change the install path.

devigned:
This PR is only waiting on getting ampz v0.4.0 into the prod storage bucket. Right now, it points to the staging bucket. Once I update it to point at the prod bucket, I'll remove the WIP tag.

@jchauncey, let me know if you need any additional info to approve the prod deploy jobs.

devigned:
@marosset, please give this another look when you have a sec.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -39,22 +47,23 @@ fi
...
 if [ -f /var/run/reboot-required ]; then
     REBOOTREQUIRED=true
+    trace_info "RebootRequired" "reboot=true"
 else
     REBOOTREQUIRED=false
 fi
 
 {{- if not NeedsContainerd}}
-cleanUpContainerd
+time_metric "CleanupContainerd" cleanUpContainerd
```
jackfrancis: @palma21 FYI, AKS Engine CSE is now instrumented for App Insights telemetry. The `time_metric` func above is a wrapper around all CSE step executions, which is disabled by default, and is a no-op.

Just FYI.

PR #2416, opened 2019-12-09
refactor: make coredns addon user-configurable

jackfrancis:
**Reason for Change**:

This PR makes the `coredns` component a user-configurable addon, exposed via the existing `kubernetesConfig.addons` interface.

This PR also includes some rationalizations to the existing `kube-dns` addon behaviors. Prior, the kube-dns/coredns gating was handled via the Kubernetes 1.11/1.12 boundary, respectively. Going forward, we will maintain both `kube-dns` and `coredns` as options, with coredns being the default.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


evmimagina:
Hi,

Does it mean we can edit now the coredns settings and change the cluster name with out it getting rewrited automatically?

Thanks,

jackfrancis:
Hi @evmimagina, probably not just w/ this change, is there an issue w/ those configuration requirements? This change will make it trivial to implement that.

evmimagina:
Hi @jackfrancis , 

The issue I'm facing is that is not possible to rewrite the default domain name for the K8S cluster (cluster.local) once it has been created (I know that when creating the AKS instance there's a parameter to set the cluster dns default name) but the thing here is that I'm using terraform to provision the AKS ( more Info: _https://www.terraform.io/docs/providers/azurerm/r/kubernetes_cluster.html_ ) …  and it seems there's no way to adjust that JUST ON coredns configmap line without it being rewrited.

What I'm doing these days is to literaly rewrite every DNS query to my desired DNS name using the coredns-custom configmap in the following way;

"
apiVersion: v1
kind: ConfigMap
metadata:
  name: coredns-custom
  namespace: kube-system  
data:
  test.server: |
  customdns_sld.com:53 {
    #log #optional for debug purposes
    rewrite stop {
        name regex (.*)\.k8s\.customdns_sld\.com.$ {1}.cluster.local
        answer name (.*)\.cluster\.local\.$ {1}.customdns_sld.com
    }
    forward . $(AKS_SERVICE_DNS_IP)
   }
"

I must say it wasn't easy to build these config lines (the documentation regarding this subject is not that good :D ) and I don't like it to make rewrites of DNS with every query to any internal service within the cluster, but... if you are wondering, why would I want to specify a desired cluster dns name? well, easy, if i want a pod to talk using SSL to an internal K8S service it's not possible without a valid domain name, right?

Regards,

jackfrancis:
Related to #2251

PR #2417, opened 2019-12-09
feat: Adding some settings for telemetry collection to api model

marosset:
**Reason for Change**:

This change adds fields needed to collect application insights telemetry data to the apimodel with a new telemetryProfile.

All collection of telemetry will be held behind the 'enableTelemetry' feature flag which must be set in order to collect telemetry.
Additionally no parts of aks-engine are currently instrumented to collect telemetry but multiple prototypes are being developed to collect timing metrics for linux and windows custom script extensions.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


pkg/engine/armvariables_test.go
```diff
@@ -143,8 +143,10 @@ func TestK8sVars(t *testing.T) {
 		"apiVersionManagedIdentity":      "2015-08-31-preview",
 		"apiVersionNetwork":              "2018-08-01",
 		"apiVersionStorage":              "2018-07-01",
+		"applicationInsightsKey":         "906ba2b3-f3b7-4cbe-ba9f-885d6982de43", // should be DefaultApplicationInsightsKey,
```
marosset: @devigned I can't seem to DefaultApplicationInsightsKey from the api pkg here. Is there something i need to do to refresh to version of the package that is getting imported?

devigned: taking a look...

devigned: I guess I'm not sure what the problem is. Looks like this unit test is passing. Can you explain a bit more or add a failing test?

PR #2418, opened 2019-12-09
chore: Add new VM skus (Ev4, ND40, ...)

serbrech:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/helpers/azureconst.go
```diff
@@ -464,44 +512,48 @@ func GetKubernetesAllowedVMSKUs() string {
         "Standard_E16-4s_v3",
         "Standard_E16-8s_v3",
         "Standard_E16_v3",
-        "Standard_E16a_v3",
```
serbrech: is that legit? some skus are gone? @mboersma @palma21

pkg/helpers/azureconst.go
```diff
@@ -605,7 +658,12 @@ func GetKubernetesAllowedVMSKUs() string {
         "Standard_ND12s",
         "Standard_ND24rs",
         "Standard_ND24s",
+        "Standard_ND40rs_v2",
```
serbrech: Here are the skus we're looking for in particular

pkg/helpers/azureconst.go
serbrech: looks legit.. https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes-memory#eav4-series

serbrech:
so, you update this line each time you add a new sku to the supported list? how is that sustainable?
https://github.com/Azure/aks-engine/blob/b89e487a3d1583ac874a1a6389d13daa30bcb5f5/pkg/engine/armvariables_test.go#L425

pkg/api/common/helper.go
```diff
@@ -103,6 +103,8 @@ func IsNvidiaEnabledSKU(vmSize string) bool {
 		"Standard_NC12s_v3":  true,
 		"Standard_NC24s_v3":  true,
 		"Standard_NC24rs_v3": true,
+		"Standard_ND40s_v3":  true,
+		"Standard_ND40rs_v2": true,
```
CecileRobertMichon: Standard_ND40rs_v2 used the Tesla P40 GPU, it should be on line 96
I don't know about Standard_ND40s_v3 (it's not listed in the docs but I assume it's also a P40 GPU since it's part of the ND series).

source https://docs.microsoft.com/en-us/azure/virtual-machines/linux/sizes-gpu

serbrech: no they use V100 as per documentation : 
https://docs.microsoft.com/en-us/azure/virtual-machines/linux/sizes-gpu#ndv2-series-preview

serbrech: and correct about the not listed one, it's in preview so not really fixed

PR #2433, opened 2019-12-11
fix: kube-dns and coredns addons UT, back-compat

jackfrancis:
**Reason for Change**:

Follow up from #2393 and #2416. UT was not fully engaged, and thus the back-compat was not implemented exactly right. This PR fixes UT, and implements back-compat according to the following recipe:

- If you are in an upgrade context and your cluster version is >= 1.12 (when coredns support first landed in aks-engine):
  - Ensure that coredns addon is set to enabled if there is no coredns addon entry in the addons array
    - The absence of coredns in the addons array indicates a "legacy" cluster config, i.e., a cluster config built using aks-engine prior to coredns being added as a user-configurable addon
  - Ensure that kube-dns addon is set to disabled
    - This is statically done, essentially to mimic the legacy implementation, which statically disables kube-dns and statically enables coredns based on an upgrade from <= 1.12 to >= 1.12

This back-compat implementation includes some permissiveness for intrepid users who *manually* edit the addons array in their pre-existing api model prior to running `aks-engine upgrade`. Specifically, we only do the above default enforcement if the `coredns` addon is not present in the api model (which is the api model signature a legacy cluster will exhibit). What this means in practice is that users who *want* to continue to use `kube-dns` on a >= 1.12 cluster may manually modify the api model accordingly. UT have been added to prove this out.

Also, ensure we account for scenarios where coredns and kube-dns could be (incorrectly) enabled together.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons_test.go
+++ b/pkg/api/addons_test.go
@@ -3256,6 +3256,10 @@ func TestSetAddonsConfig(t *testing.T) {
 					Enabled: to.BoolPtr(false),
 				},
 				{
+					Name:    common.KubeDNSAddonName,
+					Enabled: to.BoolPtr(false),
+				},
+				{
 					Name:    common.CoreDNSAddonName,
 					Enabled: to.BoolPtr(DefaultCoreDNSAddonEnabled),
 					Config: map[string]string{
@@ -4252,7 +4256,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Containers: []KubernetesContainerSpec{
 						{
 							Name:  common.CoreDNSAddonName,
-							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.15.4"][common.CoreDNSAddonName],
+							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.16.0"][common.CoreDNSAddonName],
 						},
 					},
 				},
@@ -4458,7 +4462,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Containers: []KubernetesContainerSpec{
 						{
 							Name:  common.CoreDNSAddonName,
-							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.15.4"][common.CoreDNSAddonName],
+							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.16.0"][common.CoreDNSAddonName],
 						},
 					},
 				},
@@ -5836,7 +5840,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Containers: []KubernetesContainerSpec{
 						{
 							Name:  common.CoreDNSAddonName,
-							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.15.4"][common.CoreDNSAddonName],
+							Image: "KubernetesImageBase" + K8sComponentsByVersionMap["1.14.0"][common.CoreDNSAddonName],
 						},
 					},
 				},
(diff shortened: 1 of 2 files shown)


pkg/api/addons_test.go
```diff
@@ -3255,6 +3255,10 @@ func TestSetAddonsConfig(t *testing.T) {
 					Name:    common.CloudNodeManagerAddonName,
 					Enabled: to.BoolPtr(false),
 				},
+				{
+					Name:    common.KubeDNSAddonName,
```
jackfrancis: This was an error that wasn't caught because kube-dns wasn't actually being validated in this UT.

pkg/api/addons_test.go
```diff
@@ -4252,7 +4256,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Containers: []KubernetesContainerSpec{
 						{
 							Name:  common.CoreDNSAddonName,
-							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.15.4"][common.CoreDNSAddonName],
+							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.16.0"][common.CoreDNSAddonName],
```
jackfrancis: These fixes are actually functionally no-ops, as we ship the same coredns for many versions, but it's still something we want to address.

pkg/api/addons_test.go
```diff
@@ -7213,34 +7217,27 @@ func TestSetAddonsConfig(t *testing.T) {
...
+						OrchestratorVersion: "1.13.11",
 						KubernetesConfig: &KubernetesConfig{
 							DNSServiceIP: DefaultKubernetesDNSServiceIP,
 							KubeletConfig: map[string]string{
 								"--cluster-domain": "cluster.local",
 							},
+							ClusterSubnet: DefaultKubernetesSubnet,
+							ProxyMode:     KubeProxyModeIPTables,
 							NetworkPlugin: NetworkPluginAzure,
-							Addons: []KubernetesAddon{
```
jackfrancis: The proper back-compat/upgrade flow would not have the kube-dns or coredns addons present

pkg/api/addons_test.go
```diff
@@ -7249,25 +7246,7 @@ func TestSetAddonsConfig(t *testing.T) {
 			expectedAddons: []KubernetesAddon{
 				{
 					Name:    common.HeapsterAddonName,
-					Enabled: to.BoolPtr(true),
```
jackfrancis: This change is due to increasing the version to 1.13 (which makes sense because we no longer support 1.12)

pkg/api/addons_test.go
```diff
@@ -7424,23 +7403,25 @@ func TestSetAddonsConfig(t *testing.T) {
...
-							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.15.4"][common.CoreDNSAddonName],
+							Image: specConfig.KubernetesImageBase + K8sComponentsByVersionMap["1.13.11"][common.CoreDNSAddonName],
 						},
 					},
 				},
 			},
 		},
 		{
-			name: "kube-dns enabled",
+			name: "upgrade w/ manual kube-dns enabled",
```
jackfrancis: This is the "user manually modifies his/her api model to explicitly continue using kube-dns, which is now an option since making kube-dns a user-configurable addon"

pkg/api/addons_test.go
```diff
@@ -7608,15 +7589,600 @@ func TestSetAddonsConfig(t *testing.T) {
 					},
 					Containers: []KubernetesContainerSpec{
 						{
-							Name:  common.KubeDNSAddonName,
+							Name:  "kubedns",
```
jackfrancis: Again, more errors slipping through because we weren't actually validating kube-dns in this UT.

jackfrancis:
Related to #2251

PR #2438, opened 2019-12-11
ci: use Jenkins Azure Service Principal instead of hardcoding default client ID/Secret variables in pipelines

CecileRobertMichon:
**Reason for Change**:

This allows easily customizing which subscription / SP Jenkins jobs should use. Also removes the need for client ID / client secret / tenant ID / subscription ID separate secret texts and overrides.

source: https://wiki.jenkins.io/display/JENKINS/Azure+Credentials+plugin

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


Jenkinsfile
```diff
@@ -66,16 +66,13 @@ def runJobWithEnvironment(jobCfg, jobName, version) {
...
-						def clientSecretOverride = opts?.clientSecret ? opts.clientSecret : 'AKS_ENGINE_3014546b_CLIENT_SECRET'
 						def creds = [
-								string(credentialsId: 'AKS_ENGINE_TENANT_ID', variable: 'TENANT_ID'),
-								string(credentialsId: clientIdOverride, variable: 'CLIENT_ID'),
-								string(credentialsId: clientSecretOverride, variable: 'CLIENT_SECRET'),
+								azureServicePrincipal(credentialsId: 'SERVICE_PRINCIPAL', subscriptionIdVariable: 'SUBSCRIPTION_ID', clientIdVariable: 'CLIENT_ID', clientSecretVariable: 'CLIENT_SECRET', tenantIdVariable: 'TENANT_ID'),
 								string(credentialsId: 'LOG_ANALYTICS_WORKSPACE_KEY', variable: 'LOG_ANALYTICS_WORKSPACE_KEY')
 							]
 
 						withCredentials(creds) {
```
jchauncey: instead of doing this it would be more idiomatic to call

```
withCredentials([azureServicePrincipal(params.CREDENTIALS)]) {
```

test/e2e/test_cluster_configs/windows/cse_benchmark_win_vhds.json
```diff
@@ -4,8 +4,6 @@
 	},
 	"options": {
 		"allowedOrchestratorVersions": ["latestReleasedVersion"],
-		"clientId": "AKS_ENGINE_e7b1dca4_CLIENT_ID",
```
marosset: These currently run in the same jenkins pipeline (the 1.16 matrix i believe) so we'll need to move them into two seperate pipelines.

Jenkinsfile
```diff
@@ -66,12 +66,8 @@ def runJobWithEnvironment(jobCfg, jobName, version) {
...
 					withEnv(envVars.collect{ k, v -> "${k}=${v}" }) {
 						// define any sensitive data needed for the test script
-						def clientIdOverride = opts?.clientId ? opts.clientId : 'AKS_ENGINE_3014546b_CLIENT_ID'
-						def clientSecretOverride = opts?.clientSecret ? opts.clientSecret : 'AKS_ENGINE_3014546b_CLIENT_SECRET'
 						def creds = [
-								string(credentialsId: 'AKS_ENGINE_TENANT_ID', variable: 'TENANT_ID'),
-								string(credentialsId: clientIdOverride, variable: 'CLIENT_ID'),
-								string(credentialsId: clientSecretOverride, variable: 'CLIENT_SECRET'),
+								azureServicePrincipal(params.SERVICE_PRINCIPAL),
 								string(credentialsId: 'LOG_ANALYTICS_WORKSPACE_KEY', variable: 'LOG_ANALYTICS_WORKSPACE_KEY')
```
CecileRobertMichon: @jchauncey what do you think of this? doesn't the array var make it easier to read since there are two items?

CecileRobertMichon:
/hold 
until all jobs have been modified

PR #2411, opened 2019-12-06
[WIP] feat: enable node-problem-detector addon by default

mboersma:
**Reason for Change**:
node-problem-detector gives apiserver more information to make better decisions about the cluster and (arguably) should be enabled by default.

**Issue Fixed**:
Fixes #2389

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

Issue #2389, opened 2019-12-03
Enable node-problem-detector by default


devigned:
**Describe the request**
It would be beneficial for node-problem-detector to be enabled by default.

**Explain why AKS Engine needs it**
It is or will be shortly included as an add-on for AKS Engine, but will be disabled. Thus, it won't help anyone by default.

**Describe the solution you'd like**
Enable node-problem-detector by default.

**Additional context**
https://github.com/Azure/aks-engine/pull/2371#issuecomment-561300237
#2364

PR #2444, opened 2019-12-12
test: check k8s events to see if node-problem-detector works

mboersma:
**Reason for Change**:
Adds an e2e test spec that creates a kernel message which node-problem-detector should turn into a Kubernetes event, if everything is wired up properly.

**Issue Fixed**:
Follows up on #2371
Refs #2389, #2411

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
I've tested this thoroughly with an apimodel that enables the node-problem-detector addon.

Issue #2281, opened 2019-11-07
kubelet failed to start on windows nodes with accelerated networking


yixwang:
**Describe the bug**
In one of our newly created cluster, we noticed that some windows nodes are missing from the cluster, when running `kubectl get nodes`. However, these nodes do exist and run healthy as shown on the Portal. 
Example:
Only one node show up. However, all three windows nodes in the node pool of VMSS on Portal are running ok.  
![image](https://user-images.githubusercontent.com/47333969/68372004-dee82f00-00f4-11ea-825a-ffe354f2d26f.png)
![image](https://user-images.githubusercontent.com/47333969/68372340-7ea5bd00-00f5-11ea-99a6-8c23a9272e2a.png)

After looking into kubelet.err.log on the windows nodes that are not recognized by Kubernetes. We found it has a lot of the following errors repeat many times:
```
At C:\k\kubeletstart.ps1:50 char:54
+ ...    $mgmtIPAfterNetworkCreate = Get-NetIPAddress $managementIP -ErrorA ...
+                                                     ~~~~~~~~~~~~~
    + CategoryInfo          : InvalidData: (:) [Get-NetIPAddress], ParameterBindingValidationException
    + FullyQualifiedErrorId : ParameterArgumentValidationError,Get-NetIPAddress
 
Get-NetIPAddress : Cannot validate argument on parameter 'IPAddress'. The argument is null. Provide a valid value for 
the argument, and then try running the command again.
```
Line 50 corresponds to the 3rd line in the following in the kubeletstart.ps1 file
```
for ($i=0;$i -lt 60;$i++)
    {
        $mgmtIPAfterNetworkCreate = Get-NetIPAddress $managementIP -ErrorAction SilentlyContinue
        if ($mgmtIPAfterNetworkCreate)
        {
            break
        }
        sleep -Milliseconds 1000
    }
```
Can someone help identify the root cause in this case? 

**Steps To Reproduce**

Create the cluster with the following apimodel.json
```
{
  "apiVersion": "vlabs",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorVersion": "1.15.4",
      "kubernetesConfig": {
        "networkPolicy": "azure",
        "enableRbac": true,
        "enableDataEncryptionAtRest": true,
        "loadBalancerSku": "Basic",
        "useManagedIdentity": true,
        "excludeMasterFromStandardLB": true,
        "apiServerConfig": {
[...]

PR #2450, opened 2019-12-12
feat: Move creation of 'ext' HNS network into cse for Windows

### LINKED ISSUE #2432: Move 'ext' network creation into CSE provisioning for Windows nodes
**Describe the request**
Currently an 'ext' network is created once on each Windows node running AzureCNI in the script that starts running the kubelet. This should be configured in the CSE and not in the script that starts the kubelet.

Failure to create the 'ext' network should manifest as a deployment failure. Currently it is confusing to customers to have a "successful" aks-engine / AKS node pool deployment but not have the nodes joined to the cluster because the kubelet never started.

This is also a supportability issue

Move logic to create the 'ext' network into the CSE

**Additional context**

marosset:
**Reason for Change**:

**Issue Fixed**:

Fixes #2432

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -226,15 +226,15 @@ try
 
         Write-Log "Write ca root"
         Write-CACert -CACertificate $global:CACertificate `
-                     -KubeDir $global:KubeDir
+            -KubeDir $global:KubeDir
 
         Write-Log "Write kube config"
         Write-KubeConfig -CACertificate $global:CACertificate `
-                         -KubeDir $global:KubeDir `
-                         -MasterFQDNPrefix $MasterFQDNPrefix `
-                         -MasterIP $MasterIP `
-                         -AgentKey $AgentKey `
-                         -AgentCertificate $global:AgentCertificate
+            -KubeDir $global:KubeDir `
+            -MasterFQDNPrefix $MasterFQDNPrefix `
+            -MasterIP $MasterIP `
+            -AgentKey $AgentKey `
+            -AgentCertificate $global:AgentCertificate
 
         Write-Log "Create the Pause Container kubletwin/pause"
         New-InfraContainer -KubeDir $global:KubeDir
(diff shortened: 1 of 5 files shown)


parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -275,12 +279,14 @@ try
                     -IdentitySystem "{{ GetIdentitySystem }}"
             }
 
-        } elseif ($global:NetworkPlugin -eq "kubenet") {
+            New-ExternalHnsNetwork
```
PatrickLang: just to clarify - is `New-ExternalHnsNetwork` not required for kubenet?

marosset: It isn't being created for kubenet today

marosset: So i assume it isn't

marosset: The removal is in the code block for if ($NetworkPlugin -eq "azure") { }

parts/k8s/windowsazurecnifunc.ps1
```diff
@@ -255,3 +255,38 @@ function GenerateAzureStackCNIConfig
...
+    Write-Log "Creating new HNS network `"ext`""
+    $externalNetwork = "ext"
+    $na = @(Get-NetAdapter -Physical)
+
+    if ($na.Count -eq 0) {
+        throw "Failed to find any physical network adapters"
+    }
+
+    # If there is more than one adapter, use the first adapter.
+    $managementIP = (Get-NetIPAddress -ifIndex $na[0].ifIndex -AddressFamily IPv4).IPAddress
```
PatrickLang: @pradipd - should this be smarter? @dineshgovindasamy said there may be cases where a NIC doesn't have an IP and we should choose another?

parts/k8s/kuberneteswindowssetup.ps1
marosset: Maybe @pradipd  or @dineshgovindasamy  can comment on what scenarios we should create the ext HNS network?

parts/k8s/windowsazurecnifunc.ps1
PatrickLang: This is re: https://github.com/Azure/aks-engine/issues/2281

pradipd: Yes. Actually, the (Get-NetAdapter -Physical) on L261 needs to be different for acceleratedNetworkingEnabledWindows.
I just haven't had a chance to figure out what it should be.

pradipd: Yes. Actually, the (Get-NetAdapter -Physical) on L261 needs to be different for acceleratedNetworkingEnabledWindows.
I just haven't had a chance to figure out what it should be.

parts/k8s/kuberneteswindowssetup.ps1
dineshgovindasamy: It should be created if you want to prevent that brief network disconnect on reboot

marosset: I updated to create the network for all network plugings.

parts/k8s/windowsazurecnifunc.ps1
PatrickLang: @pradipd ok - we'll consider that for a separate PR

PR #2456, opened 2019-12-13
test: improve resiliency of addons UT implementations

jackfrancis:
**Reason for Change**:

This PR will reduce the surface area of manual UT overhead when making changes to the addons implementations.

1) moves a lot of image URL strings to consts so that folks rev'ing versions can update in one place
2) optimizes the component verisons maps in `pkg/api/k8s_versions.go` so that common component versions that span Kubernetes versions resolve to a single reference
3) added reusability funcs/patterns to `pkg/api/addons_test.go` so that each test case only has to express the relevant addons configurations data required to prove the test case

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/k8s_versions.go
```diff
@@ -10,412 +10,137 @@ import (
 	"github.com/Azure/aks-engine/pkg/api/common"
 )
 
-// k8sComponentVersions defines "best known-working" component versions for each minor version of k8s
```
jackfrancis: This func comment is not true (perhaps it was at some point). In fact the only reason to keep this additional layer of indirection is to aid some UT implementations.

jackfrancis:
Related to #2251

PR #2459, opened 2019-12-16
feat: enable user-configurable platformUpdateDomainCount

jackfrancis:
**Reason for Change**:

Allow user-overridable `platformUpdateDomainCount` to accommodate Azure regions that are unable to accommodate the static default `3` (e.g., `centraluseuap`).

This PR essentially keeps the static `3` configuration in place (via defaults enforcement), i.e., it does not auto-configure the platform update domain count value based on region.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -614,6 +614,9 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade bool) {
 	if p.IsAzureStackCloud() && p.MasterProfile.PlatformFaultDomainCount == nil {
 		p.MasterProfile.PlatformFaultDomainCount = to.IntPtr(DefaultAzureStackFaultDomainCount)
 	}
+	if p.MasterProfile.PlatformUpdateDomainCount == nil {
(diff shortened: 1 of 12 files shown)


pkg/api/defaults.go
```diff
@@ -644,6 +647,9 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 		if p.IsAzureStackCloud() && profile.PlatformFaultDomainCount == nil {
 			profile.PlatformFaultDomainCount = to.IntPtr(DefaultAzureStackFaultDomainCount)
 		}
+		if profile.PlatformUpdateDomainCount == nil {
```
jackfrancis: @serbrech FYI, we are making PlatformUpdateDomainCount an api model property, and the default setting of `3` is being moved into defaults enforcement (prior, it was statically set in the ARM template). Thus, AKS will have to incorporate this defaults change, otherwise it will generate ARM templates with no PlatformUpdateDomainCount configuration, which will probably fail.

PR #2467, opened 2019-12-17
fix: apply new master node labels for k8s v1.18+ compatibility

### LINKED ISSUE #2265: node-role.kubernetes.io/master changes for 1.17
See this for the skinny:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kubernetes-dev/8ad1puP8hxo/pBfbAqyIAAAJ

mboersma:
**Reason for Change**:
To preserve the existing behavior of the deprecated `node-role.kubernetes.io/master`, future versions of Kubernetes will use these two [more targeted labels](https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-07-16-node-role-label-use.md#instructions-for-deployers).

**Issue Fixed**:
Fixes #2265

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
These do no harm to apply to earlier versions of Kubernetes, so this is the simplest implementation.

The added e2e spec will fail eventually in upgrade testing, which builds clusters with earlier versions of AKS Engine but may run tests from master. To fix that, we can skip that part of the test unless the k8s version is 1.17.1 (not released yet) or greater.

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -684,7 +684,8 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				labels := node.Metadata.Labels
 				Expect(labels).To(HaveKeyWithValue("kubernetes.io/role", role))
 				Expect(labels).To(HaveKey(fmt.Sprintf("node-role.kubernetes.io/%s", role)))
-				if role == "master" {
+				if role == "master" && common.IsKubernetesVersionGe(
+					eng.ExpandedDefinition.Properties.OrchestratorProfile.OrchestratorVersion, "1.17.1") {
 					Expect(labels).To(HaveKeyWithValue("node.kubernetes.io/exclude-from-external-load-balancers", "true"))
 					Expect(labels).To(HaveKeyWithValue("node.kubernetes.io/exclude-disruption", "true"))
 				}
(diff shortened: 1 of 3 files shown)


mboersma:
/hold

I'm running this to show it passes, then I'm going to add a commit on top to avoid it failing in upgrade e2e runs.

mboersma:
[This CI run](https://github.com/Azure/aks-engine/pull/2467/checks?check_run_id=353113538) shows that the change passes (modulo the westus2 flake for 1.15). Now I'll add a commit that should make this safe for e2e upgrade tests.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -684,6 +684,11 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				labels := node.Metadata.Labels
 				Expect(labels).To(HaveKeyWithValue("kubernetes.io/role", role))
 				Expect(labels).To(HaveKey(fmt.Sprintf("node-role.kubernetes.io/%s", role)))
+				if role == "master" && common.IsKubernetesVersionGe(
+					eng.ExpandedDefinition.Properties.OrchestratorProfile.OrchestratorVersion, "1.17.1") {
```
jackfrancis: Is this >= 1.17.1 correct? Shouldn't it be 1.18.0-alpha.1?

mboersma: Since we're applying the new labels to all k8s versions, it's arbitrary what version we check for here.

I chose 1.17.1 because it's not out yet, so this test would be effectively disabled until we have a new AKS Engine release that includes that minor k8s release. (I've tested it locally of course.) That would avoid the problems we had in the past with this test in our back-compat jobs.

PR #2471, opened 2019-12-17
refactor: make aad addon user-configurable

jackfrancis:
**Reason for Change**:

This PR creates an `aad` user-configurable addon, exposed via the existing `kubernetesConfig.addons` interface:

```
...
    "addons": [
        {
            "name": "aad",
            "enabled": true
        }
    ]
...
```

Strictly speaking, the above doesn't make sense in the absence of an `aadProfile` w/ `adminGroupID` (validation has been added for that), but this change makes it more explicit that the aad ClusterRoleBinding is being delivered via cloud-init to `/etc/kubernetes/addons` for reconciliation by `kube-addon-manager`; in addition, putting this into the user-configurable addons space makes future evolution/maintenance of AAD in the k8s application layer more tractable.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/kubernetesparams.t
```diff
@@ -1,19 +1,3 @@
-{{if .HasAadProfile}}
-    "aadTenantId": {
```
jackfrancis: The aadTenantId is not referenced anywhere (probably leftover from a prior implementation), so I've cleaned it up while I'm in this area.

Checked as well that the AKS codebase doesn't depend on this in an unexpected way.

pkg/api/vlabs/validate.go
```diff
@@ -636,9 +636,10 @@ func (a *Properties) validateAddons() error {
 				}
 			}
 
-			switch addon.Name {
-			case "cluster-autoscaler":
-				if to.Bool(addon.Enabled) {
+			// Validation for addons if they are enabled
+			if to.Bool(addon.Enabled) {
```
jackfrancis: Wrapping this entire case/switch in a "is this addon enabled?" condition fixed a really nasty bug due to my adding aad addon validation here. Essentially, the validation below is (almost) entirely only valid in the context of the addon being enabled, so we also get to remove the `if to.Bool(addon.Enabled) {` for each addon case.

There was one "if addon is disabled" validation for `cloud-node-manager`, which was moved out to a new `else` block below.

pkg/api/types_test.go
```diff
@@ -6063,6 +6063,57 @@ func TestGetKubeProxyFeatureGates(t *testing.T) {
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			hasAADAdminGroupID := test.properties.HasAADAdminGroupID()
+			if hasAADAdminGroupID != test.expectedHasAADAdminGroupID {
+				t.Errorf("expected featureGates %t, but got %t", test.expectedHasAADAdminGroupID, hasAADAdminGroupID)
```
mboersma: Should this error message say `"expected hasAADAdminGroupID %t"` instead?

pkg/api/types_test.go
```diff
@@ -6063,6 +6063,57 @@ func TestGetKubeProxyFeatureGates(t *testing.T) {
...
+			}
+			getAADAdminGroupID := test.properties.GetAADAdminGroupID()
+			if getAADAdminGroupID != test.expectedGetAADAdminGroupID {
+				t.Errorf("expected featureGates %s, but got %s", test.expectedGetAADAdminGroupID, getAADAdminGroupID)
```
mboersma: Same comment as above.

pkg/api/vlabs/types_test.go
```diff
@@ -729,3 +729,54 @@ func GetMockPropertiesWithCustomCloudProfile(name string, hasCustomCloudProfile,
...
+		},
+	}
+
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			hasAADAdminGroupID := test.properties.HasAADAdminGroupID()
+			if hasAADAdminGroupID != test.expectedHasAADAdminGroupID {
+				t.Errorf("expected featureGates %t, but got %t", test.expectedHasAADAdminGroupID, hasAADAdminGroupID)
```
mboersma: Ditto.

pkg/api/vlabs/types_test.go
```diff
@@ -729,3 +729,54 @@ func GetMockPropertiesWithCustomCloudProfile(name string, hasCustomCloudProfile,
...
+			}
+			getAADAdminGroupID := test.properties.GetAADAdminGroupID()
+			if getAADAdminGroupID != test.expectedGetAADAdminGroupID {
+				t.Errorf("expected featureGates %s, but got %s", test.expectedGetAADAdminGroupID, getAADAdminGroupID)
```
mboersma: Ditto.

jackfrancis:
Related to #2251

PR #2443, opened 2019-12-12
chore: Updating Azure NPM to v1.0.30

jaer-tsun:
**Reason for Change**:

Updates to the latest release of Azure NPM:
- removing telemetry sidecar from NPM
- set NPM resource limits

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2472, opened 2019-12-17
chore: update azure-npm-daemonset addon to v1.0.30

jackfrancis:
**Reason for Change**:

Updates to the latest release of Azure NPM:

- removing telemetry sidecar from NPM
- set NPM resource limits

Derived from #2443, implementing on top of recent master due to lots of moving parts in the addons space at present.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@jaer-tsun I think this is what we want. Specifically:

- we need to update all the azure npm specs:
  1. the spec for all versions of k8s < 1.16
  2. the spec for k8s 1.16
  3. the spec for k8s 1.17
- we need to entirely remove the reference to the telemetry container everywhere
- we need to maintain back-compat with earlier specs (incl. the existence of the telemetry container)

UT have been updated accordingly.

jackfrancis:
FYI keeping the older versions of NPM + the telemetry container in the VHD in the event that AKS continues to use those older versions while continuing to pick up new VHDs.

jackfrancis:
Validated the new Azure NPM spec w/ 1.30 against k8s 1.15, 1.16, and 1.17

PR #2478, opened 2019-12-18
feat: Configuring docker log rotation for Windows nodes

marosset:
**Reason for Change**:

Windows nodes are configured with no log rotation and this can lead to disks filling up for long running clusters.
This PR sets the log rotation setting to be the same as what is set as default for linux machines.

**Issue Fixed**:

Fixes issue #373

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
For windows nodes docker logs by default get written to c:\ProgramData\docker\containers\{container id}\{container id}-json.log
The container logs in c:\var\log\containers are symlink's of the most recent log docker logs for each container.

PR #2473, opened 2019-12-18
chore: update cloud-provider-azure components to v0.4.0

mboersma:
**Reason for Change**:
See https://github.com/kubernetes-sigs/cloud-provider-azure/releases/tag/v0.4.0

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:
This still isn't enabled unless `useCloudControllerManager` is on.

PR #2479, opened 2019-12-18
refactor: change MCR base URL to /oss/kubernetes/

mboersma:
**Reason for Change**:
It appears that we're publishing images to `mcr.microsoft.com/oss/kubernetes/` now instead of `mcr.microsoft.com/k8s/core/`. This changes that base URL in AKS Engine code to match.

**Issue Fixed**:
Unblocks #2473

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
It looks like this change will only affect `pause`, `cloud-node-manager`, and `cloud-controller-manager` in practice. I've verified that all our expected images and tags are there at the new URL.


mboersma:
I'll work around this in #2473 instead.

PR #2480, opened 2019-12-18
refactor: make cilium addon user-configurable

jackfrancis:
**Reason for Change**:

This PR makes the `cilium` networkPolicy component a user-configurable addon, exposed via the existing `kubernetesConfig.addons` interface:

```
...
    "addons": [
        {
            "name": "cilium",
            "enabled": true
        }
    ]
...
```

Strictly speaking, the above doesn't make sense in the absence of a `networkPolicy` w/ a value of `"cilium"` (validation has been added for that), but this change allows for per-container image overrides to aid future development of this surface area.

Manual testing validated that our current 1.16 and 1.17 implementations *do not work*, so validation (and documentation) has been added to that effect.

Additionally, the simple `"networkPlugin": "ciilum"` input pattern has been deprecated, as it doesn't make sense: the cilium IPAM implementation depends upon artifacts delivered by the daemonset in the spec. It's possible there's a way to decompose the current, monolithic spec into IPAM- and NetworkPolicy-distinct specs, but that is an exercise for the future. Because concretely the `networkPolicy` needs to be `"cilium"`, thus to deliver the "cilium" addon, we validate that `"networkPlugin": "ciilum"` is not allowed in the api model as input without `"networkPolicy": "ciilum"`, to avoid misleading folks that it's in fact possible to use cilium just for IPAM (again, at present, that is not possible).

**Issue Fixed**:

Related to #2251

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -307,6 +307,11 @@ ensureKubelet() {
     KUBELET_RUNTIME_CONFIG_SCRIPT_FILE=/opt/azure/containers/kubelet.sh
     wait_for_file 1200 1 $KUBELET_RUNTIME_CONFIG_SCRIPT_FILE || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart kubelet || exit $ERR_KUBELET_START_FAIL
+    {{if HasCiliumNetworkPolicy}}
+    while [ ! -f /etc/cni/net.d/05-cilium.conf ]; do
```
jackfrancis: This change is to give a little extra time for the kubelet to reconcile before returning success to the user. The current implementation of cilium is to deliver this CNI dependency via a daemonset, which means container networking doesn't work for the first few mins of the kubelet coming online.

pkg/engine/armvariables.go
```diff
@@ -210,6 +210,12 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 		cloudInitFiles["auditdRules"] = getBase64EncodedGzippedCustomScript(auditdRules, cs)
 	}
 
+	if kubernetesConfig != nil {
+		if kubernetesConfig.NetworkPlugin == NetworkPluginCilium {
+			cloudInitFiles["systemdBPFMount"] = getBase64EncodedGzippedCustomScript(systemdBPFMount, cs)
```
jackfrancis: Moved this to the `cloudInitFiles` variable object because it's more idiomatic, and it saves ARM variable overhead.

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -165,21 +165,19 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			if cfg.BlockSSHPort {
 				Skip("SSH port is blocked")
 			} else {
-				var nodeList *node.List
+				var nodes []node.Node
 				var err error
 				if !eng.ExpandedDefinition.Properties.HasLowPriorityScaleset() {
-					nodeList, err = node.GetReady()
+					nodes, err = node.GetWithRetry(1*time.Second, cfg.Timeout)
```
jackfrancis: These changes came about due to poking at cilium clusters w/ E2E and seeing flakes due to the fact that a cilium cluster configuration initiates an OS reboot (thus the nodes may go offline during these tests and we need to retry).

test/e2e/test_cluster_configs/network/cilium.json
```diff
@@ -1,59 +0,0 @@
-{
```
jackfrancis: This test doesn't make sense: there's only one cilium config, and that's for implementing NetworkPolicy (plus cilium IPAM). There is no exclusive networkPlugin=cilium cluster configuration.

test/e2e/test_cluster_configs/network/kubenet_containerd.json
```diff
@@ -19,7 +19,7 @@
 				{
 					"name": "agentpool1",
 					"count": 2,
-					"vmSize": "Standard_D2_v3",
```
jackfrancis: Cleaning up while I was here

pkg/api/vlabs/validate_test.go
```diff
@@ -1521,6 +1524,109 @@ func TestValidateAddons(t *testing.T) {
...
+						Addons: []KubernetesAddon{
+							{
+								Name:    "cilium",
+								Enabled: to.BoolPtr(true),
+							},
+						},
+					},
+				},
+			},
+			expectedErr: errors.Errorf("%s addon may only be enabled the networkPolicy=%s", common.CiliumAddonName, NetworkPolicyCilium),
```
mboersma:
```suggestion
			expectedErr: errors.Errorf("%s addon may only be enabled if the networkPolicy=%s", common.CiliumAddonName, NetworkPolicyCilium),
```

pkg/api/vlabs/validate_test.go
```diff
@@ -1521,6 +1524,109 @@ func TestValidateAddons(t *testing.T) {
...
```
mboersma:
```suggestion
			expectedErr: errors.Errorf("%s addon may only be enabled if the networkPolicy=%s", common.CiliumAddonName, NetworkPolicyCilium),
```

pkg/api/vlabs/validate_test.go
```diff
@@ -1521,6 +1524,109 @@ func TestValidateAddons(t *testing.T) {
...
```
mboersma:
```suggestion
			expectedErr: errors.Errorf("%s addon may only be enabled if the networkPolicy=%s", common.CiliumAddonName, NetworkPolicyCilium),
```

pkg/engine/engine.go
```diff
@@ -626,6 +626,7 @@ func getBase64EncodedGzippedCustomScript(csFilename string, cs *api.ContainerSer
 	templ := template.New("ContainerService template").Funcs(getContainerServiceFuncMap(cs))
 	_, err = templ.Parse(string(b))
 	if err != nil {
+		fmt.Println("here")
```
mboersma: A debugging statement was left behind.

test/e2e/test_cluster_configs/network_policy/cilium.json
```diff
@@ -10,20 +11,19 @@
...
-					"networkPolicy": "cilium",
-					"networkPlugin": "cilium"
+					"networkPolicy": "cilium"
 				}
 			},
 			"masterProfile": {
 				"count": 1,
 				"dnsPrefix": "",
-				"vmSize": "Standard_D2_v3"
+				"vmSize": "Standard_D8_v3"
```
mboersma: Why the `vmSize` change?

jackfrancis: Seeing test flakes, this change didn't move the needle, will revert back. The `GINKGO_FOCUS` addition has proven to be a valid flake reducer.

PR #2481, opened 2019-12-18
feat: update containermonitoring addon for december release

rashmichandrashekar:
December release changes for Containermonitoring addon -
feat: High scale support
feat: 1.16 support
Complete set of changes here - https://github.com/microsoft/Docker-Provider/blob/ci_feature_prod/README.md


parts/k8s/containeraddons/1.16/kubernetesmasteraddons-omsagent-daemonset.yaml
```diff
@@ -67,79 +67,58 @@ data:
...
+      run_interval 60
       log_level debug
      </source>
 
     #Kubernetes health
     <source>
      type kubehealth
      tag kubehealth.ReplicaSet
-     run_interval 60s
+     run_interval 60
```
ganga1980: is this intentional removing units (i.e. s) for this field here?

rashmichandrashekar: yes. This is related to scheduler optimization

devigned:
Just a heads up the commit message doesn't follow semantic pull request rules. Please update it when you get a chance.

rashmichandrashekar:
> Just a heads up the commit message doesn't follow semantic pull request rules. Please update it when you get a chance.

I updated it after the PR was created. The rule suggested I add it in the title, and i did add a suggested prefix "improvement". The rule check doesnt seem to be picking up this change.. could you pls run the semantic PR check again?

devigned:
@rashmichandrashekar the problem is that "improvement" is not one of accepted works (I think).

parts/k8s/containeraddons/1.16/kubernetesmasteraddons-omsagent-daemonset.yaml
ganga1980: > yes. This is related to scheduler optimization

thanks.

jackfrancis:
@rashmichandrashekar we are currently only testing the container-monitoring addon on k8s v1.13-1.15. Is that correct? Or do you expect that this addon works on 1.16 and 1.17 also? (in which case we should add test coverage for those versions)

jackfrancis:
Run `make generate` and commit generated code to address unit test failure.

rashmichandrashekar:
> @rashmichandrashekar we are currently only testing the container-monitoring addon on k8s v1.13-1.15. Is that correct? Or do you expect that this addon works on 1.16 and 1.17 also? (in which case we should add test coverage for those versions)

@jackfrancis - Are you taking about the vesions here -  
aks-engine/test/e2e/test_cluster_configs/container_monitoring.json

PR #2460, opened 2019-12-16
refactor: remove unused orchestrators

mboersma:
**Reason for Change**:
Removes code that supported the [legacy ACS service](https://azure.microsoft.com/en-us/updates/azure-container-service-will-retire-on-january-31-2020/).

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2500, opened 2019-12-23
docs: design proposal for removing old orchestrators

mboersma:
**Reason for Change**:
Proposes a staged process for removing old DCOS, Swarm, etc. code in human-reviewable chunks.

**Issue Fixed**:
Refs #2460

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:
Any comments would be highly appreciated. 😄


mboersma:
The word on the street is that ACS may receive a stay of execution, so the old orchestrator code may need to be preserved (if not updated...).

PR #2502, opened 2019-12-23
refactor: remove SwarmMode/DockerCE orchestrator

mboersma:
**Reason for Change**:
Removes code supporting the SwarmMode/DockerCE orchestrator, as a first step toward removing all old orchestrator code in AKS Engine.

**Issue Fixed**:
Refs #2500

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This is still a `size/XXL` PR. Does it seem reviewable? Others may be of a similar size.


mboersma:
/hold

This is premature: I'm still waiting for feedback on #2500, we should wait until closer to the ACS end-of-life date, and it's the holiday season. 🎄

mboersma:
The word on the street is that ACS may receive a stay of execution, so the old orchestrator code may need to be preserved (if not updated...).

PR #2504, opened 2019-12-26
feat: cleaning up old kubelet/kubeproxy logs for Windows nodes

### LINKED ISSUE #2470: Purge old kubelet/kubeproxy/other logs on Windows nodes
**Describe the request**
Currently logs created by various Kubernetes components on Windows nodes do not get purged and as a result can eat up large amounts of disk space. 

aks-engine should configure log purging for windows nodes

1. **Clean up NSSM logs**
Today kubelet/kubeproxy are run inside NSSM and log to stdout/stderr and NSSM captures log output in .out.log and .err.log files. NSSM can also be configured to roll logs at specified time interval or when logs files reach a certain size. NSSM does not do any log cleanup.
One approach would be to create a scheduled task to keep the last N logs for each service.

- Pros
  - Simple approach
- Cons
  - We want to move away from NSSM in the future
  - We have observed logs not getting written when systems are under extreme load (but have not root caused this to be an issue with NSSM)
 
2. **Have kubelet/kubeproxy log to files**
Have kubelet/kubeproxy perform logging to files instead of capturing output through NSSM.

- Pros
  - Might have more consistent logging when system is under load
  - Would help us move away from NSSM
- Cons
  - Kubelet/kubeproxy still log everything to stdout/stderr when `--logtostderr=false` is passed meaning we would either need to stop logging all output from NSSM or also do # 1 (bug?? since there is also a `--alsologtostderr` flag which is not getting set)
  - Kubelet/Kubeproxy do not honor `--log-file-max-size` (bug??) so logs do not get rotated
  - Kubelet/kubeproxy create 3 logs (.INFO, .WARNING, .ERROR) which contain duplicate information

**Additional context**
AzureCNI and related binaries do perform log rolling/purging to keep log sizes acceptable.

marosset:
**Reason for Change**:

**Issue Fixed**:

Fixes #2470 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This change increases the size of kubelet.err.log/kubeproxu.err.log to 10Mb and adds a scheduled task to keep the 5 most recent rotated logs.

### DIFF
--- a/parts/k8s/windowslogscleanup.ps1
+++ b/parts/k8s/windowslogscleanup.ps1
@@ -1,0 +1,25 @@
+<#
+.DESCRIPTION
+    This script cleans old rotated logs for various kubernetes components.
+#>
+
+$global:LogPath = "c:\k\windowslogscleanup.log"
+
+filter Timestamp { "$(Get-Date -Format o): $_" }
+
+function Write-Log ($message) {
+    $message | Timestamp | Tee-Object -FilePath $global:LogPath -Append
+}
+
+Write-Log "Entering windowslogscleanup.ps1"
+
+$logFilePrefixes = @("kubelet", "kubelet.err", "kubeproxy", "kubeproxy.err")
+
+foreach ($logFilePrefix in $logFilePrefixes) {
+    $oldLogs = [IO.Directory]::GetFiles("c:\temp\kubelogs\logs", "$($logFilePrefix)-*.log")
+    $oldLogs = $oldLogs | Sort-Object | Select-Object -SkipLast 5
+    foreach ($oldLog in $oldLogs) {
+        Write-Log "Removing $oldLog"
+        Remove-Item $oldLog
+    }
+}
--- a/parts/k8s/windowslogscleanup.ps1
+++ b/parts/k8s/windowslogscleanup.ps1
@@ -16,7 +16,7 @@ Write-Log "Entering windowslogscleanup.ps1"
 $logFilePrefixes = @("kubelet", "kubelet.err", "kubeproxy", "kubeproxy.err")
 
 foreach ($logFilePrefix in $logFilePrefixes) {
-    $oldLogs = [IO.Directory]::GetFiles("c:\temp\kubelogs\logs", "$($logFilePrefix)-*.log")
+    $oldLogs = [IO.Directory]::GetFiles("c:\k", "$($logFilePrefix)-*.log")
     $oldLogs = $oldLogs | Sort-Object | Select-Object -SkipLast 5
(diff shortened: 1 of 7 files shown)


marosset:
Yea, we'll need to find a way to deal with errors in the scheduled task.
The solution I'd like to see is to move away from nssm and have the kubelet/kubeproxy manage their own logs but as I pointed out in #2470 there are some issues with this for Windows we should look into addressing.

parts/k8s/windowslogscleanup.ps1
```diff
@@ -0,0 +1,25 @@
+<#
+.DESCRIPTION
+    This script cleans old rotated logs for various kubernetes components.
+#>
+
+$global:LogPath = "c:\k\windowslogscleanup.log"
```
PatrickLang: that's captured in https://github.com/Azure/aks-engine/blob/3a2dcb2cb89fab9b05d50ff52a64e1d2155d82ac/scripts/collect-windows-logs.ps1#L7 - good

parts/k8s/windowslogscleanup.ps1
```diff
@@ -0,0 +1,25 @@
...
+function Write-Log ($message) {
+    $message | Timestamp | Tee-Object -FilePath $global:LogPath -Append
+}
+
+Write-Log "Entering windowslogscleanup.ps1"
+
+$logFilePrefixes = @("kubelet", "kubelet.err", "kubeproxy", "kubeproxy.err")
+
+foreach ($logFilePrefix in $logFilePrefixes) {
+    $oldLogs = [IO.Directory]::GetFiles("c:\temp\kubelogs\logs", "$($logFilePrefix)-*.log")
```
PatrickLang: is `c:\temp\kubelogs\logs` right? shouldn't it just be c:\k ?

marosset: Whoops, those were paths I was using the test the powershell and scheduled task locally.

parts/k8s/windowslogscleanup.ps1
```diff
@@ -0,0 +1,25 @@
...
+function Write-Log ($message) {
+    $message | Timestamp | Tee-Object -FilePath $global:LogPath -Append
+}
+
+Write-Log "Entering windowslogscleanup.ps1"
+
+$logFilePrefixes = @("kubelet", "kubelet.err", "kubeproxy", "kubeproxy.err")
+
+foreach ($logFilePrefix in $logFilePrefixes) {
+    $oldLogs = [IO.Directory]::GetFiles("c:\k", "$($logFilePrefix)-*.log")
```
PatrickLang: 👍

PR #2513, opened 2020-01-02
fix: configure addons before setting kubelet config

aramase:
**Reason for Change**:

As part of PR #2393, the following change was made to move addon configuration after setting kubelet config - https://github.com/Azure/aks-engine/commit/691e3d3c3a837abbb4f607d5e6b040b517af0f2a#diff-fd43f080184ed15731e5d7e0d917d6e1L486-L490. This change breaks setting kubelet config as kubelet configuration checks for addon configs while setting defaults. For instance `--non-masquerade-cidr` in kubelet is set to cluster subnet if `ip-masq-agent` is not enabled. But since the addon was configured after, kubelet assumes masq-agent is not enabled and sets the `--non-masquerade-cidr` to cluster subnet. This results in `ip-masq-agent` and `kubelet` both configuring iptables which could possibly break things.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

cc @lachie83

### DIFF
--- a/pkg/api/defaults-kubelet_test.go
+++ b/pkg/api/defaults-kubelet_test.go
@@ -519,6 +519,15 @@ func TestKubeletIPMasqAgentEnabledOrDisabled(t *testing.T) {
 		t.Fatalf("got unexpected '--non-masquerade-cidr' kubelet config value %s, the expected value is %s",
 			k["--non-masquerade-cidr"], DefaultNonMasqueradeCIDR)
 	}
+
+	// No ip-masq-agent addon configuration specified, --non-masquerade-cidr should be 0.0.0.0/0
+	cs = CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
+	cs.setKubeletConfig(false)
+	k = cs.Properties.OrchestratorProfile.KubernetesConfig.KubeletConfig
+	if k["--non-masquerade-cidr"] != DefaultNonMasqueradeCIDR {
+		t.Fatalf("got unexpected '--non-masquerade-cidr' kubelet config value %s, the expected value is %s",
+			k["--non-masquerade-cidr"], DefaultNonMasqueradeCIDR)
+	}
 }
 
 func TestEnforceNodeAllocatable(t *testing.T) {
--- a/pkg/api/defaults-kubelet_test.go
+++ b/pkg/api/defaults-kubelet_test.go
@@ -48,7 +48,7 @@ func TestKubeletConfigDefaults(t *testing.T) {
 		"--max-pods":                          strconv.Itoa(DefaultKubernetesMaxPods),
 		"--network-plugin":                    NetworkPluginKubenet,
 		"--node-status-update-frequency":      K8sComponentsByVersionMap[cs.Properties.OrchestratorProfile.OrchestratorVersion]["nodestatusfreq"],
-		"--non-masquerade-cidr":               DefaultKubernetesSubnet,
+		"--non-masquerade-cidr":               DefaultNonMasqueradeCIDR,
 		"--pod-manifest-path":                 "/etc/kubernetes/manifests",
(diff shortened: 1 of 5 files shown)


pkg/api/defaults.go
```diff
@@ -480,12 +480,14 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
...
-		// Configure kubelet
-		cs.setKubeletConfig(isUpgrade)
-
 		// Configure addons
+		// This needs to be configured first as set kubelet config depends
+		// on certains default addons to be populated (eg. ip-masq-agent)
 		cs.setAddonsConfig(isUpgrade)
 
+		// Configure kubelet
+		cs.setKubeletConfig(isUpgrade)
```
CecileRobertMichon: Based on https://github.com/Azure/aks-engine/pull/2393/files#r355038175, was there a reason for changing kubelet config defaults to before addons defaults? ie. is there a dependency introduced in #2393 between addons and kubelet config defaults?

CecileRobertMichon:
Let's add a unit test that fails before this change and passes after the change to prove this fixes the issue and protect from future breaking changes.

pkg/api/defaults.go
CecileRobertMichon: I think this is one example: https://github.com/Azure/aks-engine/blob/master/pkg/api/addons.go#L576

aramase: `--cluster-domain` is always set to `cluster.local`. So not sure if we need this change.

aramase: `ip-masq-agent` addon is enabled by default. It's only explicitly disabled by setting kubernetes config/ hosted master. So instead of checking if the addon is enabled, I've changed the check to see if it's explicitly disabled and then perform the logic. 

@CecileRobertMichon lmk what you think?

pkg/api/defaults-kubelet_test.go
```diff
@@ -499,6 +519,15 @@ func TestKubeletIPMasqAgentEnabledOrDisabled(t *testing.T) {
 		t.Fatalf("got unexpected '--non-masquerade-cidr' kubelet config value %s, the expected value is %s",
 			k["--non-masquerade-cidr"], DefaultNonMasqueradeCIDR)
 	}
+
+	// No ip-masq-agent addon configuration specified, --non-masquerade-cidr should be 0.0.0.0/0
```
jackfrancis: @aramase @CecileRobertMichon Actually, this additional UT is not what we want. If I add this coverage to the commit prior to #2393 (`$ git checkout 6e76fd9e69a7ee8aa61d2d3c559919a460bad5ed`) this test fails:

```
got unexpected '--non-masquerade-cidr' kubelet config value 10.240.0.0/12, the expected value is 0.0.0.0/0
```

We want to identify exactly what cluster configuration + defaults enforcement *was* working prior to #2393, and prove that this fix restores that functionality.

aramase: @jackfrancis That is because in the unit tests `ip-masq-agent` was never enabled. So these tests will fail in all versions prior to this. But going forward it will help us spot such issues.

jackfrancis: If this is truly a fix, then something must have broken in #2393 that can be expressed as a UT, right?

aramase: This was never unit tested before. There are individual unit tests for `setKubeletConfig` and `setAddonConfig`, but none which check the desired result after running both.

aramase: With this change the ip-masq-agent enabled/disabled isn't really dependent on `setAddonConfig` anymore as the check to see if ip-masq is enabled/disabled is done outside of that while setting non-masq-cidr. So I feel this unit test is appropriate. This unit test addition is essentially increasing the coverage for the path we didn't check before.

PR #2517, opened 2020-01-06
refactor: user-configurable flannel and scheduled maintenance addons

jackfrancis:
and remove legacy addons interface

**Reason for Change**:

This PR makes the `flannel` and `scheduled-maintenance` addons into the user-configurable area, exposed via the existing `kubernetesConfig.addons` interface:

```
...
    "addons": [
        {
            "name": "flannel",
            "enabled": true
        },
        {
            "name": "scheduled-maintenance",
            "enabled": true
        }
    ]
...
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/addons.go
```diff
@@ -421,7 +421,7 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 			},
 			{
 				Name:  common.AntreaInstallCNIContainerName,
-				Image: k8sComponents[common.AntreaInstallCNIContainerName],
+				Image: k8sComponents["antrea"+common.AntreaInstallCNIContainerName],
```
jackfrancis: Both flannel and antrea deliver a "install-cni" container in their respective specs, so this additional key prefix helps to disambiguate between the two in the container image reference hash table.

PR #2521, opened 2020-01-06
chore: Updating Azure NPM to v1.0.31

jaer-tsun:
**Reason for Change**:

Updates to the latest release of Azure NPM:
https://github.com/Azure/azure-container-networking/releases/tag/v1.0.31

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jaer-tsun:
@jackfrancis 
@mboersma

jaer-tsun:
oops, let me push the npm image first...

jaer-tsun:
ok I pushed NPM v1.0.31

packer/install-dependencies.sh
```diff
@@ -313,9 +313,9 @@ for AZURE_CNI_NETWORKMONITOR_VERSION in ${AZURE_CNI_NETWORKMONITOR_VERSIONS}; do
 done
 
 AZURE_NPM_VERSIONS="
+1.0.31
```
CecileRobertMichon: this is only adding the new version to the VHD build, not using it in code... I assume you also intended to use this new npm version by default?

jaer-tsun: yup I would like to use it by default

CecileRobertMichon: you also need to change https://github.com/Azure/aks-engine/blob/master/pkg/api/k8s_versions.go#L46 then

jaer-tsun: let me check the previous pr again

mboersma:
@jaer-tsun I mirrored the azure-vnet-cni 1.0.31 binaries to our production blob store. ✔️

packer/install-dependencies.sh
```diff
@@ -313,9 +313,9 @@ for AZURE_CNI_NETWORKMONITOR_VERSION in ${AZURE_CNI_NETWORKMONITOR_VERSIONS}; do
...
 1.0.30
 1.0.29
-1.0.28
```
jackfrancis: Let's not remove 1.0.28 from the VHD as part of this change

jaer-tsun: Is it ok to remove since 1.0.28 and below had quite a few bugs?

packer/install-dependencies.sh
```diff
@@ -313,9 +313,10 @@ for AZURE_CNI_NETWORKMONITOR_VERSION in ${AZURE_CNI_NETWORKMONITOR_VERSIONS}; do
...
+v.0.28
```
jackfrancis: typo

Issue #2377, opened 2019-12-02
Be able to configure both the number of IPs or the allocatedOutboundPorts


palmerabollo:
**Describe the request**

Be able to configure both the number of IPs or the allocatedOutboundPorts.

**Explain why AKS Engine needs it**

In a cluster with >63 nodes, different services started to fail, including the cluster-autoscaler (we’re using Kubernetes v1.13.10 and cluster-autoscaler:v1.13.6 in this environment), because of this limit:
 
> W1128 14:49:18.684311 1 utils.go:457] Failed to remove node azure:///subscriptions/REDACTED/resourcegroups/REDACTED/providers/microsoft.compute/virtualmachinescalesets/REDACTED-vmss/virtualmachines/1298: Code="SpecifiedAllocatedOutboundPortsForOutboundRuleExceedsTotalNumberOfAvailablePorts" Message="Specified Allocated Outbound Ports 1008 for Outbound Rule /subscriptions/REDACTED/resourceGroups/REDACTED/providers/Microsoft.Network/loadBalancers/REDACTED/outboundRules/LBOutboundRule exceeds total number of available ports 920. Reduce allocated ports or increase number of IP addresses for outbound rule."

On a fresh cluster created with aks-engine, there are some elements that affect outbound connectivity. If we're not wrong, at the default LoadBalancer:
- There is an IP linked as Frontend IP Config called k8s-agent-ip-outbound. Each IP supports 64000 ports.
- A Backend Pool that contains all NICs for any instance in all VMSS, excluding master nodes.
- There is an outbound rule that matches the Frontend IP Config described above with the Backend Pool. This outbound rule has configured by default 1024 allocated ports, which allows up to 62 instances across all VMSS to be created (64000 ports / 1024).
 
**Describe the solution you'd like**

We would like to be able to automate the creation of K8S clusters with aks-engine that support hundreds of nodes. We don’t see any **options in aks-engine to configure more IP addresses** (it seems to be the best option) **or to configure the allocatedOutboundPorts.**

**Describe alternatives you've considered**

As a workaround, we can reduce the allocatedOutboundPorts to 512 using az cli after the cluster is created.

**Additional context**

- What's your recommendation for deploying large clusters (e.g. 300 nodes) with private instances using aks-engine?
- We don't know if agentProfile.LoadBalancerBackendAddressPoolIDs can help somehow.

CecileRobertMichon:
@palmerabollo is this a feature request for AKS or AKS Engine? 

[...]

PR #2526, opened 2020-01-07
fix: default LB AllocatedOutboundPorts to 0

### LINKED ISSUE #2558: AllocatedOutboundPorts defaults to 1024, which is too low
At present, we do not specify a value for `AllocatedOutboundPorts`, and Azure is assigning the value `1024`, which caps SNAT ports at that number.

For larger clusters this is not sufficient. A `0` default (i.e., "automatic") is preferable, as it enables more than `1024` ports in practice for large clusters.

jackfrancis:
**Reason for Change**:

This PR sets the `AllocatedOutboundPorts` configuration to `0`, as this is a preferred default configuration for LB outbound rules.

`0` is equivalent to "automatic", which means is preferable to the Azure-provided default of `1024` , which essentially caps SNAT ports at that number.

See here for more info:

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview

**Issue Fixed**:

Fixes #2558

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/examples/kubernetes.json
+++ b/examples/kubernetes.json
@@ -2,10 +2,24 @@
   "apiVersion": "vlabs",
   "properties": {
     "orchestratorProfile": {
-      "orchestratorType": "Kubernetes"
+      "orchestratorType": "Kubernetes",
+      "kubernetesConfig": {
+        "loadBalancerSku": "Standard",
+        "excludeMasterFromStandardLB": true,
+        "privateCluster": {
+          "enabled": true,
+          "jumpboxProfile": {
+            "name": "jumpbox",
+            "vmSize": "Standard_D2_v3",
+            "osDiskSizeGB": 30,
+            "username": "azureuser",
+            "publickey": ""
+          }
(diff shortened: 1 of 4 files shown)


Michael-Sinz:
Setting this to 4096 is not a good default as it will keep clusters to under 15 nodes.

Setting this to 0 will use the adaptive behavior of the LB:

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview#snatports

The option/value to set a fixed smaller number (say 32, in order to be able to reach 1000 nodes like the automatic table says) to prevent the strange networking problem when crossing a VM count boundary as describe here:

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#preallocatedports

jackfrancis:
We will do #2377 as a follow-up PR to make this configurable

PR #2534, opened 2020-01-08
chore: update go-dev tools image

mboersma:
**Reason for Change**:
See https://github.com/deis/docker-go-dev/releases/tag/v1.25.2

Specifically updates `golangci-lint` to v1.23.1 and go itself to 1.13.6.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,6 +1,6 @@
 {
 	"name": "Go for aks-engine",
-	"image": "quay.io/deis/go-dev:v1.25.0",
+	"image": "quay.io/deis/go-dev:v1.25.1",
 	"extensions": [
 		"ms-vscode.go"
 	],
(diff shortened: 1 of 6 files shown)


jackfrancis:
lol

mboersma:
Interesting--I hit these same (apparently spurious) errors when I ran this locally, then it went away and I can't reproduce them now.

The problem is that these new warnings don't look actionable:


level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: analysis skipped: errors in package: [/__w/1/s/pkg/api/defaults.go:849:20: cannot use cidrFirstIP (variable of type net.IP) as net.IP value in argument to append /__w/1/s/pkg/api/defaults.go:855:23: cannot use ips (variable of type []net.IP) as []net.IP value in assignment /__w/1/s/pkg/api/converterfromapi.go:616:26: cannot use &(azure.Environment literal) (value of type *azure.Environment) as *azure.Environment value in assignment]"

PR #2538, opened 2020-01-08
feat: Add support for non-Azure Stack custom clouds

sukhong:
**Reason for Change**:
Currently, when CustomCloudProfile is specified, aks-engine treats it as Azure Stack and executes many Azure Stack-specific logic (i.e. loading azure stack specific image, setting lower arm api versions, disabling instance metadata, etc).

This makes impossible for users to specify custom cloud endpoints without triggering Azure Stack specific logic.

This PR decouples Azure Stack logic from Custom Cloud logic by introducing an internally computed property "IsCustomCloudProfile".

The main idea is to use “IsAzureStackCloud” property for doing any Azure Stack specific works, and use “IsCustomCloudProfile” property for generating actual azure.json and azurestackcloud.json files in VMs in the end.

Sample Input cluster definition:
````
"customCloudProfile": {
      "environment": {
        "name": "CustomCloud",
        "managementPortalURL": "https://customcloud.azure.custom/"
      }
}
````
Produces **azure.json**:
````
{
    "cloud":"AzureStackCloud",
    "tenantId": "bf85dc9d-cb43-44a4-80c4-469e8c58249e",
    "subscriptionId": "...",
    ...
}
````
and produces **azurestackcloud.json**:
````
{
    "name": "CustomCloud",
    "managementPortalURL": "https://customcloud.azure.custom/",
    ...
}
````

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
1. I tried my best to have 0% impact on existing Azure Stack functionality (i.e. when CustomCloudProfile was specified with "AzureStackCloud", the behavior should be exactly same).

2. The only non-critical breaking change is that previously when location = "randomlocation", cloudname = "randomcloudname", GetTargetEnv() func used to return "AzurePublicCloud". Now it will return "AzureStackCloud". However, usage of this scenario is almost buggy and I believe this will affect no users. I had to modify one scenario in TestGetTargetEnv() unit test case to reflect this change.

3. I also wanted to rename cse_customcloud.sh script is into cse_azurestackcloud.sh (the script seems to be Azure Stack specific), but decided not to do it to reduce the change scope

4. PortalURL is not required when CustomCloudProfile name is not AzureStackCloud

### CODE REVIEW

**Comment by devigned** [comment]
@jadarsie please give this a look when you have a minute.

File: pkg/api/defaults-custom-cloud-profile.go [hunk 1]
```diff
@@ -93,100 +110,114 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
...
 
-// SetAzureStackCloudSpec sets the cloud spec for Azure Stack .
-func (p *Properties) SetAzureStackCloudSpec(params AzureStackCloudSpecParams) error {
-	if p.IsAzureStackCloud() {
-		var azureStackCloudSpec AzureEnvironmentSpecConfig
+// SetAzureCustomCloudSpec sets the cloud spec for Custom Cloud .
+func (p *Properties) SetCustomCloudSpec(params AzureCustomCloudSpecParams) error {
+	if p.IsCustomCloudProfile() {
+		var azureCustomCloudSpec AzureEnvironmentSpecConfig
 		switch p.CustomCloudProfile.DependenciesLocation {
```
honcao: You would need predefine a AzureEnvironmentSpecConfig for the new security cloud and set the azureCustomCloudSpec based on the location.

pkg/api/addons.go
```diff
@@ -26,7 +26,7 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 	specConfig := cloudSpecConfig.KubernetesSpecConfig
 	omsagentImage := "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod11012019"
 	var workspaceDomain string
-	if cs.Properties.IsAzureStackCloud() {
+	if cs.Properties.IsCustomCloudProfile() {
 		dependenciesLocation := string(cs.Properties.CustomCloudProfile.DependenciesLocation)
 		workspaceDomain = helpers.GetLogAnalyticsWorkspaceDomain(dependenciesLocation)
 		if strings.EqualFold(dependenciesLocation, "china") {
```
honcao: Please also handle the logic for the LX cloud. I assume  oms agent Image will also be hosted in LX cloud.

pkg/api/const.go
```diff
@@ -471,14 +471,14 @@ const (
...
-	// AzureStackDependenciesLocationPublic indicates to get dependencies from in AzurePublic cloud
-	AzureStackDependenciesLocationPublic = "public"
-	// AzureStackDependenciesLocationChina indicates to get dependencies from AzureChina cloud
-	AzureStackDependenciesLocationChina = "china"
-	// AzureStackDependenciesLocationGerman indicates to get dependencies from AzureGerman cloud
-	AzureStackDependenciesLocationGerman = "german"
-	// AzureStackDependenciesLocationUSGovernment indicates to get dependencies from AzureUSGovernment cloud
-	AzureStackDependenciesLocationUSGovernment = "usgovernment"
+	// AzureCustomCloudDependenciesLocationPublic indicates to get dependencies from in AzurePublic cloud
+	AzureCustomCloudDependenciesLocationPublic = "public"
```
honcao: you would need define dependencies location of LX cloud

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -22,61 +22,78 @@ type CustomCloudProfileDefaultsParams struct {
...
 	}
 	return nil
 }
 
-// SetCustomCloudProfileEnvironment retrieves the endpoints from Azure Stack metadata endpoint and sets the values for azure.Environment
+// SetCustomCloudProfileEnvironment retrieves the endpoints from metadata endpoint (when required) and sets the values for azure.Environment
 func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
 	p := cs.Properties
-	if p.IsAzureStackCloud() {
+	if p.IsCustomCloudProfile() {
```
honcao: Because the endpoints are well-known, I would suggest to pre-define the azure.Environment in aks engine code base and directly use it get the  azure.Environment based on the environment name.
Azure Stack has dynamic endpoints. we are forced to using the logic to populate the endpoints.
1. The logic to format the endpoints are different between Azure Stack and LX cloud.
2. Not all endpoints are filled since not all services are available on Azure Stack.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -93,100 +110,114 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
...
-			azureStackCloudSpec = AzureCloudSpecEnvMap[AzurePublicCloud]
+			azureCustomCloudSpec = AzureCloudSpecEnvMap[AzurePublicCloud]
 		}
 		if p.CustomCloudProfile.Environment == nil || p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix == "" {
 			return errors.New("Failed to set Cloud Spec for Azure Stack due to invalid environment")
 		}
 
-		azureStackCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
-		azureStackCloudSpec.CloudName = AzureStackCloud
+		azureCustomCloudSpec.EndpointConfig.ResourceManagerVMDNSSuffix = p.CustomCloudProfile.Environment.ResourceManagerVMDNSSuffix
```
honcao: this logic is only for Azure Stack.

pkg/api/defaults-custom-cloud-profile.go
sukhong: This is a custom cloud and location can literally be anything. I do not want to (probably not even allowed) add specific new cloud-related information in the public repo. Instead, everything should come from user-provided config...

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -93,100 +110,114 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
...
+
+		if (p.CustomCloudProfile.Environment.Name == "" || p.IsAzureStackCloud()) {
+			azureCustomCloudSpec.CloudName = AzureStackCloud
+		} else {
+			azureCustomCloudSpec.CloudName = p.CustomCloudProfile.Environment.Name
+		}
 
 		//Sets default values for telemetry PID where none is set
 		if p.CustomCloudProfile.AzureEnvironmentSpecConfig == nil {
```
honcao: those logic are Azure Stack only

pkg/api/vlabs/const.go
```diff
@@ -160,14 +160,14 @@ const (
...
-	// AzureStackDependenciesLocationGerman indicates to get dependencies from AzureGerman cloud
-	AzureStackDependenciesLocationGerman = "german"
-	// AzureStackDependenciesLocationUSGovernment indicates to get dependencies from AzureUSGovernment cloud
-	AzureStackDependenciesLocationUSGovernment = "usgovernment"
+	// AzureCustomCloudDependenciesLocationPublic indicates to get dependencies from in AzurePublic cloud
+	AzureCustomCloudDependenciesLocationPublic = "public"
+	// AzureCustomCloudDependenciesLocationChina indicates to get dependencies from AzureChina cloud
+	AzureCustomCloudDependenciesLocationChina = "china"
+	// AzureCustomCloudDependenciesLocationGerman indicates to get dependencies from AzureGerman cloud
+	AzureCustomCloudDependenciesLocationGerman = "german"
```
honcao: you would need add now dependency location for LX cloud.

pkg/api/addons.go
sukhong: I don't think we can add any LX-specific logics in a public repo...

pkg/api/const.go
sukhong: I don't think we can add any LX-specific logics in a public repo...

pkg/api/defaults-custom-cloud-profile.go
honcao: The following is the AzureChinaCloudSpec for china cloud. as you can see, all the dependencies including to images and plugins are hosted separately in Azure China cloud.
end customer will not know the values.


//AzureChinaCloudSpec is the configurations for Azure China (Mooncake)
	AzureChinaCloudSpec = AzureEnvironmentSpecConfig{
		CloudName: AzureChinaCloud,
		//DockerSpecConfig specify the docker engine download repo
		DockerSpecConfig: DockerSpecConfig{
			DockerEngineRepo:         "https://mirror.azk8s.cn/docker-engine/apt/repo/",
			DockerComposeDownloadURL: "https://mirror.azk8s.cn/docker-toolbox/linux/compose",
		},
		//KubernetesSpecConfig - Due to Chinese firewall issue, the default containers from google is blocked, use the Chinese local mirror instead
		KubernetesSpecConfig: KubernetesSpecConfig{
			KubernetesImageBase:              "gcr.azk8s.cn/google_containers/",
			TillerImageBase:                  "gcr.azk8s.cn/kubernetes-helm/",
			ACIConnectorImageBase:            "dockerhub.azk8s.cn/microsoft/",
			NVIDIAImageBase:                  "dockerhub.azk8s.cn/nvidia/",
			AzureCNIImageBase:                "dockerhub.azk8s.cn/containernetworking/",
			MCRKubernetesImageBase:           "mcr.microsoft.com/k8s/core/",
			CalicoImageBase:                  "dockerhub.azk8s.cn/calico/",
			EtcdDownloadURLBase:              "https://mirror.azk8s.cn/kubernetes/etcd",
			KubeBinariesSASURLBase:           DefaultKubernetesSpecConfig.KubeBinariesSASURLBase,
			WindowsTelemetryGUID:             DefaultKubernetesSpecConfig.WindowsTelemetryGUID,
			CNIPluginsDownloadURL:            "https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
			VnetCNILinuxPluginsDownloadURL:   "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
			VnetCNIWindowsPluginsDownloadURL: "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
			ContainerdDownloadURLBase:        "https://mirror.azk8s.cn/kubernetes/containerd/",
		},
		DCOSSpecConfig: DCOSSpecConfig{
			DCOS188BootstrapDownloadURL:     fmt.Sprintf(AzureChinaCloudDCOSBootstrapDownloadURL, "5df43052907c021eeb5de145419a3da1898c58a5"),
			DCOSWindowsBootstrapDownloadURL: "https://dcosdevstorage.blob.core.windows.net/dcos-windows",
			DCOS190BootstrapDownloadURL:     fmt.Sprintf(AzureChinaCloudDCOSBootstrapDownloadURL, "58fd0833ce81b6244fc73bf65b5deb43217b0bd7"),
			DCOS198BootstrapDownloadURL:     fmt.Sprintf(AzureChinaCloudDCOSBootstrapDownloadURL, "f4ae0d20665fc68ee25282d6f78681b2773c6e10"),
		},

		EndpointConfig: AzureEndpointConfig{
			ResourceManagerVMDNSSuffix: "cloudapp.chinacloudapi.cn",
		},
		OSImageConfig: map[Distro]AzureOSImageConfig{
			Ubuntu:            Ubuntu1604OSImageConfig,
			Ubuntu1804:        Ubuntu1804OSImageConfig,
			RHEL:              RHELOSImageConfig,
			CoreOS:            CoreOSImageConfig,
			AKSUbuntu1604:     AKSUbuntu1604OSImageConfig,
			AKS1604Deprecated: AKSUbuntu1604OSImageConfig, // for back-compat
			AKSUbuntu1804:     AKSUbuntu1804OSImageConfig,
			AKS1804Deprecated: AKSUbuntu1804OSImageConfig, // for back-compat
		},
	}

sukhong: I do not think it is allowed to put LX-specific URLs in anywhere in public clouds. I also believe all dependency plugins are pre-installed in aks-engine base images so that disconnected clouds will need to go fetch them, that is why I think it is fine to use AzurePublicCloud defaults for custom clouds for now.

pkg/api/defaults-custom-cloud-profile.go
sukhong: I don't think this logic is Azure Stack specific. This logic is simply replacing default VMDNSSuffix with user-provided one.

pkg/api/defaults-custom-cloud-profile.go
sukhong: Would it break if I set default values to empty string for non-AzureStack custom clouds?

If so, I actually do not think there is any harm for defaulting AzureTelemetryPID to AZureStackTelemetyPIDs.

pkg/api/vlabs/const.go
sukhong: We cannot add any LX-specific URLs in a public repo (got a confirmation from LX folks). Also, this logic should be generic to support all custom clouds.

For now, I think defaulting dependencies location to public for non-Azure Stack custom clouds seems fine. Later, we can add a feature where users can specify all URLs for dependency location manually, instead of inferring from a fixed location.

pkg/api/defaults-custom-cloud-profile.go
sukhong: We cannot add any LX-specific URLs in a public repo (got a confirmation from LX folks) :(

pkg/api/defaults-custom-cloud-profile.go
honcao: I am not aware of the requirement that the LX endpoints are also confidential.
and if we will rely on the VHD to include all dependencies. I am ok with current change.

pkg/api/defaults-custom-cloud-profile.go
honcao: is this case, Please exam the logic to format the endpoints to meet the LX requirement.

pkg/api/defaults-custom-cloud-profile.go
honcao: I am not sure that if the empty id is an issue for non-auzre stack environment, 
the behavior is controlled by ARM, 
Azure Stack ARM has special logic to target specific guids to send telemetry data.

sukhong:
@jadarsie Can you please take a look at this PR? We verified that everything works as expected.

pkg/api/defaults-custom-cloud-profile.go
```diff
@@ -93,100 +110,114 @@ func (cs *ContainerService) SetCustomCloudProfileEnvironment() error {
 			}
 
 			env.ManagementPortalURL = endpoints.PortalEndpoint
-			env.ResourceManagerVMDNSSuffix = fmt.Sprintf("cloudapp.%s", azsFQDNSuffix)
-			env.StorageEndpointSuffix = fmt.Sprintf("%s.%s", cs.Location, azsFQDNSuffix)
-			env.KeyVaultDNSSuffix = fmt.Sprintf("vault.%s.%s", cs.Location, azsFQDNSuffix)
+
+			if (p.IsAzureStackCloud()) {
```
jadarsie: nit, parenteses not needed

sukhong:
I need to follow up with two things:
1. Support specifying custom root ca cert and mount it to kube-controller-manager
2. AADProfile logic needs to be changed to support custom oidc-issuer-url (line 85 in default-apiserver.go)

craiglpeters:
@sukhong is this now ready for review (it also needs a rebase)

ericsuhong:
Opened new one here: https://github.com/Azure/aks-engine/pull/3063

PR #2540, opened 2020-01-08
refactor: user-configurable components

jackfrancis:
**Reason for Change**:

This PR introduces user-configurable Kubernetes components in the api model. E.g.:

```json
"kubernetesConfig": {
    "components": [
        {
            "name": "kube-controller-manager",
            "enabled" : true,
            "containers": [
                {
                    "name": "kube-controller-manager",
                    "image": "myDockerHubUser/custom-controller-manager-build:v1.17.2-dirty",
                }
            ],
            "config": {
                "command": "kube-controller-manager-custom"
            }
        }
    ]
}
```

There are no functional changes with this PR. This additional configurable surface area will make per-component development easier, and will allow for more maintainable, rapid development of component image references.

- [x] unit tests
- [x] documentation

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/manifests/kubernetesmaster-kube-controller-manager.yaml
```diff
@@ -11,25 +11,46 @@ spec:
...
+      args: [{{GetControllerManagerArgs}}]
+{{- if IsAzureStackCloud}}
+      env:
+      - name: AZURE_ENVIRONMENT_FILEPATH
+        value: "/etc/kubernetes/azurestackcloud.json"
+{{end}}
       volumeMounts:
         - name: etc-kubernetes
           mountPath: /etc/kubernetes
+{{- if IsKubernetesVersionGe "1.17.0"}}
```
mboersma: One advantage of keeping all the separate manifests was that it made it easier to diff them with the canonical upstream versions. Generally we've kept manifests as synced up as possible to facilitate this.

Just sayin', probably not worth changing this PR.

pkg/api/common/const.go
```diff
@@ -245,3 +247,26 @@ const (
...
+	// APIServerComponentName is the name of the kube-apiserver component
+	APIServerComponentName = "kube-apiserver"
+	// AddonManagerComponentName is the name of the kube-addon-manager component
+	AddonManagerComponentName = "kube-addon-manager"
+)
+
+const (
+	// AzureStackSuffix is appended to kubernetes version on Azure Stack instances
+	AzureStackSuffix = "-azs"
+	// AzureStackPrefix is appended to windows binary version for Azure Stack instances
```
mboersma:
```suggestion
	// AzureStackPrefix is prepended to windows binary version for Azure Stack instances
```

pkg/api/components.go
```diff
@@ -0,0 +1,259 @@
...
+func synthesizeComponentsConfig(components []KubernetesComponent, defaultComponent KubernetesComponent, isUpgrade bool) {
+	i := getComponentsIndexByName(components, defaultComponent.Name)
+	if i >= 0 {
+		components[i] = assignDefaultComponentVals(components[i], defaultComponent, isUpgrade)
+	}
+}
+
+func getAPIServerDefaultCommandString(cs *ContainerService) string {
+	if common.IsKubernetesVersionGe(cs.Properties.OrchestratorProfile.OrchestratorVersion, "1.17.0") {
+		return fmt.Sprintf("\"kube-apiserver\"")
```
mboersma: Shouldn't we just return strings directly from these helper functions instead of using `fmt.Sprintf`? These aren't actually using any string formatting logic.

pkg/api/components.go
```diff
@@ -0,0 +1,259 @@
...
+	}
+	return fmt.Sprintf("\"/hyperkube\", \"kube-apiserver\"")
+}
+
+func getControllerManagerDefaultCommandString(cs *ContainerService) string {
+	if common.IsKubernetesVersionGe(cs.Properties.OrchestratorProfile.OrchestratorVersion, "1.17.0") {
+		return fmt.Sprintf("\"kube-controller-manager\"")
+	}
+	return fmt.Sprintf("\"/hyperkube\", \"kube-controller-manager\"")
```
mboersma: Same comment as above RE: `fmt.Sprintf`.

parts/k8s/manifests/kubernetesmaster-kube-controller-manager.yaml
jackfrancis: Right. This approach just seems so much more maintainable. Perhaps we need UT tooling to do some go template-processed static validation against the upstream sources.

mboersma: Agree that this is more maintainable.

pkg/api/components.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+	for i := range defaultComponent.Containers {
+		c := component.GetContainersIndexByName(defaultComponent.Containers[i].Name)
+		if c < 0 {
+			component.Containers = append(component.Containers, defaultComponent.Containers[i])
+		} else {
+			if component.Containers[c].Image == "" || isUpgrade {
+				component.Containers[c].Image = defaultComponent.Containers[i].Image
+			}
+			if component.Containers[c].CPURequests == "" {
```
jackfrancis: FYI this configurable resource limits functionality is not yet implemented, but it seems likely that it may be at some point in the future, so it was optimal to functionally implement this as part of the data spec now (w/ tests), during this initial abstraction introduction. If we do want it later, it saves time having already done this foundational work; in the meantime it has no effect on any of the currently implemented component templates (and it won't be documented).

PR #2546, opened 2020-01-09
perf: upgrade AKS tunnel image in VHD

gossion:
**Reason for Change**:

The AKS tunnel images in VHD are out-of-date, we need to bake new images in the VHD

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/packer/install-dependencies.sh
+++ b/packer/install-dependencies.sh
@@ -345,16 +345,16 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-TUNNELFRONT_VERSIONS="v1.9.2-v4.0.4"
+TUNNELFRONT_VERSIONS="v1.9.2-v3.0.11 v1.9.2-v4.0.11"
 for TUNNELFRONT_VERSION in ${TUNNELFRONT_VERSIONS}; do
-    CONTAINER_IMAGE="docker.io/deis/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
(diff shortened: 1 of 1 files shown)


packer/install-dependencies.sh
```diff
@@ -345,16 +345,16 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
...
+TUNNELFRONT_VERSIONS="v1.9.2-v3.0.11 v1.9.2-v4.0.11"
 for TUNNELFRONT_VERSION in ${TUNNELFRONT_VERSIONS}; do
-    CONTAINER_IMAGE="docker.io/deis/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
+    CONTAINER_IMAGE="mcr.microsoft.com/aks/hcp/hcp-tunnel-front:${TUNNELFRONT_VERSION}"
     pullContainerImage "docker" ${CONTAINER_IMAGE}
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-KUBE_SVC_REDIRECT_VERSIONS="1.0.2"
+KUBE_SVC_REDIRECT_VERSIONS="1.0.5 1.0.7"
```
xuto2: we also have 1.0.4 in mcr, do we not need that?

xuto2: also are these images the same as that in acr? Depends on when we completely moved to mcr, some cluster may still need to use acr image when this vhd is rolled out.

gossion: According to our toggle, 1.0.4 is not used anymore.
mcr has all images including the old images, those are copied from acr. so there won't be a problem when switching to mcr.

PR #2547, opened 2020-01-09
feat: Azure Spot VMSS

### LINKED ISSUE #2391: Support for Spot VMs / VMSS
**Describe the request**
With the deprecation and replacement of Low-priority VM's and VM Scale Sets for Spot VM's / VMSS we need to be able to configure Priority=Spot and BilliingProfile.maxPrice which was introduced in apiVersion 2019-03-01.
 
Low-priority functionality already exists and should be lifecycled to Spot.

Incrementing the hard coded api version parameter to at least 2019-03-01, but 2019-07-01 (lastest) is preferred, found here:
https://github.com/Azure/aks-engine/blob/4ec6504bc6da674d98edde6bcf40c168bde99dc9/pkg/engine/armvariables.go#L151

And allow billingProfile.maxPrice to be configurable with a default value of "-1.0".

None.

**Additional context**

gtxistxgao:
**Reason for Change**:

Add the support for Azure Spot VMs

**Issue Fixed**:

Partially Fixes #2391 pending work is to remove Low Priority

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

Add new ScaleSetPriority type: "Spot"
Add new field to support the max billing feature of Spot VM: SpotMaxPrice
Deleted examples/kubernetes-vmss-low-priority folder.
Added examples/kubernetes-vmss-spot folder and the related kubernetes.json file.


pkg/api/defaults.go
```diff
@@ -627,9 +627,15 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			profile.AvailabilityProfile = VirtualMachineScaleSets
 		}
 		if profile.AvailabilityProfile == VirtualMachineScaleSets {
-			if profile.ScaleSetEvictionPolicy == "" && profile.ScaleSetPriority == ScaleSetPriorityLow {
+			if profile.ScaleSetEvictionPolicy == "" && (profile.ScaleSetPriority == ScaleSetPriorityLow || profile.ScaleSetPriority == ScaleSetPrioritySpot) {
 				profile.ScaleSetEvictionPolicy = ScaleSetEvictionPolicyDelete
 			}
+
+			if profile.ScaleSetPriority == ScaleSetPrioritySpot && profile.SpotMaxPrice == nil {
+				var maximumValueFlag float64 = -1
```
jackfrancis: essentially the default is "no max price", is that what this means?

pkg/api/types.go
```diff
@@ -1516,6 +1517,11 @@ func (a *AgentPoolProfile) IsLowPriorityScaleSet() bool {
 	return a.AvailabilityProfile == VirtualMachineScaleSets && a.ScaleSetPriority == ScaleSetPriorityLow
 }
 
+// IsSpotScaleSet returns true if the VMSS is Spot Scale Set
+func (a *AgentPoolProfile) IsSpotScaleSet() bool {
```
jackfrancis: let's add a UT for this new method

pkg/engine/armvariables.go
```diff
@@ -608,7 +608,7 @@ func getK8sAgentVars(cs *api.ContainerService, profile *api.AgentPoolProfile) ma
 	if profile.IsAvailabilitySets() {
 		agentVars[agentOffset] = fmt.Sprintf("[parameters('%s')]", agentOffset)
 		agentVars[agentAvailabilitySet] = fmt.Sprintf("[concat('%s-availabilitySet-', parameters('nameSuffix'))]", agentName)
-	} else if profile.IsLowPriorityScaleSet() {
+	} else if profile.IsLowPriorityScaleSet() || profile.IsSpotScaleSet() {
```
jackfrancis: I think there is some equivalent foo in `parts/agentparams.t` (though I'm not actually sure the template is consuming these particular ARM vars/params at present...

pkg/api/const.go
```diff
@@ -98,6 +98,8 @@ const (
 	ScaleSetPriorityRegular = "Regular"
 	// ScaleSetPriorityLow means the ScaleSet will use Low-priority VMs
 	ScaleSetPriorityLow = "Low"
+	// ScaleSetPrioritySpot means the ScaleSet will use Spot VMs
+	ScaleSetPrioritySpot = "Spot"
```
CecileRobertMichon: Let's document the new value. https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#L567 is a good place to start (there's already an entry for ScaleSetPriority).

pkg/api/defaults.go
gtxistxgao: yes. -1 means no limit. This is the doc if you're interested: https://docs.microsoft.com/en-us/rest/api/compute/virtualmachinescalesets/createorupdate#billingprofile

gtxistxgao: oh not no limit, limit up-to on-demand. ;-p

pkg/api/types.go
gtxistxgao: sure. added.

pkg/api/const.go
gtxistxgao: Thanks. Edited the doc and added explanation to the new field.

pkg/api/types.go
```diff
@@ -578,6 +578,7 @@ type AgentPoolProfile struct {
 	AvailabilityProfile                 string               `json:"availabilityProfile"`
 	ScaleSetPriority                    string               `json:"scaleSetPriority,omitempty"`
 	ScaleSetEvictionPolicy              string               `json:"scaleSetEvictionPolicy,omitempty"`
+	SpotMaxPrice                        *float64             `json:"spotMaxPrice,omitempty"`
```
CecileRobertMichon: let's also add the type to vlabs so we can test it in aks-engine E2E

CecileRobertMichon: that will require changing pkg/api/vlabs/types.go, pkg/api/vlabs/types_test.go (copy the same test as the one in api/types.go) pkg/api/convertertoapi.go and pkg/api/converterfromapi.go

pkg/api/const.go
jluk: Everyone probably already knows this - but "Low" will be deprecated at one year from now and no longer work

CecileRobertMichon: Thanks @jluk that's a good point, maybe we should add a note to the docs with the link to the deprecation notice, right now it looks like we're equally recommending both options.

devigned: perhaps, even a warning for api models which include low.

jackfrancis: Slight preference for removing low as an option altogether (w/ back-compat support for pre-existing cluster). (As a distinct, follow-up PR)

If we can prevent any aks-engine created clusters from running low pri vmss a year from now when it is officially deprecated, that would be 💯

CecileRobertMichon: +1 on removing the option in a follow up PR. We might not have time to get to the follow up right away though so in the meantime adding a small note to the docs about deprecation is a low hanging fruit.

gtxistxgao: Added "(DEPRECATED)" notation in the doc in this PR after the "Low" option. @jluk VMSS team mentioned yesterday that they will treat "Low" as "Spot" VMSS internally. So it's better to remove it early as "Low" does not give the option to set maxPrice in AKS-E.

pkg/api/types.go
gtxistxgao: Thanks Cecile! Added field to vlabs and related tests. Also added into examples/kubernetes-vmss-spot

pkg/api/const.go
jluk: Stopping more clusters from setting low sounds fine and I'll defer how to language that into engine.

For a year low is still accepted and just behaves as spot with maxPrice = -1 so not truly deprecated yet just on the path to it. Also Jag is owns Spot, not VMSS so these rules apply for both VM and VMSS.

pkg/engine/armvariables.go
gtxistxgao: not sure what should I do. do I need to add anything else?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -453,6 +453,14 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		vmssVMProfile.EvictionPolicy = compute.VirtualMachineEvictionPolicyTypes(fmt.Sprintf("[variables('%sScaleSetEvictionPolicy')]", profile.Name))
 	}
 
+	if profile.IsSpotScaleSet() {
+		vmssVMProfile.Priority = compute.VirtualMachinePriorityTypes(fmt.Sprintf("[variables('%sScaleSetPriority')]", profile.Name))
```
jackfrancis: the first two assignments here are identical to the assignments in the above `if profile.IsLowPriorityScaleSet() {` condition block. Let's expand that conditional block to be `if profile.IsLowPriorityScaleSet() || profile.IsSpotScaleSet() {`, and remove the duplicate assignment statement.

This way this new `if profile.IsSpotScaleSet() {` conditional block will only have the `BillingProfile` assignment, which is the only assignment here unique to spot scale sets.

gtxistxgao: update the code as suggested.

devigned:
Awesome work, @gtxistxgao!

PR #2548, opened 2020-01-09
feat: Use default AKS Windows OS Image for k8s

AbelHu:
**Reason for Change**:

With this change, users do not need to specify a Windows OS image when creating Windows agent pools in kubernetes. It will always uses a default Windows OS image per Azure cloud environment.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I build it and create a k8s cluster on Azure with a Windows agent pool. And I also run e2e test and all test cases passed.

E2E test results:
```
JUnit report was created: /aks-engine/test/e2e/kubernetes/junit.xml

Ran 34 of 50 Specs in 1608.712 seconds
SUCCESS! -- 34 Passed | 0 Failed | 0 Pending | 16 Skipped
PASS
```

Below is Windows agent pool I used for testing.
```
      {
        "name": "npwin",
        "count": 1,
        "vmSize": "Standard_D2_v3",
        "osType": "Windows",
        "availabilityProfile": "VirtualMachineScaleSets"
      }
```

Result of `k get no -o wide`:
```
NAME                                 STATUS   ROLES    AGE   VERSION   INTERNAL-IP    EXTERNAL-IP   OS-IMAGE                         KERNEL-VERSION      CONTAINER-RUNTIME
2185k8s01000000                      Ready    agent    10m   v1.14.7   10.240.0.96    <none>        Windows Server 2019 Datacenter   10.0.17763.864      docker://19.3.2
k8s-agentpool1-21854569-vmss000000   Ready    agent    13m   v1.14.7   10.240.0.34    <none>        Ubuntu 16.04.6 LTS               4.15.0-1064-azure   docker://3.0.8
k8s-agentpool1-21854569-vmss000001   Ready    agent    13m   v1.14.7   10.240.0.65    <none>        Ubuntu 16.04.6 LTS               4.15.0-1064-azure   docker://3.0.8
k8s-master-21854569-0                Ready    master   13m   v1.14.7   10.255.255.5   <none>        Ubuntu 16.04.6 LTS               4.15.0-1064-azure   docker://3.0.8
```

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -37,6 +37,11 @@ const (
 	ACC1604           Distro = "acc-16.04"
 )
 
+// WindowsOSVersion string consts
+const (
+	AKSWindowsServer2019 WindowsOSVersion = "aks-windows-server-2019"
+)
+
 const (
 	// SwarmVersion is the Swarm orchestrator version
 	SwarmVersion = "swarm:1.1.0"
@@ -48,8 +53,6 @@ const (
 	DockerCEDockerComposeVersion = "1.14.0"
 	// KubernetesWindowsDockerVersion is the default version for docker on Windows nodes in kubernetes
 	KubernetesWindowsDockerVersion = "19.03.2"
-	// KubernetesDefaultWindowsSku is the default SKU for Windows VMs in kubernetes
-	KubernetesDefaultWindowsSku = "Datacenter-Core-1809-with-Containers-smalldisk"
 )
 
 // validation values
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -65,7 +65,8 @@ func (cs *ContainerService) SetPropertiesDefaults(params PropertiesDefaultsParam
 	}
 
 	if cs.Properties.WindowsProfile != nil {
-		properties.setWindowsProfileDefaults(params.IsUpgrade, params.IsScale)
+		cloudSpecConfig := cs.GetCloudSpecConfig()
+		properties.setWindowsProfileDefaults(params.IsUpgrade, params.IsScale, cloudSpecConfig)
 	}
 
 	properties.setTelemetryProfileDefaults()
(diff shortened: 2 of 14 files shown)


pkg/api/const.go
```diff
@@ -35,6 +35,7 @@ const (
 	AKSUbuntu1604     Distro = "aks-ubuntu-16.04"
 	AKSUbuntu1804     Distro = "aks-ubuntu-18.04"
 	ACC1604           Distro = "acc-16.04"
+	WindowsDistro     Distro = "windows"
```
CecileRobertMichon: Not sure re-using the `Distro` property makes sense here, since "distro" is [commonly known to mean Linux distribution.](https://en.wikipedia.org/wiki/Linux_distribution).

pkg/engine/params.go
```diff
@@ -221,11 +221,26 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
...
+				addValue(parametersMap, "agentWindowsSku", cloudSpecConfig.OSImageConfig[api.WindowsDistro].ImageSku)
+			} else {
+				addValue(parametersMap, "agentWindowsSku", properties.WindowsProfile.WindowsSku)
+			}
+			if properties.WindowsProfile.ImageVersion == "" {
+				addValue(parametersMap, "agentWindowsVersion", cloudSpecConfig.OSImageConfig[api.WindowsDistro].ImageVersion)
+			} else {
+				addValue(parametersMap, "agentWindowsVersion", properties.WindowsProfile.ImageVersion)
+			}
 		}
```
marosset: Please add some unit tests to validate this logic

pkg/api/const.go
marosset: I think I'd be ok overloading here here.
Lets be more specific so we can support multiple releases of windows server.
How about AKSWindowsServer2019 here to match the

pkg/engine/params.go
marosset: Also, the properties.WindowsProfile properties here all get defaults in [https://github.com/Azure/aks-engine/blob/f9dd0e808e569787a7bc61119234fb1f403cef0c/pkg/api/defaults.go#L696-L720](https://github.com/Azure/aks-engine/blob/f9dd0e808e569787a7bc61119234fb1f403cef0c/pkg/api/defaults.go#L696-L720) so they should never be empty string.

---
In reply to: [364890111](https://github.com/Azure/aks-engine/pull/2548#discussion_r364890111) [](ancestors = 364890111)

pkg/api/const.go
CecileRobertMichon: That doesn't feel great to me... any way we could rename the property to something more generic or create a separate string type for Windows? @jackfrancis thoughts?

If we do decide to stick with this we also need to change the comment above the "Distro" type - right now it's `Distro represents Linux distro to use for Linux VMs`.

pkg/engine/params.go
AbelHu: Sorry for missing that. I think that I can remove setWindowsProfileDefaults.

marosset: Instead of removing setWindowsProfileDefaults, why not just keep the aks-engine code the same but remove these fields from the apimodel?

marosset: If the goal here is to just use the aks-engine defaults i think that can already be accomplished by leaving WindowsPublisher, WIndowsOffer, WindowsSKU, and ImageVersion out of the WindowsProfile block in the cluster definition/apimodel.

pkg/api/const.go
AbelHu: I will update the comment for the `Distro` type and rename `WindowsDistro` to `AKSWindowsServer2019`.
I also will add `Distro` in `WindowsProfile` so users can specify it.

marosset: I'm still not sure we need any changes to get the desired behavior.

If you specify the following in the cluster definition you'll get the default up-to-date values.
"windowsProfile": {
      "adminUsername": "azureuser",
      "adminPassword": "####",
      "sshEnabled": true,
      "enableAutomaticUpdates": false
    }
I believe AKS was using hard-coded values because at some point in early 2019 there was some stability issues with windows containers and AKS wanted to hard-code the image versions instead of specifying "latest"

Today we manage the defaults in aks-engine and before updating the defaults to point to new image versions.

Are there cases where this won't work? I just want to understand the use cases here.

@PatrickLang what do you think here?

pkg/engine/params.go
AbelHu: @marosset , you are right. If users want to use the default Windows OS image in aks-engine, they only need to remove the specified variables in my apimodel.
I want to use this PR for below purpose:
1. Keep consistent: Specify default image for Windows as Linux does
2. User can easily specify a Windows Image by Distro name.
Please see the update in my PR.

pkg/api/defaults.go
```diff
@@ -692,29 +692,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
 	windowsProfile := p.WindowsProfile
 	if !isUpgrade && !isScale {
-		if windowsProfile.WindowsPublisher == "" {
-			windowsProfile.WindowsPublisher = AKSWindowsServer2019OSImageConfig.ImagePublisher
-		}
-		if windowsProfile.WindowsOffer == "" {
-			windowsProfile.WindowsOffer = AKSWindowsServer2019OSImageConfig.ImageOffer
-		}
-		if windowsProfile.WindowsSku == "" {
-			windowsProfile.WindowsSku = AKSWindowsServer2019OSImageConfig.ImageSku
```
marosset: Please keep this logic in here if windowsProfile.Distro is not set.
Our documented guidance to customers is to set WindowsPublisher, WindowsOffer, and WindowsSku if they want to get the lastest tested version of an image.
https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#marketplace-images

If we remove this logic here ImageVersion users following this guidance will likely be broken.

pkg/engine/params.go
marosset: Can the logic here be moved into GetWindowsPublisher(), GetWindowsOffer(), GetWindowsSku(), GetImageVersion() functions on WindowsProfile to keep things cleaner here?

pkg/api/defaults.go
AbelHu: @marosset The document may not be accurate. Even if users set `windowsPublisher`, `windowsOffer`, and `windowsSku`, the `latest` version will not be used when the values of those 3 variables equal to the values of `AKSWindowsServer2019OSImageConfig` or `WindowsServer2019OSImageConfig`.
Different Azure environments may have different default OS images so we may need to set default values by Azure environments.
I prefer to set `ImageVersion` to `latest` if all of `windowsPublisher`, `windowsOffer`, and `windowsSku` are specified but `ImageVersion` is empty in setWindowsProfileDefaults. And we can update the doc in the link you provided. How do you think?

pkg/api/const.go
AbelHu: After checking the code, I introduce one new variable `WindowsOSVersion` for Windows OS images because `DistroValues` are defined in defaults.go and used in many places for Linux.

pkg/engine/params.go
AbelHu: Update it.

pkg/api/defaults.go
AbelHu: The old code supports to set `latest` so I just leave the code there in this PR. If we do not want users to use `latest`, we can remove it later.

AbelHu:
Still need to refine the PR so I will close this and create a new PR later.

PR #2550, opened 2020-01-10
[WIP] fix: enable SLB with private clusters

jackfrancis:
**Reason for Change**:

This PR introduces outbound Standard LB + public IPs for private cluster scenarios, to enable SLB-enabled outbound internet access from master and node vms.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/armresources.go
+++ b/pkg/engine/armresources.go
@@ -47,9 +47,10 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		armResources = append(armResources, userAssignedID, msiRoleAssignment)
 	}
 
+	// Create Standard Load Balancer resource spec for AKS Engine
 	if cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku &&
-		!isHostedMaster &&
-		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() {
+		!isHostedMaster && // skip for AKS
+		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() { // skip if we have a user-configurable Load Balancer ID to join agent pools to
 		isForMaster := false
 		publicIPAddress := CreatePublicIPAddress(isForMaster)
 		loadBalancer := CreateAgentLoadBalancer(cs.Properties, true)
--- a/pkg/engine/armresources.go
+++ b/pkg/engine/armresources.go
@@ -47,10 +47,9 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		armResources = append(armResources, userAssignedID, msiRoleAssignment)
 	}
 
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
+	if cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku &&
 		!isHostedMaster &&
-		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() &&
-		cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku {
+		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() {
 		isForMaster := false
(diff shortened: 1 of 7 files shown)


pkg/engine/armresources.go
```diff
@@ -47,10 +47,9 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
...
 	}
 
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
+	if cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku &&
 		!isHostedMaster &&
-		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() &&
-		cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku {
+		!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs() {
 		isForMaster := false
 		publicIPAddress := CreatePublicIPAddress(isForMaster)
```
devigned: You obviously didn't change this, but when ever I see bool flags passed into functions, it just screams for two functions. One for creating a pub IP for a master and another for an agent pool. I think this is indicated by the desire to name the bool being passed into `CreatePublicIPAddress`.

I think the oddest part of this function and others like it is that if you called it twice and add the resources  to the template, you would have overlapping resources (two resources with the same name). It just feels odd and opaque.

jackfrancis: Right. To be clear the state of this PR is "sort of working prototype". Once we get to "definitely working prototype" then it will be appropriate to do some of your suggested code interface improvements.

pkg/engine/armresources.go
```diff
@@ -47,10 +47,9 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
...
 		loadBalancer := CreateAgentLoadBalancer(cs.Properties, true)
```
devigned: `CreateAgentLoadBalancer` feels super opaque. Seems like it should take the publicIP and outbound rules rather than api properties and a bool.
1) The `bool` param isn't even used in the function (AFACT)
2) the properties are just passed to `createOutboundRules`

pkg/engine/armresources.go
devigned: I'll refrain. Just perusing and venting as it's top of mind.

jackfrancis: Please don't refrain, these comments are being valuably added to the public record! 😋

pkg/engine/armvariables.go
```diff
@@ -457,21 +457,21 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
-				masterVars["agentLbBackendPoolName"] = "[parameters('masterEndpointDNSNamePrefix')]"
-			}
-			masterVars["masterPublicIPAddressName"] = "[concat(parameters('orchestratorName'), '-master-ip-', variables('masterFqdnPrefix'), '-', parameters('nameSuffix'))]"
-			masterVars["masterLbID"] = "[resourceId('Microsoft.Network/loadBalancers',variables('masterLbName'))]"
-			masterVars["masterLbIPConfigID"] = "[concat(variables('masterLbID'),'/frontendIPConfigurations/', variables('masterLbIPConfigName'))]"
-			masterVars["masterLbIPConfigName"] = "[concat(parameters('orchestratorName'), '-master-lbFrontEnd-', parameters('nameSuffix'))]"
-			masterVars["masterLbName"] = "[concat(parameters('orchestratorName'), '-master-lb-', parameters('nameSuffix'))]"
 			masterVars["kubeconfigServer"] = "[concat('https://', variables('masterFqdnPrefix'), '.', variables('location'), '.', parameters('fqdnEndpointSuffix'))]"
 		}
+		masterVars["masterPublicIPAddressName"] = "[concat(parameters('orchestratorName'), '-master-ip-', variables('masterFqdnPrefix'), '-', parameters('nameSuffix'))]"
```
jackfrancis: We are now referencing these ARM variables in private cluster + SLB scenarios, so we move them out of the else block above (that else block is equivalent to "do this if not in a private cluster scenario)

jackfrancis:
See #2572

PR #2557, opened 2020-01-10
feat: Add Node Pool

### LINKED ISSUE #1611: Support aks-engine add and remove agent pools
**Describe the request**
I think that would be really great after the standard loadbalancer is really stable to support aks-engnie add/remove agent-pool commands.

Also aks-engine upgrade single agent pool. that would be great for testing new version of k8s first before applying it on all agent pools.

This already supported with the hosted aks.

Manage the agent pools with aks-engine commands.

Apply this manually with arm templates, but this will not work with the single agent-pool upgrade from example.

**Additional context**

DavidLangworthy:
This PR adds a sub command to add a single node pool.  It makes it easier to add to existing clusters and gets past the 12 node pool limit that bubbles up from ARM.  This is a resubmit of an earlier PR that went stale.

Fixes #1611


jackfrancis:
@DavidLangworthy invading this space a bit to ensure we merge this super useful addition w/ E2E coverage

Also: preferring the "--node-pool" option as this matches with the customary semantics for this kind of thing (e.g., node pools instead of agent pools).

Thanks again for this!

DavidLangworthy:
@jackfrancis Looks great.  Thank you!

Issue #2370, opened 2019-11-26
containerd integration using deprecated "cri" binary


cpuguy83:
**Describe the bug**

For containerd we are fetching the deprecated standalone cri-containerd binary.
CRI is now compiled into containerd directly, this should be used instead.

**Additional context**

https://github.com/Azure/aks-engine/blob/f8f373d241780a7945e40cdc9dc0d4b326d334b2/parts/k8s/cloud-init/artifacts/cse_install.sh#L173-L178

From cri-containerd repo:

> Note: The standalone cri-containerd binary is end-of-life. cri-containerd is transitioning from a standalone binary that talks to containerd to a plugin within containerd. This github branch is for the cri plugin. See standalone-cri-containerd branch for information about the standalone version of cri-containerd.

I already have some packaging for containerd, but have not tested extensively or finalized the pipeline to publish such packages.

CecileRobertMichon:
Thanks @cpuguy83. Is this ready to be fixed or do we need to wait for 
> I already have some packaging for containerd, but have not tested extensively or finalized the pipeline to publish such packages.

?

cpuguy83:
We could use the version of containerd that we package with our moby packages, and it may even be worth it to have access to docker (for users) in the containerd-based setups.

PR #2559, opened 2020-01-10
Use moby packages to get containerd.

cpuguy83:
This has the downside of not being able to control the version of
containerd but should work otherwise until we split containerd packaging
from moby.

--- 

Just posting to see if this is worth it.
Related to #2370

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -134,6 +134,11 @@ installMoby() {
     fi
 }
 
+installContainerd() {
+	installMoby
+}
+
+
 installKataContainersRuntime() {
     echo "Adding Kata Containers repository key..."
     ARCH=$(arch)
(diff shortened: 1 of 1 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -134,6 +134,11 @@ installMoby() {
     fi
 }
 
+installContainerd() {
+	installMoby
```
CecileRobertMichon: we already call `installMoby` in `installContainerRuntime` at https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_install.sh#L112

It might make sense to change the condition for that call from 
```
        installMoby
    fi
```
to
```
    if [[ "$CONTAINER_RUNTIME" == "docker" || "$CONTAINER_RUNTIME" == "kata-containers" || "$CONTAINER_RUNTIME" == "containerd" ]]; then
        installMoby
    fi
```
To have all calls of `installMoby` in one place and make the code easier to read. Otherwise it's not clear what the intentions are in terms of when to install Moby.

CecileRobertMichon:
closing after @cpuguy83 confirmed we don't want this merged as-is.

PR #2569, opened 2020-01-13
fix: don't expire nonexistent passwords

### LINKED ISSUE #2277: Password expired when not using a password
**Describe the bug**
Logging into an aks-engine master deployed a while ago, it reports that the password must be changed, but we're using only SSH keys, not a password, so we don't know the current password and thus can no longer log in.

**Steps To Reproduce**
Use an apimodel with data such as:
```
    "linuxProfile": {
      "adminUsername": "<username>",
      "ssh": {
        "publicKeys": [
          {
            "keyData": "ssh-rsa <base64> <username>@<hostname>"
          }
        ]
      }
```

**Expected behavior**
There is no password, or at least it doesn't expire.

**AKS Engine version**
0.38.2

**Kubernetes version**
1.15.0

**Additional context**
Here's the console output I get logging in with the SSH key:
```
WARNING: Your password has expired.
You must change your password now and login again!
Changing password for <username>.
(current) UNIX password:
```

I do get this error for clusters deployed on the following dates:
2019-07-29
2019-07-16
2019-06-12

but not for a cluster deployed on:
2019-09-15

so the password appears to be expiring sometime between 52 and 100 days after cluster deployment.

mboersma:
**Reason for Change**:
None of the Linux accounts on an AKS Engine node are intended to allow username/password login, yet having an expiration policy can apparently prevent access after that time passes even if public key authentication is used.

This disables inactivity and time-passed expiration (99999 is represented as "never" in the relevant Linux tools) to avoid users being inadvertently locked out of accounts.

**Issue Fixed**:
Fixes #2277

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cis.sh
+++ b/parts/k8s/cloud-init/artifacts/cis.sh
@@ -58,12 +58,12 @@ setPWExpiration() {
   grep 'PASS_MIN_DAYS' /etc/login.defs && exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   sed -i "s|INACTIVE=||g" /etc/default/useradd || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   grep 'INACTIVE=' /etc/default/useradd && exit $ERR_CIS_APPLY_PASSWORD_CONFIG
-  echo 'PASS_MAX_DAYS 90' >> /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
-  grep 'PASS_MAX_DAYS 90' /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
+  echo 'PASS_MAX_DAYS 99999' >> /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
+  grep 'PASS_MAX_DAYS 99999' /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   echo 'PASS_MIN_DAYS 7' >> /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   grep 'PASS_MIN_DAYS 7' /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cis.sh
```diff
@@ -58,12 +58,12 @@ setPWExpiration() {
   grep 'PASS_MIN_DAYS' /etc/login.defs && exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   sed -i "s|INACTIVE=||g" /etc/default/useradd || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
   grep 'INACTIVE=' /etc/default/useradd && exit $ERR_CIS_APPLY_PASSWORD_CONFIG
-  echo 'PASS_MAX_DAYS 90' >> /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
-  grep 'PASS_MAX_DAYS 90' /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
+  echo 'PASS_MAX_DAYS 99999' >> /etc/login.defs || exit $ERR_CIS_APPLY_PASSWORD_CONFIG
```
jackfrancis: Strictly speaking this change will expire passwords on vms sometime in 2293 or 2294, depending on when this PR merges. Are we assuming that climate change-induced geographical events will predict that Azure data centers won't remain viable in their current locales well before that time, and thus mooting any long-term eventual password expiration?

mboersma: That is exactly what we're predicting. Don't sell your MSFT stock though...at least until 2280.

jackfrancis:
We need to have a plan for building a new VHD and updating the `pwquality-validate.sh` accordingly after that VHD is referenced in aks-engine.

CecileRobertMichon:
I think this change would need editing https://docs.microsoft.com/en-us/azure/aks/security-hardened-vm-host-image because `5.4.1.1 | Ensure password expiration is 90 days or less` won't be applicable anymore.

PR #2572, opened 2020-01-14
fix: enable SLB with private clusters

### LINKED ISSUE #2549: Private cluster support for SLB
At present, aks-engine template generation accepts a private cluster + SLB cluster configuration scenario, but it doesn't do what a user might think it does (e.g., provide outbound internet access via SLB-enforced LB + SNAT).

We should implement something like the above, and then document the outbound access ramifications for customers who explicitly want their private cluster implementations to restrict outbound as well as inbound access to the cluster.

jackfrancis:
**Reason for Change**:

This PR introduces outbound Standard LB + public IPs for private cluster scenarios, to enable SLB-enabled outbound internet access from master and node vms.

Cluster confguration | Before PR (current) | After PR
------------ |------------ | -------------
Private cluster + 1 master + Standard LB + VMSS node pools | No Standard LB provisioned for master or node pools, nodes configured for `"loadBalancerSku": "Standard"` [👎 **NON-FUNCTIONAL cluster config**] | master outbound via master NIC NSG, Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👍 **WORKING cluster config**]
Private cluster + multiple masters + Standard LB + VMSS node pools  | Internal Standard LB provisioned for masters (no outbound internet access), no Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👎 **NON-FUNCTIONAL cluster config**]  | Standard LB provisioned for masters outbound egress, Standard internal LB provisioned for internal apiserver load balancing, and Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👍 **WORKING cluster config**]
Private cluster + 1 master + Basic LB + VMSS agent node pools  | No LBs provisioned for masters or node pools, nodes configured for `"loadBalancerSku": "Basic"` [👍 **WORKING cluster config**] | no change
Private cluster + multiple masters + Basic LB + VMSS node pools  | Internal Basic LB provisioned for masters, no LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Basic"` [👍 **WORKING cluster config**] | no change
Private cluster + 1 master + Standard LB + VMAS node pools | No Standard LB provisioned for master or node pools, nodes configured for `"loadBalancerSku": "Standard"` [👎 **NON-FUNCTIONAL cluster config**] | master outbound via master NIC NSG, Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👍 **WORKING cluster config**]
Private cluster + multiple masters + Standard LB + VMAS node pools  | Internal Standard LB provisioned for masters (no outbound internet access), no Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👎 **NON-FUNCTIONAL cluster config**] | Standard LB provisioned for masters outbound egress, Standard internal LB provisioned for internal apiserver load balancing, and Standard LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Standard"` [👍 **WORKING cluster config**]
Private cluster + 1 master + Basic LB + VMAS node pools  | No LBs provisioned for masters or node pools, nodes configured for `"loadBalancerSku": "Basic"` [👍 **WORKING cluster config**] | no change
Private cluster + multiple masters + Basic LB + VMAS node pools  | Internal Basic LB provisioned for masters, no LB provisioned for node pools, nodes configured for `"loadBalancerSku": "Basic"` [👍 **WORKING cluster config**] | no change

**Issue Fixed**:

Fixes #2549 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/armresources.go
```diff
@@ -47,13 +47,13 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		armResources = append(armResources, userAssignedID, msiRoleAssignment)
 	}
 
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
-		!isHostedMaster &&
+	if !isHostedMaster &&
```
jackfrancis: This change creates the Standard LB for node pools in a private cluster scenario. Prior to this PR that LB was not being created in a private cluster scenario.

pkg/engine/armresources_test.go
```diff
@@ -591,7 +591,7 @@ func TestGenerateARMResourcesWithVMSSAgentPool(t *testing.T) {
 	}
 }
 
-func getNICIPConfigs(n int) *[]network.InterfaceIPConfiguration {
+func getNICIPConfigs(n int, skipMasterLB bool) *[]network.InterfaceIPConfiguration {
```
jackfrancis: updated this convenience func so we could more easily UT scenarios that don't build a master LB (e.g., single master private cluster)

pkg/engine/armvariables.go
```diff
@@ -430,6 +430,21 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
 	} else {
+		if cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku && hasAgentPool {
+			masterVars["agentPublicIPAddressName"] = "[concat(parameters('orchestratorName'), '-agent-ip-outbound')]"
+			masterVars["agentLbID"] = "[resourceId('Microsoft.Network/loadBalancers',variables('agentLbName'))]"
+			masterVars["agentLbIPConfigID"] = "[concat(variables('agentLbID'),'/frontendIPConfigurations/', variables('agentLbIPConfigName'))]"
+			masterVars["agentLbIPConfigName"] = "[concat(parameters('orchestratorName'), '-agent-outbound')]"
+			masterVars["agentLbName"] = "[parameters('masterEndpointDNSNamePrefix')]"
+			masterVars["agentLbBackendPoolName"] = "[parameters('masterEndpointDNSNamePrefix')]"
+		}
+		if !(cs.Properties.OrchestratorProfile.IsPrivateCluster() && cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.BasicLoadBalancerSku) {
```
jackfrancis: private cluster + basic LB does not need these vars (which serve the master LB and public IP resources), because:

- private cluster + basic LB + 1 master uses NIC outbound rules for master outbound access
- private cluster + basic LB + multiple masters uses internal master LB for outbound access (doesn't need a public IP)

pkg/engine/ipaddresses.go
```diff
@@ -12,12 +12,15 @@ import (
 // When it's for master, this public ip address is created and added to the loadbalancer's frontendIPConfigurations
 // and it's created with the fqdn as name.
 // When it's for agent, this public ip address is created and added to the loadbalancer's frontendIPConfigurations.
-func CreatePublicIPAddress(isForMaster bool) PublicIPAddressARM {
+func CreatePublicIPAddress(isForMaster, includeDNS bool) PublicIPAddressARM {
```
jackfrancis: The new `includeDNS` param supports the private cluster + Standard LB scenario, which requires a public IP for master outbound access SNAT, but we don't attach a DNS record to it (that DNS record is for conveniently exposing inbound req's, e.g., apiserver, ssh to master)

pkg/engine/loadbalancers.go
```diff
@@ -140,73 +108,115 @@ func CreateLoadBalancer(prop *api.Properties, isVMSS bool) LoadBalancerARM {
...
-					ID: to.StringPtr("[concat(variables('masterLbID'),'/probes/tcpHTTPSProbe')]"),
-				},
-			},
-		}
-		*loadBalancer.LoadBalancer.LoadBalancerPropertiesFormat.LoadBalancingRules = append(*loadBalancer.LoadBalancer.LoadBalancerPropertiesFormat.LoadBalancingRules, udpRule)
-	}
-
-	if isVMSS {
-		inboundNATPools := []network.InboundNatPool{
+	if !prop.OrchestratorProfile.IsPrivateCluster() {
```
jackfrancis: Because we are now re-using this master LB for private cluster + Standard LB scenarios (in order to grant outbound internet access to the master(s)) we remove the LB attributes that don't make sense in a private cluster scenario, e.g.,, inbound TCP 443 and 22 access.

pkg/engine/loadbalancers.go
```diff
@@ -259,8 +269,8 @@ func createOutboundRules(prop *api.Properties) *[]network.OutboundRule {
 	}
 }
 
-func CreateAgentLoadBalancer(prop *api.Properties, isVMSS bool) LoadBalancerARM {
-
+// CreateStandardLoadBalancerForNodePools returns an ARM resource for the Standard LB that has all nodes in its backend pool
```
jackfrancis: A more accurate func name

pkg/engine/networkinterfaces.go
```diff
@@ -29,13 +29,11 @@ func CreateNetworkInterfaces(cs *api.ContainerService) NetworkInterfaceARM {
 	}
 
 	lbBackendAddressPools := []network.BackendAddressPool{}
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() {
```
jackfrancis: These if not conditions are not necessary as `CreateMasterVMNetworkInterfaces` is never called in a private cluster configuration flow.

pkg/engine/networkinterfaces.go
```diff
@@ -305,8 +304,7 @@ func createAgentVMASNetworkInterface(cs *api.ContainerService, profile *api.Agen
 	} else {
 		dependencies = append(dependencies, "[variables('vnetID')]")
 	}
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
-		profile.LoadBalancerBackendAddressPoolIDs == nil &&
+	if profile.LoadBalancerBackendAddressPoolIDs == nil &&
```
jackfrancis: Add the Standard LB for node pools for private cluster + SLB + VMAS configuration

pkg/engine/networkinterfaces.go
```diff
@@ -352,8 +350,7 @@ func createAgentVMASNetworkInterface(cs *api.ContainerService, profile *api.Agen
 					)
 				}
 			} else {
-				if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
-					cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku &&
+				if cs.Properties.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == api.StandardLoadBalancerSku &&
```
jackfrancis: Add the NIC to the Standard LB backend pool for node pools for private cluster + SLB + VMAS configuration

pkg/engine/armresources.go
devigned: maybe, 

```go
// IsNotAKSAndRequiresASLB should have a solid comment
if cs.Properties.IsNotAKSAndRequiresASLB() {
// ...
}
```
Secondarily, I think this is wrong `!cs.Properties.AnyAgentHasLoadBalancerBackendAddressPoolIDs()`. If there are multiple agent pools and only one of them has a backend pool ID, then we wouldn't make a load balancer and the other agent pools, besides the one that has a backend pool specified, would not be exposed to traffic.

pkg/engine/ipaddresses.go
devigned: code nit:
- There are now two bools on the signature. Perhaps, the function could be decomposed.

Also, what happens in the function if someone calls `CreatePublicIPAddress(false, true)`? That seems like it would be subtly incorrect creating a public IP for the agent, but using the master dns settings. 

```go
func GetMasterDNSSettings() *network.PublicIPAddressDNSSettings { ... }
func CreateMasterPublicIPAddress() PublicIPAddressARM { ... }
func CreateAgentPublicIPAddress() PublicIPAddressARM { ... }
func createPublicIPAddress(name string) { ... }
```
or
```go
func SetMasterDNS(ip PublicIPAddressARM) PublicIPAddressARM { ... }
func CreatePublicIPAddress(name string) PublicIPAddressARM { ... }
```

pkg/engine/masterarmresources.go
```diff
@@ -155,12 +163,14 @@ func createKubernetesMasterResourcesVMSS(cs *api.ContainerService) []interface{}
 		masterResources = append(masterResources, internalLb)
 	}
 
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() {
-		isForMaster := true
-		publicIPAddress := CreatePublicIPAddress(isForMaster)
-		loadBalancer := CreateLoadBalancer(cs.Properties, true)
-		masterResources = append(masterResources, publicIPAddress, loadBalancer)
+	isForMaster := true
+	includeDNS := true
```
devigned: maybe?
```go
includeDNS := !cs.Properties.OrchestratorProfile.IsPrivateCluster()
```

pkg/engine/armresources.go
```diff
@@ -47,13 +47,18 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
 		armResources = append(armResources, userAssignedID, msiRoleAssignment)
 	}
 
-	if !cs.Properties.OrchestratorProfile.IsPrivateCluster() &&
+	// Create the Standard Load Balancer resource spec, so long as:
```
jackfrancis: @devigned @CecileRobertMichon does this comment clarify things?

pkg/engine/armresources_test.go
jackfrancis: Well, this was hear before I invited myself :)

But mainly:

https://raw.githubusercontent.com/Azure/azure-sdk-for-go/v36.2.0/services/network/mgmt/2018-08-01/network/models.go

Sorry, yeah, that file is literally too large for github to handle. Search for `type InterfacePropertiesFormat struct {` and look for the `IPConfigurations` property

pkg/engine/ipaddresses.go
jackfrancis: Agree this should be refactored. I think the addition of a 2nd bool to this func was a lighter footprint than decomposing this func, or creating a distinct "create public IP with no DNS" func.

Also agree that this invites a pathology, I think this approach is the least amount of code updates overall.

Would love to see some targeted, iterative refactoring in this surface area generally, these interfaces are pretty statically implemented at present. :/

pkg/engine/armvariables.go
CecileRobertMichon: can this be added as a code comment?

pkg/engine/armresources_test.go
devigned: @jhendrixMSFT do you know why slices are `*[]type` rather than just `[]type`?

devigned: @jackfrancis for expediency, I guess we should just leave it as is. Just seems like an unneeded indirection.

pkg/engine/ipaddresses.go
devigned: Sounds good. Let's defer and refactor when not pressed for time.

devigned:
/lgtm cancel

@CecileRobertMichon looks like you had a ? or comment to be resolved. I'll defer to you.

pkg/engine/armresources_test.go
jhendrixMSFT: Sorry for the delay.  It's so that we can send empty arrays over-the-wire (there are tests for this in the atuorest test suite).  The other option would be to remove the `,omitempty` from the tag however I'm not sure if this has other unwanted side-effects (I'm looking into this in our track 2 code generator).

devigned: Thank you for the explanation, @jhendrixMSFT.

Seems like removing `,omitempty`, as you point out, would be equivalent functionality with less verbosity. +1 from me

jhendrixMSFT: I looked into this a bit more, unfortunately it breaks the scenario where you need to differentiate between omitting the slice vs sending an empty slice (same goes for maps).  I think we're going to have to live with this.

I keep coming back to having a builder pattern that removes the need to create these struct literals.  I haven't had a chance to explore this in depth, and it might introduce other complexities that offset the gains for pointer-to-type.

devigned: Thank you for jogging my memory. I forgot we had to support empty slices b/c they are significant to some services 😞 . 

Perhaps, those services could mark these aberrant fields in the OpenAPI spec with something like `"x-non-conforming-evil-ptr-slice-bc-my-service-needs-empty-slices": true`, and then all the rest of the APIs can be prettier?? What do you think??

jhendrixMSFT: I like the idea however I think getting all services to correctly update their swaggers would be a hard sell.  Plus I think there are legitimate cases where you actually want to send the empty slice, so getting it right might be tricky.

Another option would be to maintain our own marshalling stack that applies the semantics we want.  We'd have to do it for both JSON and XML, and TBH I don't think the ends justify the cost (maybe it's cheaper than I envision).

PR #2573, opened 2020-01-14
fix: template error when there is no SPN in AKS MSI clusters

norshtein:
**Reason for Change**:

This PR fixes a defeat in https://github.com/Azure/aks-engine/pull/1935/files. In that PR, it always adds `servicePrincipalClientId` and `servicePrincipalClientSecret` in parameters map for AKS clusters which enables user assigned identity(call it AKS MSI cluster below). However, this will cause a defeat: for AKS MSI clusters, if its `servicePrincipalProfile` is nil or its clientID is an empty string, the generated ARM template will have syntax error:

> Template deployment reached terminal state with error: Code="InvalidTemplate" Message="Deployment template validation failed: 'The value for the template parameter 'servicePrincipalClientId' at line '1' and column '14858' is not provided. Please see https://aka.ms/resource-manager-parameter-files for usage details.'." AdditionalInfo=[{"info":{"lineNumber":1,"linePosition":14858,"path":"properties.template.parameters.servicePrincipalClientId"},"type":"TemplateViolation"}]

We have to provide a default value for AKS MSI clusters if its servicePrincipalProfile is empty. This PR fixes the bug.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/params_k8s.go
+++ b/pkg/engine/params_k8s.go
@@ -128,8 +128,7 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 		}
 
 		if kubernetesConfig == nil ||
-			!kubernetesConfig.UseManagedIdentity ||
-			properties.IsHostedMasterProfile() {
+			!kubernetesConfig.UseManagedIdentity {
 			servicePrincipalProfile := properties.ServicePrincipalProfile
 
 			if servicePrincipalProfile != nil {
@@ -152,6 +151,17 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 						addValue(parametersMap, "servicePrincipalObjectId", servicePrincipalProfile.ObjectID)
 					}
 				}
+			}
+		}
+		// The cluster is AKS-hosted cluster and it uses managed identity
+		if kubernetesConfig.UseManagedIdentity && properties.IsHostedMasterProfile() {
+			servicePrincipalProfile := properties.ServicePrincipalProfile
+			if servicePrincipalProfile == nil || servicePrincipalProfile.ClientID == "" || servicePrincipalProfile.Secret == "" {
+				addValue(parametersMap, "servicePrincipalClientId", "msi")
+				addValue(parametersMap, "servicePrincipalClientSecret", "msi")
+			} else {
+				addValue(parametersMap, "servicePrincipalClientId", servicePrincipalProfile.ClientID)
+				addValue(parametersMap, "servicePrincipalClientSecret", servicePrincipalProfile.Secret)
 			}
 		}
 


jackfrancis:
@yangl900 @xuto2 @palma21 could you kindly review this change?

norshtein:
Please do not merge this PR for now. It may cause another potential problem. Let me rethink about it...

norshtein:
Hi @jackfrancis @mboersma , do we have a convenient way to not use any placeholder(currently the placeholder is "msi") for `servicePrincipalClientId` and `servicePrincipalClientSecret` in `azure.json` in agent node?  For some aks addons, they will check whether `servicePrincipalClientId` or `servicePrincipalClientSecret` is an empty string, if not, it will try to authenticate using service principal firstly(even if `useManagedIdentityExtension` is `true`), and the placeholder "msi" may cause addons cannot work correctly

pkg/engine/params_k8s.go
```diff
@@ -154,6 +153,17 @@ func assignKubernetesParameters(properties *api.Properties, parametersMap params
 				}
 			}
 		}
+		// The cluster is AKS-hosted cluster and it uses managed identity
+		if kubernetesConfig.UseManagedIdentity && properties.IsHostedMasterProfile() {
+			servicePrincipalProfile := properties.ServicePrincipalProfile
+			if servicePrincipalProfile == nil || servicePrincipalProfile.ClientID == "" || servicePrincipalProfile.Secret == "" {
+				addValue(parametersMap, "servicePrincipalClientId", "msi")
```
xuto2: Isn't it safer to error out in this case? Burying AKS default behavior in aks-engine will cause confusion down the road

xuto2:
can we do this validation+error or default value assignment in AKS code before it calls AKS-Engine? It doesn't seem like this should be part of AKS-Engine.

pkg/engine/params_k8s.go
norshtein: The SPN will soon disappear in AKS when managed identity is enabled. That is `servicePrincipalProfile == nil || servicePrincipalProfile.ClientID == "" || servicePrincipalProfile.Secret == ""`  will become true soon. We can't produce an error in this case otherwise all clusters enabled managed identity will not work.

norshtein: I got your point. It makes sense to add the default value in aks-rp instead of aks-engine, I will add this in aks-rp.

craiglpeters:
@norshtein is this still pending, or is it being dropped? If you're dropping, please close

norshtein:
PR is not needed. Closed.

PR #2588, opened 2020-01-16
feat: add support for Kubernetes 1.16.6

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#changelog-since-v1164
and https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.16.md#changelog-since-v1165

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Apprently the 1.16.5 hyperkube was also tagged as "1.16.6-beta.0", and that later tag won in the build:

```
Expected
    <string>: v1.16.5
to equal
    <string>: v1.16.6-beta.0


$ ORCHESTRATOR_RELEASE=1.16 akse-deploy 
$ k get no
NAME                                 STATUS   ROLES    AGE   VERSION
k8s-agentpool1-26399701-vmss000000   Ready    agent    53s   v1.16.6-beta.0
k8s-agentpool1-26399701-vmss000001   Ready    agent    59s   v1.16.6-beta.0
k8s-master-26399701-0                Ready    master   59s   v1.16.6-beta.0
```

That's very sad. I really don't want to hack up a test to work around this build error, nor do I want to wait for 1.16.6 in hopes that it will fix it.

mboersma:
/hold

See https://groups.google.com/forum/?nomobile=true#!topic/kubernetes-dev/Q0zNysVxTVE for a summary of current discussion on the git version problem that is blocking this PR. I think there will be a v1.16.6 soon that fixes this issue, so let's update this PR to that release when that happens.

vhd/packer/install-dependencies.sh
```diff
@@ -424,6 +424,7 @@ echo "  - busybox" >> ${VHD_LOGS_FILEPATH}
 K8S_VERSIONS="
 1.17.2
 1.17.1
+1.16.6
```
jackfrancis: Let's remove the 1.16.1 versions from the VHD spec

vhd/packer/configure-windows-vhd.ps1
```diff
@@ -81,6 +81,7 @@ function Get-FilesToCacheOnVHD
             "https://acs-mirror.azureedge.net/wink8s/v1.16.1-1int.zip",
             "https://acs-mirror.azureedge.net/wink8s/v1.16.2-1int.zip",
             "https://acs-mirror.azureedge.net/wink8s/v1.16.4-1int.zip",
+            "https://acs-mirror.azureedge.net/wink8s/v1.16.6-1int.zip",
```
jackfrancis: @marosset Do we want to keep all these Windows versions pre-downloaded into the VHD, even those that aren't supported in the supported list above?

marosset: I think we usually just have supported versions here.

marosset: It won't hurt anything tho, we have plenty of space on the VHD currently.

vhd/packer/install-dependencies.sh
mboersma: I removed 1.16.2 as well, since it's been disabled forever.

PR #2589, opened 2020-01-16
feat: add support for Kubernetes 1.15.9

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1157
and https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#changelog-since-v1158

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

See https://groups.google.com/forum/?nomobile=true#!topic/kubernetes-dev/Q0zNysVxTVE for a summary of current discussion on the git version problem that is blocking this PR. I think there will be a v1.15.9 soon that fixes this issue, so let's update this PR to that release when that happens.

vhd/packer/install-dependencies.sh
```diff
@@ -428,6 +428,7 @@ K8S_VERSIONS="
 1.16.2
 1.16.1
 1.16.1-azs
+1.15.9
```
jackfrancis: Let's remove the two 1.15.4 releases in the VHD spec

Issue #2545, opened 2020-01-08
Automate internal hosting (e.g., acs-mirror) of dependent artifacts (e.g., Azure CNI)


jackfrancis:
Our current acs-mirror platform is manually managed. Let's use a standard upstream-maintained pipeline to automate this instead.

PR #2593, opened 2020-01-16
build: fetch azure-vnet-cni from kubernetesartifacts storage

mboersma:
**Reason for Change**:
Moves the download source for `azure-cni` archives to our preferred production blob storage.

This is second step in an effort to move off the `acs-mirror` blob storage (and replace manual steps with automated pipelines).

**Issue Fixed**:
Refs #2545

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Note that the cni-plugins archive is a different beast not tamed in this PR. This is just azure-vnet-cni.

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -330,11 +330,11 @@
       "type": "string"
     },
     "vnetCniLinuxPluginsURL": {
-      "defaultValue": "https://acs-mirror.azureedge.net/cni/azure-vnet-cni-linux-amd64-latest.tgz",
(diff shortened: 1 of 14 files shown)


mboersma:
/hold

I still need to work out an issue with the backend pipeline that populates this, and then backfill all the old versions that are expected.

mboersma:
/hold cancel

I promoted these versions, which should cover all references to azure-vnet-cni-* in the current code:

v1.0.31
v1.0.30
v1.0.29
v1.0.28
v1.0.17

For each one, the azure-cni pipeline publishes two archive files following this URL pattern:

https://kubernetesartifacts.azureedge.net/azure-cni/v1.0.30/binaries/azure-vnet-cni-linux-amd64-v1.0.30.zip
https://kubernetesartifacts.azureedge.net/azure-cni/v1.0.30/binaries/azure-vnet-cni-windows-amd64-v1.0.30.zip

PR #2597, opened 2020-01-17
feat: Set the default Windows sku and version when upgrading a VMSS

AbelHu:
**Reason for Change**:

This PR sets default WindowsSku and WindowsVersion when upgrading a VMSS.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -714,7 +714,26 @@ func (p *Properties) setWindowsProfileDefaults(isUpgrade, isScale bool) {
 				windowsProfile.ImageVersion = "latest"
 			}
 		}
+	} else if isUpgrade {
+		// Image reference publisher and offer only can be set when you create the scale set so we keep the old values.
+		// Reference: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#create-time-properties
+		if windowsProfile.WindowsPublisher == AKSWindowsServer2019OSImageConfig.ImagePublisher && windowsProfile.WindowsOffer == AKSWindowsServer2019OSImageConfig.ImageOffer {
+			if windowsProfile.ImageVersion == "" {
+				windowsProfile.ImageVersion = AKSWindowsServer2019OSImageConfig.ImageVersion
+			}
+			if windowsProfile.WindowsSku == "" {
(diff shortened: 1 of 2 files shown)


pkg/api/defaults.go
```diff
@@ -714,7 +714,26 @@ func (p *Properties) setWindowsProfileDefaults(isUpgrade, isScale bool) {
...
+	} else if isUpgrade {
+		// Image reference publisher and offer only can be set when you create the scale set so we keep the old values.
+		// Reference: https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-upgrade-scale-set#create-time-properties
+		if windowsProfile.WindowsPublisher == AKSWindowsServer2019OSImageConfig.ImagePublisher && windowsProfile.WindowsOffer == AKSWindowsServer2019OSImageConfig.ImageOffer {
+			if windowsProfile.ImageVersion == "" {
+				windowsProfile.ImageVersion = AKSWindowsServer2019OSImageConfig.ImageVersion
+			}
+			if windowsProfile.WindowsSku == "" {
+				windowsProfile.WindowsSku = AKSWindowsServer2019OSImageConfig.ImageSku
+			}
```
xuto2: Just curious: does this mean current windows upgrade is not working, or someone is picking up the value for us?

AbelHu: Current RP hardcoded a version so it will always set `ImageVersion ` and `WindowsSku` so they are always not empty.

gtxistxgao: This will block user upgrade their images even in upgrade operation. Does this version of AKS-E guarantees to support all the history published windows images?

AbelHu: @gtxistxgao I cannot fully understand your question. This change only impacts those users who have created a cluster with `AKSWindowsServer2019OSImageConfig` or `WindowsServer2019OSImageConfig`. And aks-engine only fills `ImageVersion` and `WindowsSku` when they are not empty.
aks-engine only sets the latest default `ImageVersion` and `ImageSku`. This is similar for creating logic.

gtxistxgao: I see. My misunderstanding. This echos the contract to use tested image version.

xuto2: So this is really making the windows upgrade with the latest AKS-Engine supported image. Thanks for the info @AbelHu @gtxistxgao

PR #2606, opened 2020-01-19
feat: multi AI telemetry keys

### LINKED ISSUE #2575: Enable telemetry to log to more than one App Insights resource (Linux)
Users of aks-engine should be able to take advantage of telemetry too. A user should be able to add their own AppInsights key as well as the aks-engine telemetry endpoint. These telemetry endpoints should be able to be turned on or off independent of each other. That is to say, a user should be able to opt out of sending telemetry to aks-engine and send to only their AppInsights resource.

related to #2415 and #2400, /cc @marosset and @craiglpeters

devigned:
**Reason for Change**:

This pull request enables users and aks-engine to have Application Insights telemetry keys. 

The telemetry key is now injected into the build rather than being hard coded as a const. This should allow us to override the ldflags for a release build to inject a telemetry key upon release.

**Issue Fixed**:

fixes #2575 

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/docs/topics/telemetry.md
+++ b/docs/topics/telemetry.md
@@ -1,9 +1,10 @@
-# [WIP]
+# Telemetry
 
-## Motivation
+## Custom Script Extensions
 
-The aks-engine team is working towards collecting telemetry from custom script extensions (CSEs) used to configure Kubernetes nodes.
-This data will be used to collect aggregate, non-identifiable data to help us answer gather the following information:
+The AKS Engine team has instrumented custom script extensions (CSEs) so collection of telemetry may 
+be enabled. This instrumentation is used to collect aggregate, non-identifiable data to help us 
+answer gather the following information:
 
 - Timing metrics for various operations performed during CSE execution
 - Metrics on failures encountered during CSE execution
@@ -12,15 +13,23 @@ This data will be used to collect aggregate, non-identifiable data to help us an
   - Node pool counts and VM Sku data
   - Which versions of VHDs are still in use
 
-This data will be used to monitor the health of cluster deployments (including AKS clusters) that are deployed with aks-engine as well as to help us prioritize future investments/feature work in the tool.
+This data will be used to monitor the health of cluster deployments (including AKS clusters) that 
+are deployed with AKS Engine as well as to help us prioritize future investments/feature work in 
+the tool.
 
 ## Configuration
 
 Collection of all telemetry is currently **disabled** by default.
 
-Telemetry can be enabled by setting the `EnableTelemetry` feature flag to `true` in the apimodel.json.
+Telemetry can be enabled by setting the `enableTelemetry` feature flag to `true` in the 
+apimodel.json. When `enableTelemetry` is set to `true`, telemetry will be sent to the AKS Engine 
+Application Insights cluster.
 
-Telemetry can be routed to different application insights instance be specifying `telemetryProfile.applicationInsightsKey` to the instrumentation key of your application insights instance.
+Telemetry can be routed to an additional Application Insights instance by specifying 
+`telemetryProfile.applicationInsightsKey` with the value of the instrumentation key of your 
+Application Insights instance. 
+
+That means you can see and use the same data that is being sent to the AKS Engine team.
 
 ``` javascript
 {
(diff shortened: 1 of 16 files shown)


marosset:
Can you update with the new behavior (where a an application key is no longer required to get 'system' telemetry)?

pkg/api/defaults.go
```diff
@@ -774,7 +776,7 @@ func (p *Properties) setTelemetryProfileDefaults() {
 	}
 
 	if len(p.TelemetryProfile.ApplicationInsightsKey) == 0 {
-		p.TelemetryProfile.ApplicationInsightsKey = DefaultApplicationInsightsKey
+		p.TelemetryProfile.ApplicationInsightsKey = telemetry.AKSEngineAppInsightsKey
```
marosset: Should this be empty string since TelemetryProfile.ApplicationInsightsKey is only used for user keys?

docs/topics/telemetry.md
```diff
@@ -12,15 +13,23 @@ This data will be used to collect aggregate, non-identifiable data to help us an
...
 
 Collection of all telemetry is currently **disabled** by default.
 
-Telemetry can be enabled by setting the `EnableTelemetry` feature flag to `true` in the apimodel.json.
+Telemetry can be enabled by setting the `enableTelemetry` feature flag to `true` in the 
+apimodel.json. When `enableTelemetry` is set to `true`, telemetry will be sent to the aks-engine 
+Application Insights cluster.
 
-Telemetry can be routed to different application insights instance be specifying `telemetryProfile.applicationInsightsKey` to the instrumentation key of your application insights instance.
+Telemetry can be routed to an additional Application Insights instance by specifying
```
craiglpeters: Seems we should have a statement about where the data goes by default before we introduce an additional sink

devigned: Just before this, I say the following:
> When `enableTelemetry` is set to `true`, telemetry will be sent to the aks-engine 
 Application Insights cluster.

Should this be rephrased?

docs/topics/telemetry.md
```diff
@@ -1,9 +1,10 @@
-# [WIP]
+# Telemetry
 
-## Motivation
+## Custom Script Extensions
 
-The aks-engine team is working towards collecting telemetry from custom script extensions (CSEs) used to configure Kubernetes nodes.
-This data will be used to collect aggregate, non-identifiable data to help us answer gather the following information:
+The aks-engine team has enabled collecting telemetry from custom script extensions (CSEs) used to
```
marosset: We should reword this to something like 'The aks-engine team has instrumented custom script extensions (CSEs) so collection of telemetry may be enabled'

This seems more neutral instead of leading this this is enabled and then later saying it is actually disabled by default.

Thoughts?

docs/topics/telemetry.md
```diff
@@ -35,3 +44,21 @@ Telemetry can be routed to different application insights instance be specifying
...
+
+### Collection Settings and Results
+
+aks-engine has the concept of a system defined aks-engine Application Insights key as well as a user
+defined key. The table below describes where telemetry is logged based on the configuration
+settings.
+
+| enableTelemetry  | applicationInsightsKey set | result |
+| ------------- | ------------- | ------------- |
+| true | true | both user and aks-e telemetry is tracked |
```
marosset: i think we should stick with aks-engine instread of aks-e unless other docs refer to aks-e.
@CecileRobertMichon thoughts?

(same for below refs)

CecileRobertMichon: `AKS Engine` is the preferred wording in docs

pkg/api/defaults.go
```diff
@@ -774,7 +775,7 @@ func (p *Properties) setTelemetryProfileDefaults() {
 	}
 
 	if len(p.TelemetryProfile.ApplicationInsightsKey) == 0 {
```
marosset: I wonder if we should rename this something like UserApplicationInsightsKey/SecondaryApplicationInsightsKey (or anything that would differential it from the one our team will use to collect data) for code clarity

devigned: All of the settings in the api-model are for users. I think it makes sense for a user with the current name. Really, they should not need to know about the system key. (IMO)

marosset: Agreed, but code maintainers should be able to easily distinguish.
Whatever makes that easiest.

docs/topics/telemetry.md
devigned: Updated all references to be `AKS Engine`

PR #2613, opened 2020-01-22
chore: Bump moby to 3.0.10

cpuguy83:
**Reason for Change**:

Update moby to 3.0.9

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -240,7 +240,7 @@
       }
     },
     "mobyVersion": {
-      "defaultValue": "3.0.8",
+      "defaultValue": "3.0.10",
       "metadata": {
         "description": "The Azure Moby build version"
       },
@@ -252,7 +252,8 @@
          "3.0.5",
          "3.0.6",
          "3.0.7",
-         "3.0.8"
+         "3.0.8",
+         "3.0.10"
        ],
       "type": "string"
     },
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -37312,7 +37312,7 @@ installMoby() {
         if [[ "${MOBY_CLI}" == "3.0.4" ]]; then
             MOBY_CLI="3.0.3"
         fi
-        apt_get_install 20 30 120 moby-engine=${MOBY_VERSION} moby-cli=${MOBY_CLI} --allow-downgrades || exit $ERR_MOBY_INSTALL_TIMEOUT
+        apt_get_install 20 30 120 moby-engine=${MOBY_VERSION}* moby-cli=${MOBY_CLI}* --allow-downgrades || exit $ERR_MOBY_INSTALL_TIMEOUT
     fi
 }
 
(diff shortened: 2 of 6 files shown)


parts/k8s/kubernetesparams.t
```diff
@@ -240,7 +240,7 @@
       }
     },
     "mobyVersion": {
-      "defaultValue": "3.0.8",
+      "defaultValue": "3.0.9",
```
CecileRobertMichon: you also need to update the VHD script to install 3.0.9 (installdependencies.sh)

cpuguy83: Fixed... I think?

parts/k8s/kubernetesparams.t
```diff
@@ -240,7 +240,7 @@
       }
     },
     "mobyVersion": {
-      "defaultValue": "3.0.8",
+      "defaultValue": "3.0.9+azure-38",
```
CecileRobertMichon: why the `+azure-38` tag?

cpuguy83: `+azure` is an extra hint to users that it's not some really old version of docker.
The `-38` is the version of the package, essentially. So we can change something package related only and not bump engine versions.

I was hoping `apt-get install moby-engine=3.0.9` would work with this scheme it it apparently wants the full version only.
This is likely to get even more complicated as we move away from static compilation and have packages targeted for specific distro versions.

cpuguy83: What I'm going to recommend is that we create a metapackage like `aks-engine-moby` which can hold all the specific dependencies for a given version.
So we'd a package with the desired simple version (e.g. 3.0.9), inside the package it's just a list of dependencies and their versions. This also fixes the issue of having to keep CLI and engine versions in lock step.

cpuguy83: WDYT?

pkg/engine/templates_generated.go
```diff
@@ -37312,7 +37312,7 @@ installMoby() {
         if [[ "${MOBY_CLI}" == "3.0.4" ]]; then
             MOBY_CLI="3.0.3"
         fi
-        apt_get_install 20 30 120 moby-engine=${MOBY_VERSION} moby-cli=${MOBY_CLI} --allow-downgrades || exit $ERR_MOBY_INSTALL_TIMEOUT
+        apt_get_install 20 30 120 moby-engine=${MOBY_VERSION}* moby-cli=${MOBY_CLI}* --allow-downgrades || exit $ERR_MOBY_INSTALL_TIMEOUT
```
cpuguy83: PTAL.
The glob here matches the latest version matching the glob.
So, for instance, if we happen to have something like `3.0.10+azure-1` and `3.0.10+azure-2` available, a value of `3.0.10` in the spec will get the `-2`.

cpuguy83:
I've changed the version to 3.0.10 which includes a version of runc with CVE mitigations.
I've opted to not add 3.0.9 to the list of available versions since we never had it merged in and it has a known issue.

cpuguy83:
> fatal error: concurrent map read and map write

cpuguy83:
Looks like there's some parallel tests that are using a shared object?

jackfrancis:
@cpuguy83 If you're curious the "k8s_non_vhd_deployment" pr-e2e job validates this change (as does the "AKS Engine VHD Pipeline" job, although that only validates that the VHD CI script successfully traverses your changes to install the new version of moby, it doesn't actually run k8s E2E tests).

PR #2617, opened 2020-01-22
build: fetch cni-plugins from kubernetesartifacts storage

mboersma:
**Reason for Change**:
Moves the download source for `cni-plugins` archives to our preferred production blob storage.

This is third step in an effort to move off the `acs-mirror` blob storage (and replace manual steps with automated pipelines).

**Issue Fixed**:
Refs #2545

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -93,7 +93,7 @@ var (
 		EtcdDownloadURLBase:              "https://acs-mirror.azureedge.net/github-coreos",
 		KubeBinariesSASURLBase:           "https://acs-mirror.azureedge.net/wink8s/",
 		WindowsTelemetryGUID:             "fb801154-36b9-41bc-89c2-f4d4f05472b0",
-		CNIPluginsDownloadURL:            "https://acs-mirror.azureedge.net/cni/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
+		CNIPluginsDownloadURL:            "https://kubernetesartifacts.azureedge.net/cni-plugins/" + CNIPluginVer + "/binaries/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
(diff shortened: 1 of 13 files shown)


pkg/api/azenvtypes.go
```diff
@@ -93,7 +93,7 @@ var (
 		EtcdDownloadURLBase:              "https://acs-mirror.azureedge.net/github-coreos",
 		KubeBinariesSASURLBase:           "https://acs-mirror.azureedge.net/wink8s/",
 		WindowsTelemetryGUID:             "fb801154-36b9-41bc-89c2-f4d4f05472b0",
-		CNIPluginsDownloadURL:            "https://acs-mirror.azureedge.net/cni/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
+		CNIPluginsDownloadURL:            "https://kubernetesartifacts.azureedge.net/cni-plugins/" + CNIPluginVer + "/binaries/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
```
jackfrancis: I think we want to sync up w/ mooncake config owners to ensure we don't inadvertently make their existing acs-mirror mirror (sorry) stale.

@feiskyer @andyzhangx do you maintain `https://mirror.azk8s.cn`? If so, do you know if it is manually maintained, or does it pull from acs-mirror automatically?

mboersma: Ideally, we could have the backend pipeline also populate `mirror.azk8s.cn`. cc: @jchauncey.

andyzhangx: @jackfrancis
it's automatically populated from https://github.com/containernetworking/plugins, all binaries are under:
https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/

jackfrancis: Thanks @andyzhangx! @mboersma I think this means we are good to go here. As we continue to do this work to migrate off of acs-mirror, let's continue to check in w/ mooncake folks to ensure we don't break those mirrors; sounds like we're in good shape

PR #2620, opened 2020-01-23
test: configurable pvc E2E tests

jackfrancis:
**Reason for Change**:

This PR modifies the PVC E2E test implementation so that it only runs if a `TEST_PVC` env var is set to `true` (in addition to other, pre-existing conditional criteria).

We will ship this change w/ a change to the "everything" test cluster configuration, so that that cluster config gets PVC test coverage.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@chewong FYI

PR #2624, opened 2020-01-24
fix: added resource limits for networkmonitor daemonset

tamilmani1989:
**Reason for Change**:

This PR addresses defining resource request and limits for network monitor daemonset running in Azure CNI clusters.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/addons.go
```diff
@@ -306,8 +306,12 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
...
 		Containers: []KubernetesContainerSpec{
 			{
-				Name:  common.AzureCNINetworkMonitorAddonName,
-				Image: specConfig.AzureCNIImageBase + k8sComponents[common.AzureCNINetworkMonitorAddonName],
+				Name:           common.AzureCNINetworkMonitorAddonName,
+				Image:          specConfig.AzureCNIImageBase + k8sComponents[common.AzureCNINetworkMonitorAddonName],
+				CPURequests:    "30m",
+				MemoryRequests: "25Mi",
+				CPULimits:      "200m",
+				MemoryLimits:   "512Mi",
```
jackfrancis: These limits numbers seem high, is that what's being used in practice?

tamilmani1989: which one is higher cpu or memory?

jackfrancis: Both, but memory more so.

These are sidecar "health check" style bash scripts that run as daemonsets, right? I think using up a half-gig of memory in particular as required overhead, per-node, just to use Azure CNI, seems wrong. @mboersma @devigned  thoughts?

tamilmani1989: yes this daemonset would not consume 512M. I just gave that as limit but potentially it should not hit that. Same goes for cpu

jackfrancis: The purpose of a resource limit is to define the upper bound of a container beyond which a pod containing it may be terminated (depending on the density of the node on which the pod is scheduled). Let's set the limit to that number. It should not be, I don't think, over an order of magnitude greater than the expected working footprint (in this case `25Mi`). A good rule of thumb is to define your requests with a little bit of overhead beyond the stable state that your stress tests prove are the norm (say, 2x); and then ideally your limits value would be identical. Otherwise the configuration signals that the underlying runtime is not stable.

Again, would love thoughts from @devigned and @mboersma. This particular runtime doesn't sound like one that needs a lot of tolerance for memory peaks and valleys, and so setting a higher limit would (1) prevent optimal node pod scheduling density for folks running Azure CNI (they may have to absorb 512Mi *just* for a health check) and (2) obfuscate memory leaks in the underlying code as a result of excessive tolerance.

Also cc @CecileRobertMichon @marosset @ritazh

devigned: I like the suggestion of 2x more than stable state as general guidance. Just keep it as low as reasonable and be able to recover if the container gets killed for going over that limit.

mboersma:
@tamilmani1989 I'm sorry this PR has been here so long. Do you have time to rebase it and to revise the limit numbers?

tamilmani1989:
> @tamilmani1989 I'm sorry this PR has been here so long. Do you have time to rebase it and to revise the limit numbers?

Sorry @mboersma For delayed response. Yes i will change mem to 256M and rebase it. will do by this week

tamilmani1989:
@mboersma addressed comments

mboersma:
@tamilmani1989 the generated code needs to be recreated with `make generate` and committed.

Also, these numbers are less than before, but still seem like a [WAG](https://en.wikipedia.org/wiki/Scientific_wild-ass_guess) 😄. Unless network monitor has sudden huge bursts of memory usage, setting the memory limit an order of magnitude more than the memory request may be overkill. Is there memory profiling data that shows a usage pattern like this? If not, we think 2x (`"50Mi"`) would be safer.. 

Similarly for CPU usage: does it need to burst from 30 to 200? If the request values are similar to real-world numbers, and the usage is relatively stable, then setting the limit values excessively high doesn't help the Kubernetes scheduler make good decisions.

tamilmani1989:
@mboersma I think 256 Mb should be good, i saw memory peak to 90Mb  when 250 pods are created.  Regarding cpu usage, what do you consider as ideal? 100?

mboersma:
@tamilmani1989 @jackfrancis what if we went with:

```go
				CPURequests:    "30m",
				MemoryRequests: "25Mi",
				CPULimits:      "100m",
				MemoryLimits:   "100Mi",
```

That seems generous enough given the lack of data. 😄 And `MemoryLimits` a bit more than the peak observed in load testing sounds like what we are aiming for.

Let me know what you think, I'm happy to push a commit on top of this.

tamilmani1989:
> @tamilmani1989 @jackfrancis what if we went with:
> 
> ```go
> 				CPURequests:    "30m",
> 				MemoryRequests: "25Mi",
> 				CPULimits:      "100m",
> 				MemoryLimits:   "100Mi",
> ```
> 
> That seems generous enough given the lack of data. 😄 And `MemoryLimits` a bit more than the peak observed in load testing sounds like what we are aiming for.
> 
> Let me know what you think, I'm happy to push a commit on top of this.

I saw network monitor peaked to 100 Mi once. So its better if we have upper bound as 150 atleast

mboersma:
@tamilmani1989 we talked about this yesterday and the numbers you have are fine with us. We can always adjust in the future based on more data. It's important to establish requests and limits in the first place.

I rebased this on master and added the same changes to the 1.18 and 1.19 manifests. It should be ready to merge now after CI passes. Thanks for your patience! 😄

PR #2631, opened 2020-01-27
ci: work around linter by building binary first

mboersma:
**Reason for Change**:
#2534 broke the CI unit test step because of a `golangci-lint` bootstrap issue. This works around it by building the binary first, which I've verified fixes it locally.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Back to the drawing board.

PR #2638, opened 2020-01-28
fix: scale operation looks for windows name for linux pool

### LINKED ISSUE #2637: scale operations on linux node pools throw superfluous error
Scale operations throws a superfluous error message:

`ERRO[0005] resource name was missing from identifier`

This doesn't (seemingly) affect the actual scale operation, but it is unsettling.

jackfrancis:
**Reason for Change**:

This PR ensures we don't parse for a Windows name during scale operations unless we are running in the context of a Windows node pool.

This fixes a regression introduced in #2483

**Issue Fixed**:

Fixes #2637

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -268,10 +268,9 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					continue
 				}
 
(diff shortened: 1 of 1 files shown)


jackfrancis:
See #2643

PR #2643, opened 2020-01-28
fix: make Windows scaling work

### LINKED ISSUE #2637: scale operations on linux node pools throw superfluous error

### LINKED ISSUE #2642: scaling a Windows VMSS node pool creates a new VMSS
Create a new cluster with a Windows pool named "poolwin". You'll get a VMSS node pool like this:

```
$ k get nodes -o json
NAME                                 STATUS   ROLES    AGE     VERSION   INTERNAL-IP      EXTERNAL-IP   OS-IMAGE                         KERNEL-VERSION      CONTAINER-RUNTIME
4128k8s03000000                      Ready    <none>   18s     v1.16.6   10.239.48.4      <none>        Windows Server 2019 Datacenter   10.0.17763.864      docker://19.3.2
4128k8s03000001                      Ready    <none>   16s     v1.16.6   10.239.48.35     <none>        Windows Server 2019 Datacenter   10.0.17763.864      docker://19.3.2
...
```

(The `4128k8s03000000` and `4128k8s03000001` nodes above are part of a VMSS named `4128k8s03`)

When you use `aks-engine scale` and pass in the `--node-pool poolwin` arg, instead of scaling the above `4128k8s03` VMSS, it creates a new node pool backed by a VMSS named `4128k8s00`.

Ostensibly the string foo that translates a meaningful string (e.g., "poolwin") to a condensed VMSS name string (e.g., "4128k8s03") is not the same foo for deploy and scale.

jackfrancis:
**Reason for Change**:

This PR will make Windows scaling actually work. The current scaling implementation uses an incorrect substring parsing implementation to derive the "pool index suffix", which is used at cluster creation time to compose the name of a Windows VMSS resource. Ultimately, this incorrect parsing has two bad side-effects:

- it prevents the scale operation from correctly identifying the VMSS resource when the `--node-pool` argument value is the pool name of a Windows node pool
- the above has the side-effect of creating a *new* Windows node pool with the desired `--new-node-count`, instead of failing the scale operation

Additionally, this PR fixes a bug which was causing VHD-backed Linux node pools from entering the first bad side-effect above, and logging a spuriuos error `ERRO[0004] resource name was missing from identifier` for such Linux node pool scaling scenarios.

**Issue Fixed**:

Fixes #2642
Fixes #2637 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -374,8 +375,8 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				return errors.Wrap(err, "failed to get VMSS list in the resource group")
 			}
 			for _, vmss := range vmssListPage.Values() {
-				vmName := *vmss.Name
-				if !sc.vmInAgentPool(vmName, vmss.Tags) {
+				vmssName := *vmss.Name
+				if !sc.vmInAgentPool(vmssName, vmss.Tags) {
 					continue
 				}
 
@@ -389,10 +390,13 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					}
 				}
 
-				osPublisher := vmss.VirtualMachineProfile.StorageProfile.ImageReference.Publisher
-				if osPublisher != nil && (strings.EqualFold(*osPublisher, "MicrosoftWindowsServer") || strings.EqualFold(*osPublisher, "microsoft-aks")) {
-					_, _, winPoolIndex, _, err = utils.WindowsVMNameParts(vmName)
-					log.Errorln(err)
+				if sc.agentPool.OSType == api.Windows {
+					winPoolIndexStr := vmssName[len(vmssName)-2:]
+					var err error
+					winPoolIndex, err = strconv.Atoi(winPoolIndexStr)
+					if err != nil {
+						return errors.Wrap(err, "failed to get Windows pool index from VMSS name")
+					}
 				}
 
 				currentNodeCount = int(*vmss.Sku.Capacity)


jackfrancis:
Tested with VMSS scale up and down against a cluster with a single Windows node pool.

cmd/scale.go
```diff
@@ -374,8 +375,8 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				return errors.Wrap(err, "failed to get VMSS list in the resource group")
 			}
 			for _, vmss := range vmssListPage.Values() {
-				vmName := *vmss.Name
-				if !sc.vmInAgentPool(vmName, vmss.Tags) {
+				vmssName := *vmss.Name
```
jackfrancis: This is a code hygiene change: we are re-using the VMSS name as a variable, not the name of an individual vm instance. Naming the variable `vmssName` is clearer.

cmd/scale.go
```diff
@@ -389,10 +390,13 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					}
 				}
 
-				osPublisher := vmss.VirtualMachineProfile.StorageProfile.ImageReference.Publisher
-				if osPublisher != nil && (strings.EqualFold(*osPublisher, "MicrosoftWindowsServer") || strings.EqualFold(*osPublisher, "microsoft-aks")) {
-					_, _, winPoolIndex, _, err = utils.WindowsVMNameParts(vmName)
-					log.Errorln(err)
+				if sc.agentPool.OSType == api.Windows {
+					winPoolIndexStr := vmssName[len(vmssName)-2:]
```
jackfrancis: This is bad. See:

https://github.com/Azure/aks-engine/blob/master/pkg/api/types.go#L916

Essentially, we bake a two digit integer into the pool name, which we are only able to determine conclusively at cluster creation time by enumerating through the api model array. There's nothing JSON-breaking that would prevent the api model from being changed over time in such a way where the `agentPoolProfiles` array items get re-ordered. So the only way we can determine the index that cluster creation appended into the VMSS name is to get the last two chars and convert it to an int.

And then we use that int starting in Line 466 below.

I think this solution is the most practical way forward without a significant refactor.

jackfrancis:
@jadarsie Can you confirm that Azure Stack windows VMSS names look like this:

https://github.com/Azure/aks-engine/blob/master/pkg/api/types.go#L916

E.g.: `4128k8s00 `

PR #2644, opened 2020-01-28
build: fetch etcd from MCR container image

mboersma:
**Reason for Change**:
Uses Microsoft's container image builds to install `etcd` and `etcdctl`, instead of requiring a tarball from the acs-mirror storage.

This is the fourth step in an effort to move off the acs-mirror blob storage (and replace manual steps with automated pipelines).

**Issue Fixed**:
Refs #2545 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This may take slightly longer than the equivalent "download and untar" function it replaces, is my guess.


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -83,14 +83,14 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
-if [[ -n "${MASTER_NODE}" ]] && [[ -z "${COSMOS_URI}" ]]; then
-    time_metric "InstallEtcd" installEtcd
-fi
-
 {{- if not HasCoreOS}}
 time_metric "InstallContainerRuntime" installContainerRuntime
 {{end}}
 
+if [[ -n "${MASTER_NODE}" ]] && [[ -z "${COSMOS_URI}" ]]; then
+    time_metric "InstallEtcd" installEtcd "docker"
```
jackfrancis: ooh ooh we can compare before/after etcd installation methods cc: @devigned

PR #2649, opened 2020-01-30
chore: use azure containerd packages

### LINKED ISSUE #2370: containerd integration using deprecated "cri" binary

cpuguy83:
**Reason for Change**:

**Issue Fixed**:

Fixes #2370

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

This is a POC.
Looking for feedback on the implementation.
Also need tests to run, currently having issues deploying things.

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -24,6 +24,10 @@ removeMoby() {
     apt-get purge -y moby-engine moby-cli
 }
 
+removeContainerd() {
+    apt-get purge -y moby-containerd moby-runc
+}
+
 installEtcd() {
     CURRENT_VERSION=$(etcd --version | grep "etcd Version" | cut -d ":" -f 2 | tr -d '[:space:]')
     if [[ "$CURRENT_VERSION" == "${ETCD_VERSION}" ]]; then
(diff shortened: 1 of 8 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -227,18 +231,14 @@ installContainerd() {
...
-        rm -Rf /usr/bin/containerd
-        rm -Rf /var/lib/docker/containerd
-        rm -Rf /run/docker/containerd
-        if [[ ! -f "$CONTAINERD_DOWNLOADS_DIR/${CONTAINERD_TGZ_TMP}" ]]; then
-            downloadContainerd
-        fi
-        tar -xzf "$CONTAINERD_DOWNLOADS_DIR/$CONTAINERD_TGZ_TMP" -C /
-        sed -i '/\[Service\]/a ExecStartPost=\/sbin\/iptables -P FORWARD ACCEPT -w' /etc/systemd/system/containerd.service
-        echo "Successfully installed cri-containerd..."
+        removeContainerd
```
jackfrancis: Shall I infer that this is a "best practice, just-in-case" type operation? In other words, the most deterministic way to install a particular version of the `moby-containerd` is to ensure no pre-existing `moby-containerd` and `moby-runc` packages aren't present on the OS already?

Also, is `moby-runc` installed as a dependency of `moby-containerd`? Or is it just a conflicting package and so we need to ensure it's removed before we install `moby-containerd`?

cpuguy83: I copied this pattern from `installMoby` is all.

`moby-runc` is a dependency for `moby-containerd` and is automatically installed with `moby-containerd`.

jackfrancis:
@cpuguy83 Are the `moby-engine` and `moby-cli` installed packages able to co-exist with `moby-containerd`? If so, might we consider pre-installing `moby-containerd` on the VHD so we don't have to fetch all the apt artifacts during bootstrap runtime?

cpuguy83:
> Are the moby-engine and moby-cli installed packages able to co-exist with moby-containerd

`moby-cli` yes, `moby-engine`, not yet...
`moby-containerd` explicitly lists `moby-engine` as a conflict right now.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -227,18 +231,14 @@ installContainerd() {
...
+        echo "deb [arch=amd64,arm64,armhf] https://packages.microsoft.com/ubuntu/${UBUNTU_RELEASE}/multiarch/prod testing main" > /tmp/microsoft-prod-testing.list
+        retrycmd_if_failure 10 5 10 cp /tmp/microsoft-prod.list /etc/apt/sources.list.d/ || exit $ERR_MOBY_APT_LIST_TIMEOUT
```
cpuguy83: Oops

parts/k8s/cloud-init/artifacts/cse_install.sh
jackfrancis: Got it.

Should we invoke `removeMoby` at the beginning of the containerd installation process to ensure we don't ship both packages?

Likewise we should probably invoke `removeContainerd` during moby installation.

PR #2651, opened 2020-01-30
refactor: extract common script parameters into function

CecileRobertMichon:
**Reason for Change**:

Refactor `provisionScriptParametersCommon` into a function so it can be unit tested. Also use functions to get the values for most variable instead of relying on ARM template functions (--> more unit tests).
This is also a first step towards potentially allow other projects to call this exported function to get the variables necessary for running the bootstrap script.
The same refactor will follow for `provisionScriptParametersMaster`.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/types.go
```diff
@@ -2134,6 +2167,71 @@ func (cs *ContainerService) GetAzureProdFQDN() string {
 	return FormatProdFQDNByLocation(cs.Properties.MasterProfile.DNSPrefix, cs.Location, cs.Properties.GetCustomCloudName())
 }
 
+// GetProvisionScriptParametersCommon returns the environment variables needed to run the Linux bootstrap scripts
+// Ensure that the clientSecret parameter is surrounded by single quotes to protect against special characters
+func (cs *ContainerService) GetProvisionScriptParametersCommon(location, resourceGroup, tenantID, subscriptionID, clientID, clientSecret, apiServerCertificate, kubeletPrivateKey, clusterKeyVaultName string) string {
```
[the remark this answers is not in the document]
CecileRobertMichon: Yep good call

pkg/api/types.go
```diff
@@ -2134,6 +2167,84 @@ func (cs *ContainerService) GetAzureProdFQDN() string {
...
+	KubeletPrivateKey    string
+	ClusterKeyVaultName  string
+}
+
+// GetProvisionScriptParametersCommon returns the environment variables needed to run the Linux bootstrap scripts
+// Ensure that the clientSecret parameter is surrounded by single quotes to protect against special characters
+func (cs *ContainerService) GetProvisionScriptParametersCommon(input ProvisionScriptParametersInput) string {
+	cloudSpecConfig := cs.GetCloudSpecConfig()
+	kubernetesConfig := cs.Properties.OrchestratorProfile.KubernetesConfig
+	provisionScriptParametersCommonString := "" +
```
serbrech: nit:
You could use strings.Builder

```
var b strings.Builder
b.WriteString("...")
b.WriteString("...")
return b.String()
```

pkg/api/types.go
```diff
@@ -1935,6 +1946,28 @@ func (p *Properties) IsNVIDIADevicePluginEnabled() bool {
...
+// GetKubernetesVersion returns the cluster Kubernetes version, with the Azure Stack suffix if Azure Stack Cloud.
+func (p *Properties) GetKubernetesVersion() string {
+	if p.IsAzureStackCloud() {
+		return p.OrchestratorProfile.OrchestratorVersion + AzureStackSuffix
+	}
+	return p.OrchestratorProfile.OrchestratorVersion
+}
+
+// GetKubernetesHyperkubeSpec returns the string to use for the Kubernetes hyperkube image.
+func (p *Properties) GetKubernetesHyperkubeSpec() string {
```
jackfrancis: Looking forward to rebasing on top of this comming and using function in my own refactor PR here:

https://github.com/Azure/aks-engine/pull/2540/files#diff-9d1171fa68cde0d400f802311a34ceccR205

:)

pkg/engine/armvariables.go
```diff
@@ -78,70 +78,17 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
 	hasCosmosEtcd := masterProfile != nil && masterProfile.HasCosmosEtcd()
-	isIPv6DualStackFeatureEnabled := cs.Properties.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack")
-
-	kubernetesVersion := orchProfile.OrchestratorVersion
-	if cs.Properties.IsAzureStackCloud() {
-		kubernetesVersion = orchProfile.OrchestratorVersion + AzureStackSuffix
-	}
-
-	provisionScriptParametersCommonString := "[concat(" +
-		"'ADMINUSER=',parameters('linuxAdminUsername')," +
```
jackfrancis: Should we audit all params under `parts/k8s/kubernetesparams.t` & `pkg/engine/params_k8s` (& maybe `pkg/engine/params.go`) and remove all ARM params handling for params that are no longer being used?

CecileRobertMichon: I did that and removed 2 of them, the rest were still being used one way or another. I think we can reduce further with more refactoring but I stopped there for the purpose of this PR.

pkg/api/types.go
CecileRobertMichon: Yeah that's a good point. It does make the concatenation more efficient but will make the code harder to read because it will require putting each WriteString on a new line (ie. double the lines it's on now) and the string for the env var name won't be next to its value anymore. Thoughts?

pkg/engine/armvariables.go
```diff
@@ -78,70 +78,17 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
-		"' CLOUDPROVIDER_RATELIMIT_BUCKET_WRITE=',parameters('cloudproviderConfig').cloudProviderRatelimitBucketWrite," +
-		"' LOAD_BALANCER_DISABLE_OUTBOUND_SNAT=',toLower(string(parameters('cloudproviderConfig').cloudProviderDisableOutboundSNAT))," +
-		"' USE_MANAGED_IDENTITY_EXTENSION=',variables('useManagedIdentityExtension')," +
-		"' USE_INSTANCE_METADATA=',variables('useInstanceMetadata')," +
-		"' LOAD_BALANCER_SKU=',variables('loadBalancerSku')," +
-		"' EXCLUDE_MASTER_FROM_STANDARD_LB=',variables('excludeMasterFromStandardLB')," +
-		"' MAXIMUM_LOADBALANCER_RULE_COUNT=',variables('maximumLoadBalancerRuleCount')," +
-		"' CONTAINER_RUNTIME=',parameters('containerRuntime')," +
-		"' CONTAINERD_DOWNLOAD_URL_BASE=',parameters('containerdDownloadURLBase')," +
-		"' POD_INFRA_CONTAINER_SPEC=',parameters('kubernetesPodInfraContainerSpec')," +
```
CecileRobertMichon: Deleted this var that wasn't used anywhere @jackfrancis fyi

pkg/api/types.go
serbrech: I think it would allow you to break that method in a smaller one: 
`AddParameter(builder, name, value)`

jackfrancis: How about something like this?

https://play.golang.org/p/-oi0IU1_N36

pkg/api/types.go
```diff
@@ -2186,63 +2186,78 @@ type ProvisionScriptParametersInput struct {
...
+		"NETWORK_API_VERSION":                  APIVersionNetwork,
+		"NETWORK_MODE":                         kubernetesConfig.NetworkMode,
+		"KUBE_BINARY_URL":                      kubernetesConfig.CustomKubeBinaryURL,
+	}
+
+	keys := make([]string, 0)
+	for k := range parameters {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
jackfrancis: ooh sorted

CecileRobertMichon: yeah you can't range through a map in order in go so it was that or rewrite the unit tests... But I think having a sorted result is nice so that's bonus

PR #2653, opened 2020-01-31
fix: persist new pool after addpool operation

jackfrancis:
**Reason for Change**:

This PR properly persists the additional node pool after a successful `aks-engine addpool` operation, so that subsequent operations (e.g., scale, upgrade) will be able to know about the new pool via the canonical api model JSON.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/addpool.go
+++ b/cmd/addpool.go
@@ -44,7 +44,6 @@ type addPoolCmd struct {
 	client           armhelpers.AKSEngineClient
 	locale           *gotext.Locale
 	nameSuffix       string
-	nodePoolIndex    int
 	logger           *log.Entry
 	apiserverURL     string
 	kubeconfig       string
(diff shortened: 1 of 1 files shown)


jackfrancis:
@DavidLangworthy for a look as he's the original generous contributor of this feature

cmd/addpool.go
```diff
@@ -44,7 +44,6 @@ type addPoolCmd struct {
 	client           armhelpers.AKSEngineClient
 	locale           *gotext.Locale
 	nameSuffix       string
-	nodePoolIndex    int
```
jackfrancis: `nodePoolIndex` is (AFAICT) unused at this point, probably the easiest implementation is to just append the new pool to the end of the existing `AgentPoolProfiles` array. (There shouldn't be a need to keep track of any array indices)

DavidLangworthy: Looks fine to me.  I don't remember using it anywhere. 
Thanks

PR #2655, opened 2020-01-31
 build: fetch k8s Windows .zip from kubernetesartifacts storage

mboersma:
**Reason for Change**:
Moves the download source for the `v1.17.2-1int.zip`-named Windows support archives for Kubernetes to our preferred artifact storage. This is integrated with the Azure Pipelines build for Kubernetes (thanks to @sozercan!).

Integrating the Azure Stack variant is still a `TODO:` so this leaves the build script and Makefile target for that intact.

**Issue Fixed**:
Refs #2545

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

Need to follow up on the availability of 1.13 and 1.18 artifacts, seems like CI is complaining.

sozercan:
@mboersma which ones are we missing?

mboersma:
@sozercan none of them? 😅  I did some `wget` tests and it looks like everything is backfilled as expected, so the CI failure must be something unrelated after all.

mboersma:
Kubernetes v1.13.11 and v1.18.0-alpha.2 are the only two versions that aren't supported by the Windows VHD, and those are the ones that are failing. So I think that's a clue that the fallback code to get the Windows .zip needs to be updated.

PR #2657, opened 2020-01-31
test: wait for "Ready" nodes, where appropriate

jackfrancis:
**Reason for Change**:

Use `Ready` as a required node status before determining to test it. This is more sane in general, and specifically, should help improve test flakes when running E2E immediately after running `aks-engine upgrade`, which may finish before all deleted vms have been unregistered as nodes in the cluster.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/cluster.sh
```diff
@@ -240,7 +240,8 @@ if [ "${UPGRADE_CLUSTER}" = "true" ]; then
       --auth-method client_secret \
       --client-id ${AZURE_CLIENT_ID} \
       --client-secret ${AZURE_CLIENT_SECRET} || exit 1
-
+    # wait a few mins for added/deleted vms to sync with the k8s api node registration state
+    sleep 5m
```
mboersma: Five minutes? That's like...forever.

jackfrancis: I know, and it didn't even work!

Alternate approach incoming...

PR #2659, opened 2020-01-31
refactor: cleanup duplicate functions and move apiversions to consts

CecileRobertMichon:
**Reason for Change**:

pre-refactor refactor: separating all non related changes in #2651 to make reviews easier and reverting possible if necessary.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/masteroutputs.t
+++ b/parts/masteroutputs.t
@@ -6,7 +6,7 @@
       "value": ""
 {{end}}
     }
-{{if AnyAgentUsesAvailabilitySets}}
+{{if HasVMASAgentPool}}
     ,
     "agentStorageAccountSuffix": {
       "type": "string",
--- a/pkg/api/mocks.go
+++ b/pkg/api/mocks.go
@@ -84,6 +84,8 @@ func CreateMockContainerService(containerServiceName, orchestratorVersion string
 		NetworkPolicy:           DefaultNetworkPolicy,
 		EtcdVersion:             DefaultEtcdVersion,
 		MobyVersion:             DefaultMobyVersion,
+		ContainerdVersion:       DefaultContainerdVersion,
+		LoadBalancerSku:         DefaultLoadBalancerSku,
 		KubeletConfig:           make(map[string]string),
 		ControllerManagerConfig: make(map[string]string),
 	}
@@ -231,4 +233,28 @@ func GetMockPropertiesWithCustomCloudProfile(name string, hasCustomCloudProfile,
 		p.CustomCloudProfile.AuthenticationMethod = ClientSecretAuthMethod
 	}
 	return p
+}
+
+func getMockAddon(name string) KubernetesAddon {
+	return KubernetesAddon{
+		Name: name,
+		Containers: []KubernetesContainerSpec{
+			{
+				Name:           name,
+				CPURequests:    "50m",
+				MemoryRequests: "150Mi",
+				CPULimits:      "50m",
+				MemoryLimits:   "150Mi",
+			},
+		},
+		Pools: []AddonNodePoolsConfig{
+			{
+				Name: "pool1",
+				Config: map[string]string{
+					"min-nodes": "3",
+					"max-nodes": "3",
+				},
+			},
+		},
+	}
 }
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -925,18 +925,16 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile, index int) string {
 	return vmPrefix
 }
 
-// AnyAgentUsesVirtualMachineScaleSets checks whether any of the agents in the AgentPool use VMSS or not
-func (p *Properties) AnyAgentUsesVirtualMachineScaleSets() bool {
-	for _, agentProfile := range p.AgentPoolProfiles {
-		if agentProfile.IsVirtualMachineScaleSets() {
-			return true
-		}
+// GetVMType returns the type of VM "vmss" or "standard" to be passed to the cloud provider
+func (p *Properties) GetVMType() string {
+	if p.HasVMSSAgentPool() {
+		return VMSSVMType
 	}
-	return false
+	return StandardVMType
 }
 
-// AnyAgentUsesAvailabilitySets checks whether any of the agents in the AgentPool use VMAS or not
-func (p *Properties) AnyAgentUsesAvailabilitySets() bool {
+// HasVMASAgentPool checks whether any of the agents in the AgentPool use VMAS or not
+func (p *Properties) HasVMASAgentPool() bool {
 	for _, agentProfile := range p.AgentPoolProfiles {
 		if agentProfile.IsAvailabilitySets() {
 			return true
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1048,15 +1048,18 @@ func (p *Properties) GetVirtualNetworkName() string {
 // GetSubnetName returns the subnet name of the cluster based on its current configuration.
 func (p *Properties) GetSubnetName() string {
 	var subnetName string
-	if p.IsHostedMasterProfile() {
-		if p.AreAgentProfilesCustomVNET() {
-			subnetName = strings.Split(p.AgentPoolProfiles[0].VnetSubnetID, "/")[DefaultSubnetNameResourceSegmentIndex]
+
+	if !p.IsHostedMasterProfile() {
+		if p.MasterProfile.IsCustomVNET() {
+			subnetName = strings.Split(p.MasterProfile.VnetSubnetID, "/")[DefaultSubnetNameResourceSegmentIndex]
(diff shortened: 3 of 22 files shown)


parts/masteroutputs.t
```diff
@@ -6,7 +6,7 @@
       "value": ""
 {{end}}
     }
-{{if AnyAgentUsesAvailabilitySets}}
+{{if HasVMASAgentPool}}
```
CecileRobertMichon: renamed `AnyAgentUsesAvailabilitySets` to `HasVMASAgentPool` to match the VMSS equivalent

pkg/api/mocks.go
```diff
@@ -232,3 +234,27 @@ func GetMockPropertiesWithCustomCloudProfile(name string, hasCustomCloudProfile,
 	}
 	return p
 }
+
+func getMockAddon(name string) KubernetesAddon {
```
CecileRobertMichon: moved this to mocks file (was in test file)

pkg/api/types.go
```diff
@@ -925,18 +925,16 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile, index int) string {
 	return vmPrefix
 }
 
-// AnyAgentUsesVirtualMachineScaleSets checks whether any of the agents in the AgentPool use VMSS or not
-func (p *Properties) AnyAgentUsesVirtualMachineScaleSets() bool {
```
CecileRobertMichon: removing this function has it is a duplicate of `HasVMSSAgentPool`

pkg/api/types.go
```diff
@@ -925,18 +925,16 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile, index int) string {
...
-	for _, agentProfile := range p.AgentPoolProfiles {
-		if agentProfile.IsVirtualMachineScaleSets() {
-			return true
-		}
+// GetVMType returns the type of VM "vmss" or "standard" to be passed to the cloud provider
+func (p *Properties) GetVMType() string {
```
CecileRobertMichon: new function to get VMType

pkg/api/types.go
```diff
@@ -925,18 +925,16 @@ func (p *Properties) GetAgentVMPrefix(a *AgentPoolProfile, index int) string {
...
+	if p.HasVMSSAgentPool() {
+		return VMSSVMType
 	}
-	return false
+	return StandardVMType
 }
 
-// AnyAgentUsesAvailabilitySets checks whether any of the agents in the AgentPool use VMAS or not
-func (p *Properties) AnyAgentUsesAvailabilitySets() bool {
+// HasVMASAgentPool checks whether any of the agents in the AgentPool use VMAS or not
```
CecileRobertMichon: renamed `AnyAgentUsesAvailabilitySets` to `HasVMASAgentPool` to match the VMSS equivalent

pkg/engine/armvariables.go
```diff
@@ -145,23 +144,22 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
 		"masterFqdnPrefix":              "[tolower(parameters('masterEndpointDNSNamePrefix'))]",
-		"apiVersionCompute":             "2019-07-01",
-		"apiVersionDeployments":         "2018-06-01",
-		"apiVersionStorage":             "2018-07-01",
-		"apiVersionKeyVault":            "2018-02-14",
-		"apiVersionNetwork":             "2018-08-01",
-		"apiVersionManagedIdentity":     "2015-08-31-preview",
-		"apiVersionAuthorizationUser":   "2018-09-01-preview",
-		"apiVersionAuthorizationSystem": "2018-01-01-preview",
+		"apiVersionCompute":             api.APIVersionCompute,
```
CecileRobertMichon: moved all apiVersions to constants

pkg/engine/armvariables.go
```diff
@@ -398,6 +391,7 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 	}
 
 	masterVars["nsgID"] = "[resourceId('Microsoft.Network/networkSecurityGroups',variables('nsgName'))]"
+	masterVars["subnetName"] = cs.Properties.GetSubnetName()
```
CecileRobertMichon: use `GetSubnetName()` function instead of embedding logic in armvariables file

pkg/api/types_test.go
```diff
@@ -6571,97 +6555,7 @@ func TestAnyAgentUsesAvailabilitySets(t *testing.T) {
...
-			expected: true,
-		},
-	}
-
-	for _, test := range tests {
-		test := test
-		t.Run(test.name, func(t *testing.T) {
-			t.Parallel()
-			ret := test.p.AnyAgentUsesVirtualMachineScaleSets()
+			ret := test.p.HasVMASAgentPool()
```
jackfrancis: Wrong substitution?

pkg/engine/armvariables.go
```diff
@@ -413,16 +407,16 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
 		masterVars["storageAccountBaseName"] = ""
 	}
 
-	if cs.Properties.AnyAgentUsesVirtualMachineScaleSets() {
+	masterVars["vmType"] = cs.Properties.GetVMType()
+
+	if cs.Properties.HasVMSSAgentPool() {
 		masterVars["primaryAvailabilitySetName"] = ""
-		masterVars["vmType"] = "vmss"
```
jackfrancis: yuck, thanks

pkg/api/types_test.go
CecileRobertMichon: no it's just that I removed the whole test for the function I deleted so the diff appears weird. Test for AnyAgentUsesVirtualMachineScaleSets is completely gone, test for HasVMASAgentPool starts on line 6476 and continues here.

jackfrancis: or maybe the diff is confusing here

pkg/engine/armvariables.go
jackfrancis: it's not clear to me that the existing GetSubnetName() returns the equivalent string responses as the above foo

CecileRobertMichon: yeah I think you're right, looking into it... I assumed the existing function did the same thing but looks like it didn't

jackfrancis: I think the existing implementation depends upon ARM params usage patterns, so maybe that means it's a candidate for a future refactor phase.

pkg/api/types.go
```diff
@@ -1050,15 +1048,18 @@ func (p *Properties) GetVirtualNetworkName() string {
 // GetSubnetName returns the subnet name of the cluster based on its current configuration.
 func (p *Properties) GetSubnetName() string {
```
CecileRobertMichon: @jackfrancis I fixed the function to look more like the armvariables logic so it's easier to compare and we were indeed missing the condition `if !p.IsHostedMasterProfile() && !p.MasterProfile.IsCustomVNET() && p.MasterProfile.IsVirtualMachineScaleSets()` then `subnetName = "subnetmaster"` 

Hopefully the tests at https://github.com/Azure/aks-engine/pull/2659/files#diff-de120f39e8d8448b9aa430e66c997a97R4820 help bring more confidence that it actually does the same thing...

That being the naming logic doesn't really make sense and I don't see why vmss masters need a different subnet name but I'm guessing that's just the natural evolution of the two templates being in separate files previously, and changing the behavior would mean risking breaking upgrade/scale so I'm going to stay away from that for now.

CecileRobertMichon: I recommend looking at the file directly https://github.com/Azure/aks-engine/blob/01e8d7720d6483b95467294e2561cdebef260dd0/pkg/api/types.go#L1049, the diff is pretty hard to read...

pkg/engine/engine.go
```diff
@@ -744,10 +744,7 @@ func getClusterAutoscalerAddonFuncMap(addon api.KubernetesAddon, cs *api.Contain
 			return api.GetClusterAutoscalerNodesConfig(addon, cs)
 		},
 		"GetVMType": func() string {
-			if cs.Properties.AnyAgentUsesVirtualMachineScaleSets() {
-				return base64.StdEncoding.EncodeToString([]byte("vmss"))
-			}
-			return base64.StdEncoding.EncodeToString([]byte("standard"))
```
CecileRobertMichon: @jackfrancis do you have any idea why we are base64 encoding the result here?

CecileRobertMichon: looks like it's being used as data in addon yaml specs, I renamed the function so it's clear

jackfrancis: Right. We should rename this to something like `GetVMTypeForClusterAutoscaler`

jackfrancis: `GetBase64EncodedVMType` also works :)

PR #2661, opened 2020-01-31
fix: don't expire adminUser or etcd accounts

### LINKED ISSUE #2277: Password expired when not using a password

mboersma:
**Reason for Change**:
Removes password expiration for the Linux `adminUser` account, since it uses only public key auth. Same for the non-interactive `etcd` account. Leaves the default rotation policy intact for accounts that may be created with passwords.

```shell
$ sudo cat /etc/shadow  # before
root:*LOCK*:14600::::::
daemon:*:18268:0:99999:7:::
bin:*:18268:0:99999:7:::
sys:*:18268:0:99999:7:::
...
statd:*:18277:0:99999:7:::
azureuser:!:18292:7:90:7:30::
etcd:!:18292:7:90:7:30::

$ sudo cat /etc/shadow  # after
root:*LOCK*:14600::::::
daemon:*:18268:0:99999:7:::
bin:*:18268:0:99999:7:::
sys:*:18268:0:99999:7:::
...
statd:*:18277:0:99999:7:::
azureuser:!:18292:0:99999:7:::
etcd:!:18292:0:99999:7:::
```

**Issue Fixed**:
Closes #2569 
Fixes #2277

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -23,9 +23,16 @@ systemctlEnableAndStart() {
     fi
 }
 
+configureAdminUser(){
+    chage -E -1 -I -1 -m 0 -M 99999 "${ADMINUSER}"
+    chage -l "${ADMINUSER}"
+}
+
 configureEtcdUser(){
-    useradd -U "etcd"
-    id "etcd"
+    useradd -U etcd
+    chage -E -1 -I -1 -m 0 -M 99999 etcd
+    chage -l etcd
+    id etcd
 }
 
 configureSecrets(){
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -52,6 +52,8 @@ else
     REBOOTREQUIRED=false
 fi
 
+time_metric "ConfigureAdminUser" configureAdminUser
+
 {{- if not NeedsContainerd}}
 time_metric "CleanupContainerd" cleanUpContainerd
 {{end}}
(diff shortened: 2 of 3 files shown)


jackfrancis:
Cool, so essentially we continue to "harden" during VHD CI the existing password expiration requirements, but do "last mile exceptions" (during node creation) for known accounts that don't have passwords.

mboersma:
Yes, basically this amounts to making just a "last-mile exception" for the Linux admin user, so the password expiration policy won't ever prevent SSH'ing in with public key auth.

I made the same change for the etcd user basically out of a desire for consistency with other system-defined accounts. But it's not meant to be a login account, so leaving the default policy intact and allowing its login eventually to be disabled is (apparently) harmless and (arguably) more consistent.

jackfrancis:
I like including etcd. The fact that password expiration disables interactive logins for accounts without a password suggests there could be some other side-effect for non-user accounts (like etcd that) that need to remain fully operational forever.

PR #2662, opened 2020-01-31
fix: update CSE moby version substring match

jackfrancis:
**Reason for Change**:

Follow-up from #2613. Now that we are delivering a `+azure` suffix w/ the internal Azure builds moby, we need to strip that from the version substring fetcher in CSE.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@cpuguy83 FYI, should have caught this in the PR review

PR #2683, opened 2020-02-05
feat: install bpftrace + bcc by default

alexeldeib:
**Reason for Change**:
This PR adds base layer of bpf debugging tools to aks-engine vhd + and cse install script. We'd like every cluster we bootstrap to have these tools available to do deep performance analysis and debugging, both ad-hoc and via automation. While we'd ultimately like to get bpftrace binaries on the system, we can immediately get benefit by using the lower-level bcc tools now. These also have much quicker install time, so it's easy to add them into the normal install path.

update: added bpftrace binary! for automation use cases and programmable interaction with the data, we will still want to keep the bcc libs installed. so this change can be purely additive.

**Issue Fixed**:
n/a

**Requirements**:
  
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -61,6 +61,13 @@ ERR_AZURE_STACK_GET_ARM_TOKEN=120 {{/* Error generating a token to use with Azur
 ERR_AZURE_STACK_GET_NETWORK_CONFIGURATION=121 {{/* Error fetching the network configuration for the node */}}
 ERR_AZURE_STACK_GET_SUBNET_PREFIX=122 {{/* Error fetching the subnet address prefix for a subnet ID */}}
 
+{{/* BCC/BPF-related error codes */}}
+ERR_IOVISOR_KEY_DOWNLOAD_TIMEOUT=126 {{/* Timeout waiting to download IOVisor repo key */}}
+ERR_IOVISOR_APT_KEY_TIMEOUT=127 {{/* Timeout waiting for IOVisor apt-key */}}
+ERR_BCC_INSTALL_TIMEOUT=128 {{/* Timeout waiting for bcc install */}}
+ERR_BPFTRACE_BIN_DOWNLOAD_FAIL=129 {{/* Failed to download bpftrace binary */}}
+ERR_BPFTRACE_TOOLS_DOWNLOAD_FAIL=130 {{/* Failed to download bpftrace default programs */}}
+
 OS=$(sort -r /etc/*-release | gawk 'match($0, /^(ID_LIKE=(coreos)|ID=(.*))$/, a) { print toupper(a[2] a[3]); exit }')
 UBUNTU_OS_NAME="UBUNTU"
 RHEL_OS_NAME="RHEL"
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -8,7 +8,9 @@ CNI_DOWNLOADS_DIR="/opt/cni/downloads"
 CONTAINERD_DOWNLOADS_DIR="/opt/containerd/downloads"
 K8S_DOWNLOADS_DIR="/opt/kubernetes/downloads"
 APMZ_DOWNLOADS_DIR="/opt/apmz/downloads"
+BPFTRACE_DOWNLOADS_DIR="/opt/bpftrace/downloads"
 UBUNTU_RELEASE=$(lsb_release -r -s)
+UBUNTU_CODENAME=$(lsb_release -c -s)
 
 removeEtcd() {
     if [[ $OS == $COREOS_OS_NAME ]]; then
@@ -169,6 +171,18 @@ installNetworkPlugin() {
     rm -rf $CNI_DOWNLOADS_DIR &
 }
 
+installBcc() {
+    echo "Installing BCC tools..."
+    IOVISOR_KEY_TMP=/tmp/iovisor-release.key
+    IOVISOR_URL=https://repo.iovisor.org/GPG-KEY
+    retrycmd_if_failure_no_stats 120 5 25 curl -fsSL $IOVISOR_URL > $IOVISOR_KEY_TMP || exit $ERR_IOVISOR_KEY_DOWNLOAD_TIMEOUT
+    wait_for_apt_locks
+    retrycmd_if_failure 30 5 30 apt-key add $IOVISOR_KEY_TMP || exit $ERR_IOVISOR_APT_KEY_TIMEOUT
+    echo "deb https://repo.iovisor.org/apt/${UBUNTU_CODENAME} ${UBUNTU_CODENAME} main" > /etc/apt/sources.list.d/iovisor.list
+    apt_get_update || exit $ERR_APT_UPDATE_TIMEOUT
+    apt_get_install 120 5 25 bcc-tools libbcc-examples linux-headers-$(uname -r) || exit $ERR_BCC_INSTALL_TIMEOUT
+}
+
 downloadCNI() {
     mkdir -p $CNI_DOWNLOADS_DIR
     CNI_TGZ_TMP=$(echo ${CNI_PLUGINS_URL} | cut -d "/" -f 5)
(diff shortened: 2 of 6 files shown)


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -52,6 +52,9 @@ ERR_CIS_ASSIGN_ROOT_PW=111 {{/* Error assigning root password in CIS enforcement
 ERR_CIS_ASSIGN_FILE_PERMISSION=112 {{/* Error assigning permission to a file in CIS enforcement */}}
 ERR_PACKER_COPY_FILE=113 {{/* Error writing a file to disk during VHD CI */}}
 ERR_CIS_APPLY_PASSWORD_CONFIG=115 {{/* Error applying CIS-recommended passwd configuration */}}
+ERR_IOVISOR_KEY_DOWNLOAD_TIMEOUT=116 {{/* Timeout waiting to download IOVisor repo key */}}
```
CecileRobertMichon: 110-120 codes are all for CIS related errors, can you move these new ones to bigger numbers? We try to keep exit codes ordered by theme

alexeldeib: Is that better? I was wondering why there is a gap 😄

CecileRobertMichon: Yes, an thank you for adding that comment :D we should really add comments for all of them to clarify at some point...

you need to run make generate to make CI happy

alexeldeib:
dat ci bump 😅

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -169,6 +170,18 @@ installNetworkPlugin() {
...
 }
 
+installBcc() {
+    echo "Installing BCC tools..."
+    IOVISOR_KEY_TMP=/tmp/iovisor-release.key
+    IOVISOR_URL=https://repo.iovisor.org/GPG-KEY
+    retrycmd_if_failure_no_stats 120 5 25 curl -fsSL $IOVISOR_URL > $IOVISOR_KEY_TMP || exit $ERR_IOVISOR_KEY_DOWNLOAD_TIMEOUT
+    wait_for_apt_locks
+    retrycmd_if_failure 30 5 30 apt-key add $IOVISOR_KEY_TMP || exit $ERR_IOVISOR_APT_KEY_TIMEOUT
+    echo "deb https://repo.iovisor.org/apt/${UBUNTU_CODENAME} ${UBUNTU_CODENAME} main" > /etc/apt/sources.list.d/iovisor.list
```
juan-lee: do we need to do anything special to make this idempotent? If the bcc tools are already installed on the vhd, we'd want to skip these steps, right?

alexeldeib: apt should just no-op the steps which is desired, I could do a better job cleaning up the files though. I'm not sure it's worth the extra logic to do something like `find /usr/lib -name libbcc.so.0` to detect the installed lib over letting apt handle this itself. L180 I used '>' over '>>' to force the file contents to be fresh/correct.

alexeldeib: let me know what you prefer?

CecileRobertMichon: you could just move the call to `installBcc` to inside the `installDeps` function which only gets called for non-VHD deployments

alexeldeib: I thought that's what `"$FULL_INSTALL_REQUIRED" = "true"` was for, is that not what this curently does? The only reason I didn't put it inside `installDeps` was to keep the additional apt repo separate like the other drivers. I can just nudge it in there though

CecileRobertMichon: oh yeah where it's at right now is perfectly fine, didn't see you already put it inside the full install clause 👍

alexeldeib:
@CecileRobertMichon @juan-lee added bpftrace if you'd like to re-review. The only reason I marked wip is for the prod artifacts

alexeldeib:
@mboersma I hear you may be good to review this as well re: blob downloads -- seemed fairly straightforward, but appreciate your 👀

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -209,6 +223,39 @@ ensureAPMZ() {
...
 
+# install mostly-static bpftrace binary linked only against glibc > 2.23
+# it should be comptaible with any linux distro using a new enough kernel.
+# this may in the future require filtering the host OS if we encounter compatibility issues.
+# we also install the most commonly used bpftrace scripts by default.
+installBpftrace() {
+    local version=$1
+    local bpftrace_bin="bpftrace"
+    local bpftrace_tools="bpftrace-tools.tar"
+    local bpftrace_url="https://upstreamartifacts.blob.core.windows.net/$bpftrace_bin/$version"
```
mboersma: Maybe this should be the CDN URL instead? `https://upstreamartifacts.azureedge.net/$bpftrace_bin/$version`

alexeldeib: fixed and tested with this change

alexeldeib:
removing wip as i've validated with the prod artifacts

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -209,6 +223,39 @@ ensureAPMZ() {
     ln -Ffs "$bin_path" "$apmz_filepath" # symlink apmz into /usr/local/bin/apmz
 }
 
+# install mostly-static bpftrace binary linked only against glibc > 2.23
+# it should be comptaible with any linux distro using a new enough kernel.
+# this may in the future require filtering the host OS if we encounter compatibility issues.
+# we also install the most commonly used bpftrace scripts by default.
+installBpftrace() {
+    local version=$1
```
CecileRobertMichon: if the version isn't user configurable, why make it an input? Right now we have to update it in two places for version updates (the vhd script and cse main). Unless there are plans to make it configurable, I think it makes more sense to keep the version inside the install function.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -209,6 +223,39 @@ ensureAPMZ() {
     ln -Ffs "$bin_path" "$apmz_filepath" # symlink apmz into /usr/local/bin/apmz
 }
 
+# install mostly-static bpftrace binary linked only against glibc > 2.23
```
CecileRobertMichon: I believe we're tight on cloud-init space and we generally try to avoid comments that aren't critical in this file, @jackfrancis to confirm

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -85,6 +86,10 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
     time_metric "EnsureAuditD" ensureAuditD
 fi
 
+if [[ "$FULL_INSTALL_REQUIRED" = "true" ]]; then
+    time_metric "InstallBpftrace" installBpftrace "v0.9.4"
```
CecileRobertMichon: does this work with non-ubuntu distros, eg. coreos?

alexeldeib: yes

parts/k8s/cloud-init/artifacts/cse_install.sh
alexeldeib: ironically the answer to your other comment is this line 😅 but understandable, will remove

parts/k8s/cloud-init/artifacts/cse_main.sh
alexeldeib: it will work on any linux kernel with glibc > 2.23 compile with BPF enabled (i.e., very broad)

parts/k8s/cloud-init/artifacts/cse_install.sh
alexeldeib: i'll add back if jack negatively confirms

parts/k8s/cloud-init/artifacts/cse_install.sh
alexeldeib: good point, fixed

alexeldeib:
sorry, just cleaning up git history...

alexeldeib:
@jackfrancis @mboersma any additional thoughts here?

jackfrancis:
@mbearup is there anything in this changeset that will make your effort more difficult? Mainly FYI this will definitely collide with your PR.

dalehamel:
Actually it looks like the failure for kprobe is specific to CI, it is also disabled on travis I see. 

I think that the runtime tests may not have been running at all before, and your addition to add the disable tests actually activated them lol.

I found a couple of other nitpicks that I'll send to you later as a patch of suggestions @alexeldeib thanks for being patient. Once those are in i think we can get this merged.

mboersma:
I'm going to merge this since it's been approved for several days. @mbearup my apologies if this creates a rebase hell for you. ♻️

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -61,6 +61,13 @@ ERR_AZURE_STACK_GET_ARM_TOKEN=120 {{/* Error generating a token to use with Azur
 ERR_AZURE_STACK_GET_NETWORK_CONFIGURATION=121 {{/* Error fetching the network configuration for the node */}}
 ERR_AZURE_STACK_GET_SUBNET_PREFIX=122 {{/* Error fetching the subnet address prefix for a subnet ID */}}
 
+{{/* BCC/BPF-related error codes */}}
+ERR_IOVISOR_KEY_DOWNLOAD_TIMEOUT=126 {{/* Timeout waiting to download IOVisor repo key */}}
+ERR_IOVISOR_APT_KEY_TIMEOUT=127 {{/* Timeout waiting for IOVisor apt-key */}}
```
CecileRobertMichon: Just realized these are actually reserved exit codes so they shouldn't be used

> "According to the above table, exit codes 1 - 2, 126 - 165, and 255 [1] have special meanings, and should therefore be avoided for user-specified exit parameters. Ending a script with exit 127 would certainly cause confusion when troubleshooting (is the error code a "command not found" or a user-defined one?). However, many scripts use an exit 1 as a general bailout-upon-error. Since exit code 1 signifies so many possible errors, it is not particularly useful in debugging."

from http://www.tldp.org/LDP/abs/html/exitcodes.html

PR #2687, opened 2020-02-06
Add template/options for using Shared Image Gallery

mbearup:
Add config options for arbitrary Debian-based images

**Reason for Change**:
Adds a template and fixes for using images from shared image gallery.

**Issue Fixed**:
None

**Requirements**:
N/A

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -73,6 +73,32 @@ NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
 NVIDIA_CONTAINER_RUNTIME_VERSION=2.0.0
 
+configure_prerequisites() {
+    if which apt > /dev/null; then
+        # Install required packages; only for apt-based systems
+        apt update -qq
+        apt install -y curl gpg jq nfs-common
```
jackfrancis: Let's move `curl`, `gpg`, and `nfs-common` into the `for apt_package` enumeration in the `installDeps` func (in `cse_install.sh`). `jq` is already there, so we should already get that on distros that don't pre-install it.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -73,6 +73,32 @@ NVIDIA_DOCKER_VERSION=2.0.3
...
+    fi
+
+    modprobe br_netfilter overlay
```
jackfrancis: Can this be integrated into the existing `configureCNI` func? (in `cse_config.sh`)

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -73,6 +73,32 @@ NVIDIA_DOCKER_VERSION=2.0.3
...
+    check_and_fix_kernel_parameter /proc/sys/net/ipv4/ip_forward net.ipv4.ip_forward 1
+}
+
+check_and_fix_kernel_parameter() {
```
jackfrancis: I did a quick audit of the code to see if we can use this new `check_and_fix_kernel_parameter` func to generalize existing foo, didn't find any. If you could double-check that, 👍. (If you find any opportunities, feel free to refactor to use this new func.)

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -41,7 +41,11 @@ installEtcd() {
             docker run --rm --entrypoint cat ${CONTAINER_IMAGE} /usr/local/bin/etcd > "$path/etcd"
             docker run --rm --entrypoint cat ${CONTAINER_IMAGE} /usr/local/bin/etcdctl > "$path/etcdctl"
         else
-            img unpack -o "$path" ${CONTAINER_IMAGE}
+            # img unpack requires a non-existent dirctory
+            tmpdir=/root/etcd${RANDOM}
```
jackfrancis: prefer `mktemp` ?

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -41,7 +41,11 @@ installEtcd() {
...
+            img unpack -o ${tmpdir} ${CONTAINER_IMAGE}
+            mv ${tmpdir}/usr/local/bin/etcd ${tmpdir}/usr/local/bin/etcdctl /usr/bin
```
jackfrancis: let's re-use the existing `$path` var instead of referring to `/usr/bin` explicitly, even though it's somewhat academic (i.e., this stuff hasn't been validated on coreos at all)

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -89,23 +90,25 @@ fi
...
+{{- if NeedsContainerd}}
+time_metric "InstallContainerd" installContainerd
+{{end}}
+
 if [[ -n "${MASTER_NODE}" ]] && [[ -z "${COSMOS_URI}" ]]; then
     {{- if IsDockerContainerRuntime}}
     CLI_TOOL="docker"
     {{else}}
     CLI_TOOL="img"
+    # This codepath requires img, which is not installed until
```
jackfrancis: You can remove this comment (though it was useful for my code review!)

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -426,6 +426,11 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
   content: |
     #!/bin/bash
     set -x
+    if [[ ! -s /etc/environment ]]; then
+        # /etc/environment is empty, which will break subsequent sed commands"
```
jackfrancis: Let's use go templating comments to save cloud-init overhead :)

E.g.:

```
{{- /* /etc/environment is empty, which will break subsequent sed commands"
       Append a blank line... */}}
```

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: Done!

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: Looks like br_netfilter is already there. It's possible overlay isn't even needed. Will run a test...

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: Looks like the codebase already uses sysctl-d-60-CIS.conf for all the sysctl needs.I could move the setting there, but don't recall if it will get priority over

parts/k8s/cloud-init/artifacts/cse_install.sh
mbearup: So that's what I tried, but note the comment above - `img unpack` requires a *non-existent* directory. Once `mktemp` creates the directory, `img` will refuse to unpack to that path. Believe me I tried :) As I said earlier, I honestly don't know how this ever worked, and I couldn't find an `img` option to unpack to an existing directory.

parts/k8s/cloud-init/artifacts/cse_install.sh
mbearup: Done!

parts/k8s/cloud-init/masternodecustomdata.yml
mbearup: Done!

parts/k8s/cloud-init/artifacts/cse_main.sh
mbearup: Done

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: Correction, a few lines before we run `for apt_package` we run `curl`...
```
installDeps() {
    retrycmd_if_failure_no_stats 120 5 25 curl -fsSL https://packages.microsoft.com
...
    for apt_package in apache2-utils ... curl ...
```
So this will require some improvisation.

mbearup: I went ahead and added curl to the image so simplify things. However I uncovered a different issue. installDeps isn't actually getting called...

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: I've removed these - br_netfilter is already there and I see no difference with overlay missing.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: So this is tricky... /etc/sysctl.d/99-sysctl.conf is symlinked to /etc/sysctl.conf where the ip_forward setting resides. So the settings in 60-CIS do not supersede it.
With this setting there by default (and thus present at boot time in /proc) there's no alternative but to remove it from /etc/sysctl.conf and update /proc. I welcome recommendations.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: OK, resolved the issue with installDeps - it's around detecting Debian derivatives as Debian while retaining Ubuntu-specific funtionality.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
mbearup: The current fix works, so unless there are stylistic comments I'll consider this resolved.

PR #2690, opened 2020-02-06
add the missing images in vhd for AKS.

andyliuliming:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


andyliuliming:
@xuto2  nice.

packer/install-dependencies.sh
```diff
@@ -448,6 +423,28 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
   fi
 done
 
+# pull patched hyperkube image for AKS
+# this is used by kube-proxy
+PATCHED_HYPERKUBE_IMAGES="
+v1.12.8_v0.0.5
+v1.13.10_v0.0.5
+v1.13.11_v0.0.5
```
jackfrancis: We should probably delete the duplicate versions in the `K8S_VERSIONS` array above, right? Otherwise the VHD won't have enough storage space (limited to 30 GB of local storage). Each of these additional hyperkube images is roughly 1 GB in size.

packer/install-dependencies.sh
```diff
@@ -104,6 +99,20 @@ for DASHBOARD_VERSION in ${DASHBOARD_VERSIONS}; do
...
 done
 
+NEW_DASHBOARD_VERSIONS="2.0.0-beta8"
+for DASHBOARD_VERSION in ${NEW_DASHBOARD_VERSIONS}; do
+    CONTAINER_IMAGE="mcr.microsoft.com/oss/kubernetes/kubernetesui/dashboard:v${DASHBOARD_VERSION}"
+    pullContainerImage "docker" ${CONTAINER_IMAGE}
+    echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
+done
+
+NEW_DASHBOARD_METRICS_SGRAPER_VERSIONS="1.0.2"
```
xuto2: s/SGRAPER/SCRAPER/

packer/install-dependencies.sh
```diff
@@ -297,8 +247,7 @@ done
 
 AZURE_CNIIMAGEBASE="mcr.microsoft.com/containernetworking"
 AZURE_CNI_NETWORKMONITOR_VERSIONS="
-0.0.6
```
xuto2: 0.0.6 was there 23days ago, it means there are possible clusters with supported k8s version and still requires it for the scale up/upgrade work until we upgrade all ccps. Let's keep it for a while and then remove it.

packer/install-dependencies.sh
```diff
@@ -336,20 +284,50 @@ for NVIDIA_DEVICE_PLUGIN_VERSION in ${NVIDIA_DEVICE_PLUGIN_VERSIONS}; do
     echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
-TUNNELFRONT_VERSIONS="v1.9.2-v4.0.4"
+TUNNELFRONT_VERSIONS="v1.9.2-v4.0.11"
```
xuto2: We need "v1.9.2-v3.0.11 v1.9.2-v4.0.11" based on the recent change from Guoxun

https://github.com/Azure/aks-engine/pull/2546/files

packer/install-dependencies.sh
```diff
@@ -448,6 +423,28 @@ for KUBERNETES_VERSION in ${K8S_VERSIONS}; do
...
+# pull patched hyperkube image for AKS
+# this is used by kube-proxy
+PATCHED_HYPERKUBE_IMAGES="
+v1.12.8_v0.0.5
+v1.13.10_v0.0.5
+v1.13.11_v0.0.5
+v1.13.12_f0.0.2
+v1.14.6_v0.0.5
+v1.14.7_v0.0.5
+v1.14.8_f0.0.4
```
xuto2: how about 1.14.8_f0.02?

packer/install-dependencies.sh
```diff
@@ -394,25 +381,13 @@ for KMS_PLUGIN_VERSION in ${KMS_PLUGIN_VERSIONS}; do
...
-    pullContainerImage "docker" ${CONTAINER_IMAGE}
-    echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
-done
-
 pullContainerImage "docker" "busybox"
 echo "  - busybox" >> ${VHD_LOGS_FILEPATH}
 
 K8S_VERSIONS="
 1.17.0
 1.16.4
```
xuto2: We've moved to all to MCR, I don't think we need the hyperkube images from gcr now.

xuto2: @gossion, can you confirm?

gossion: that is right. We have moved all to MCR. 
So this VHD is only for AKS, right? if that is true, we should remove all gcr, dockerhub, quay images - per our egress policy, no traffic should go to gcr by default.

xuto2: in fact we need to pull these from mcr as we're now seeing error due to below image missing
mcr.microsoft.com/oss/kubernetes/hyperkube:v1.16.4
mcr.microsoft.com/oss/kubernetes/hyperkube:v1.15.7

packer/install-dependencies.sh
```diff
@@ -394,25 +381,13 @@ for KMS_PLUGIN_VERSION in ${KMS_PLUGIN_VERSIONS}; do
...
-1.16.2
 1.16.1
-1.16.1-azs
```
xuto2: The below two can also be removed
1.15.4-azs
1.14.7-azs

packer/install-dependencies.sh
```diff
@@ -266,11 +218,9 @@ done
 
 CORE_DNS_VERSIONS="
 1.6.5
```
xuto2: I see error in pull 1.6.6, so maybe we can baker that as well, since it's just 40MB.
mcr.microsoft.com/oss/kubernetes/coredns:1.6.6

xuto2: and 
mcr.microsoft.com/oss/kubernetes/autoscaler/cluster-proportional-autoscaler:1.3.0_v0.0.5

packer/install-dependencies.sh
xuto2: @gossion yes, this is AKS "private" VHD. One edge case is the scale up for VMAS, which would still refer to old setting(from the pod, say kubeproxy), and when it uses the new VHD, it will need to pull images from GCR. But for VMAS scaling operation, we only need to support the 1.13+ vesrions.

packer/install-dependencies.sh
```diff
@@ -50,12 +50,7 @@ if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
   overrideNetworkConfig
 fi
 
-ETCD_VERSION="3.3.15"
-ETCD_DOWNLOAD_URL="https://acs-mirror.azureedge.net/github-coreos"
-installEtcd
-echo "  - etcd v${ETCD_VERSION}" >> ${VHD_LOGS_FILEPATH}
-
-MOBY_VERSION="3.0.8"
+MOBY_VERSION="3.0.10"
```
xuto2: This requires changes in installMoby (https://github.com/Azure/aks-engine/pull/2613/files) and it was just adapted by AKS-E master 10 days ago(https://github.com/Azure/aks-engine/pull/2660). I prefer we wait until the next VHD, unless there is fixes required in 3.0.10.

packer/install-dependencies.sh
xuto2: docker pull mcr.microsoft.com/oss/kubernetes/kubernetesui/dashboard:v2.0.0-beta
Error response from daemon: manifest for mcr.microsoft.com/oss/kubernetes/kubernetesui/dashboard:v2.0.0-beta not found: manifest unknown: manifest unknown
tony@tonyxu-HP-Z240:~/MyWork/go/src/github.com/Azure/agentbaker$ docker pull mcr.microsoft.com/oss/kubernetes/kubernetesui/metrics-scraper:v1.0.2
Error response from daemon: manifest for mcr.microsoft.com/oss/kubernetes/kubernetesui/metrics-scraper:v1.0.2 not found: manifest unknown: manifest unknown

packer/install-dependencies.sh
jackfrancis: This spec will not work with duplicate images, we don't have enough storage. Can't build a VHD from this spec. Every duplicated image in `K8S_VERSIONS` above that has an equivalent here needs to be removed to free up space.

packer/install-dependencies.sh
xuto2: I will fix this in anotherPR

PR #2694, opened 2020-02-07
docs: proposal for Azure locations + SKUs automation

mboersma:
**Reason for Change**:
Proposes a solution for some AKS Engine data structures needing to be updated.

**Issue Fixed**:
See #1894

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


docs/design/automate-skus.md
```diff
@@ -0,0 +1,124 @@
...
+A python script will regenerate some, but not all of the data structures.
+
+## Goals
+
+- On a regular schedule, notify the AKS Engine team if new locations or SKUs exist
+- Automate the creation of the pull request with the new locations or SKUs
+
+## Non-goals
+
+## Proposal: scriptable aks-engine commands
```
jackfrancis: I love that this proposal also standardizes the way we deliver this information to users.

docs/design/automate-skus.md
```diff
@@ -0,0 +1,124 @@
...
+#### Pros
+
+- Location and SKU support data is always current
+- Live Azure as the one source of truth
+
+#### Cons
+
+- Requires significant refactoring and implementation effort
+- Making API calls during `aks-engine generate` may be unwanted behavior
+- Somewhat slower and less reliable than using the data in code
```
jackfrancis: I agree that the most concrete downside is absorbing runtime API flakes for all aks-engine ARM operations. Removing that contingency is worth the additional "staleness" of the consts that we statically maintain in code, especially if we reduce the stale time by automating much of the process of updating those static consts.

docs/design/automate-skus.md
```diff
@@ -0,0 +1,124 @@
...
+		"centraluseuap",
+		"chinaeast",
+...
+
+```
+
+### Example SKUs output
+
+```shell
+$ aks-engine get-skus -o human
```
marosset: Nice

docs/design/automate-skus.md
```diff
@@ -0,0 +1,124 @@
...
+## Goals
+
+- On a regular schedule, notify the AKS Engine team if new locations or SKUs exist
+- Automate the creation of the pull request with the new locations or SKUs
+
+## Non-goals
+
+## Proposal: scriptable aks-engine commands
+
+New developer-only commands in the `aks-engine` binary generate the necessary Go source code. The commands replace the existing python script and extend it to cover the accelerated networking capability list.
```
marosset: Would it make sense to add another make target instead of and aks-engine commands?
If we do that it might be easier to periodically run the job in a devops pipeline or jenkins job to get a signal of when we need to perform updates.

marosset: That signal could be that there were diffs between the checked in generated files and newly generated files.

mboersma: Yes, the presence of a diff in the code should be considered the signal that we need a PR. I forgot to put this explicitly in the doc, but :+1:.

PR #2716, opened 2020-02-12
ci: use "k create deployment" instead of "k run" if 1.16 or greater

haofan-ms:
**Reason for Change**:

"kubectl run" command is deprecated, and that case issue when running the "port-forward" e2e test using kubernetes 1.16+. This PR use the "kubectl create deployment" to create deployment and test result shows the previous issue is gone.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/deployment/deployment.go
+++ b/test/e2e/kubernetes/deployment/deployment.go
@@ -117,7 +117,9 @@ func CreateLinuxDeploy(image, name, namespace, miscOpts string) (*Deployment, er
 	var commandTimeout time.Duration
 	var cmd *exec.Cmd
 	overrides := `{ "spec":{"template":{"spec": {"nodeSelector":{"beta.kubernetes.io/os":"linux"}}}}}`
-	if miscOpts != "" {
+	if miscOpts == "1.16+" {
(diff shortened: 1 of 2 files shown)


test/e2e/kubernetes/deployment/deployment.go
```diff
@@ -117,7 +117,9 @@ func CreateLinuxDeploy(image, name, namespace, miscOpts string) (*Deployment, er
 	var commandTimeout time.Duration
 	var cmd *exec.Cmd
 	overrides := `{ "spec":{"template":{"spec": {"nodeSelector":{"beta.kubernetes.io/os":"linux"}}}}}`
-	if miscOpts != "" {
+	if miscOpts == "1.16+" {
+		cmd = exec.Command("k", "create", "deployment", name, "-n", namespace, "--image", image)
```
mboersma: One problem with just switching to `kubectl create deployment` is that there is no `--overrides` flag to customize the deployment created. So don't have a way to specify a `nodeSelector` that ensures a workload lands on a Linux node. I think we may have to explicitly create a deployment manifest to get that behavior.

There are two other places in `deployment.go` that do `kubectl run` and then wait for a deployment to show up, so those would need to be fixed as well.

PR #2739, opened 2020-02-18
fix: init Azure clients consistently

mboersma:
**Reason for Change**:
While adding new Azure client support to our `AzureClient`, I noticed that not all client objects are being initialized the same way. This appears to be an oversight, so I fixed it.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/azureclient.go
+++ b/pkg/armhelpers/azureclient.go
@@ -393,6 +393,7 @@ func getClient(env azure.Environment, subscriptionID, tenantID string, armAuthor
 	c.disksClient.PollingDuration = DefaultARMOperationTimeout
 	c.groupsClient.PollingDuration = DefaultARMOperationTimeout
 	c.interfacesClient.PollingDuration = DefaultARMOperationTimeout
+	c.msiClient.PollingDuration = DefaultARMOperationTimeout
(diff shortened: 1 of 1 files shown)


jackfrancis:
What exactly change? (diff is a bit obtuse here)

mboersma:
Sorry, the diff is hard to parse. I alpha-sorted each section so I could do a visual comparison and see which were missing. (I would have made this a better refactor but there aren't any existing interfaces to use here and it looks awkward to create one.)

In summary:
- `c.availabilitySetsClient and c.virtualMachineExtensionsClient` weren't getting `authorizer` set as other clients did
- `c.msiClient, c.ServicePrincipalsClient, c.VirtualMachineExtensionsClient, c.VirtualMachineImagesClient, and c.workspacesClient` weren't getting `PollingDuration` set as other clients did
- Several clients weren't getting `RequestInspector` set in `AddAcceptLanguages`
- Several clients weren't getting `RequestInspector` set in `AddAuxiliaryTokens`

With the exception of those clients using `graphAuthorizer`, it looks like all these init patterns should be identical for the set of clients, and this PR makes that so.

PR #2752, opened 2020-02-20
feat: use kubernetesImageBase as the components prefix on Azure Stack

jadarsie:
**Reason for Change**:
Recent changes to how AKSe handles component images will prevent Azure Stack users from using their old cluster definitions moving forward. Being able to reuse the cluster definition would be ideal.

If that's the expected behavior, this PR is meant to fix that.
If users are supposed to change their cluster definitions, then I will update documentation.

**I will push tests later today once I know code changes are needed**
**I know this is currently breaking UTs**

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/api/components.go
+++ b/pkg/api/components.go
@@ -220,7 +220,9 @@ func getComponentDefaultContainerImage(component string, cs *ContainerService) s
 	hyperkubeImageBase := kubernetesImageBase
 	hyperkubeImage := hyperkubeImageBase + k8sComponents[common.Hyperkube]
 	if cs.Properties.IsAzureStackCloud() {
-		hyperkubeImage = hyperkubeImage + common.AzureStackSuffix
+		hyperkubeImage = cs.Properties.OrchestratorProfile.KubernetesConfig.KubernetesImageBase +
+			k8sComponents[common.Hyperkube] +
+			common.AzureStackSuffix
 	}
 	if kubernetesConfig.CustomHyperkubeImage != "" {
 		hyperkubeImage = kubernetesConfig.CustomHyperkubeImage


pkg/api/components.go
```diff
@@ -220,7 +220,9 @@ func getComponentDefaultContainerImage(component string, cs *ContainerService) s
 	hyperkubeImageBase := kubernetesImageBase
 	hyperkubeImage := hyperkubeImageBase + k8sComponents[common.Hyperkube]
 	if cs.Properties.IsAzureStackCloud() {
-		hyperkubeImage = hyperkubeImage + common.AzureStackSuffix
+		hyperkubeImage = cs.Properties.OrchestratorProfile.KubernetesConfig.KubernetesImageBase +
```
jackfrancis: Why do we want to use `cs.Properties.OrchestratorProfile.KubernetesConfig.KubernetesImageBase` instead of `cs.GetCloudSpecConfig().KubernetesSpecConfig.KubernetesImageBase` for Azure Stack?

jadarsie: I guess that's part of the question. 

Right now, `cs.GetCloudSpecConfig().KubernetesSpecConfig.KubernetesImageBase` points to gcr but azure stack components should use the custom hyperkube image hosted in MCR.

Would it be better to update `cs.GetCloudSpecConfig().KubernetesSpecConfig.KubernetesImageBase`? I could not easy verify if changing that will break something else.

jackfrancis: Does `cs.GetCloudSpecConfig().KubernetesSpecConfig.KubernetesImageBase` not point to the `CustomCloudProfile` equivalent? That's what we want, right?

jadarsie: It does not, and it is probably not what we need if I understand correctly.
`KubernetesImageBase` is also used as the prefix for addons.

jackfrancis: I think this is bugz. I think we want `GetCloudSpecConfig()` to return the `AzureEnvironmentSpecConfig` that is inside config object `CustomCloudProfile`. I assume Azure Stack delivers a valid `KubernetesImageBase` value therein, correct?

jadarsie:
Closing, taking a different direction.

PR #2755, opened 2020-02-20
chore: rationalize AKS Engine VHD config

jackfrancis:
**Reason for Change**:

This PR brings the VHD spec up-to-date.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/packer/install-dependencies.sh
+++ b/vhd/packer/install-dependencies.sh
@@ -51,7 +51,7 @@ if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
   overrideNetworkConfig
 fi
 
-apmz_version="v0.4.0"
+apmz_version="v0.5.0"
 ensureAPMZ "${apmz_version}"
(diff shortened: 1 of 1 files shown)


vhd/packer/install-dependencies.sh
```diff
@@ -422,6 +305,7 @@ pullContainerImage "docker" "busybox"
 echo "  - busybox" >> ${VHD_LOGS_FILEPATH}
 
 K8S_VERSIONS="
+1.18.0-alpha.5
```
mboersma: We haven't previously put pre-releases in the VHD. Is this for optimizing e2e tests?

jackfrancis: Yeah, I thought it would make sense to start doing this to enable us to level up our testing standards for pre-release versions.

marosset: Should we do the same for windows?

jackfrancis: I vote yes!

PR #2759, opened 2020-02-20
feat: Enabling SSH on windows nodes by default

marosset:
**Reason for Change**:

SSH has been supported on Windows since 1809/WS2019.
Since Windows 1803 is no longer being supported by MS this PR installs/configures SSH for windows agent pools unless WindowsProfile.SSHEnables is explicitly set to false.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


marosset:
@jackfrancis @jadarsie @ksubrmnn  PTAL
I reactivated this old PR which enables SSH by default for Windows agent pools

pkg/api/defaults.go
```diff
@@ -749,6 +749,11 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 func (p *Properties) setWindowsProfileDefaults(isUpgrade, isScale bool) {
 	windowsProfile := p.WindowsProfile
 	if !isUpgrade && !isScale {
+		if windowsProfile.SSHEnabled == nil {
+			sshEnabled := DefaultWindowsSshEnabled
+			windowsProfile.SSHEnabled = &sshEnabled
```
jackfrancis, line 754: you can skip the extra temporary var and use `windowsProfile.SSHEnabled = to.BoolPtr(DefaultWindowsSshEnabled )`

pkg/api/const.go
```diff
@@ -485,6 +485,8 @@ const (
 	DefaultNonMasqueradeCIDR = "0.0.0.0/0"
 	// DefaultKubeProxyMode is the default KubeProxyMode value
 	DefaultKubeProxyMode KubeProxyMode = KubeProxyModeIPTables
+	// DefaultWindowsSshEnabled is the default windowsProfile.sshEnabled value
+	DefaultWindowsSshEnabled = true
```
jackfrancis, line 489: nit: prefer `DefaultWindowsSSHEnabled` (SSH instead of Ssh)

marosset: Yea... this failed linting.
Cleaning this up now.

pkg/api/types.go
```diff
@@ -1751,6 +1751,14 @@ func (w *WindowsProfile) GetWindowsSku() string {
 	return KubernetesDefaultWindowsSku
 }
 
+// GetSSHEnabled gets it ssh should be enabled for Windows nodes
+func (w *WindowsProfile) GetSSHEnabled() bool {
+	if w.SSHEnabled != nil {
+		return *w.SSHEnabled
+	}
+	return DefaultWindowsSSHEnabled
```
jackfrancis, line 1759: This is probably not what we want. Are we assuming that this method is only invoked in execution flows after default enforcement flows have been traversed? (And if so, I still don’t think we want this method on this pointer object in this way.)

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -298,7 +298,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 					dockerVersionCmd := fmt.Sprintf("\"docker version\"")
 					for _, n := range nodes {
 						if n.IsWindows() {
-							if eng.ExpandedDefinition.Properties.WindowsProfile != nil && !eng.ExpandedDefinition.Properties.WindowsProfile.SSHEnabled {
+							if eng.ExpandedDefinition.Properties.WindowsProfile != nil && !eng.ExpandedDefinition.Properties.WindowsProfile.GetSSHEnabled() {
```
jackfrancis, line 301: for example, the method invocation here will operate against a ContainerService object post-template generation (i.e., we can rely upon the actual data in the object and not have to worry about highjacking the underlying value w/ the default const value)

pkg/api/types.go
marosset: This is following the same behavior as GetEnabledWindowsUpdate directly below.

This was done this way in order to not perform updates to the stored apimodel.json in upgrade/scale scenarios but still be able to enforce a default value.

What do you have in mind for this logic?

jackfrancis: So these methods are just for the E2E test runner?

test/e2e/kubernetes/kubernetes_test.go
marosset: What about the case where we are working with an template generated with an older version of aks-engine which may not have a value for sshEnabled stored? 
If we call GetSSHEnabled() we are guaranteed to never have a nil value.

pkg/api/types.go
marosset: And to protection when operating on older templates.

marosset: (I think - I initially wrote this ~2 months ago)

jackfrancis: Cool. So, I think the best way to do this to protect from future bugs (that method may be re-used in some critical cluster creation flow and then have unexpected results) is to do the default substitution in the test code itself, or maybe a a helper function in the e2e surface area that replaces nil values w/ defaults.

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Good point. I think we have the runner call a func/method in its own area that wraps nil values underneath appropriate defaults (true/false)

pkg/api/types.go
marosset: Let me take a look at this again tomorrow and see if this can be cleaned up some.

marosset: so I took a look and if there is an apimodel.json that does not have sshEnabled set is passed in to aks-engine generate the resulting apimodel.json in _output does persist the default value in code.

marosset: The logic in defaults.go will set a value for w.SSHEnabled for generate operations but will not set a value for aks-engine upgrade or aks-engine scale operations - which looks like the behavior we want.
I see a unit test that verifies apimodel.json is not modified during those commands.
@jackfrancis i'm not following why you are against having a function first do a nil-check before returning a value - can you help me understand your concerns here?

jackfrancis: I think my main concern is that `GetSSHEnabled` is a accessor method that wraps the SSHEnabled bool pointer property. So I would expect a nil value of that to be falsy.

marosset: This PR changed SSHEnabled property from a bool to a bool pointer so we could get an extra state - set to false, set to true, or unset by the user).
In the case where it is unset by the user wouldn't it make sense to set it to a default value instead of false?

jackfrancis: Right, perhaps I'm being too pedantic about that. It just seems that an accessor should return the truth, and if we want nil values to be semantically meaningful we should do that semantic enforcment so that it's sticky. I.e., so that by the time we use the accessors we can just have them be accessors, rather than being overloaded with that default value hijacking.

I think concretely what feels right to me is to set the default values during the defaults enforcement flow, and then confirm that the accessor is always being used only after that flow has been traversed and the data state representations have been rationalized (nil values converted to defaults, etc)

PR #2772, opened 2020-02-24
 feat: "aks-engine get-skus" command

mboersma:
**Reason for Change**:
Provides a hidden developer-oriented command that can recreate the `pkg/helpers/azure_skus_const.go` file:

```console
$ aks-engine get-skus --client-id=${ID} --client-secret=${SECRET}
Name                    Storage Account Type  Accelerated Networking Support
Standard_A0             Standard_LRS          false
Standard_A1             Standard_LRS          false
Standard_A10            Standard_LRS          false
Standard_A11            Standard_LRS          false
Standard_A1_v2          Standard_LRS          false
Standard_A2             Standard_LRS          false
Standard_A2_v2          Standard_LRS          false
Standard_A2m_v2         Standard_LRS          false
Standard_A3             Standard_LRS          false
...
$ aks-engine get-skus --client-id=${ID} --client-secret=${SECRET} -o code
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT license.

package helpers

// GetKubernetesAllowedVMSKUs provides the allowed sizes for Kubernetes agent VMs.
//
// Code generated for package helpers by aks-engine DO NOT EDIT. (@generated)
//
// To generate this code, run the command:
//   aks-engine get-skus --output=code

type VMSku struct {
	Name                  string
	StorageAccountType    string
	AcceleratedNetworking bool
}

var VMSkus = []VMSku{
	{
		Name:                  "Standard_A0",
		StorageAccountType:    "Standard_LRS",
		AcceleratedNetworking: false,
	},
	{
		Name:                  "Standard_A1",
		StorageAccountType:    "Standard_LRS",
		AcceleratedNetworking: false,
	},
...
...

```

**Issue Fixed**:
Refs #1894
See [automate_skus.md](https://github.com/Azure/aks-engine/blob/master/docs/design/automate-skus.md)

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

This needs more unit tests.

mboersma:
Codecov failed to report in but I think this is still dropping unit test coverage too much.

jackfrancis:
copy pasta alert in PR description :)

pkg/armhelpers/azurestack/azureclient.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/Azure/azure-sdk-for-go/services/apimanagement/mgmt/2017-03-01/apimanagement"
 	"github.com/Azure/azure-sdk-for-go/services/authorization/mgmt/2015-07-01/authorization"
 	"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2017-03-30/compute"
+	compute2 "github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-07-01/compute"
```
jackfrancis: @jadarsie Are you O.K. with this hack?

jadarsie: `2019-07-01` will likely fail, give me a few to check which version to use

jadarsie: `mgmt/2017-03-30/compute` has a `ResourceSkusClient` type, why the need to import `compute2`?

serbrech: but isn't that just for that specific sku generartion command? should not have any impact on how aks-engine works on stack, right?

mboersma: That's right, it's just to support the `aks-engine get-skus` command, which is hidden by default and unlikely to be run on Azure Stack.

I think the problem with the `mgmt/2017-03-30/compute` client was that it has a different func signature for the same call, and bridging the two types was proving difficult. But it's been a while, so I'll dust off this PR and take another look.

mboersma: I don't think it's worth bending over backwards (as I already did) to bridge the two API versions for Azure Stack. I also don't think there's a use case for this hidden, developer-only command in that environment. I suggest we leave it officially unimplemented on Azure Stack and keep the code simpler.

pkg/helpers/helpers.go
```diff
@@ -109,58 +109,6 @@ func CreateSSH(rg io.Reader, s *i18n.Translator) (privateKey *rsa.PrivateKey, pu
...
-		"Standard_F4s_v2", "Standard_F64s_v2", "Standard_F72s_v2", "Standard_F8", "Standard_F8s",
-		"Standard_F8s_v2", "Standard_L16s_v2", "Standard_L32s_v2", "Standard_L48s_v2", "Standard_L64s_v2",
-		"Standard_L80s_v2", "Standard_L8s_v2", "Standard_M128", "Standard_M128-32ms", "Standard_M128-64ms",
-		"Standard_M128m", "Standard_M128ms", "Standard_M128s", "Standard_M16-4ms", "Standard_M16-8ms",
-		"Standard_M16ms", "Standard_M208ms_v2", "Standard_M208s_v2", "Standard_M32-16ms", "Standard_M32-8ms",
-		"Standard_M32ls", "Standard_M32ms", "Standard_M32ts", "Standard_M416ms_v2", "Standard_M416s_v2",
-		"Standard_M64", "Standard_M64-16ms", "Standard_M64-32ms", "Standard_M64ls", "Standard_M64m",
-		"Standard_M64ms", "Standard_M64s", "Standard_M8-2ms", "Standard_M8-4ms", "Standard_M8ms":
-		return true
-	// grandfathered SKUs with accelerated networking enabled
```
mboersma, line 151: This was unreachable cruft AFAICT. You can't provision a cluster unless it's in the SKU list.

mboersma:
In the end, this added four SKUs:

 'Standard_NV16as_v4',
 'Standard_NV32as_v4',
 'Standard_NV4as_v4',
 'Standard_NV8as_v4',

PR #2774, opened 2020-02-24
fix: only configAddons if addons enabled

jackfrancis:
**Reason for Change**:

This PR fixes an anomolous invocation of `configAddons` during CSE, which only exists according to a particular set of conditions. E.g.:

```
+ time_metric ConfigAddons configAddons
+ local name start end diff tags
+ name=ConfigAddons
+ shift
++ apmz time unixnano
+ start=1582571554614609524
+ configAddons
/opt/azure/containers/provision.sh: line 69: configAddons: command not found
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -185,9 +185,11 @@ time_metric "ConfigureK8sCustomCloud" configureK8sCustomCloud
 
 time_metric "ConfigureCNI" configureCNI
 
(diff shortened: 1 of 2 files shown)


jackfrancis:
The new condition in `cse_main.sh` pairs with this in `cse_config.sh`:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_config.sh#L444

PR #2781, opened 2020-02-25
feat: add support for single stack IPv6

aramase:
**Reason for Change**:

- Add support for single-stack IPv6 cluster in order to run conformance tests and graduate IPv6 to beta in kubernetes.

API model - https://gist.github.com/aramase/a19fcbf1f309dab99600ee9c0d700fce

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/vlabs/validate.go
```diff
@@ -1255,6 +1255,24 @@ func (k *KubernetesConfig) Validate(k8sVersion string, hasWindows, ipv6DualStack
...
+			return errors.Errorf("could not validate version %s", k8sVersion)
+		}
+		minVersion, err := semver.Make("1.18.0-alpha.4")
+		if err != nil {
+			return errors.New("could not validate version")
+		}
+		if sv.LT(minVersion) {
+			return errors.Errorf("IPv6 single stack not available in kubernetes version %s", k8sVersion)
+		}
+		// singel stack IPv6 feature is currently only supported with kubenet
```
aojea: singel/single/

aramase: thanks! done

jackfrancis:
@aramase Is this ready for final review?

aramase:
@jackfrancis yes, this is ready for review.

aramase:
@jackfrancis Please review this when you get a chance. I would like to get it merged and then setup conformance job in prow for single stack IPv6 cluster.  Need that for graduation of IPv6 to beta.

parts/k8s/addons/1.18/kubernetesmasteraddons-kube-proxy-daemonset.yaml
```diff
@@ -9,6 +9,15 @@ data:
       kubeconfig: /var/lib/kubelet/kubeconfig
     clusterCIDR: "{{ContainerConfig "cluster-cidr"}}"
     mode: "{{ContainerConfig "proxy-mode"}}"
+    {{- if ContainerConfig "bind-address"}}
```
jackfrancis: So go template if operates against strings according to `""` = false and any non-empty string = true ?

pkg/api/vlabs/validate.go
```diff
@@ -454,12 +454,12 @@ func (a *Properties) validateAgentPoolProfiles(isUpdate bool) error {
...
 		// validate os type is linux if dual stack feature is enabled
-		if a.FeatureFlags.IsIPv6DualStackEnabled() {
+		if a.FeatureFlags.IsIPv6DualStackEnabled() || a.FeatureFlags.IsIPv6OnlyEnabled() {
 			if agentPoolProfile.OSType == Windows {
-				return errors.Errorf("Dual stack feature is supported only with Linux, but agent pool '%s' is of os type %s", agentPoolProfile.Name, agentPoolProfile.OSType)
+				return errors.Errorf("Dual stack and single stack IPv6 feature is supported only with Linux, but agent pool '%s' is of os type %s", agentPoolProfile.Name, agentPoolProfile.OSType)
 			}
 			if agentPoolProfile.Distro == CoreOS {
-				return errors.Errorf("Dual stack feature is currently supported only with Ubuntu, but agent pool '%s' is of distro type %s", agentPoolProfile.Name, agentPoolProfile.Distro)
+				return errors.Errorf("Dual stack and single stack IPv6 feature is currently supported only with Ubuntu, but agent pool '%s' is of distro type %s", agentPoolProfile.Name, agentPoolProfile.Distro)
```
jackfrancis: Should we add additional validation to ensure that we don't allow *both* `EnableIPv6DualStack` and `EnableIPv6Only` feature flags at the same time? I don't think that makes sense, and that could be an edge case vector for some of the `||` conditions in the code.

pkg/api/addons.go
```diff
@@ -677,6 +678,13 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		},
 	}
 
+	// set host network to true for single stack IPv6 as the the nameserver is currently
+	// IPv4 only. By setting it to host network, we can leverage the host routes to successfully
+	// resolve dns.
+	if cs.Properties.FeatureFlags.IsFeatureEnabled("EnableIPv6Only") {
```
jackfrancis: We should add UT coverage to `addons_test.go` for these ipv6 scenarios. Lemme know if you want help, it's 4k+ lines of UT coverage :)

parts/k8s/addons/1.18/kubernetesmasteraddons-kube-proxy-daemonset.yaml
aramase: @jackfrancis correct, if it's `""` then it's ignored.

pkg/api/vlabs/validate.go
aramase: Will add a check to ensure it doesn't happen.

pkg/api/addons.go
aramase: Let me take a stab at it, will ping you if you I need help :)

pkg/api/addons_test.go
```diff
@@ -3834,17 +3992,22 @@ func concatenateDefaultAddons(addons []KubernetesAddon, version string) []Kubern
...
 
 func overwriteDefaultAddons(addons []KubernetesAddon, version, kubernetesImageBase, kubernetesImageBaseType string) []KubernetesAddon {
+	overrideAddons := make(map[string]KubernetesAddon)
+	for _, addonOverride := range addons {
+		overrideAddons[addonOverride.Name] = addonOverride
+	}
+
 	var ret []KubernetesAddon
 	defaults := getDefaultAddons(version, kubernetesImageBase, kubernetesImageBaseType)
-	for _, addonOverride := range addons {
```
aramase: @jackfrancis With this logic, if `overwriteDefaultAddons` is called with more than 1 overwrite addon, it would result in duplicate values  for `overwriteAddons[1:]`. So I've updated the logic to just do a single pass of the array.

pkg/api/addons.go
aramase: @jackfrancis added the tests, PTAL!

parts/k8s/addons/1.18/kubernetesmasteraddons-kube-proxy-daemonset.yaml
jackfrancis: Looked up and verified in the actual docs as well :)

pkg/api/addons_test.go
jackfrancis: yikes, thanks!

PR #2794, opened 2020-02-26
[WIP] feat: install the azure.json file only on master node

### LINKED ISSUE #2747: Support a configuration that does not add service principal to worker nodes
**Describe the request**
The current behavior of AKS-engine is to push service principal credentials to all nodes in the cluster.
In a multi tenant scenario, it is desirable to disallow azure credentials from the nodes.
One such scenario is a case where some nodes in the cluster expose external-facing API endpoints, and thus can be more easily compromised. Once compromised, it's better that the attacker won't have access to the ARM resources the service principal has access to.

More specifically, it would be beneficial to have the service principal available to the master node, but not to the worker nodes

This feature will provide an additional configuration option for customers that will provide a secure cluster configuration for customers that can get to a situation where untrusted code runs on worker nodes.

We would like to be able to configure this by agent pool. as part of agent pool configuration we would like to have the following config key
`install-service-identity=false `  (for example).
for all vms in agent pool that have this configuration key set to false,  service principal won't be provisioned. We do want that the Master node will have the service principle installed. 

N/A

**Additional context**
In our case we are going to run 3rd party OSS software on the worker node - code that we did not write and we do not trust. if the node is compromised we want to make sure there is no privilege escalation in this case. 
If the node has a service principle installed on it the hacker will be able to use it to access ARM resources and we want to prevent this from occurring. 
The application that runs on worker nodes does not require access to any ARM resource.

shliderm:
feat: install the Azure.json file only on Master node.

**Reason for Change**:

Having the Azure.json file on all the nodes in the cluster might not be necessary for some cases and can be a security risk for other cases (if the node is compromised -it is better that the attacker won't have access to the ARM resources the service principal has access to)

**Issue Fixed**:

"Fixes #2747"

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This is work in progress, we will want to also add a feature flag to the agent pools definition  that will be set to 'true' for agent pools that we do want to install the azure.json on them (for example infra-nodes that are used for managing the resources etc.)

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -165,18 +165,20 @@ configureK8s() {
     chmod 0644 "${APISERVER_PUBLIC_KEY_PATH}"
     chown root:root "${APISERVER_PUBLIC_KEY_PATH}"
 
-    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
-    touch "${AZURE_JSON_PATH}"
-    chmod 0600 "${AZURE_JSON_PATH}"
-    chown root:root "${AZURE_JSON_PATH}"
-
-    set +x
     echo "${KUBELET_PRIVATE_KEY}" | base64 --decode > "${KUBELET_PRIVATE_KEY_PATH}"
     echo "${APISERVER_PUBLIC_KEY}" | base64 --decode > "${APISERVER_PUBLIC_KEY_PATH}"
     {{/* Perform the required JSON escaping */}}
     SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET//\\/\\\\}
     SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET//\"/\\\"}
-    cat << EOF > "${AZURE_JSON_PATH}"
+
+    if [[ -n "${MASTER_NODE}" ]]; then
+        AZURE_JSON_PATH="/etc/kubernetes/azure.json"
+        touch "${AZURE_JSON_PATH}"
+        chmod 0600 "${AZURE_JSON_PATH}"
+        chown root:root "${AZURE_JSON_PATH}"
+
+        set +x
+        cat << EOF > "${AZURE_JSON_PATH}"
 {
     "cloud":"{{GetTargetEnvironment}}",
     "tenantId": "${TENANT_ID}",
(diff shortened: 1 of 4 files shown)


jadarsie:
Azure Stack still needs that file, please take that into account.

pkg/api/defaults-kubelet_test.go
```diff
@@ -213,38 +213,38 @@ func TestKubeletConfigUseCloudControllerManager(t *testing.T) {
...
 func TestKubeletConfigCloudConfig(t *testing.T) {
 	// Test default value and custom value for --cloud-config
-	cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
-	cs.setKubeletConfig(false)
-	k := cs.Properties.OrchestratorProfile.KubernetesConfig.KubeletConfig
-	if k["--cloud-config"] != "/etc/kubernetes/azure.json" {
-		t.Fatalf("got unexpected '--cloud-config' kubelet config default value: %s",
-			k["--cloud-config"])
-	}
+	// cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
```
devigned: Please remove rather than comment.

pkg/api/defaults-kubelet_test.go
```diff
@@ -213,38 +213,38 @@ func TestKubeletConfigUseCloudControllerManager(t *testing.T) {
...
-	// Test default value and custom value for --azure-container-registry-config
-	cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
-	cs.setKubeletConfig(false)
-	k := cs.Properties.OrchestratorProfile.KubernetesConfig.KubeletConfig
-	if k["--azure-container-registry-config"] != "/etc/kubernetes/azure.json" {
-		t.Fatalf("got unexpected '--azure-container-registry-config' kubelet config default value: %s",
-			k["--azure-container-registry-config"])
-	}
+	//  Test default value and custom value for --azure-container-registry-config
+	// 	cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
```
devigned: Please remove rather than comment.

ritazh:
Would it make more sense to make this optional for now? I think there are few places in aks-engine right now that depend on the /etc/kubernetes/azure.json file on the agent nodes. e.g. `kubeletConfig`, pod identity. The e2e jobs should give us good signals on what needs to be updated to work with this change.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -165,18 +165,20 @@ configureK8s() {
...
     SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET//\"/\\\"}
-    cat << EOF > "${AZURE_JSON_PATH}"
+
+    if [[ -n "${MASTER_NODE}" ]]; then
+        AZURE_JSON_PATH="/etc/kubernetes/azure.json"
+        touch "${AZURE_JSON_PATH}"
+        chmod 0600 "${AZURE_JSON_PATH}"
+        chown root:root "${AZURE_JSON_PATH}"
+
+        set +x
```
CecileRobertMichon: this needs to stay befor the private key and service principal manipulations above to avoid printing secrets in CSE logs

shliderm: @CecileRobertMichon  by "this" you mean `set +x` ?  thanks for your comment.

CecileRobertMichon: yes, `set +x`

It disables printing commands to the logs (in this case the commands themselves contain sensitive information)

sukhong:
Please note that air-gapped clouds also need this file.

Issue #1046, opened 2019-04-13
Top nodes is broken after rotate-certs


CecileRobertMichon:
```
Azure Container Cluster using the Kubernetes Orchestrator with a linux agent pool 
  should be able to get nodes metrics
  /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:672

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

$ k top nodes

• Failure [609.057 seconds]
Azure Container Cluster using the Kubernetes Orchestrator
/go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:89
  with a linux agent pool
  /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:618
    should be able to get nodes metrics [It]
    /go/src/github.com/Azure/aks-engine/test/e2e/kubernetes/kubernetes_test.go:672

    Expected
        <bool>: false
    to be true
```

E2E fails after rotating certs because `top nodes` doesn't work.

CecileRobertMichon:
This was with k8s 1.9, need to check if the same applies to k8s versions that don't use heapster.

CecileRobertMichon:
Confirmed top nodes works after rotation on k8s 1.14. I suspect this is a heapster vs metrics server issue.

CecileRobertMichon:
keep open

avdams:
Hi @CecileRobertMichon ,
Can I pick this issue up ?

CecileRobertMichon:
@avdams yes please! Let me know if you need any help

PR #2799, opened 2020-02-27
feat: Enabling Secret Rotation on Azure Stack 

deaborch:
**Reason for Change**:

As discussed, this PR takes a look at the current implementation of the secret rotation feature and adapts it to work on azure stack as well as improving its usability on azure as well. 

**Issue Fixed**:

This PR looks at the current implemetation of the secret rotation and adapts it to work on azure stack. In addition to this, the PR strives to improve its usage on azure as well. 

Note
- This only deals with rotating the self-signed certificates used in the creation of the Kubernetes cluster 
- The PR does not provide a means of rotating the service principle
- This PR does not provide support for customer provided certificates
- This PR does not apply to any customer provided secret added post deployment
- The certificate rotation does not change the Certificate Authotity used in the creation of the cluster

Certificates Rotated
- Etcd Certificates 
- ApiServer Certificates 
- Kubelet Certificates 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/rotate_certs.go
+++ b/cmd/rotate_certs.go
@@ -6,6 +6,7 @@ package cmd
 import (
 	"bytes"
 	"context"
+	"encoding/pem"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -94,21 +95,6 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 
 	var err error
 
-	if err = rcc.getAuthArgs().validateAuthArgs(); err != nil {
-		return errors.Wrap(err, "failed to get validate auth args")
-	}
-
-	if rcc.client, err = rcc.authProvider.getClient(); err != nil {
-		return errors.Wrap(err, "failed to get client")
-	}
-
-	ctx, cancel := context.WithTimeout(context.Background(), armhelpers.DefaultARMOperationTimeout)
-	defer cancel()
-	_, err = rcc.client.EnsureResourceGroup(ctx, rcc.resourceGroupName, rcc.location, nil)
-	if err != nil {
-		return errors.Wrap(err, "ensuring resource group")
-	}
-
 	// load the cluster configuration.
 	if _, err = os.Stat(rcc.apiModelPath); os.IsNotExist(err) {
 		return errors.Errorf("specified api model does not exist (%s)", rcc.apiModelPath)
@@ -131,6 +117,29 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "parsing the api model")
 	}
 
+	if rcc.containerService.Properties.IsAzureStackCloud() {
+		writeCustomCloudProfile(rcc.containerService)
+		err = rcc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{IsUpgrade: false, IsScale: false})
+		if err != nil {
+			return errors.Wrap(err, "error parsing the api model")
+		}
+	}
+
+	if err = rcc.getAuthArgs().validateAuthArgs(); err != nil {
+		return errors.Wrap(err, "failed to get validate auth args")
+	}
+
+	if rcc.client, err = rcc.authProvider.getClient(); err != nil {
+		return errors.Wrap(err, "failed to get client")
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), armhelpers.DefaultARMOperationTimeout)
+	defer cancel()
+	_, err = rcc.client.EnsureResourceGroup(ctx, rcc.resourceGroupName, rcc.location, nil)
+	if err != nil {
+		return errors.Wrap(err, "ensuring resource group")
+	}
+
 	if rcc.outputDirectory == "" {
 		if rcc.containerService.Properties.MasterProfile != nil {
 			rcc.outputDirectory = path.Join("_output", rcc.containerService.Properties.MasterProfile.DNSPrefix)
@@ -148,8 +157,22 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 
 	log.Infoln("Generating new certificates")
 
-	// reset the certificateProfile and use the exisiting certificate generation code to generate new certificates.
-	rcc.containerService.Properties.CertificateProfile = &api.CertificateProfile{}
+	// reset the certificateProfile and use the existing certificate generation code to generate new certificates.
+	ca := ""
+	caKey := ""
+
+	if err = validatePem(rcc.containerService.Properties.CertificateProfile.CaCertificate); err == nil {
+		ca = rcc.containerService.Properties.CertificateProfile.CaCertificate
+	}
+
+	if err = validatePem(rcc.containerService.Properties.CertificateProfile.CaPrivateKey); err == nil {
+		caKey = rcc.containerService.Properties.CertificateProfile.CaPrivateKey
+	}
+
+	rcc.containerService.Properties.CertificateProfile = &api.CertificateProfile{
+		CaCertificate: ca,
+		CaPrivateKey:  caKey,
+	}
 	certsGenerated, _, err := rcc.containerService.SetDefaultCerts(api.DefaultCertParams{
 		PkiKeySize: helpers.DefaultPkiKeySize,
 	})
@@ -161,6 +184,13 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 		return errors.Errorf("specified ssh filepath does not exist (%s)", rcc.sshFilepath)
 	}
 	rcc.setSSHConfig()
+
+	log.Infoln("Rotating etcd certificates")
+
+	err = rcc.rotateEtcd()
+	if err != nil {
+		return errors.Wrap(err, "rotating etcd cluster")
+	}
 
 	log.Infoln("Rotating apiserver certificate")
 
@@ -176,13 +206,6 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "rotating kubelet")
 	}
 
-	log.Infoln("Rotating etcd certificates")
-
-	err = rcc.rotateEtcd(ctx)
-	if err != nil {
-		return errors.Wrap(err, "rotating etcd cluster")
-	}
-
 	log.Infoln("Updating kubeconfig")
 	err = rcc.updateKubeconfig()
 	if err != nil {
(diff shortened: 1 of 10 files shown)


cmd/rotate_certs.go
```diff
@@ -359,7 +373,7 @@ func (rcc *rotateCertsCmd) getKubeClient() (armhelpers.KubernetesClient, error)
 }
 
 // Rotate etcd CA and certificates in all of the master nodes.
-func (rcc *rotateCertsCmd) rotateEtcd(ctx context.Context) error {
+func (rcc *rotateCertsCmd) rotateEtcd() error {
```
devigned: @jackfrancis or @CecileRobertMichon any ideas on why we don't use `scp` rather than ssh'ing into the machine and executing the bash script?

cmd/rotate_certs.go
```diff
@@ -389,11 +403,11 @@ func (rcc *rotateCertsCmd) rotateEtcd(ctx context.Context) error {
 		}
 	}
 
-	log.Infoln("Rebooting all nodes... This might take a few minutes")
-	err := rcc.rebootAllNodes(ctx)
-	if err != nil {
-		return errors.Wrap(err, "rebooting the nodes")
-	}
+	// log.Infoln("Rebooting all nodes... This might take a few minutes")
```
devigned: If it is no longer of use, better to remove the code rather than just comment it out. We'll always have the git history.

cmd/rotate_certs.go
```diff
@@ -131,6 +116,29 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
...
 
+	if rcc.containerService.Properties.IsAzureStackCloud() {
+		writeCustomCloudProfile(rcc.containerService)
+		err = rcc.containerService.Properties.SetAzureStackCloudSpec(api.AzureStackCloudSpecParams{IsUpgrade: false, IsScale: false})
+		if err != nil {
+			return errors.Wrap(err, "error parsing the api model")
+		}
+	}
+
+	if err = rcc.getAuthArgs().validateAuthArgs(); err != nil {
```
devigned: Why validate after writing the custom cloud profile?

cmd/rotate_certs.go
```diff
@@ -149,7 +157,13 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 	log.Infoln("Generating new certificates")
 
 	// reset the certificateProfile and use the exisiting certificate generation code to generate new certificates.
-	rcc.containerService.Properties.CertificateProfile = &api.CertificateProfile{}
+	ca := rcc.containerService.Properties.CertificateProfile.CaCertificate
```
devigned: Is there a chance CaCertificate or CaPrivateKey will be empty?

cmd/rotate_certs.go
```diff
@@ -389,11 +403,11 @@ func (rcc *rotateCertsCmd) rotateEtcd(ctx context.Context) error {
 		}
 	}
 
-	log.Infoln("Rebooting all nodes... This might take a few minutes")
-	err := rcc.rebootAllNodes(ctx)
```
devigned: Must the nodes be rebooted? Could we just restart some services?

cmd/rotate_certs.go
deaborch: The first step initialy was to create a new empty certificate profile. This meant that the previous CA certs will not be present and therefore regenerated. This is why i temporarily store the CA details before running the certificate generation

cmd/rotate_certs.go
deaborch: They don't have to be rebooted, that is why i left that part of the code out. However i will be interested to know if anyone has any reason why it should be

CecileRobertMichon:
Does this fix https://github.com/Azure/aks-engine/issues/1046?

cmd/rotate_certs.go
CecileRobertMichon: I don't remember a good reason, but possibly because the nodes don't have a public IP so we need to SSH to the master first to be able to access the nodes

cmd/rotate_certs.go
deaborch: Here we are following the deploy logic

pkg/armhelpers/mockclients.go
```diff
@@ -301,6 +301,17 @@ func (mkc *MockKubernetesClient) ListAllPods() (*v1.PodList, error) {
 	return &v1.PodList{}, nil
 }
 
+//ListAllSystemPods returns all Pods running in system namespace
```
jadarsie: nit, space after `//`

deaborch:
> Does this fix #1046?

@CecileRobertMichon  just checked on azure stack and `kubectl top nodes` doesn't work regardless of secret rotation.

jadarsie:
Would it be possible to include rotate-certs as part of the e2e?

cmd/rotate_certs.go
devigned: @jackfrancis or @mboersma do you folks know why nodes needed to be rebooted?

devigned:
@deaborch please add an e2e test for rotation and I think we can get this pulled in.

cmd/rotate_certs.go
jackfrancis: nope

jackfrancis:
@deaborch it looks like this uses the existing CLI verb, and thus the existing documentation is still valid?

https://github.com/Azure/aks-engine/blob/master/docs/topics/certificaterotation.md

deaborch:
@jackfrancis yes the documentation will not change at all

cmd/rotate_certs.go
```diff
@@ -148,8 +157,22 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 
 	log.Infoln("Generating new certificates")
 
-	// reset the certificateProfile and use the exisiting certificate generation code to generate new certificates.
-	rcc.containerService.Properties.CertificateProfile = &api.CertificateProfile{}
+	// reset the certificateProfile and use the existing certificate generation code to generate new certificates.
+	ca := ""
+	caKey := ""
+
+	if err = validatePerm(rcc.containerService.Properties.CertificateProfile.CaCertificate); err == nil {
```
jadarsie: Is it OK to continue if `validatePerm` returns an error?

cmd/rotate_certs.go
```diff
@@ -451,6 +472,15 @@ func (rcc *rotateCertsCmd) rotateKubelet() error {
 			}
 		}
 	}
+
+	for _, host := range rcc.masterNodes {
+		log.Debugf("Restarting kubelet on node %s", host.Name)
+		out, err := rcc.sshCommandExecuter("sudo systemctl restart kubelet", rcc.masterFQDN, host.Name, "22", rcc.sshConfig)
```
jadarsie: It would be nice if this had retry logic

cmd/rotate_certs.go
```diff
@@ -513,3 +543,21 @@ func executeCmd(command, masterFQDN, hostname string, port string, config *ssh.C
...
+func contains(arr []string, str string) bool {
+	for _, a := range arr {
+		if a == str {
+			return true
+		}
+	}
+	return false
+}
+
+func validatePerm(data string) error {
```
jadarsie: `Pem` instead of `Perm`?

cmd/rotate_certs.go
```diff
@@ -513,3 +543,21 @@ func executeCmd(command, masterFQDN, hostname string, port string, config *ssh.C
...
+	block, _ := pem.Decode([]byte(data))
+	if block == nil {
+		return errors.New("failed to parse certificate PEM")
```
jadarsie: `PEM certificate`

cmd/rotate_certs_test.go
```diff
@@ -413,3 +472,16 @@ func makeTmpDir(t *testing.T) (string, func()) {
...
+	t.Parallel()
+
+	g := NewGomegaWithT(t)
+
+	array := []string{"label1", "label2"}
+	result := contains(array, "label1")
+	g.Expect(result).Should(BeTrue())
+	result = contains(array, "test")
+	g.Expect(result).ShouldNot(BeTrue())
+
```
jadarsie: Remove black line

cmd/rotate_certs_test.go
```diff
@@ -136,23 +180,39 @@ func TestDeleteAllPods(t *testing.T) {
...
-	err = rcc.deleteAllPods()
+	err = rcc.deleteKubeSystemPods()
 	g.Expect(err).To(HaveOccurred())
 	g.Expect(err.Error()).To(ContainSubstring("failed to delete pod"))
 
 	mockClient.MockKubernetesClient.FailDeletePod = false
-	err = rcc.deleteAllPods()
+	err = rcc.deleteKubeSystemPods()
 	g.Expect(err).NotTo(HaveOccurred())
+
```
jadarsie: blank

cmd/rotate_certs.go
deaborch: Would be nice to have but we will first need a way to check if it happened or not using ssh

cmd/rotate_certs.go
deaborch: If it returns an error then an empty string will be used. This would mean the CA Certificates will be regenerated. Rotating the Ca certificates is not supported at this time so we may need to fail fast if the root certificates cannot be validated

cmd/rotate_certs.go
jadarsie: it can be done using ssh
create a follow up issue please to address this eventually

cmd/rotate_certs.go
jadarsie: Please make sure documentation explains if it does not, or at least log a message to stdout

test/e2e/cluster.sh
```diff
@@ -146,6 +146,24 @@ else
...
+    ${DEV_IMAGE} \
+    ./bin/aks-engine rotate-certs \
+    --subscription-id ${AZURE_SUBSCRIPTION_ID} \
+    --api-model _output/$RESOURCE_GROUP/apimodel.json \
+    --apiserver "$RESOURCE_GROUP.$REGION.cloudapp.azure.com" \
+    --location $REGION \
+    --resource-group $RESOURCE_GROUP \
+    --auth-method client_secret \
+    --client-id ${AZURE_CLIENT_ID} \
+    --ssh _output/$RESOURCE_GROUP/$RESOURCE_GROUP-ssh \
```
jackfrancis: too many `$RESOURCE_GROUP` instances here, should be:

`--ssh _output/$RESOURCE_GROUP-ssh \`

PR #2802, opened 2020-02-27
chore: force image base to MCR if target cloud is Azure Stack

jadarsie:
**Reason for Change**:
Azure Stack's hyperkube images are now hosted in a different MCR endpoint. This PR fixes upgrade for clusters that pulled from the old MCR endpoint.

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


cmd/upgrade.go
```diff
@@ -323,3 +324,13 @@ func isVMSSNameInAgentPoolsArray(vmss string, cs *api.ContainerService) bool {
 	}
 	return false
 }
+
+func azureStackImageBaseOverrides(kc *api.KubernetesConfig) {
```
jackfrancis: @CecileRobertMichon do you have any thoughts on doing this override here? Is this preferable to the pattern we have currently, where we pass along upgrade context through the code flow and do upgrade-specific mutations during defaults enforcement?

CecileRobertMichon: this has the advantage of also correcting the use of deprecated fields during create. So if someone has a stored apimodel with a deprecated value, their creates would still succeed. We should probably not do silent overrides though, I would vote for adding a log line here that warns the user that the image is being overridden. 

The other thing is whatever we choose it's better to have consistency. If we have "isUpgrade" flags in a bunch of places and then do overrides that makes it easier to introduce bugs because we can end up having dead code that we think is setting a value but then it ends up getting set to something else later with an override. Which also means `upgrade.go` might not be the best place to put this because it's isolated from the other value defaults so someone changing the defaults code might not realize it exists.

CecileRobertMichon: > this has the advantage of also correcting the use of deprecated fields during create

actually, that's not true because we're doing this in the upgrade file :)

PR #2804, opened 2020-02-28
feat: Add IsCredentialAutoGenerated for WindowsProfile

AbelHu:
**Reason for Change**:

AKS plans to auto generate Windows username and password if Windows agent pools can be added to this cluster later. We need to distinguish that the Windows credential is set by users from that the Windows credential is auto generated.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/types.go
```diff
@@ -207,18 +207,19 @@ type CustomNodesDNS struct {
...
+	ImageVersion              string            `json:"imageVersion"`
+	WindowsImageSourceURL     string            `json:"windowsImageSourceURL"`
+	WindowsPublisher          string            `json:"windowsPublisher"`
+	WindowsOffer              string            `json:"windowsOffer"`
+	WindowsSku                string            `json:"windowsSku"`
+	WindowsDockerVersion      string            `json:"windowsDockerVersion"`
+	Secrets                   []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled                bool              `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates    *bool             `json:"enableAutomaticUpdates,omitempty"`
+	IsCredentialAutoGenerated *bool             `json:"isCredentialAutoGenerated,omitempty"`
```
marosset: Please add in alphabetical order (just inserting the new fields is  OK)

marosset: Since this is a nullable field please consider adding a function similar to https://github.com/Azure/aks-engine/blob/3dd077cb5035f4ea2cbf7dec420fa2c351da94ee/pkg/api/types.go#L1705
Which gets the value or returns an appropriate default value to ensure safe and consistent access to the field.

AbelHu: I added this as same as that aks-engine does for Linux. It seems like that no one else uses it except AKS so let us only add one field but not introduce more functions here. How do you think, @marosset ?

https://github.com/Azure/aks-engine/blob/master/pkg/api/types.go#L178-L189
```
type LinuxProfile struct {
	AdminUsername string `json:"adminUsername"`
	SSH           struct {
		PublicKeys []PublicKey `json:"publicKeys"`
	} `json:"ssh"`
	Secrets               []KeyVaultSecrets   `json:"secrets,omitempty"`
	Distro                Distro              `json:"distro,omitempty"`
	ScriptRootURL         string              `json:"scriptroot,omitempty"`
	CustomSearchDomain    *CustomSearchDomain `json:"customSearchDomain,omitempty"`
	CustomNodesDNS        *CustomNodesDNS     `json:"CustomNodesDNS,omitempty"`
	IsSSHKeyAutoGenerated *bool               `json:"isSSHKeyAutoGenerated,omitempty"`
}
```

AbelHu: Added GetIsCredentialAutoGenerated

PR #2832, opened 2020-03-04
chore: default to large cluster settings on Azure Stack

jadarsie:
**Reason for Change**:
Default AKSe settings do not play well with Azure Stack limits. This PR changes defaults for Azure Stack to reduce the load to ARM.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -306,6 +306,22 @@ const (
 	DefaultKubernetesCloudProviderRateLimitBucketWrite = DefaultKubernetesCloudProviderRateLimitBucket
 )
 
+// Azure Stack configures all clusters as if they were large clusters.
+const (
+	DefaultAzureStackKubernetesCloudProviderBackoffRetries       = 6
+	DefaultAzureStackKubernetesCloudProviderBackoffJitter        = 1.0
+	DefaultAzureStackKubernetesCloudProviderBackoffDuration      = 5
+	DefaultAzureStackKubernetesCloudProviderBackoffExponent      = 1.5
+	DefaultAzureStackKubernetesCloudProviderRateLimitQPS         = 3.0
+	DefaultAzureStackKubernetesCloudProviderRateLimitQPSWrite    = 3.0
+	DefaultAzureStackKubernetesCloudProviderRateLimitBucket      = 10
+	DefaultAzureStackKubernetesCloudProviderRateLimitBucketWrite = 10
+	DefaultAzureStackKubernetesNodeStatusUpdateFrequency         = "1m"
+	DefaultAzureStackKubernetesCtrlMgrRouteReconciliationPeriod  = "1m"
+	DefaultAzureStackKubernetesCtrlMgrNodeMonitorGracePeriod     = "5m"
+	DefaultAzureStackKubernetesCtrlMgrPodEvictionTimeout         = "1m"
+)
+
 const (
 	//AzureEdgeDCOSBootstrapDownloadURL is the azure edge CDN download url
 	AzureEdgeDCOSBootstrapDownloadURL = "https://dcosio.azureedge.net/dcos/%s/bootstrap/%s.bootstrap.tar.xz"
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -267,7 +267,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		// Enforce sane cloudprovider backoff defaults.
-		o.KubernetesConfig.SetCloudProviderBackoffDefaults()
+		a.SetCloudProviderBackoffDefaults()
 
 		if o.KubernetesConfig.CloudProviderRateLimit == nil {
 			o.KubernetesConfig.CloudProviderRateLimit = to.BoolPtr(DefaultKubernetesCloudProviderRateLimit)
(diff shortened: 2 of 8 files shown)


pkg/api/const.go
```diff
@@ -306,6 +306,22 @@ const (
...
+	DefaultAzureStackKubernetesCloudProviderBackoffDuration      = 5
+	DefaultAzureStackKubernetesCloudProviderBackoffExponent      = 1.5
+	DefaultAzureStackKubernetesCloudProviderRateLimitQPS         = 3.0
+	DefaultAzureStackKubernetesCloudProviderRateLimitQPSWrite    = 3.0
+	DefaultAzureStackKubernetesCloudProviderRateLimitBucket      = 10
+	DefaultAzureStackKubernetesCloudProviderRateLimitBucketWrite = 10
+	DefaultAzureStackKubernetesNodeStatusUpdateFrequency         = "1m"
+	DefaultAzureStackKubernetesCtrlMgrRouteReconciliationPeriod  = "1m"
+	DefaultAzureStackKubernetesCtrlMgrNodeMonitorGracePeriod     = "5m"
+	DefaultAzureStackKubernetesCtrlMgrPodEvictionTimeout         = "1m"
```
rjaini: Today default value is "--pod-eviction-timeout=5m0s". Lets better check why is that value ?.

rjaini: https://github.com/Azure/aks-engine/blob/feecc8b4f6173c68d6ee15792036e5e0faed03c6/pkg/api/const.go#L399

pkg/api/types.go
```diff
@@ -2167,19 +2179,32 @@ func (k *KubernetesConfig) RequiresDocker() bool {
...
 	if k.CloudProviderBackoffRetries == 0 {
 		k.CloudProviderBackoffRetries = DefaultKubernetesCloudProviderBackoffRetries
+		if p.IsAzureStackCloud() {
+			k.CloudProviderBackoffRetries = DefaultAzureStackKubernetesCloudProviderBackoffRetries
+		}
 	}
 	if k.CloudProviderBackoffMode != CloudProviderBackoffModeV2 {
 		if k.CloudProviderBackoffExponent == 0 {
 			k.CloudProviderBackoffExponent = DefaultKubernetesCloudProviderBackoffExponent
+			if p.IsAzureStackCloud() {
```
rjaini: This can be removed.

pkg/api/types.go
```diff
@@ -2167,19 +2179,32 @@ func (k *KubernetesConfig) RequiresDocker() bool {
...
+				k.CloudProviderBackoffExponent = DefaultAzureStackKubernetesCloudProviderBackoffExponent
+			}
 		}
 		if k.CloudProviderBackoffJitter == 0 {
 			k.CloudProviderBackoffJitter = DefaultKubernetesCloudProviderBackoffJitter
+			if p.IsAzureStackCloud() {
```
rjaini: This can be removed.

pkg/api/const.go
```diff
@@ -306,6 +306,22 @@ const (
 	DefaultKubernetesCloudProviderRateLimitBucketWrite = DefaultKubernetesCloudProviderRateLimitBucket
 )
 
+// Azure Stack configures all clusters as if they were large clusters.
+const (
+	DefaultAzureStackKubernetesCloudProviderBackoffRetries       = 6
+	DefaultAzureStackKubernetesCloudProviderBackoffJitter        = 1.0
```
rjaini: Defaults for 1.14.0 onward for cloudProviderBackoffMode is v2
https://github.com/Azure/aks-engine/blob/beb8ad935ff3b4e37c17bbbe129562e28372068a/docs/topics/clusterdefinitions.md#L78

Hence based on Azure Cloud provider code CloudProviderBackoffExponent and CloudProviderBackoffJitter is not needed to be set as it is no op.

rjaini: Code is now deprecated https://github.com/msazurestackworkloads/kubernetes/blob/release-1.14/pkg/cloudprovider/providers/azure/azure.go#L317

pkg/api/const.go
jadarsie: FYI, it aligns with the large clusters [guideline](https://github.com/Azure/aks-engine/blob/master/docs/howto/kubernetes-large-clusters.md)

pkg/api/types.go
jadarsie: I would rather keep it even if regular defaults and large-cluster defaults match.
Regular defaults can then evolve independently.

pkg/api/const.go
rjaini: The docs is for all K8s version. New behavior is different from 1.14 which is officially supported on Azure Stack Hub.
We should check once with Jack if we need to update the large cluster guideline page. As it might be stale and may need some updates.

pkg/api/const.go
jadarsie: I would rather keep both Azure and Azure Stack flows consistent.
Worst case seems to be a warning in the logs.

pkg/api/const.go
jadarsie: @jackfrancis is the guideline stale?

pkg/api/defaults.go
```diff
@@ -267,7 +267,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		// Enforce sane cloudprovider backoff defaults.
-		o.KubernetesConfig.SetCloudProviderBackoffDefaults()
+		a.SetCloudProviderBackoffDefaults()
```
jackfrancis: Ah yes, the eminently significant `a` variable.

PR #2839, opened 2020-03-04
chore: update go compiler to 1.14.2

mboersma:
**Reason for Change**:
Bumps the default compiler to go 1.14.2 and syncs golangci-lint with the local version.

See https://github.com/deis/docker-go-dev/releases/tag/v1.26.2

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Previous versions of golangci-lint have caused problems in CI before, so let's watch for that.


.golangci.yml
```diff
@@ -4,6 +4,7 @@ run:
     - test/i18n
     - pkg/test
   skip-files:
+    - pkg/api/agentPoolOnlyApi/v20180331/apiloader_test.go
```
mboersma, line 7: Versions of `golangci-lint` since v1.23.7 have choked on a legal struct literal in this file. I tried many refactorings, then gave up and added it to the "ignored by linter" list.

mboersma:
I *think* I have finally gotten the current version of `golangci-lint` to behave, but given that the previous failures were a bit flaky (and that I can almost never reproduce it locally), let's give this a few more retries in CI.

mboersma:
Importantly, Go 1.14.2 (as well as 1.13.10) [has a fix](https://github.com/golang/go/issues?q=milestone%3AGo1.14.2+label%3ACherryPickApproved) for the dreaded "PowerRegisterSuspendResumeNotification on Azure App Services" bug for which #2976 was intended as a temporary fix.

PR #2843, opened 2020-03-04
chore: don't re-run provision.sh if already called. 

paulgmiller:
Because commandToExecute changes with every upgrade (since the kubernetes version env variable changes) CSE are run when AKS tried to do an instance upgrade then reimage (which vmss is recommending as preferred over creating and deleting repeatedly). 

Rerunning the CSE's seems to hit a file watch timeout maybe 1:8 or 1:16 times. I have not found the file responsible just the error code.

However it seems like the CSE's were not made to be idempotent and maybe shoudl just abort if they've already run to completion. So maybe it's a good idea in general to check the provision.complete sentinal file that already gets emited at the end of the script?

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -1,13 +1,13 @@
 #!/bin/bash
 ERR_FILE_WATCH_TIMEOUT=6 {{/* Timeout waiting for a file */}}
-ERR_ALREADY_RUN=116 {{/*Have already run and not sure if idempotent*/}
 
 set -x
-echo $(date),$(hostname), startcustomscript>>/opt/m
-if [ -s /opt/azure/containers/provision.complete ]; then
-      exit $ERR_ALREADY_RUN
+if [ -f /opt/azure/containers/provision.complete ]; then
+      echo "Already ran to success exiting..."
+      exit 0
 fi
 
+echo $(date),$(hostname), startcustomscript>>/opt/m
 
 for i in $(seq 1 3600); do
     if [ -s {{GetCSEHelpersScriptFilepath}} ]; then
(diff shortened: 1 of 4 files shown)


paulgmiller:
Still investigating here but creating a PR as it was easier and more informative than an issue maybe?

pkg/engine/vmextensions.go
```diff
@@ -86,6 +86,7 @@ func CreateCustomScriptExtension(cs *api.ContainerService) VirtualMachineExtensi
 			AutoUpgradeMinorVersion: to.BoolPtr(true),
 			Settings:                &map[string]interface{}{},
 			ProtectedSettings: &map[string]interface{}{
+				//note that any change to this property will trigger a CSE rerun on upgrade as well as reimage.
```
jackfrancis: This inline comment in a struct literal expression is breaking lint. gofmt should put you on the right path

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/bash
 ERR_FILE_WATCH_TIMEOUT=6 {{/* Timeout waiting for a file */}}
+ERR_ALREADY_RUN=116 {{/*Have already run and not sure if idempotent*/}
 
 set -x
 echo $(date),$(hostname), startcustomscript>>/opt/m
+if [ -s /opt/azure/containers/provision.complete ]; then
+      exit $ERR_ALREADY_RUN
```
jackfrancis: Do we actually want to return an error here, or is this just for gathering error rate data? As you've identified, the bootstrap scripts as-is are not implemented idempotently, so something unexpected causes them to re-run "accidentally", perhaps we should just return zero immediately.

paulgmiller: To me success is fine though it might be infuriating if you hit this path get a 0 and don't understand why. Ideally there's be multiple success code or I could just echo ALREADY RUN

jackfrancis: Let's move this exit statement to run before the datestamp echo statement

jackfrancis:
@paulgmiller make generate and commit generated code to appease UT check

paulgmiller:
don't suppose there's unittests for the bash script I can improve :)

jackfrancis:
@paulgmiller feel free to add some! :) j/k

We do use shellcheck, so there's something

jackfrancis:
Thanks @paulgmiller!

PR #2849, opened 2020-03-06
chore: don't require azure.json on node vms

jackfrancis:
**Reason for Change**:

This PR supports empty string (`""`) values for `kubeletConfig` configuration properties, to reflect that those values are interpreted significantly by the kubelet runtime. Specifically, when these three properties are empty strings, it tells the kubelet runtime that it is not running in an azure cloudprovider context (e.g., on a master vm):

```
"--cloud-provider":                  "",
"--cloud-config":                    "",
"--azure-container-registry-config": ""
```

To accommodate the significance of the above configuration set, this PR delivers the `/etc/kubernetes/azure.json` file to node vms only if those above three configuration properties are not *all empty*. In other words, if any one of the above properties is not `""`, then `/etc/kubernetes/azure.json` is paved on the vm just as it is prior to this change.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -165,14 +165,19 @@ configureK8s() {
     chmod 0644 "${APISERVER_PUBLIC_KEY_PATH}"
     chown root:root "${APISERVER_PUBLIC_KEY_PATH}"
 
-    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
-    touch "${AZURE_JSON_PATH}"
-    chmod 0600 "${AZURE_JSON_PATH}"
-    chown root:root "${AZURE_JSON_PATH}"
-
-    set +x
     echo "${KUBELET_PRIVATE_KEY}" | base64 --decode > "${KUBELET_PRIVATE_KEY_PATH}"
     echo "${APISERVER_PUBLIC_KEY}" | base64 --decode > "${APISERVER_PUBLIC_KEY_PATH}"
+    configureKubeletServerCert
+    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
+    if [[ -n "${MASTER_NODE}" ]]; then
+        if [[ "${ENABLE_AGGREGATED_APIS}" = True ]]; then
+            generateAggregatedAPICerts
+        fi
+    else
+        {{- /* If we are a node vm then we only proceed w/ local azure.json configuration if cloud-init has pre-paved that file */}}
+        wait_for_file 1 1 $AZURE_JSON_PATH || return
+    fi
+
     {{/* Perform the required JSON escaping */}}
     SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET//\\/\\\\}
     SERVICE_PRINCIPAL_CLIENT_SECRET=${SERVICE_PRINCIPAL_CLIENT_SECRET//\"/\\\"}
@@ -218,16 +223,9 @@ configureK8s() {
 EOF
     set -x
     if [[ "${CLOUDPROVIDER_BACKOFF_MODE}" = "v2" ]]; then
-        sed -i "/cloudProviderBackoffExponent/d" /etc/kubernetes/azure.json
-        sed -i "/cloudProviderBackoffJitter/d" /etc/kubernetes/azure.json
+        sed -i "/cloudProviderBackoffExponent/d" $AZURE_JSON_PATH
+        sed -i "/cloudProviderBackoffJitter/d" $AZURE_JSON_PATH
     fi
-    if [[ -n "${MASTER_NODE}" ]]; then
-        if [[ "${ENABLE_AGGREGATED_APIS}" = True ]]; then
-            generateAggregatedAPICerts
-        fi
-    fi
-
-    configureKubeletServerCert
 }
 
 configureCNI() {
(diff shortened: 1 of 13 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -165,14 +165,19 @@ configureK8s() {
...
 
-    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
-    touch "${AZURE_JSON_PATH}"
-    chmod 0600 "${AZURE_JSON_PATH}"
-    chown root:root "${AZURE_JSON_PATH}"
-
-    set +x
     echo "${KUBELET_PRIVATE_KEY}" | base64 --decode > "${KUBELET_PRIVATE_KEY_PATH}"
     echo "${APISERVER_PUBLIC_KEY}" | base64 --decode > "${APISERVER_PUBLIC_KEY_PATH}"
+    configureKubeletServerCert
```
jackfrancis: Moved this up here to accommodate the "return early if no azure.json" logic. It's not related to the azure.json paving, so can be done in any order.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -165,14 +165,19 @@ configureK8s() {
...
+    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
+    if [[ -n "${MASTER_NODE}" ]]; then
+        if [[ "${ENABLE_AGGREGATED_APIS}" = True ]]; then
+            generateAggregatedAPICerts
```
jackfrancis: Ditto moved this up earlier as well.

pkg/api/defaults-kubelet.go
```diff
@@ -33,7 +33,7 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		switch key {
 		case "--anonymous-auth", "--client-ca-file":
 			if !to.Bool(o.KubernetesConfig.EnableSecureKubelet) { // Don't add if EnableSecureKubelet is disabled
-				staticLinuxKubeletConfig[key] = ""
+				delete(staticLinuxKubeletConfig, key)
```
jackfrancis: Now that we are no longer deleting empty string keys in the `removeKubeletFlags` func we do it here.

pkg/api/defaults-kubelet.go
```diff
@@ -270,13 +270,6 @@ func removeKubeletFlags(k map[string]string, v string) {
 			delete(k, key)
 		}
 	}
-
-	// Get rid of keys with empty string values
```
jackfrancis: We don't remove empty string valued keys because they are significant

pkg/api/defaults-kubelet_test.go
```diff
@@ -2166,3 +2167,77 @@ func TestReadOnlyPort(t *testing.T) {
 	}
 
 }
+
+func TestRemoveKubeletFlags(t *testing.T) {
```
jackfrancis: This never had UT coverage

pkg/api/types.go
```diff
@@ -1638,6 +1638,36 @@ func (a *AgentPoolProfile) IsUbuntuNonVHD() bool {
 	return a.IsUbuntu() && !a.IsVHDDistro()
 }
 
+// RequiresCloudproviderConfig returns true if the azure.json cloudprovider config should be delivered to the nodes in this pool
+func (a *AgentPoolProfile) RequiresCloudproviderConfig() bool {
+	if a.KubernetesConfig != nil && a.KubernetesConfig.KubeletConfig != nil {
+		if v, ok := a.KubernetesConfig.KubeletConfig["--cloud-provider"]; ok {
```
jackfrancis: Open to a better way to implement this logic. The UT cases below tell the truth in terms of proving that we're doing what we want to do.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -165,14 +165,19 @@ configureK8s() {
...
+        fi
+    else
+        {{- /* If we are a node vm then we only proceed w/ local azure.json configuration if cloud-init has pre-paved that file */}}
```
jackfrancis: To be clear, we are *not* pre-paving `azure.json` for master vms (note that there is no diff to `masternodecustomdata.yml`. The existing code (specifically `cat << EOF > "${AZURE_JSON_PATH}"`) works as-is whether or not the file is there (the `> "${AZURE_JSON_PATH}"` part). So we are just using the pre-existence of `azure.json`, in a non-master flow only, to determine whether or not we will include the `azure.json` on this node. If the "check for file" test fails, we return immediately, because that tells us we aren't including `azure.json` on this node.

This works because:

1) we are paving `azure.json` on nodes, if needed, prior to paving *this* file itself (`cse_config.sh` in code, and `/opt/azure/containers/provision_configs.sh` on the local vm fs)
2) we are waiting for *this* file in `cse_main.sh` before executing code in *this* file, and because the cloud-init `write_files` property is a yaml sequence, those will be processed in order. Thus, by the time we are in *this* execution flow, *if* `azure.json` should be here, it will be here.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -165,14 +165,19 @@ configureK8s() {
     chmod 0644 "${APISERVER_PUBLIC_KEY_PATH}"
     chown root:root "${APISERVER_PUBLIC_KEY_PATH}"
 
-    AZURE_JSON_PATH="/etc/kubernetes/azure.json"
-    touch "${AZURE_JSON_PATH}"
-    chmod 0600 "${AZURE_JSON_PATH}"
-    chown root:root "${AZURE_JSON_PATH}"
-
-    set +x
```
shliderm: I think we still need this `set +x` here to avoid printing secrets in CSE logs

jackfrancis: Yep, thanks!

PR #2858, opened 2020-03-06
feat: collect Windows CSE logs during log collection

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


scripts/collect-windows-logs.ps1
```diff
@@ -40,6 +40,7 @@ Invoke-WebRequest -UseBasicParsing https://raw.githubusercontent.com/Microsoft/S
 & 'c:\k\debug\collectlogs.ps1' | write-Host
 $netLogs = Get-ChildItem (Get-ChildItem -Path c:\k\debug -Directory | Sort-Object LastWriteTime -Descending | Select-Object -First 1).FullName | Select-Object -ExpandProperty FullName
 $paths += $netLogs
+$paths += "c:\AzureData\CustomDataSetupScript.log"
```
jackfrancis: nit: in the future let's process this and windows cse scripts via go templating so we can refer to this string programmatically (as a common reference that cse + log collection script can both use)

marosset: We intentionally don't want this script to be processed via go templating. 
We want users to be able to curl this and run it if it isn't already on the node.

PR #2862, opened 2020-03-09
feat: Azure CNI dual stack support

tamilmani1989:
**Reason for Change**:

This PR adds support for Azure CNI dual stack. The v4 IPs will be in vnet and v6 ips are non-vnet and follow udr solution as like kubenet. The node ipam controller will carve unique ipv6 subnet for each node and CNI then assigns ipv6 ips from that subnet to pods. The cni dual stack is currently supported only for bridge mode and not with any network policies. This PR can be merged after releasing new cni with ipv6 support.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/defaults-cloud-controller-manager.go
```diff
@@ -5,18 +5,27 @@ package api
...
 
 	"github.com/Azure/aks-engine/pkg/api/common"
 )
 
 func (cs *ContainerService) setCloudControllerManagerConfig() {
 	o := cs.Properties.OrchestratorProfile
+	isAzureCNIDualStack := cs.Properties.IsAzureCNIDualStack()
+	clusterCidr := o.KubernetesConfig.ClusterSubnet
+	if isAzureCNIDualStack {
+		clusterSubnets := strings.Split(clusterCidr, ",")
```
jackfrancis: shouldn't we be splitting on `"."` (period)?, as the `ClusterSubnet` string property will be expressed as a normal CIDR string.

pkg/api/defaults-cloud-controller-manager.go
```diff
@@ -5,18 +5,27 @@ package api
...
+		if len(clusterSubnets) > 1 {
```
jackfrancis: I think we want `if len(clusterSubnets) > 0` here if the reason is to not index a non-existent array item.

jackfrancis: sorry, ignore this!

pkg/api/defaults-cloud-controller-manager.go
```diff
@@ -5,18 +5,27 @@ package api
...
+			clusterCidr = clusterSubnets[1]
+		}
+	}
 	staticCloudControllerManagerConfig := map[string]string{
-		"--allocate-node-cidrs":         strconv.FormatBool(!o.IsAzureCNI()),
-		"--configure-cloud-routes":      strconv.FormatBool(o.RequireRouteTable()),
+		"--allocate-node-cidrs":         strconv.FormatBool(!o.IsAzureCNI() || isAzureCNIDualStack),
+		"--configure-cloud-routes":      strconv.FormatBool(o.RequireRouteTable() || isAzureCNIDualStack),
```
jackfrancis: Let's move this additional `isAzureCNIDualStack` criteria into the `RequireRouteTable` method itself, then we'll get that new criteria for free in all the other areas that depend upon `RequireRouteTable()`

pkg/api/defaults-cloud-controller-manager.go
tamilmani1989: Nope not splitting ip address here. Dual stack can have 2 subnets (v4 and v6) comma separated. Splitting v4 and v6 subnet here

pkg/api/defaults-cloud-controller-manager.go
tamilmani1989: Yes i thought about this but there is one concern. Need to move RequireRouteTable as function of Properties instead of OrchestratorProfile . Is that fine?

func (p *Properties) IsAzureCNIDualStack() bool {
	o := p.OrchestratorProfile
	f := p.FeatureFlags
	return o.IsAzureCNI() && f.IsFeatureEnabled("EnableIPv6DualStack")
}

pkg/api/defaults.go
```diff
@@ -419,7 +426,13 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			if !cs.Properties.MasterProfile.IsCustomVNET() {
 				if cs.Properties.OrchestratorProfile.IsAzureCNI() {
 					// When VNET integration is enabled, all masters, agents and pods share the same large subnet.
-					cs.Properties.MasterProfile.Subnet = cs.Properties.OrchestratorProfile.KubernetesConfig.ClusterSubnet
+					cs.Properties.MasterProfile.Subnet = o.KubernetesConfig.ClusterSubnet
+					o := cs.Properties.OrchestratorProfile
```
jackfrancis: Why are we re-assigning `o` here?

pkg/api/defaults-cloud-controller-manager.go
jackfrancis: Good catch. Yes, that's O.K.

pkg/api/defaults.go
tamilmani1989: yes not needed. fixed

pkg/api/defaults-cloud-controller-manager.go
tamilmani1989: just confirming this OK is meant for moving RequireRouteTable as function of Properties?

jackfrancis: Yep, I am definitely encouraging you to refactor RequireRouteTable in order to support using the isAzureCNIDualStack Properties method. Hopefully it won't be too tricky to update all usages of RequireRouteTable. If so, lemme know.

tamilmani1989: done

tamilmani1989:
> /azp run pr-e2e

thanks.may succeed depending on region the cluster was created. New NRP rollout is still going on.

tamilmani1989:
@jackfrancis  can you please approve and merge if there are no comments? @aramase left a comment saying changes look good. Also CNI dual stack version is merged to master now.

jackfrancis:
@tamilmani1989 let's change the note in L99 of this file: `examples/dualstack/README.md` to remove the "supported only with kubenet" part.

tamilmani1989:
@jackfrancis  done

jackfrancis:
@tamilmani1989 we are seeing regressions in cluster configurations after this merged. I'll revert and we can triage what's going on w/ a follow-up PR

tamilmani1989:
@jackfrancis what regressions are happening? Let me try to test it.

PR #2864, opened 2020-03-09
feat: Windows containerd + azure-cni

ksubrmnn:
**Reason for Change**:

Once completed, this will install ContainerD and configure Kubernetes to use it on Windows with Azure CNI

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
/hold do-not-merge


parts/k8s/windowscontainerdfunc.ps1
```diff
@@ -27,7 +27,11 @@ function Install-Containerd
 {
     Param(
       [Parameter(Mandatory = $true)][string]
```
marosset: can you 'format document' this file in VScode?
There are some inconsistencies around spacing between the different params in this param block.

pkg/api/vlabs/validate.go
```diff
@@ -1677,9 +1677,6 @@ func (a *Properties) validateContainerRuntime() error {
 
 	// TODO: These validations should be relaxed once ContainerD and CNI plugins are more readily available
 	if containerRuntime == Containerd && a.HasWindows() {
```
marosset: Should we add any validation around k8s version here?
Will containerd work with versions older than 1.18?

ksubrmnn: I'm not sure. Who could I ask?

marosset: I think we'll need to try it out 🤣

marosset:
@ksubrmnn I think you need to remake the generated files after formatting the files with containerd functions

ksubrmnn:
> @ksubrmnn I think you need to remake the generated files after formatting the files with containerd functions

Yep, realized that then got on a call mid make bootstrap. Should be fixed!

PR #2865, opened 2020-03-10
fix: address containerd errata

jackfrancis:
**Reason for Change**:

This PR addresses three observed issues w/ the Azure-curated containerd implementation:

- Ensure that we are using the v2 configuration on agent as well as master nodes
- Ensure that we configure iptables for IP forwarding
- Ensure that we configure (via sysctl) the IPv4 stack for IP forwarding

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -193,6 +193,21 @@ write_files:
 {{end}}
 
 {{if NeedsContainerd}}
+- path: /etc/sysctl.d/11-containerd.conf
+  permissions: "0644"
+  owner: root
+  content: |
+    net.ipv4.ip_forward = 1
+    #EOF
+
+- path: /etc/systemd/system/containerd.service.d/exec_start.conf
+  permissions: "0644"
+  owner: root
+  content: |
+    [Service]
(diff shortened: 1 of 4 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -193,6 +193,21 @@ write_files:
...
+  content: |
+    net.ipv4.ip_forward = 1
+    #EOF
+
+- path: /etc/systemd/system/containerd.service.d/exec_start.conf
+  permissions: "0644"
+  owner: root
+  content: |
+    [Service]
+    ExecStartPre=/sbin/iptables -P FORWARD ACCEPT
```
cpuguy83: WDYT if instead of changing the default forwarding policy if we do something like:

```
-A FORWARD -i azure0 -o azure0 -j ACCEPT
-A FORWARD -i azure0 ! -o azure0 -j ACCEPT
```

This way we only forward packets for azure0.

jackfrancis: Yes, let's do this as a follow-up for both docker and containerd (as far as I can tell this additional targeted hardening is appropriate for any CRI).

Issue #2855, opened 2020-03-06
default master.availabilityProfile to 'AvailabilitySet' and make it optional on Azure Stack.


rjaini:
Make availabilityProfile  optional field and assign value as 'AvailabilitySet'  for Azure Stack.

jackfrancis:
Fixed w/ #2866

PR #2866, opened 2020-03-10
fix: make master availabilityProfile an optional field for Azure Stack

jadarsie:
**Reason for Change**:
PR #2717 made master.availabilityProfile a required field.

**Issue Fixed**:
#2855

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -258,10 +258,11 @@ const (
 	DefaultAzureStackUseInstanceMetadata = false
 	// DefaultAzureStackAcceleratedNetworking set to false as Azure Stack today doesn't support accelerated networking
 	DefaultAzureStackAcceleratedNetworking = false
-	DefaultAzureStackAvailabilityProfile   = AvailabilitySet
-	// DefaultAzureStackFaultDomainCount set to 3 as Azure Stack today has minimum 4 node deployment.
+	// DefaultAzureStackAvailabilityProfile set to AvailabilitySet as VMSS clusters are not suppored on Azure Stack
+	DefaultAzureStackAvailabilityProfile = AvailabilitySet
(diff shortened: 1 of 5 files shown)


pkg/api/const.go
```diff
@@ -256,13 +256,11 @@ const (
 const (
 	// DefaultUseInstanceMetadata set to false as Azure Stack today doesn't support instance metadata service
 	DefaultAzureStackUseInstanceMetadata = false
-
 	// DefaultAzureStackAcceleratedNetworking set to false as Azure Stack today doesn't support accelerated networking
 	DefaultAzureStackAcceleratedNetworking = false
-
+	DefaultAzureStackAvailabilityProfile   = AvailabilitySet
```
rjaini: nit: Good to add a comment for consistency purpose.

pkg/api/defaults.go
```diff
@@ -599,6 +599,9 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade bool) {
 	// set default to VMAS for now
 	if p.MasterProfile.AvailabilityProfile == "" {
 		p.MasterProfile.AvailabilityProfile = AvailabilitySet
+		if p.IsAzureStackCloud() {
```
rjaini: Do we need this ?. As line 601 is already setting to AvailabilitySet.

rjaini: And I dont think so AKSe support VMSS on master ?.

jadarsie: Not needed now because Az and AzS match at this point in time.
It will be handy if Az's default change.

pkg/api/const.go
jackfrancis: yes, please add a comment here to the exported const

Issue #2868, opened 2020-03-10
kube-system pods are not running in AKS-Engine Dual Stack with kubenet


tamilmani1989:
**Describe the bug**

**Steps To Reproduce**

Used kubernetes cluster config from https://raw.githubusercontent.com/Azure/aks-engine/master/examples/dualstack/kubernetes.json and created a dual stack cluster.

**kubectl get pods -n kube-system
No resources found in kube-system namespace.**

**Expected behavior**
kubectl get pods -n kube-system should show some kubernetes service pods in running state

**AKS Engine version**
Pulled latest master and build it

**Kubernetes version**
1.16

**Additional context**
saw this error message in kubelet logs:

kube-controller-manager-k8s-master-94730297-0" is forbidden: no providers available to validate pod request
7c97625fa9d6)"), skipping: failed to "StartContainer" for "kube-controller-manager" with CrashLoopBackOff: "back-off 1m20s restarting failed container=kube-c
net does not have netConfig. This is most likely due to lack of PodCIDR
net does not have netConfig. This is most likely due to lack of PodCIDR
kube-controller-manager-k8s-master-94730297-0" is forbidden: no providers available to validate pod request
7c97625fa9d6)"), skipping: failed to "StartContainer" for "kube-controller-manager" with CrashLoopBackOff: "back-off 1m20s restarting failed container=kube-c
net does not have netConfig. This is most likely due to lack of PodCIDR
net does not have netConfig. This is most likely due to lack of PodCIDR
net does not have netConfig. This is most likely due to lack of PodCIDR
kube-controller-manager-k8s-master-94730297-0" is forbidden: no providers available to validate pod request
7c97625fa9d6)"), skipping: failed to "StartContainer" for "kube-controller-manager" with CrashLoopBackOff: "back-off 1m20s restarting failed container=kube-c
net does not have netConfig. This is most likely due to lack of PodCIDR
net does not have netConfig. This is most likely due to lack of PodCIDR
kube-controller-manager-k8s-master-94730297-0" is forbidden: no providers available to validate pod request
7c97625fa9d6)"), skipping: failed to "StartContainer" for "kube-controller-manager" with CrashLoopBackOff: "back-off 1m20s restarting failed container=kube-c
net does not have netConfig. This is most likely due to lack of PodCIDR

aramase:
Tried to repro using the master branch, but unable to repro it on a new cluster using master branch. Will try and take a look at existing cluster to see what the issue is.

[...]

PR #2882, opened 2020-03-11
feat: allow iptables mode for dualstack 1.18+

aramase:
**Reason for Change**:

- `iptables` proxier support for dualstack has been added in 1.18.0-alpha.2. The PR allows using `iptables` proxy mode in 1.18+ versions.
- Set different cluster subnet for 1.16 dualstack cluster
- Update dualstack docs

**Issue Fixed**:

fixes https://github.com/Azure/aks-engine/issues/2868

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -218,8 +218,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				}
 				// ipv4 and ipv6 subnet for dual stack
 				if cs.Properties.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack") {
-					defaultKubernetesClusterSubnetIPv6 := cs.getDefaultKubernetesClusterSubnetIPv6()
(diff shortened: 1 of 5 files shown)


pkg/api/defaults.go
```diff
@@ -218,7 +218,8 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				}
 				// ipv4 and ipv6 subnet for dual stack
 				if cs.Properties.FeatureFlags.IsFeatureEnabled("EnableIPv6DualStack") {
-					o.KubernetesConfig.ClusterSubnet = strings.Join([]string{DefaultKubernetesClusterSubnet, DefaultKubernetesClusterSubnetIPv6}, ",")
+					defaultKubernetesClusterSubnetIPv6 := cs.getDefaultKubernetesClusterSubnetIPv6()
```
jackfrancis: I think we can just invoke `cs.getDefaultKubernetesClusterSubnetIPv6()` in L222, we don't need the `defaultKubernetesClusterSubnetIPv6` variable

aramase: updated!

PR #2890, opened 2020-03-12
fix: fixing nssm logging in windows CSE

marosset:
**Reason for Change**:

nssm.exe output when captured in powershell contains null characters which mangle the log files.
This PR removes the null characters before they get logged.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
Before change parts of the CSE log would look like
image

After the change the same part of the log looks like
image

marosset:
This was really starting to bug me.

PR #2891, opened 2020-03-12
chore: optimize CSE payload

jackfrancis:
**Reason for Change**:

This is a housekeeping PR to remove unnecessary bytes from CSE rendered payloads, to prevent data limit threshold exceptions.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -14,11 +14,9 @@ systemctlEnableAndStart() {
     RESTART_STATUS=$?
     systemctl status $1 --no-pager -l > /var/log/azure/$1-status.log
     if [ $RESTART_STATUS -ne 0 ]; then
-        echo "$1 could not be started"
         return 1
     fi
     if ! retrycmd_if_failure 120 5 25 systemctl enable $1; then
-        echo "$1 could not be enabled by systemctl"
         return 1
     fi
 }
@@ -261,7 +259,7 @@ configureCNIIPTables() {
     fi
 }
 
-{{if NeedsContainerd}}
+{{- if NeedsContainerd}}
 ensureContainerd() {
     wait_for_file 1200 1 /etc/systemd/system/containerd.service.d/exec_start.conf || exit $ERR_FILE_WATCH_TIMEOUT
     wait_for_file 1200 1 /etc/containerd/config.toml || exit $ERR_FILE_WATCH_TIMEOUT
@@ -269,6 +267,7 @@ ensureContainerd() {
 }
 {{end}}
 
+{{- if IsDockerContainerRuntime}}
 ensureDocker() {
     DOCKER_SERVICE_EXEC_START_FILE=/etc/systemd/system/docker.service.d/exec_start.conf
     wait_for_file 1200 1 $DOCKER_SERVICE_EXEC_START_FILE || exit $ERR_FILE_WATCH_TIMEOUT
@@ -296,14 +295,15 @@ ensureDocker() {
     wait_for_file 1200 1 $DOCKER_MONITOR_SYSTEMD_FILE || exit $ERR_FILE_WATCH_TIMEOUT
     systemctlEnableAndStart docker-monitor.timer || exit $ERR_SYSTEMCTL_START_FAIL
 }
+{{end}}
 
-{{if EnableEncryptionWithExternalKms}}
+{{- if EnableEncryptionWithExternalKms}}
 ensureKMS() {
     systemctlEnableAndStart kms || exit $ERR_SYSTEMCTL_START_FAIL
 }
 {{end}}
 
-{{if IsIPv6Enabled}}
+{{- if IsIPv6Enabled}}
 ensureDHCPv6() {
     wait_for_file 3600 1 {{GetDHCPv6ServiceCSEScriptFilepath}} || exit $ERR_FILE_WATCH_TIMEOUT
     wait_for_file 3600 1 {{GetDHCPv6ConfigCSEScriptFilepath}} || exit $ERR_FILE_WATCH_TIMEOUT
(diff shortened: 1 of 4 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -14,11 +14,9 @@ systemctlEnableAndStart() {
     RESTART_STATUS=$?
     systemctl status $1 --no-pager -l > /var/log/azure/$1-status.log
     if [ $RESTART_STATUS -ne 0 ]; then
-        echo "$1 could not be started"
         return 1
     fi
     if ! retrycmd_if_failure 120 5 25 systemctl enable $1; then
-        echo "$1 could not be enabled by systemctl"
```
mboersma: Without this logging, we won't know which `if` clause returned the 1 value. Should we use a different return value, or keep the logging?

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -380,14 +380,14 @@ createKubeManifestDir() {
...
-    chown $ADMINUSER:$ADMINUSER $KUBECONFIGDIR
-    chown $ADMINUSER:$ADMINUSER $KUBECONFIGFILE
-    chmod 700 $KUBECONFIGDIR
-    chmod 600 $KUBECONFIGFILE
+    local DIR=/home/$ADMINUSER/.kube
+    local FILE=$DIR/config
+    mkdir -p $DIR
+    touch $FILE
+    chown $ADMINUSER:$ADMINUSER $DIR
+    chown $ADMINUSER:$ADMINUSER $FILE
```
mboersma: This could be one command:
`chown $ADMINUSER:$ADMINUSER $DIR $FILE`

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -121,33 +112,30 @@ installGPUDrivers() {
...
         ;;
     esac
 
     local PACKAGES="make gcc dkms"
     wait_for_apt_locks
     retrycmd_if_failure 30 5 3600 apt-get -y install $PACKAGES  || exit $ERR_SGX_DRIVERS_INSTALL_TIMEOUT
 
-    local SGX_DRIVER
-    SGX_DRIVER=$(basename $SGX_DRIVER_URL)
+    local DRIVER=$(basename $URL)
```
mboersma: These could be one statement BTW:
`local DRIVER=$(basename $URL) OE_DIR=/opt/azure/containers/oe`

Doesn't save that many characters, but TIL more about `local`. 📖

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -121,33 +112,31 @@ installGPUDrivers() {
...
     esac
 
     local PACKAGES="make gcc dkms"
     wait_for_apt_locks
     retrycmd_if_failure 30 5 3600 apt-get -y install $PACKAGES  || exit $ERR_SGX_DRIVERS_INSTALL_TIMEOUT
 
-    local SGX_DRIVER
-    SGX_DRIVER=$(basename $SGX_DRIVER_URL)
+    local DRIVER
+    DRIVER=$(basename $URL)
```
mboersma: Could be one line:
`local DRIVER=$(basename URL)`

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: Granted having human messages is nicer, but I think the entire story is here:

```+ systemctlEnableAndStart rpcbind
+ systemctl_restart 100 5 30 rpcbind
+ retries=100
+ wait_sleep=5
+ timeout=30
+ svcname=rpcbind
++ seq 1 100
+ for i in '$(seq 1 $retries)'
+ timeout 30 systemctl daemon-reload
+ timeout 30 systemctl restart rpcbind
+ break
+ RESTART_STATUS=0
+ systemctl status rpcbind --no-pager -l
+ '[' 0 -ne 0 ']'
+ retrycmd_if_failure 120 5 25 systemctl enable rpcbind
+ retries=120
+ wait_sleep=5
+ timeout=25
+ shift
+ shift
+ shift
++ seq 1 120
+ for i in '$(seq 1 $retries)'
+ timeout 25 systemctl enable rpcbind
Synchronizing state of rpcbind.service with SysV init with /lib/systemd/systemd-sysv-install...
Executing /lib/systemd/systemd-sysv-install enable rpcbind
+ break
+ echo Executed '"systemctl' enable 'rpcbind"' 1 times
```

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: Oh I am so on this one.

parts/k8s/cloud-init/artifacts/cse_config.sh
mboersma: I guess `set -x` has us covered. 👍

parts/k8s/cloud-init/artifacts/cse_install.sh
jackfrancis: it can't be! shellcheck complains that there are two evaluations that happen as a result, and you mask one (the `$()` shell out)

jackfrancis: I mean, it can be, but we'd have to add a shellcheck comment which would be more bytes.

Also, shellcheck is right.

mboersma: D'oh!

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -157,13 +138,10 @@ configureKubeletServerCert() {
...
     KUBELET_PRIVATE_KEY_PATH="/etc/kubernetes/certs/client.key"
     touch "${KUBELET_PRIVATE_KEY_PATH}"
-    chmod 0600 "${KUBELET_PRIVATE_KEY_PATH}"
-    chown root:root "${KUBELET_PRIVATE_KEY_PATH}"
-
     APISERVER_PUBLIC_KEY_PATH="/etc/kubernetes/certs/apiserver.crt"
     touch "${APISERVER_PUBLIC_KEY_PATH}"
-    chmod 0644 "${APISERVER_PUBLIC_KEY_PATH}"
-    chown root:root "${APISERVER_PUBLIC_KEY_PATH}"
+    chmod 0644 "${KUBELET_PRIVATE_KEY_PATH}" "${APISERVER_PUBLIC_KEY_PATH}"
```
mboersma: This changes ownership of `${KUBELET_PRIVATE_KEY_PATH}` from 0600 to 0644, which may be problematic.

jackfrancis: yep, that's why you're here :)

PR #2815, opened 2020-03-02
fix: use ntp instead of systemd-timesyncd in 18.04

jackfrancis:
**Reason for Change**:

This PR changes the 18.04-LTS bootstrap implementation (both VHD and non-VHD) to use the `ntp` package instead of the built-in `systemd-timesyncd`. This is due to a hypothesis that this time synchronization configuration is more reliable.

**Issue Fixed**:

Fixes #1306

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2893, opened 2020-03-12
test: add test config to use gen2 images for agent pools

### LINKED ISSUE #2846: Test Gen2 Ubuntu 18.04 and Windows base images
**Describe the request**
Users need the capabilities of the [DC2 series](https://docs.microsoft.com/en-us/azure/virtual-machines/dcv2-series) VMs which are very large and support confidential compute and AI workloads. The DC2 series only works with  [Gen2 VM images](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/generation-2)

Some users need to manage their own clusters, and thus can't use a managed service, but have workloads that require DC2-series VMs

Test the Gen2 Ubuntu and Windows base images in AKS Engine. Provisioning will become faster and more reliable once there is an [AKS Engine VHD](https://github.com/Azure/aks-engine/issues/2601), but until then it'd be useful to make sure it works with the base images

**Additional context**

marosset:
**Reason for Change**:

Add test config which uses gen2 images for both linux and windows agent pools.

**Issue Fixed**:

Fixes #2846 2846

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
I was able to complete a successful deployment with the above config with aks-engine locally but i want to confirm the job can run in jenkins too

marosset:
I pointed a Jenkin's job to my branch and was able to run a successful test pass with INCLUDE_JOB_REGEX set to 'v1.17/gen2-vhds'

@jackfrancis  how do we want to include this test config into our existing matrix - should we treat it the same as the no-outbound tests (where we don't run this config as part of the normal version matrices?) and just create another periodic job that runs pretty in-frequently?

jackfrancis:
@marosset rather than ship a new api model in the `test/e2e/test_cluster_configs` directory, you could incorporate the E2E changes in this PR and we could then set up custom pipelines that use 18.04-gen2-backed node pools across the set of cluster configs:

https://github.com/Azure/aks-engine/pull/2815

See:

- https://github.com/Azure/aks-engine/pull/2815/files#diff-843229aab4c6145c7e5a11eb844015a5R104
- https://github.com/Azure/aks-engine/pull/2815/files#diff-ca2c1db78f7cfcd48ee1894b1e89f8d5R57
- https://github.com/Azure/aks-engine/pull/2815/files#diff-ca2c1db78f7cfcd48ee1894b1e89f8d5R312

I think this PR is going to merge faster than #2815 :) I'll rebase #2815 on top of master once this lands.

examples/kubernetes-non-vhd-distros.json
```diff
@@ -0,0 +1,64 @@
...
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes"
+    },
+    "masterProfile": {
+      "count": 1,
+      "dnsPrefix": "",
+      "vmSize": "Standard_D2_v3",
+      "distro": "ubuntu"
+    },
+    "agentPoolProfiles": [
```
jackfrancis: @marosset FYI, moved the gen2 E2E coverage to this api model, which is used by the PR E2E spec. So now all PRs will validate these 4 non-VHD flavors.

examples/kubernetes-non-vhd-distros.json
```diff
@@ -0,0 +1,64 @@
...
+        "distro": "ubuntu-18.04"
+      },
+      {
+        "name": "pool1804gen2",
+        "count": 1,
+        "vmSize": "Standard_D2s_v3",
+        "distro": "ubuntu-18.04-gen2"
+      },
+      {
+        "name": "agentwin",
```
marosset: nit: can you add gen2 to the name?

pkg/api/vlabs/const.go
```diff
@@ -30,6 +30,7 @@ const (
 const (
 	Ubuntu            Distro = "ubuntu"
 	Ubuntu1804        Distro = "ubuntu-18.04"
```
marosset: It bugs me that we have to define these consts twice...

marosset:
Other than a few nits looks good to me.
I cannot approve because I was the original author of the PR :)

pkg/api/vlabs/const.go
jackfrancis: Yes it's called a brittle dependency hierarchy :)

PR #2900, opened 2020-03-13
[WIP] feat: allow upgrade of single node pools

jadarsie:
**Reason for Change**:
Allow granular `upgrade` operations through a new `--node-pool` flag.

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests


cmd/upgrade.go
```diff
@@ -192,6 +200,19 @@ func (uc *upgradeCmd) loadCluster() error {
 }
 
 func (uc *upgradeCmd) validateTargetVersion() error {
+	cpVersion := uc.containerService.Properties.OrchestratorProfile.OrchestratorVersion
+	if uc.isSinglePoolUpgrade() {
+		// Make sure node pool is upgraded to the control plane version
+		// Is it OK for --force to be able to bypass this validation?
```
jadarsie: Feedback would be appreciated here

jackfrancis: I think you should be able to upgrade a discrete node pool to any version. It shouldn't need to match the version in another node pool, or the version running on the control plane.

Thoughts?

jadarsie: The assumption is that most people would like to upgrade to the CP version. This check will prevent potential mistakes. If you must upgrade to a different version, then it can be done by using `--force`.

Another question is if `--force` should allow agent pool upgrades to a version higher than the CP version.

jackfrancis: `--force` is only meant to bypass version validation (aks-engine support map) and enable downgrade (i.e., allow the target version to be <= current version)

I think we need to enable node pools to be upgraded independent of control plane. There will be reasons why it makes sense to run a newer/different version on a node compared to control plane. That should be an easy gesture, not an override gesture, which would suggest that that's dangerous, which it's not.

jadarsie: sounds good, testing with that changes

PR #2903, opened 2020-03-13
chore: update azure-policy addon versions

sozercan:
**Reason for Change**:

- Updates Azure Policy addon (gatekeeper + policy) to latest versions

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


examples/addons/azure-policy/azure-policy.json
```diff
@@ -43,4 +43,4 @@
       "secret": ""
     }
   }
-}
+}
```
ritazh: nit: newline

parts/k8s/addons/azure-policy-deployment.yaml
```diff
@@ -77,261 +83,243 @@ spec:
...
-        controller-tools.k8s.io: "1.0"
+        gatekeeper.sh/system: "yes"
     spec:
       containers:
       - args:
-        - --auditInterval={{ContainerConfig "auditInterval"}}
-        - --constraintViolationsLimit={{ContainerConfig "constraintViolationsLimit"}}
         - --port=8443
+        - --logtostderr
+        - --exempt-namespace=kube-system
```
ritazh: maybe add another issue to track `disable-cert-rotation` and cert generation to work with the addonmanager

sozercan: do we want this to be a part of `rotate-certs`?
https://github.com/Azure/aks-engine/blob/master/docs/topics/certificaterotation.md

examples/addons/azure-policy/README.md
```diff
@@ -93,12 +93,25 @@ kubectl get pods -n kube-system
...
+### Option 1
+
+- Update `apimodel.json` and set `azure-policy` addon `enabled` to `false`.
+- Run `aks-engine upgrade` with the updated `apimodel.json`
+- Run `kubectl delete deployments.apps -n kube-system gatekeeper-controller-manager azure-policy` to remove running deployments and pods
+
+### Option 2
+
+- SSH into AKS-Engine Kubernetes master node
+- Run `kubectl delete -f /etc/kubernetes/addons/azure-policy-deployment.yaml && sudo rm /etc/kubernetes/addons/azure-policy-deployment.yaml` to remove all resources created by addon
```
jackfrancis: In this option we'll also have to manually delete the deployments and apps resources, as removing the spec from the kube-addon-manager origin won't induce that deletion automatically. (Unless I'm wrong about that and you proved that the deletion does happen automatically?)

parts/k8s/addons/azure-policy-deployment.yaml
```diff
@@ -77,261 +83,243 @@ spec:
...
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
   labels:
     control-plane: controller-manager
-    controller-tools.k8s.io: "1.0"
+    gatekeeper.sh/system: "yes"
     kubernetes.io/cluster-service: "true"
     addonmanager.kubernetes.io/mode: Reconcile
```
jackfrancis: Is there ever a reason why a user would manually maintain the addon outside of aks-engine cluster operations? Like, increase replica count? If so, we should set `addonmanager.kubernetes.io/mode` to `EnsureExists`, so that those user-initiated customizations won't be overridden by kube-addon-manager.

parts/k8s/addons/azure-policy-deployment.yaml
jackfrancis: This sounds separate from rotating cluster certificates, which is what the above doc is meant to describe.

examples/addons/azure-policy/README.md
sozercan: yes, `kubectl delete -f /etc/kubernetes/addons/azure-policy-deployment.yaml` would delete the deployed resources in this case (there is `&&` in there)

parts/k8s/addons/azure-policy-deployment.yaml
sozercan: we were thinking (as a follow-up) that aks-engine can generate gatekeeper cert and then users can manually rotate using `rotate-certs`. This is similar to behavior in AKS: https://docs.microsoft.com/en-us/azure/aks/certificate-rotation

parts/k8s/addons/azure-policy-deployment.yaml
sozercan: in this case, we want to curate the deployment arguments as much as possible to prevent any potential issues

examples/addons/azure-policy/README.md
```diff
@@ -93,12 +93,25 @@ kubectl get pods -n kube-system
...
+| image          | no       | image                             | "mcr.microsoft.com/oss/open-policy-agent/gatekeepe:v3.1.0-beta.7" |
 | cpuRequests    | no       | cpu requests for the container    | "100m"                                               |
 | memoryRequests | no       | memory requests for the container | "256Mi"                                              |
-| cpuLimits      | no       | cpu limits for the container      | "100m"                                               |
+| cpuLimits      | no       | cpu limits for the container      | "1000m"                                              |
 | memoryLimits   | no       | memory limits for the container   | "512Mi"                                              |
 
+## Disable Azure Policy Add-on
+
+### Option 1
```
RamyasreeChakka: @sozercan You mentioned user has to clean up constraint templates and constraints also on uninstall. Should we add that step too?

sozercan: afaik you dont need to cleanup gk created resource, when user reenables, it should apply on top of existing ones if they chose not to cleanup manually

parts/k8s/addons/azure-policy-deployment.yaml
sozercan: added #2927

PR #2904, opened 2020-03-14
test: create functions in network policy specific file and let tests call it.

csfmomo:
**Reason for Change**:
Functions created in network policy specific file for tests to consume. Increase code reusablity.

**Requirements**:

        test: Testing 💚 
- [ ] includes documentation
- [ ] adds unit tests
- [X ] tested upgrade from previous version

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -1499,10 +1499,10 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				eng.HasNetworkPolicy("cilium") || eng.HasNetworkPolicy("antrea") {
 				nsDev, nsProd := "development", "production"
 				By("Creating development namespace")
-				namespaceDev, err := namespace.CreateIfNotExist(nsDev)
+				namespaceDev, err := namespace.CreateNamespaceDeleteIfExist(nsDev)
 				Expect(err).NotTo(HaveOccurred())
 				By("Creating production namespace")
-				namespaceProd, err := namespace.CreateIfNotExist(nsProd)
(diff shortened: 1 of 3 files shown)


test/e2e/kubernetes/namespace/namespace.go
```diff
@@ -41,8 +41,18 @@ func CreateIfNotExist(name string) (*Namespace, error) {
 	n, err := Get(name)
```
jackfrancis: Can we rename this function to `CreateDeleteIfExists` to reflect the new behavior?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -704,6 +704,13 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			By("Ensuring that we have functional DNS resolution from a linux container")
 			validateDNSLinuxName := "validate-dns-linux"
 			validateDNSLinuxNamespace := "default"
+			// Create a kubectl version check to avoid api version mismatch which due to deployment failure in next step
```
jackfrancis: FYI, this is fine as an informational check, but the `k` util is already designed to use the version of `kubectl` that the API returns in a call to the cluster. Is that not working in your experience?

csfmomo: In my local test with out ssh enabled, I found my kubectl version comes from https://github.com/Azure/aks-engine/blob/master/scripts/k#L40
It's usually older than test cluster's version which result in k create * failure. 
I didn't experience auto cluster call and match in local so I add this check. I have to manually override it. However, from pipeline validation, it seems kubectl version can match.

test/e2e/kubernetes/namespace/namespace.go
csfmomo: Sure, let me rename it.

PR #2880, opened 2020-03-11
feat: configurable sysctl.d configuration

jackfrancis:
**Reason for Change**:

This PR will allow user-configurable sysctl.d Linux kernel runtime configuration.

No kernel configuration changes are landing here; all pre-existing aks-engine kernel params tuning has been standardized into the new `sysctldConfig` interface:

```
"kubernetesConfig": {
    "sysctldConfig": {
        "net.ipv4.tcp_keepalive_time": "120",
        "net.ipv4.tcp_keepalive_intvl": "75",
        "net.ipv4.tcp_keepalive_probes": "9"
    }
}
```

(The above is an example of passing custom config.)

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2905, opened 2020-03-16
feat: add UltraSSD support

andyzhangx:
**Reason for Change**:

This PR adds UltraSSD support for both AgentPool and master node, details about [ UltraSSD](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-types#ultra-ssd-preview)

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

cc @kkmsft


jackfrancis:
Thanks @andyzhangx! There appear to be missing implementations for both agentPoolProfile and masterProfile. Here's a recent PR for reference that added a property to both master and agent type definitions:

https://github.com/Azure/aks-engine/pull/2880

We want to make sure we similarly update both configuration objects for this changes as well, on these vectors:

- api and vlabs type definitions
- converter "to"
- converter "from"
- validation, if appropriate
- defaults enforcement
- ARM template composition
- documentation

pkg/api/types.go
```diff
@@ -569,6 +569,7 @@ type MasterProfile struct {
 	AvailabilityZones         []string          `json:"availabilityZones,omitempty"`
 	SinglePlacementGroup      *bool             `json:"singlePlacementGroup,omitempty"`
 	AuditDEnabled             *bool             `json:"auditDEnabled,omitempty"`
+	UltraSSDEnabled           *bool             `json:"ultraSSDEnabled,omitempty"`
```
andyzhangx: @jackfrancis looks like I need to remove this `UltraSSDEnabled` from MasterProfile since master node does not need `UltraSSDEnabled`, and I have already added `UltraSSDEnabled` for AgentPoolProfile.

jackfrancis: We don't think there's any story where UltraSSDEnabled would be useful for a master vm?

andyzhangx: correct, UltraSSDEnabled is only for user's workload on agent node, let me remove that property from master node first.

jackfrancis: To be clear, ultraSSDEnabled is only needed for agent nodes according to the requirements you're developing against, right? There's no reason why it wouldn't work on a master vm, correct?

@ritazh @craiglpeters FYI, it sounds like we'll introduce this functionality on agent nodes first of all. But assuming there's nothing that forbids it from being useful on a master node, we want to be aware that that chunk of work will be in addition (and after) this chunk of work from @andyzhangx.

andyzhangx: @jackfrancis finally I also added this flag to master node, it's only a small fix:
https://github.com/Azure/aks-engine/pull/2905/commits/f4612ab96d11c93aff2c7eb9de49b4b0ed868f07

andyzhangx:
> * api and vlabs type definitions
done
> * converter "to"
done
> * converter "from"
done
> * validation, if appropriate
not necessary since it's a bool value
> * defaults enforcement
default is false which is compatible
> * ARM template composition
what is this? 
> * documentation
done

@jackfrancis PTAL

docs/topics/clusterdefinitions.md
```diff
@@ -725,6 +725,7 @@ A cluster can have 0 to 12 agent pool profiles. Agent Pool Profiles are used for
...
+| ultraSSDEnabled                                            | no                                                             | Enable UltraSSD feature for each node VM. More details about [Ultra disk](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/disks-types#ultra-ssd-preview).                                                                                                                                                                                                                                                                                                                                                                |
```
jackfrancis: Thanks for adding to masterProfile! Could you also add this documentation note under the `masterProfile` section?

pkg/api/converterfromapi.go
```diff
@@ -520,6 +520,7 @@ func convertMasterProfileToVLabs(api *MasterProfile, vlabsProfile *vlabs.MasterP
 	vlabsProfile.SinglePlacementGroup = api.SinglePlacementGroup
 	vlabsProfile.CosmosEtcd = api.CosmosEtcd
 	vlabsProfile.AuditDEnabled = api.AuditDEnabled
+	vlabsProfile.UltraSSDEnabled = api.UltraSSDEnabled
```
jackfrancis: We need the equivalent conversion boilerplate for agentPoolProfiles as well.

pkg/engine/virtualmachines.go
```diff
@@ -541,6 +541,12 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 		}
 	}
 
+	if to.Bool(profile.UltraSSDEnabled) {
```
jackfrancis: Let's add this equivalent foo to the `CreateMasterVM` func so that the master virtualMachine resources have the same `UltraSSDEnabled` configuration

pkg/engine/virtualmachinescalesets.go
```diff
@@ -688,6 +688,12 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		}
 	}
 
+	if to.Bool(profile.UltraSSDEnabled) {
```
jackfrancis: Ditto here, let's add the equivalent of this inside the `CreateMasterVMSS` func

andyzhangx:
@jackfrancis PTAL

jackfrancis:
@yangl900 @xuto2 @palma21 FYI UltraSSD support has landed in AKS Engine w/ functional E2E test coverage

@ritazh @craiglpeters FYI

andyzhangx:
> @yangl900 @xuto2 @palma21 FYI UltraSSD support has landed in AKS Engine w/ functional E2E test coverage
> 
> @ritazh @craiglpeters FYI

thanks for the code review, this feature is required by some customers, next we need to integrate into AKS.

PR #2906, opened 2020-03-16
fix: scale cannot find VM index for the VMAS+VHD+Linux case

jadarsie:
**Reason for Change**:
Scaling fails if the target pool is a VMAS pool of VHD-based linux nodes.
VM `osPublisher` is not enought to distinguish between Linux and Windows.

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -269,7 +269,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					continue
 				}
 
-				if vm.OsProfile != nil && vm.OsProfile.WindowsConfiguration != nil {
+				if sc.agentPool.OSType == api.Windows {
 					_, _, winPoolIndex, index, err = utils.WindowsVMNameParts(vmName)
 				} else {
 					_, _, index, err = utils.K8sLinuxVMNameParts(vmName)
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -269,8 +269,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					continue
 				}
 
-				osPublisher := vm.StorageProfile.ImageReference.Publisher
-				if osPublisher != nil && (strings.EqualFold(*osPublisher, "MicrosoftWindowsServer") || strings.EqualFold(*osPublisher, "microsoft-aks")) {
+				if vm.OsProfile != nil && vm.OsProfile.WindowsConfiguration != nil {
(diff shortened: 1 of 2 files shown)


cmd/scale.go
```diff
@@ -269,8 +269,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 					continue
 				}
 
-				osPublisher := vm.StorageProfile.ImageReference.Publisher
-				if osPublisher != nil && (strings.EqualFold(*osPublisher, "MicrosoftWindowsServer") || strings.EqualFold(*osPublisher, "microsoft-aks")) {
+				if vm.OsProfile != nil && vm.OsProfile.WindowsConfiguration != nil {
```
marosset: would it be safer to use sc.agentPool.OsType here?

marosset:
/lgtm
@mboersma or @jackfrancis can you take a look too? - this logic has been fragile

jackfrancis:
@jadarsie Let's ensure we protect any fixes to VMAS + scale by adding `"SCALE_CLUSTER": true` to this test config:

https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/availabilityset.json#L3

(under the env config object)

marosset:
Lets add a windows node pool to the ^^^ config

test/e2e/test_cluster_configs/availabilityset.json
```diff
@@ -51,6 +53,16 @@
 					],
 					"availabilityProfile": "AvailabilitySet",
 					"vnetSubnetId": "/subscriptions/SUB_ID/resourceGroups/RG_NAME/providers/Microsoft.Network/virtualNetworks/VNET_NAME/subnets/SUBNET_NAME"
+				},
+				{
+					"name": "poolwin",
+					"count": 2,
+					"vmSize": "Standard_D2_v3",
+					"osType": "Windows",
```
marosset: You need a windowsProfile section with at least admin undername and password

			"windowsProfile": {
				"adminUsername": "azureuser",
				"adminPassword": "replacepassword1234$",
				"enableAutomaticUpdates": false,
				"sshEnabled": true
			}
https://github.com/Azure/aks-engine/blob/0e54eceb35b6abde5c5f9ad2ba34826ba42c23e6/test/e2e/test_cluster_configs/everything.json#L128

jadarsie: oh right, always bite me

PR #2908, opened 2020-03-16
fix: apply_network_config: false for 18.04-LTS

jackfrancis:
**Reason for Change**:

Apply `apply_network_config: false` cloud-init configuration to 18.04-LTS vms always, not just in a VHD CI scenario.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -220,7 +223,11 @@ if [[ -n "${MASTER_NODE}" ]]; then
     fi
     time_metric "WriteKubeConfig" writeKubeConfig
     if [[ -z "${COSMOS_URI}" ]]; then
-        time_metric "EnsureEtcd" ensureEtcd
+        if [ "$FULL_INSTALL_REQUIRED" = "true" ]; then
```
CecileRobertMichon: why is this change needed? It breaks the assumption that you can override the etcd version in the VHD by specifying a custom version.

jackfrancis: `ensureEtcd` depends upon eth0 having the correct IP address. In 18.04-LTS + non-VHD eth0 does not have the correct address.

CecileRobertMichon: what about the VHD scenario? with this change we would be skipping ensureEtcd if using VHD

jackfrancis: Yeah that's a bug will fix

parts/k8s/cloud-init/artifacts/label-nodes.service
```diff
@@ -5,4 +5,6 @@ After=kubelet.service
 Restart=always
 RestartSec=60
 ExecStart=/bin/bash /opt/azure/containers/label-nodes.sh
+[Install]
```
jackfrancis: @mboersma FYI, these missing lines were preventing `systemctl enable` from working (and a systemd bug that returned 0 upon failure was hiding this from us)

examples/kubernetes-ubuntu-distros.json
```diff
@@ -16,6 +16,12 @@
         "count": 2,
         "vmSize": "Standard_D2_v3",
         "distro": "ubuntu"
+      },
```
jackfrancis: With these changes we should be able to E2E test ubuntu 18.04-LTS non-VHD

parts/k8s/cloud-init/artifacts/cse_main.sh
CecileRobertMichon: @jackfrancis this wasn't fixed before the PR merged

PR #2910, opened 2020-03-16
fix: Error on preserve metadata failure. don't cordon on upgrades. 

paulgmiller:
**Reason for Change**:

https://msazure.visualstudio.com/CloudNativeCompute/_wiki/wikis/CloudNativeCompute.wiki/47472/Preserving-TaintsLabelsAnnotations-across-Upgrade

In an AKS upgrade (which steals liberally from this code) we had an upgrade leave nodes cordoned because drain when through but copying labels/taints timed out. 
Seems like we could achieve what we want without cordoning at all. 

In addition we decided during a design review to be more aggressive about failing and not doing anything if labels/taints/annotations haven't changed

We'd like to bring up nodes with the right taints/labels/annotations but VMSS makes that hard as there is one model and they can be unique per node.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


paulgmiller:
@jackfrancis @chengliangli0918  let me know if there are others owners for upgrade in aks-engien (is  @JackQuincy  still involved). Also curious if this should hold till after lock down or if you only want to see one of the 3 fixes I'm doiing here (1 failing 2 no-oping if no new labels/taints/annotations, 3 for this copy only drain after don't cordon/uncordon at all)

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -749,31 +743,49 @@ func (ku *Upgrader) copyCustomPropertiesToNewNode(client armhelpers.KubernetesCl
...
+	var cordonDrainTimeout time.Duration
+	if ku.cordonDrainTimeout == nil {
+		cordonDrainTimeout = defaultCordonDrainTimeout
+	} else {
+		cordonDrainTimeout = *ku.cordonDrainTimeout
+	}
+	err := operations.JustDrainNodeWithClient(client, ku.logger, newNodeName, cordonDrainTimeout)
+	if err != nil {
+		ku.logger.Warningf("Error draining agent VM %s. Proceeding with copying node properties. Error: %v", newNodeName, err)
+		//would like to return err but draining is even more likely to fail
```
paulgmiller: anyone want to vote for erroring out here?

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -782,11 +794,13 @@ func (ku *Upgrader) copyCustomNodeProperties(client armhelpers.KubernetesClient,
 		for k, v := range oldNode.Annotations {
 			if _, ok := newNode.Annotations[k]; !ok {
```
paulgmiller: this has diverged from Aks upgrade in two ways. 
1) AKS doesn't copy 
2) AKS always overwrites new with old and just replacing names. It doesn't keep any that happened to already have the same name no new or were only on new but not on old. 

Should we consolidate?

paulgmiller: if we move to what AKS does then we need a differenent way to see if things have changed.

paulgmiller: Aks and aks-engine label/annotation copy have divereged too much so I gave up on this.

pkg/operations/kubernetesupgrade/upgrader.go
paulgmiller: In aks-engine I have a work item to add an agentpool setting on when to fail based on drain.

Issue #2743, opened 2020-02-19
SGX driver installation fails


jackfrancis:
It appears that recent updates to Ubuntu 18.04-LTS have broken the current SGX driver installation:

```
# /opt/azure/containers/oe/sgx_linux_x64_driver_1.12_c110012.bin
Unpacking Intel SGX Driver ... done.
Verifying the integrity of the install package ... done.
Installing Intel SGX Driver ...
/tmp/sgx-driver-fJlmmM /home/azureuser
install -d /opt/intel/sgxdriver/package
install -d /opt/intel/sgxdriver/scripts
cp -r package/* /opt/intel/sgxdriver/package
install scripts/* /opt/intel/sgxdriver/scripts
/home/azureuser

Kernel preparation unnecessary for this kernel.  Skipping...

Building module:
cleaning build area...
'make' sign KDIR=/lib/modules/5.0.0-1031-azure/build...(bad exit status: 2)
ERROR (dkms apport): binary package for sgx: 1.12 not found
Error! Bad return status for module build on kernel: 5.0.0-1031-azure (x86_64)
Consult /var/lib/dkms/sgx/1.12/build/make.log for more information.
```

PR #2914, opened 2020-03-17
fix: parse SGX driver url and compare checksum

Francis-Liu:
**Reason for Change**:
When intel-sgx driver upgrades, old url would become invalid, thus breaking provision process. This change tries to determine what's the latest driver version by paring the checksum file.

**Issue Fixed**:
Prevent future issues as of #2743

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Tested on both Ubuntu 16.04 and 18.04

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -112,31 +112,39 @@ installGPUDrivers() {
 }
 
 installSGXDrivers() {
-    local URL
-    case $UBUNTU_RELEASE in
-    "18.04")
-        URL="https://download.01.org/intel-sgx/latest/dcap-latest/linux/distro/ubuntuServer18.04/sgx_linux_x64_driver_1.21.bin"
-        ;;
-    "16.04")
-        URL="https://download.01.org/intel-sgx/latest/dcap-latest/linux/distro/ubuntuServer16.04/sgx_linux_x64_driver_1.21.bin"
-        ;;
-    "*")
-        exit $ERR_SGX_DRIVERS_INSTALL_TIMEOUT
-        ;;
-    esac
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -112,31 +112,41 @@ installGPUDrivers() {
...
+    sgx_driver_sha256sum_expected="${tmp_array[0]}"
+    sgx_driver_remote_path="${tmp_array[1]}"
+    sgx_driver_url="${sgx_driver_folder_url}/${sgx_driver_remote_path}"
+    sgx_driver=$(basename "$sgx_driver_url")
+
+    retrycmd_if_failure 10 10 120 curl -fsSL -O "${sgx_driver_url}" || exit $ERR_SGX_DRIVERS_INSTALL_TIMEOUT
+    read -ra tmp_array <<< "$(sha256sum ./"$sgx_driver")"
+    sgx_driver_sha256sum_real="${tmp_array[0]}"
+    if [ "$sgx_driver_sha256sum_real" != "$sgx_driver_sha256sum_expected" ]; then
+        # The checksum value is incorrect in download.01.org when the following line of code is written.
```
jackfrancis: Unfortunately we are short for space in the cloud-init that we deliver, so could we remove these comments?

In fact, if we're ignoring a mismatch, what is the purpose of logging this?

Francis-Liu: Sure, I can remove the comments. Out of curiosity, could you elaborate on what short for space means? From what you suggested, it gives me an impression that the cloud-init script hit a limitation of text length.

This ignoring of checksum is temporary due the the checksum file is wrong at the moment. After Intel uploads a correct checksum file, I will submit another PR to return a checksum error.

jackfrancis: Sure thing: Azure has a 64k limit for the customData property which delivers the cloud-init payload. All of the files that aks-engine dispatches to the VM are via cloud-init, and thus they all have to fit under 64k (translates to around 87k w/ gzip+base64-encoding)

PR #2919, opened 2020-03-18
fix: check network from node to k8s api server in CSE

bowang-666:
**Reason for Change**:

Sometimes agent nodes can not reach k8s api server due to custom network settings. It will cause "nodenotready" failures during create/scale/upgrade. This pr will check the connection from node to k8s api and raise error code if can't connect.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Run `make generate` and commit generated code to pass unit test lint.

PR #2854, opened 2020-03-06
feat: install Windows csi proxy during cluster creation

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2930, opened 2020-03-19
feat: installing csi-proxy for windows at node deployment time

marosset:
**Reason for Change**:

Redo of #2854
Add support for installing csi-proxy for windows at node-deloyment time.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


marosset:
/cc @kkmsft @andyzhangx @jackfrancis @mboersma 
Re-doing this PR since it got merged pre-maturely due to bad Tide config.

I also updated the powershell to set kubelet service to depend on csi-proxy-server service if enabled (per KK's feedback on other PR)

pkg/api/vlabs/types.go
```diff
@@ -162,6 +162,8 @@ type CustomNodesDNS struct {
 type WindowsProfile struct {
 	AdminUsername          string            `json:"adminUsername,omitempty"`
 	AdminPassword          string            `json:"adminPassword,omitempty"`
+	CSIProxyURL            string            `json:"csiProxyURL"`
```
jackfrancis: I think we should `omitempty` here. I've also noticed other properties are probably missing this, and filed a backlog issue here:

https://github.com/Azure/aks-engine/issues/2933

marosset: @jackfrancis should I add omitempty only for vlabs/types.go or pkg/api/types.go as well?

jackfrancis: to be consistent and cautious I always duplicate it in both versions, tho AFAIK the vlabs is the only one that matters in terms of user api model impact

marosset: k - will update

marosset: updated

PR #2935, opened 2020-03-19
chore: improve --client-id flag validation message

jadarsie:
**Reason for Change**:
CLI error message should indicate that the `--client-id` format is invalid instead of implying that the value may be missing.

**Requirements**:
- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -372,7 +372,7 @@ func TestValidateAuthArgs(t *testing.T) {
 			expected: errors.New(`--certificate-path and --private-key-path must be specified when --auth-method="client_certificate"`),
 		},
 		{
-			name: "ClientCertificateAuthExpectsCertificatePath",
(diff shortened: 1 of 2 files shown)


cmd/root_test.go
```diff
@@ -289,6 +291,111 @@ func TestGetAzureStackClientWithClientSecret(t *testing.T) {
 	}
 }
 
+func TestValidateAuthArgs(t *testing.T) {
```
jackfrancis: Thanks for adding this validation coverage!

What should we be doing if `AuthMethod` == "" ?

jadarsie: It is validated by `getClient` later on, but I can a validation here as well

https://github.com/Azure/aks-engine/blob/ce337aa0ef13fccb771cf31807b6adffdf519312/cmd/root.go#L214-L219

Issue #2695, opened 2020-02-07
Deprecate keyvault flexvol and switch to csi driver


ritazh:
**Describe the request**

**Explain why AKS Engine needs it**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

ritazh:
Pending csi driver repo move to kubernetes-sigs

craiglpeters:
@aramase the move has happened :)

PR #2936, opened 2020-03-19
feat: add csi-secrets-store addon

aramase:
**Reason for Change**:

- Adds csi-secrets-store addon
- Enabled by default instead of keyvault-flexvolume for 1.16+ clusters
- If both `keyvault-flexvolume` and `csi-secrets-store` addon enabled simultaneously, then will result in an error
```
Error: validating API model after populating values: Both keyvault-flexvolume and csi-secrets-store addons are enabled, only one of these may be enabled on a cluster
```
- If `keyvault-flexvolume` addon enabled explicitly for 1.16 clusters, a warning message will be displayed and `csi-secrets-store` addon will not be enabled.
```
WARN[0000] keyvault-flexvolume add-on will be DEPRECATED in favor of csi-secrets-store addon for 1.16+ 
```

- Disable keyvault-flexvol addon for 1.16+ clusters
- Failure to enable `csi-secrets-store` addon for < 1.16 clusters
```
Error: validating API model after populating values: csi-secrets-store add-on can only be used in 1.16+
```

**Issue Fixed**:

fixes https://github.com/Azure/aks-engine/issues/2695

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


docs/topics/clusterdefinitions.md
```diff
@@ -105,7 +105,7 @@ $ aks-engine get-versions
...
-| [keyvault-flexvolume](../../examples/addons/keyvault-flexvolume/README.md)                                | true                                                                                 | as many as linux agent nodes    | Access secrets, keys, and certs in Azure Key Vault from pods                                                                                                                                                                   |
+| [keyvault-flexvolume](../../examples/addons/keyvault-flexvolume/README.md)                                | true                                                                                 | as many as linux agent nodes    | Access secrets, keys, and certs in Azure Key Vault from pods. This solution will be **DEPRECATED** in favor of csi-secrets-store addon                                                                                                                                                                  |
```
jackfrancis: enabled by default should be false now, right?

parts/k8s/addons/kubernetesmasteraddons-secrets-store-csi-driver-daemonset.yaml
```diff
@@ -0,0 +1,287 @@
+apiVersion: storage.k8s.io/v1beta1
```
jackfrancis: Because we're installing this by default and it's so much larger in terms of its yaml byte payload compared to keyvault-flexvol, we may run into cloud-init data limits. We'll want to test this with other data-heavy configurations (e.g., calico).

If we do find that this touches the limit we can schedule some cloud-init payload optimization refactor work, but that would become a blocker to this landing.

@ritazh FYI

docs/topics/clusterdefinitions.md
ritazh: we prob dont want to just remove it without some kind of deprecation message for awhile right?

pkg/api/addons.go
```diff
@@ -962,6 +1009,13 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 			o.KubernetesConfig.Addons[i].Enabled = to.BoolPtr(true)
 		}
 	}
+
+	// Enable keyvault flexvolume addon during upgrade for 1.16 or greater scenarios, unless explicitly disabled
+	if isUpgrade && common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") && !o.KubernetesConfig.IsAddonDisabled(common.KeyVaultFlexVolumeAddonName) {
```
jackfrancis: you should be able to do an `IsAddonEnabled` instead of the double negative

docs/topics/clusterdefinitions.md
ritazh: maybe output it as a warning message in the cli? wdyt?

ritazh: @jackfrancis how do we support removal of addon features without impacting users now?

jackfrancis: We normally only worry about this for upgrade scenarios, in which case we (unless there's a good reason to statically override) prefer the previous config even if a new cluster at the same version would produce a different default in a similar functional surface area.

And for new clusters, if we change a default between 1.15 and 1.16, we don't customarily warn users, as our position is if the user doesn't provide an explicit dissenting configuration, he/she assents to the platform defaults (obviously this means we have to be good about what we choose as defaults).

I think in this case we should make sure that we don't allow *both* addons to be present simultaneously, right?

And in terms of deprecation, yeah, if keyvault-flexvolume functionally works in versions beyond where we set csi-secrets-store as the default, we should probably still allow it with a log warning message at cluster creation time saying "this addon will be deprecated in a future release" or something like that. I think that matches the Kubernetes pattern as well, for example, in controller-manager pod logs.

ritazh: so sounds like we should:
- upgrade: keep flexvol and output deprecation message (for how many releases?)
- new clusters 1.16+: default to csi driver and output deprecation message for flexvol

WDYT?

jackfrancis: I like it (we only output the deprecation message if the user has explicitly chosen to enable keyvault-flexvol, I assume you mean)

aramase: That sounds good to me!

ritazh: so for upgrading clusters with flexvol, how many releases should we wait until we remove support for it?

pkg/api/addons.go
```diff
@@ -770,6 +772,50 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
...
+				Name:           common.CSINodeDriverRegistrarContainerName,
+				Image:          specConfig.MCRKubernetesImageBase + k8sComponents[common.CSINodeDriverRegistrarContainerName],
+				CPURequests:    "10m",
+				MemoryRequests: "20Mi",
+				CPULimits:      "200m",
+				MemoryLimits:   "200Mi",
+			},
+			{
+				Name:           common.CSISecretsStoreDriverContainerName,
+				Image:          "docker.io/deislabs/secrets-store-csi:v0.0.8",
```
ritazh: we prob need a mcr for this at some point and make sure it's part of the vhd script

pkg/api/addons.go
aramase: thanks, done!

docs/topics/clusterdefinitions.md
aramase: Updated the PR!

- `keyvault-flexvolume` addon will only be enabled by default for <= 1.15
- `csi-secrets-store` addon will only be enabled by default for 1.16+ clusters
- When upgrading from <= 1.15 to 1.16+, 
    - If `keyvault-flexvolume` addon was enabled, it will still be enabled after upgrade.
    - If `csi-secrets-store` addon doesn't exist before upgrade to 1.16+, it will not be explicitly enabled during the upgrade.
- If `keyvault-flexvolume` addon is explicitly enabled for a new 1.16+ cluster, a warning message will be shown and the addon will be enabled.

```
WARN[0000] keyvault-flexvolume add-on will be DEPRECATED in favor of csi-secrets-store addon for 1.16+ 
```

parts/k8s/addons/kubernetesmasteraddons-secrets-store-csi-driver-daemonset.yaml
aramase: Going to test it now with calico enabled.

aramase: @jackfrancis I was able to successfully deploy 1.16 cluster with `csi-secrets-store` and `calico` enabled. Any other heavy config that you are aware of we should test with?

docs/topics/clusterdefinitions.md
```diff
@@ -105,7 +105,7 @@ $ aks-engine get-versions
...
-| [keyvault-flexvolume](../../examples/addons/keyvault-flexvolume/README.md)                                | true                                                                                 | as many as linux agent nodes    | Access secrets, keys, and certs in Azure Key Vault from pods                                                                                                                                                                   |
+| [keyvault-flexvolume](../../examples/addons/keyvault-flexvolume/README.md)                                | true (for <=1.15 clusters)                                                                                 | as many as linux agent nodes    | Access secrets, keys, and certs in Azure Key Vault from pods. This solution will be **DEPRECATED** in favor of csi-secrets-store addon                                                                                                                                                                  |
```
jackfrancis: let's change the default langage to "false (true for <= 1.15 clusters)"

parts/k8s/addons/kubernetesmasteraddons-secrets-store-csi-driver-daemonset.yaml
jackfrancis: Thanks for validating! I'm testing across the 1.17 matrix, will report back if any cluster configs complain about customData limits.

pkg/api/addons.go
```diff
@@ -189,8 +189,10 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 	}
 
 	defaultKeyVaultFlexVolumeAddonsConfig := KubernetesAddon{
-		Name:    common.KeyVaultFlexVolumeAddonName,
-		Enabled: to.BoolPtr(DefaultKeyVaultFlexVolumeAddonEnabled && !cs.Properties.HasCoreOS() && !cs.Properties.IsAzureStackCloud()),
+		Name: common.KeyVaultFlexVolumeAddonName,
+		// key-vault flexvolume solution will be deprecated in favor of secrets-store-csi-driver for 1.15+
```
jackfrancis: s/1.15+/1.16+

pkg/api/addons.go
jackfrancis: Let's add the remaining three container images to `install-dependencies.sh` that aren't there already with this PR.

@jchauncey, do we have an easy path towards setting up an MCR pipeline that will effectively consume `secrets-store-csi` releases at the above deislabs container registry endpoint?

pkg/api/addons.go
```diff
@@ -962,6 +1009,13 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 			o.KubernetesConfig.Addons[i].Enabled = to.BoolPtr(true)
 		}
 	}
+
+	// Enable keyvault flexvolume addon during upgrade for 1.16 or greater scenarios, unless explicitly disabled
```
jackfrancis: I don't think we need any special logic here. We should just be able to use whatever value is already in the api model during upgrade.

pkg/api/addons.go
aramase: The remaining 3 container images are already in `install_dependencies.sh`. All CSI components share these images. For the MCR pipeline there currently isn't one. I'll set it up.

pkg/api/vlabs/validate.go
```diff
@@ -700,6 +700,9 @@ func (a *Properties) validateAddons() error {
 					if a.HasCoreOS() {
 						return errors.New("flexvolume add-ons not currently supported on coreos distro. Please use Ubuntu")
 					}
+					if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.16.0") {
+						log.Warnf("%s add-on will be DEPRECATED in favor of csi-secrets-store addon for 1.16+", addon.Name)
```
ritazh: So if user enables this and default creates secrets csi, then that means they will have both solutions running in the cluster?

pkg/api/vlabs/validate.go
```diff
@@ -700,6 +700,9 @@ func (a *Properties) validateAddons() error {
...
+					}
```
jackfrancis: Should we add validation that fails if *both* keyvault-flexvol and cse-secrets-store addons are enabled at the same time? Or is that actually a sane configuration?

aramase: In terms of functionality, both the solutions offer the same and pod configuration to consume the addon is different.  Having both of them run at the same time will not result in any side effects. If we do decide to add the validation then we need to request the user explicitly disable `csi-secrets-store` addon which is default in 1.16+. Do we have similar experience for any other addon in aks-engine today?

jackfrancis: We don't let users enable both kube-dns and coredn simultaneously, is an example.

ritazh: yes I think validation should fail if both are enabled.

examples/addons/csi-secrets-store/README.md
```diff
@@ -0,0 +1,84 @@
...
+
+[The CSI Secrets Store](https://github.com/kubernetes-sigs/secrets-store-csi-driver) integrates secrets stores with Kubernetes via a [Container Storage Interface (CSI)](https://kubernetes-csi.github.io/docs/) volume. 
+
+With the CSI secrets store and the azure provider installed, developers can access application-specific secrets, keys, and certs stored in Azure Key Vault directly from their pods.
+
+This addon is available for 1.15+ clusters
+
+**Note** This addon is enabled by default for 1.16+ clusters
+
+To enable on 1.15 cluster, add this addon to the apimodel and also set the feature gate `CSIInlineVolume=true` for API server and kubelet config
```
ritazh: Should we remove this whole thing if we only want to support 1.16+ for secrets store csi?

jackfrancis: +1

pkg/api/vlabs/validate.go
aramase: cool! I wanted to make sure we don't introduce a new experience just for this add-on. I'll add that validation.

parts/k8s/addons/kubernetesmasteraddons-secrets-store-csi-driver-daemonset.yaml
jackfrancis: Validation passed, so we don't have to worry about this (for now!)

pkg/api/addons.go
jchauncey: How is that image being built? Can we just add that to the list of csi pipelines we already have?

aramase: Updated the PR with the mcr image for driver

pkg/api/addons.go
```diff
@@ -770,6 +772,58 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
...
+			},
+		},
+	}
+
+	// volume lifecycle mode needs to be explicitly set for 1.16+
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") {
+		defaultSecretsStoreCSIDriverAddonsConfig.Config = map[string]string{"volume-lifecycle-mode": "Ephemeral"}
+	}
+
+	// If we have any explicit keyvault-flexvolume or csi-secrets-store configuration in the addons array
```
jackfrancis: Why would this scenario happen?

Would it make more sense to just combine vlabs validation (fail an api model if both addons are configured) with additional defaults foo, like:

`Enabled: !o.KubernetesConfig.IsAddonEnabled(common.KeyVaultFlexVolumeAddonName) && to.BoolPtr(DefaultSecretStoreCSIDriverAddonEnabled && common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0")),`

aramase: makes sense! Have updated the PR.

PR #2938, opened 2020-03-19
fix: check network to k8s api for vmss

bowang-666:
**Reason for Change**:

We need to also check network connection to k8s api for vmss. Preivous change: https://github.com/Azure/aks-engine/pull/2919

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/virtualmachinescalesets.go
```diff
@@ -720,7 +720,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 		} else {
 			registry = `mcr.microsoft.com 443`
 		}
-		outBoundCmd = `retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; ERR_OUTBOUND_CONN_FAIL=50; retrycmd_if_failure 50 1 3 ` + ncBinary + ` -vz ` + registry + ` || exit $ERR_OUTBOUND_CONN_FAIL;`
+		outBoundCmd = `retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; ERR_OUTBOUND_CONN_FAIL=50; ERR_K8S_API_CONN_FAIL=51; retrycmd_if_failure 50 1 3 ` + ncBinary + ` -vz ` + registry + `; if [ $? -ne 0 ];then exit $ERR_OUTBOUND_CONN_FAIL;fi; retrycmd_if_failure 50 1 3 ` + ncBinary + ` -vz variables('kubernetesAPIServerIP') 443 || exit $ERR_K8S_API_CONN_FAIL;`
```
xuto2: nit: s/ERR_K8S_API_CONN_FAIL/ERR_K8S_API_SERVER_CONN_FAIL/g

jackfrancis:
@xuto2 @yangl900 strictly speaking, we could put this k8s apiserver check at the *end* of CSE instead of the beginning. In other words, there is nothing in the CSE code (let me know if I'm wrong) that won't work until the apiserver is responsive. And so by delaying the start of CSE until we get that response, we may end up delaying the overall cluster operation by 15-20 seconds or so (the time it takes to run CSE in a VHD-backed context).

xuto2:
@jackfrancis  AKS does many things after creating ccp and before adding nodes, including setup dns for api server, deploy non-VMSS/non-VMAS resources in ARM, which take a while. So for most case, by the time of cse execution, our api server is already up and this check is a precondition check that we want to fail earlier if node cannot even talk to the api server.

I'm not sure about AKS-Engine scenarios though. Is it possible master node is not up when agent node is running cse when an AKS-Engine cluster is created?

bowang-666:
@jackfrancis I think for AKS usage it makes more sense to put it in beginning (so we catch error earlier) than in the end. Does the engine has cases that'd want this check to be in the end?

bowang-666:
Actually we discussed and it makes more sense to put it in end, I will make the change.

pkg/engine/virtualmachinescalesets.go
```diff
@@ -721,6 +722,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 			registry = `mcr.microsoft.com 443`
 		}
 		outBoundCmd = `retrycmd_if_failure() { r=$1; w=$2; t=$3; shift && shift && shift; for i in $(seq 1 $r); do timeout $t ${@}; [ $? -eq 0  ] && break || if [ $i -eq $r ]; then return 1; else sleep $w; fi; done }; ERR_OUTBOUND_CONN_FAIL=50; retrycmd_if_failure 50 1 3 ` + ncBinary + ` -vz ` + registry + ` || exit $ERR_OUTBOUND_CONN_FAIL;`
+		outBoundK8sCmd = `;ERR_K8S_API_SERVER_CONN_FAIL=51; retrycmd_if_failure 50 1 3 ` + ncBinary + ` -vz variables('kubernetesAPIServerIP') 443 || exit $ERR_K8S_API_SERVER_CONN_FAIL;`
```
jackfrancis: I think this might be easier to maintain in the CSE script itself. Like, at the bottom of `cse_main.sh` you could add something like:

```
{{IsHostedMaster}}
retrycmd_if_failure 50 1 3 nc  -vz $APISERVER_IP 443 || exit $ERR_K8S_API_SERVER_CONN_FAIL
{{end}}
```

Using the above you'd need to add a `APISERVER_IP` env variable (see under `pkg/engine/armvariables.go`).

Do you have any thoughts @CecileRobertMichon ?

bowang-666: Does that mean we'll be also doing this for master?

jackfrancis: No. `IsHostedMaster` will return false for *all* CSE contexts where there is a master VM (e.g., an AKS Engine scenario)

CecileRobertMichon: +1 on Jack's comment

bowang-666: ok I'm moving it to cse

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -253,6 +253,14 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
 fi
 {{end}}
 
+{{- if IsHostedMaster }}
+NCLIB="nc"
```
jackfrancis: You can get rid of this as we don't have to support coreos. I.e., you can just statically put nc in the command statement, it doesn't need to be in a variable

PR #2941, opened 2020-03-20
chore: return distinct CSE err for GPU drivers runtime configuration

jackfrancis:
**Reason for Change**:

This PR introduces a distinct CSE error code so that we can better track GPU drivers failures during node bootstrapping. The new CSE 86 error will indicate that GPU driver configuration failed during node bootstrapping. This three GPU CSE errors are now:

- 84
  - == the driver processes/daemons did not launch successfully
- 85
  - == the driver prerequisites were not downloaded successfully (this should never happen in a VHD-backed cluster scenario
- 86
  - == N-series-specific driver installation did not complete successfully (these are steps that we must require at runtime during cluster provisioning because they require an N series VM SKU)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@yangl900 @xuto2 @qike-ms FYI

PR #2923, opened 2020-03-18
feat: add support for Kubernetes 1.18.0-rc.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#changelog-since-v1180-beta2

**Issue Fixed**:
Closes #2879

**Requirements**:

- [x] Kubernetes artifacts published by Azure Pipelines
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #2957, opened 2020-03-24
feat: add support for Kubernetes 1.18.0

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.18.md

**Issue Fixed**:
Closes #2923

**Requirements**:

- [ ] Kubernetes artifacts built and pushed by Azure Pipelines
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This will fail until Kubernetes 1.18.0 is actually released ~today~ *tomorrow* and our pipeline does its thing.

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -178,11 +178,12 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.17.2":         false,
 	"1.17.3":         true,
 	"1.17.4":         true,
-	"1.18.0-alpha.1": true,
-	"1.18.0-alpha.2": true,
-	"1.18.0-alpha.3": true,
-	"1.18.0-alpha.5": true,
-	"1.18.0-beta.1":  true,
+	"1.18.0-alpha.1": false,
+	"1.18.0-alpha.2": false,
+	"1.18.0-alpha.3": false,
+	"1.18.0-alpha.5": false,
+	"1.18.0-beta.1":  false,
+	"1.18.0":         true,
 }
 
 // GetDefaultKubernetesVersion returns the default Kubernetes version, that is the latest patch of the default release
--- a/pkg/api/vlabs/validate_test.go
+++ b/pkg/api/vlabs/validate_test.go
@@ -598,7 +598,7 @@ func Test_KubernetesConfig_Validate(t *testing.T) {
 			ProxyMode:     "iptables",
 		}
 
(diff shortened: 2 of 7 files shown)


mboersma:
/hold

Still awaiting release and publication. 🏃

pkg/api/common/versions.go
```diff
@@ -178,11 +178,12 @@ var AllKubernetesSupportedVersions = map[string]bool{
...
-	"1.18.0-alpha.1": true,
-	"1.18.0-alpha.2": true,
-	"1.18.0-alpha.3": true,
-	"1.18.0-alpha.5": true,
-	"1.18.0-beta.1":  true,
+	"1.18.0-alpha.1": false,
+	"1.18.0-alpha.2": false,
+	"1.18.0-alpha.3": false,
+	"1.18.0-alpha.5": false,
+	"1.18.0-beta.1":  false,
```
mboersma: Should we add `false` entries for 1.18.0-beta.2 and 1.18.0-rc.1? AKS Engine never supported either one due to bugs, but we could add them in order to reference the issues.

pkg/api/vlabs/validate_test.go
```diff
@@ -598,7 +598,7 @@ func Test_KubernetesConfig_Validate(t *testing.T) {
 			ProxyMode:     "iptables",
 		}
 
-		if err := c.Validate(k8sVersion, false, true, false); err == nil && !common.IsKubernetesVersionGe(k8sVersion, "1.18.0-alpha.2") {
+		if err := c.Validate(k8sVersion, false, true, false); err == nil && !common.IsKubernetesVersionGe(k8sVersion, "1.18.0") {
```
mboersma: @aramase I assume the specific alpha version is intentional here. We usually remove code references to pre-release versions when the final ships, but if you think it's useful to be specific about which alpha version introduced support, I can revert this.

aramase: @mboersma I used alpha.2 intentionally because this was not available in alpha.1. But we can now switch to using `1.18.0`.

pkg/api/common/versions.go
jackfrancis: Let's do, good call.

mboersma:
Update: the 1.18 release is postponed until tomorrow (Weds. 3/25).

PR #2962, opened 2020-03-25
fix: use upstreamartifacts CDN URL for apmz

mboersma:
**Reason for Change**:
Changes the URL for `apmz` to reference the Azure CDN, for consistency and also to facilitate locking down egress rules. (We reference the upstreamartifacts.azureedge.net URL consistently otherwise, so this looks like an oversight.)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
cc: @palma21

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -219,7 +219,7 @@ downloadAzureCNI() {
 
 ensureAPMZ() {
     local version=$1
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -219,7 +219,7 @@ downloadAzureCNI() {
 
 ensureAPMZ() {
```
CecileRobertMichon: @xuto2 would it make sense to also exclude AKS from this install since it isn't using CSE telemetry?

PR #2967, opened 2020-03-25
feat: adding  windows defender process exclusions

marosset:
Adding Windows defender process exclusions for kubelet.exe and csi-proxy-server.exe to reduce unnecessary cpu load on Windows nodes

**Reason for Change**:

This is an ask from @MikkelHegn from AKS team

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
Given current constraints in Azure, we may not get the 1.15 e2e to pass for a while, so I think we can consider this green enough to merge. 💚

marosset:
Thanks Matt

PR #2982, opened 2020-03-27
feat: add support for Kubernetes 1.19.0-alpha.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#v1190-alpha1

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:
I added 1.19 to the PR mix, but we can drop that commit if need be to keep our cloud footprint minimal.


mboersma:
Had a passing run with 1.19 on the second try, and it has passed e2e for me locally. So I'll discard the 1.19 e2e testing commit now.

PR #2745, opened 2020-02-19
docs: get-logs proposal

jadarsie:
Proposal to discuss features, scope and command interface.

PR #2987, opened 2020-03-30
feat: "aks-engine get-logs" command

jadarsie:
**Reason for Change**:
People new to Kubernetes (or Linux) may find challenging the retrieval of relevant logs to troubleshoot an AKS Engine operation failure.

Hence, it would be convenient to that audience to embedded the log collection inside AKS Engine itself.

**Issue Fixed**:
Related to #2745

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jadarsie:
@jackfrancis @mboersma will work in docs/UTs/e2e if this is going in the right direction, please let me know how does it look

jadarsie:
FYI @ksubrmnn in case you can find reusable snippets
https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/474

docs/topics/get-logs.md
```diff
@@ -0,0 +1,47 @@
...
+
+### SSH Sessions
+
+A valid SSH private key is always required to stablish a SSH session to the cluster Linux nodes. Windows credentials are stored in the apimodel and will be loaded from there.
+
+### Log Collection Scripts
+
+_TODO Specify starting AKSe version._ The [log collection script](/scripts/collect-logs.sh) will be available in Linux nodes that were provisioned using a flavor of the `aks-ubuntu-*` distro. If you chose a different distro for your linux pool or used an older version of AKS Engine for the initial deployment, then you can use your own script by setting [parameter](#Parameters) `--linux-script`.
+
+The default distro for Windows node pools will also include the [log collection script](/scripts/collect-windows-logs.ps1). There is no support to pass your own custom script at this point.
```
xdarsie, line 23: > There is no support to pass your own custom script at this point
@marosset what's your opinion about this?

docs/topics/get-logs.md
```diff
@@ -0,0 +1,47 @@
...
+
+At a high level, it works by stablishing a SSH session into each node, executing a `log collection` script that collects and zips relevant files, and downloading the zip file to your local computer.
+
+### SSH Sessions
+
+A valid SSH private key is always required to stablish a SSH session to the cluster Linux nodes. Windows credentials are stored in the apimodel and will be loaded from there.
+
+### Log Collection Scripts
+
+_TODO Specify starting AKSe version._ The [log collection script](/scripts/collect-logs.sh) will be available in Linux nodes that were provisioned using a flavor of the `aks-ubuntu-*` distro. If you chose a different distro for your linux pool or used an older version of AKS Engine for the initial deployment, then you can use your own script by setting [parameter](#Parameters) `--linux-script`.
```
xdarsie, line 21: > then you can use your own script
Will update this with a disclaimer

docs/topics/get-logs.md
```diff
@@ -0,0 +1,47 @@
...
+
+The `aks-engine get-logs` command can be useful to troubleshoot issues with your cluster. It will produce, collect and download to your workstation a set of files that include node configuration, cluster state and configuration, and provision log files.
+
+At a high level, it works by stablishing a SSH session into each node, executing a `log collection` script that collects and zips relevant files, and downloading the zip file to your local computer.
+
+### SSH Sessions
+
+A valid SSH private key is always required to stablish a SSH session to the cluster Linux nodes. Windows credentials are stored in the apimodel and will be loaded from there.
+
+### Log Collection Scripts
```
xdarsie, line 19: > then you can use your own script

@jackfrancis what's your opinion about allowing custom scripts?

jadarsie, line 19: > then you can use your own script

@jackfrancis what's your opinion about allowing custom scripts?

docs/topics/get-logs.md
jadarsie, line 21: > then you can use your own script

Will update this with a disclaimer that this is at-your-own-risk + expected zip file location

docs/topics/get-logs.md
jadarsie, line 23: > There is no support to pass your own custom script at this point

@marosset what's your opinion about this?

docs/topics/get-logs.md
jackfrancis: Out of scope for this initial spike, I think. Probably don't need to mention it in here until it's actually a thing.

jadarsie: What needs to happen so we can include this? I will very much like to have this option available when I work with CSS so they can collect ad-hoc data.

docs/topics/get-logs.md
marosset: For Windows I think we should encourage updates to collect-windows-logs.ps1 for now but we can consider passing in arbitrary scripts in the future if requested.

pkg/helpers/units.go
```diff
@@ -0,0 +1,29 @@
...
+// Licensed under the MIT license.
+
+package helpers
+
+import (
+	"fmt"
+	"math"
+)
+
+func Bytes(s uint64) string {
```
jackfrancis, line 11: Where is this func being called?

scripts/collect-logs.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+source /opt/azure/containers/provision_source.sh
+
+clusterInfo() {
+    FIRST_MASTER_READY=$(kubectl get nodes | grep k8s-master | grep Ready | sort | head -n 1 | cut -d ' ' -f 1)
+    if [[ "${FIRST_MASTER_READY}" == "${HOSTNAME}" ]]; then
+        retrycmd_if_failure_no_stats 3 5 120 kubectl cluster-info dump --namespace=kube-system --output-directory=${OUTDIR}/cluster-info
```
jackfrancis, line 8: FYI `retrycmd_if_failure_no_stats` has been changed to `retrycmd_no_stats` !

pkg/helpers/units.go
jadarsie: I removed the consumer code and forgot to delete it

docs/topics/get-logs.md
```diff
@@ -0,0 +1,47 @@
...
+
+All the commands in this guide require both the Azure CLI and `aks-engine`. Follow the [quickstart guide](../tutorials/quickstart.md) before continuing.
+
+This guide assumes you already have deployed a cluster using `aks-engine`. For more details on how to do that see [deploy](../tutorials/deploy.md).
+
+## Retrieving Logs
+
+The `aks-engine get-logs` command can be useful to troubleshoot issues with your cluster. It will produce, collect and download to your workstation a set of files that include node configuration, cluster state and configuration, and provision log files.
+
+At a high level, it works by stablishing a SSH session into each node, executing a `log collection` script that collects and zips relevant files, and downloading the zip file to your local computer.
```
jackfrancis, line 13: s/stablishing/establishing

docs/topics/get-logs.md
jackfrancis, line 21: Let's just write the docs to be truthful as of right now. I.e., we provide a sample script at `./scripts/collect-logs.sh`.

At a later date, when we have the script baked into the VHD, we can then add add'l input requirements: if `--linux-script` is omitted, and the distro is not `aks-ubuntu-*`, then fail the command saying "you need to provide a valid `--linux-script` arg with your own script. And we can then update the docs accordingly.

docs/topics/get-logs.md
```diff
@@ -0,0 +1,47 @@
...
+
+### SSH Sessions
+
+A valid SSH private key is always required to stablish a SSH session to the cluster Linux nodes. Windows credentials are stored in the apimodel and will be loaded from there.
```
marosset, line 17: Do you want to add a comment that says SSH needs to be enabled on Windows nodes by setting windowsProfile.sshEnabled = true?

(I opened a PR to default ssh to be enabled for Windows but it went stale after nobody reviewed it  :()

marosset: https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#windowsprofile

test/e2e/test_cluster_configs/base.json
```diff
@@ -2,6 +2,7 @@
 	"env": {
 		"SCALE_CLUSTER": true,
 		"UPGRADE_CLUSTER": true,
+		"GET_CLUSTER_LOGS": true,
```
marosset, line 5: If this works for Windows can we add a windows profile?

jackfrancis: +1

cmd/get_logs.go
```diff
@@ -0,0 +1,352 @@
...
+		return errors.New("--location flag does not match api-model location")
+	}
+
+	lauth, err := helpers.PublicKeyAuth(glc.linuxSSHPrivateKeyPath)
+	if err != nil {
+		return errors.Wrap(err, "creating linux ssh config")
+	}
+	glc.linuxSSHConfig = helpers.SSHClientConfig(glc.cs.Properties.LinuxProfile.AdminUsername, lauth)
+
+	if glc.cs.Properties.WindowsProfile != nil {
```
marosset, line 143: Add a check to make sure ssh is enabled on the windows nodes too

cmd/get_logs.go
```diff
@@ -0,0 +1,352 @@
...
+	stdout, err := session.StdoutPipe()
+	if err != nil {
+		return "", errors.Wrap(err, "opening ssh session stdout pipe")
+	}
+
+	var cmd string
+	if isLinuxNode(node) {
+		cmd = "bash -c \"cat /tmp/logs.zip > /dev/stdout\""
+	} else {
+		cmd = "type %TEMP%"
```
marosset, line 318: What is this trying to do?

jadarsie: Command `type path\to\zip` pipes the zip file content to the SSH session's stdout.

test/e2e/cluster.sh
```diff
@@ -353,3 +353,19 @@ if [ "${SCALE_CLUSTER}" = "true" ]; then
     -e ADD_NODE_POOL_INPUT=${ADD_NODE_POOL_INPUT} \
     ${DEV_IMAGE} make test-kubernetes || exit 1
 fi
+
+if [ "${GET_CLUSTER_LOGS}" = "true" ]; then
```
jackfrancis, line 357: O.K., so actually, to make this easier to test, let's move this condition into the condition in L115

And then add the actual get-logs statement after the RESOURCE_GROUP and REGION vars have been created, but before the random tags foo.

Basically, by doing that, we don't have to set upgrade/scale/addpool to true in order to test get-logs.

test/e2e/cluster.sh
```diff
@@ -112,6 +112,22 @@ docker run --rm \
 -e GINKGO_SKIP="${GINKGO_SKIP}" \
 "${DEV_IMAGE}" make test-kubernetes || exit 1
 
+if [ "${GET_CLUSTER_LOGS}" = "true" ]; then
```
jackfrancis, line 115: So, this if statement needs to be incorporated into the if statement below:

`if [ "${UPGRADE_CLUSTER}" = "true" ] || [ "${SCALE_CLUSTER}" = "true" ] || [ -n "$ADD_NODE_POOL_INPUT" ]; then`

e.g.:

`if [ "${UPGRADE_CLUSTER}" = "true" ] || [ "${SCALE_CLUSTER}" = "true" ] || [ -n "$ADD_NODE_POOL_INPUT" ] || [ "${GET_CLUSTER_LOGS}" = "true" ]; then`

We can't put it above that, because the `$RESOURCE_GROUP` and `$REGION` variables haven't yet been defined.

jadarsie: sorry, my bad, I was rushing

cmd/get_logs.go
```diff
@@ -0,0 +1,358 @@
...
+			}
+			if err := glc.loadAPIModel(); err != nil {
+				return errors.Wrap(err, "loading API model")
+			}
+			return glc.run()
+		},
+	}
+	command.Flags().StringVarP(&glc.location, "location", "l", "", "Azure location where the cluster is deployed (required)")
+	command.Flags().StringVarP(&glc.apiModelPath, "api-model", "m", "", "path to the generated apimodel.json file (required)")
+	command.Flags().StringVar(&glc.apiserverURI, "apiserver", "", "apiserver endpoint (required)")
```
jackfrancis, line 75: Should we change this to `--ssh-endpoint`? Basically, it's not obvious that the Kubernetes apiserver FQDN would also be an ssh listener endpoint. Maybe:

`command.Flags().StringVar(&glc.sshEndpoint, "ssh-endpoint", "", "FQDN, or IP address, of an ssh listener that has access to ssh into all nodes in the cluster")`

jadarsie: I agree, should be more generic. What about `--ssh-bastion` or `--ssh-jumpbox` or similar? 

`endpoint` is probably safer as it is a generic term, but `bastion`/`jumpbox`/`proxy` are terms usually [associated](https://www.google.com/search?q=ssh+proxy+command) to this behavior.

I do not mind either one, pick your favorite.

jackfrancis: How about `--ssh-host`?

jadarsie: perfect

cmd/get_logs.go
```diff
@@ -72,14 +72,14 @@ func newGetLogsCmd() *cobra.Command {
 	}
 	command.Flags().StringVarP(&glc.location, "location", "l", "", "Azure location where the cluster is deployed (required)")
 	command.Flags().StringVarP(&glc.apiModelPath, "api-model", "m", "", "path to the generated apimodel.json file (required)")
-	command.Flags().StringVar(&glc.apiserverURI, "apiserver", "", "apiserver endpoint (required)")
-	command.Flags().StringVar(&glc.linuxSSHPrivateKeyPath, "linux-ssh-private-key", "", "path to a valid private ssh key to access the cluster's Linux nodes (required)")
-	command.Flags().StringVar(&glc.linuxScriptPath, "linux-script", "", "path to the log collection script to execute on the cluster's Linux nodes")
+	command.Flags().StringVar(&glc.sshHostURI, "sshHost", "", "FQDN, or IP address, of an SSH listener that can reach all nodes in the cluster (required)")
```
jackfrancis, line 75: One last nit! (I promise)

Let's make it `--ssh-host` instead of `-sshHost` to be more consistent with other > 1 word concatenation arg expressions.

Thanks for doing this!

PR #2988, opened 2020-03-30
docs: include Basic LB SKU limitations in Azure Stack page

jadarsie:
**Reason for Change**:
Call out limits of using the Basic LB SKU.

**Requirements**:
- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/azure-stack.md
+++ b/docs/topics/azure-stack.md
@@ -178,8 +178,17 @@ If you need to expose more than 5 services, then the recommendation is to route 
 
 Currently, Azure Stack Hub only supports the `Basic` load balancer SKU.
 This SKU [limits](https://docs.microsoft.com/en-us/azure/load-balancer/concepts-limitations#skus)
-the backend pool endpoints to virtual machines in a single availability set or virtual machine scale set.
+the backend pool endpoints to virtual machines in a single availability set (or virtual machine scale set).
 Therefore, all Kubernetes services of type `LoadBalancer` should be deployed on the same agent pool.
+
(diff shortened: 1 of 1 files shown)


docs/topics/azure-stack.md
```diff
@@ -184,19 +170,26 @@ To work around this issue, do not delete `LoadBalancer` services as part of your
...
+The `Basic` load balancer SKU available on Azure Stack Hub limits the number of frontend IPs to 5. That implies that each cluster's agents pool is limited to 5 public IPs.
 
 If you need to expose more than 5 services, then the recommendation is to route traffic to those services using an Ingress controller.
 
+### Single agent pool
+
+Currently, Azure Stack Hub only supports the `Basic` load balancer SKU.
+This SKU [limits](https://docs.microsoft.com/en-us/azure/load-balancer/concepts-limitations#skus)
+the backend pool endpoints to virtual machines in a single availability set or virtual machine scale set.
+Therefore, all Kubernetes services of type `LoadBalancer` should be deployed on the same agent pool.
```
jackfrancis: I think we should clarify here that the user may have to do some discovery (which VMSS is in my LoadBalancer's backend pool?) to identify exactly which pool those workloads need to go to. It seems in practice the "first" VMSS pool is the one that Basic uses, but I'm not sure how "first" is determined, so it's probably safest to instruct the user to (1) create the svc (2) introspect the underlying Azure LB to figure out which VMSS is behind it (3) correlate that VMSS with the nodes in the pool.

There's not really a clean way to have aks-engine "automatically add labels/taints" to a pool that would work w/ the dynamic nature of the cluster (e.g., scale), so it will be hard to concisely document that. But just FYI we may want to mention this as well.

jadarsie: thx for the feedback! I added some pointers that I assume will be enough

PR #2998, opened 2020-03-31
chore: use Standard LoadBalancer as default

### LINKED ISSUE #2952: Use Standard LB by default
We need to first ensure that these issues are addressed:

- https://github.com/Azure/aks-engine/issues/1970
- https://github.com/Azure/aks-engine/issues/2033

And test all known-working cluster configurations against Standard to surface any cluster configurations that don't play well w/ Standard LB.

In favor of changing:

- Standard is the default in AKS
- No Azure SLA for Basic LoadBalancer

In favor of keeping Basic:

- Fewer Azure IaaS resources to maintain (e.g., more LoadBalancer resources when using Standard)
- No Azure IaaS cost

jackfrancis:
**Reason for Change**:

This PR changes default cluster configuration enforcement so that Standard is the LoadBalancer SKU type if the user does not declare a `loadBalancerSku` value in the `kubernetesConfig` configuration object.

`"Basic"` LoadBalancer configuration is no longer appropriate as a default for cluster configurations, as it precludes running LoadBalancer-backed Kubernetes services behind multiple node pool-enabled clusters.

Folks who want to continue using the legacy `"Basic"` LoadBalancer type will have to explicitly declare so in their api model:

```
{
    "apiVersion": "vlabs",
    "properties": {
	"orchestratorProfile": {
			"kubernetesConfig": {
				"loadBalancerSku": "Basic",
...
```

**Issue Fixed**:

Fixes #2952 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/examples/dualstack/kubernetes.json
+++ b/examples/dualstack/kubernetes.json
@@ -8,8 +8,6 @@
             "orchestratorType": "Kubernetes",
             "orchestratorRelease": "1.16",
             "kubernetesConfig": {
-                "loadBalancerSku": "Standard",
-                "excludeMasterFromStandardLB": true,
                 "clusterSubnet": "10.244.0.0/16,fc00::/8",
                 "serviceCidr": "10.0.0.0/16,fd00::/108",
                 "dnsServiceIP": "10.0.0.10",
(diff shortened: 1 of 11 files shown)


jackfrancis:
@craiglpeters @ritazh FYI

PR #3003, opened 2020-04-01
chore: change csi-proxy binary name

ZeroMagic:
**Reason for Change**:

The csi-proxy binary name has been modified in the upstream. After decompression of the latest archive, the path has been changed. 
https://kubernetesartifacts.azureedge.net/csi-proxy/master/binaries/csi-proxy.tar.gz

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/windowscsiproxyfunc.ps1
+++ b/parts/k8s/windowscsiproxyfunc.ps1
@@ -12,24 +12,24 @@ function New-CsiProxyService {
     DownloadFileOverHttp -Url $CsiProxyPackageUrl -DestinationPath $binaryPackage
 
     tar -xzf $binaryPackage -C $tempdir
-    cp "$tempdir\build\server.exe" "$KubeDir\csi-proxy-server.exe"
+    cp "$tempdir\bin\csi-proxy.exe" "$KubeDir\csi-proxy.exe"
 
     del $tempdir -Recurse
 
-    & "$KubeDir\nssm.exe" install csi-proxy-server "$KubeDir\csi-proxy-server.exe" | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server AppDirectory "$KubeDir" | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server AppRestartDekay 5000 | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server Description csi-proxy-server | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server Start SERVICE_DEMAND_START | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server ObjectName LocalSystem | RemoveNulls
-    & "$KubeDir\nssm.exe" set csi-proxy-server Type SERVICE_WIN32_OWN_PROCESS | RemoveNulls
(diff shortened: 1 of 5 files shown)


kkmsft:
Thank you, @ZeroMagic !

Nit: may be we can consider the name used to bring up the service also as csi-proxy.exe instead of csi-proxy.service.exe. @marosset  ? 

LGTM otherwise.

andyzhangx:
> Thank you, @ZeroMagic !
> 
> Nit: may be we can consider the name used to bring up the service also as csi-proxy.exe instead of csi-proxy.service.exe. @marosset ?
> 
> LGTM otherwise.

use same name as `csi-proxy.exe` is better IMO, no need to rename

marosset:
Do we have a new package published using the new file name?
If so can you update https://github.com/Azure/aks-engine/blob/master/docs/topics/csi-proxy-windows.md

parts/k8s/windowscsiproxyfunc.ps1
```diff
@@ -12,7 +12,7 @@ function New-CsiProxyService {
     DownloadFileOverHttp -Url $CsiProxyPackageUrl -DestinationPath $binaryPackage
 
     tar -xzf $binaryPackage -C $tempdir
-    cp "$tempdir\build\server.exe" "$KubeDir\csi-proxy-server.exe"
```
andyzhangx, line 15: We may use following 2 copy commands to make it compatible:
```
cp "$tempdir\build\server.exe" "$KubeDir\csi-proxy.exe"
cp "$tempdir\build\csi-proxy.exe" "$KubeDir\csi-proxy.exe"
```

kkmsft: Since alpha release is the first release for CSI proxy and it has csi-proxy.exe, I don't think we have to be backward compatible here.

andyzhangx: https://kubernetesartifacts.azureedge.net/csi-proxy/master/binaries/csi-proxy.tar.gz is already updated, and has the `csi-proxy.exe`, it already does not work with current aks-engine v0.48.0

andyzhangx: we may not need code level compatibility, @ZeroMagic will make a new csi-proxy.tar.gz  which contains following two binaries as a workaround, then both aks-engine master and v0.48.0 could work:
```
$tempdir\build\server.exe
$tempdir\build\csi-proxy.exe
```
In the future, we won't use master branch csi-proxy binary, we should always use published releases, e.g. https://kubernetesartifacts.azureedge.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz

andyzhangx: cc @chewong current azure disk csi driver on windows won't work due to this issue, @ZeroMagic will prepare a new csi-proxy.tar.gz as a workaround

andyzhangx: https://k8scsi.blob.core.windows.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz works on aks-engine master and v0.48.0

chewong: Thanks for the fix!

PR #3018, opened 2020-04-03
fix: do some clean up before retry to install the gpu driver.

andyliuliming:
**Reason for Change**:

when we doing stress test for the scaling of the GPU nodes.
we found in some cases, we will meet this error:
the module already there, and the retry will not succeed if we do not do some clean up.

so the fix is to do some clean up before the reinstallation.

tests done:
1. ran the scripts snip in the box, and the retry works.
2. did the private change in the RP, and run the stress GPU nodes scaling test, and it passed.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -507,10 +507,10 @@ installGPUDriversRun() {
     fi
     {{- /* we need to append the date to the end of the file because the retry will override the log file */}}
     local log_file_name="/var/log/nvidia-installer-$(date +%s).log"
-    sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent \
-        --kernel-name=${KERNEL_NAME} \
+    sh $GPU_DEST/nvidia-drivers-$GPU_DV -s \
+        -k=${KERNEL_NAME} \
         --log-file-name=${log_file_name} \
-        --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}"
+        -a --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}"
     exit $?
 }
 
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -487,6 +487,25 @@ configAddons() {
 }
 {{end}}
 {{- if HasNSeriesSKU}}
+
+installGPUDriversRun() {
+    # if there's no file under the module folder, means the installation not succeeds
+    # so need to do some cleanup here.
+    NVIDIA_DKMS_DIR="/var/lib/dkms/nvidia/${GPU_DV}"
+    NVIDIA_DKMS_MODULE_DIR="${NVIDIA_DKMS_DIR}/*azure/x86_64/module"
+    if [ -d "${NVIDIA_DKMS_DIR}" ]; then
+        if [ -z "$(ls -A ${NVIDIA_DKMS_MODULE_DIR})" ]; then
+            echo "the dkms folder exists, but the module does not exists, we need to do the clean up first before retry to install."
+            # we do not use the /usr/local/nvidia/bin/nvidia-uninstall directly, because the 
+            # nvidia-uninstall itself may not exists either in this case
+            rm -rf "${NVIDIA_DKMS_DIR}"
+        fi
+    fi
+    
+    sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}"
+    exit $?
+}
+
 configGPUDrivers() {
   {{/* only install the runtime since nvidia-docker2 has a hard dep on docker CE packages. */}}
   {{/* we will manually install nvidia-docker2 */}}
@@ -507,7 +526,9 @@ configGPUDrivers() {
   retrycmd_if_failure 120 5 25 pkill -SIGHUP dockerd || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
   retrycmd_if_failure 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  retrycmd_if_failure 3 1 600 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  export -f installGPUDriversRun
+  export GPU_DEST GPU_DV
+  retrycmd_if_failure 3 1 600 bash -c installGPUDriversRun || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   echo "${GPU_DEST}/lib64" >/etc/ld.so.conf.d/nvidia.conf
   retrycmd_if_failure 120 5 25 ldconfig || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   umount -l /usr/lib/x86_64-linux-gnu
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -493,19 +493,22 @@ installGPUDriversRun() {
     when you upgrade the GPU driver version, please help check whether the retry installation issue is gone,
     if yes please help remove the clean up logic here too */}}
     set -x
-    NVIDIA_DKMS_DIR="/var/lib/dkms/nvidia/${GPU_DV}"
-    NVIDIA_DKMS_MODULE_DIR="${NVIDIA_DKMS_DIR}/*azure/x86_64/module"
+    MODULE_NAME="nvidia"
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -487,6 +487,25 @@ configAddons() {
 }
 {{end}}
 {{- if HasNSeriesSKU}}
+
+installGPUDriversRun() {
```
jackfrancis, line 491: Why do we need to do this sometimes?

andyliuliming: the RCA is not yet very clear, it should be the issue of the driver installer.
when the first try of installation failed, it will leave a dirty directory there, and the retry will not works if there's dirty directory.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -487,6 +487,25 @@ configAddons() {
...
+    # if there's no file under the module folder, means the installation not succeeds
```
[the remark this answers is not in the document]
andyliuliming: thanks, the comments is put into the go template now.

andyliuliming:
@mboersma @jackfrancis could you please help do a review again? some customer are pending on this :)
need to cherry pick to the aks-release branch after this merged. thanks in advance.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -507,7 +534,9 @@ configGPUDrivers() {
   retrycmd_if_failure 120 5 25 pkill -SIGHUP dockerd || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
   retrycmd_if_failure 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  retrycmd_if_failure 3 1 600 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  export -f installGPUDriversRun
+  export GPU_DEST GPU_DV
```
xuto2, line 538: do we need to export this?

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -487,6 +487,33 @@ configAddons() {
...
+        if [ -x "$(command -v dkms)" ]; then
+          dkms remove -m ${MODULE_NAME} -v ${GPU_DV} -k ${KERNEL_NAME}
+        else
+          echo "the dkms tool is not found, so we just remove the directory to have a try"
+          rm -rf "${NVIDIA_DKMS_DIR}"
+        fi
+    fi
+    {{- /* we need to append the date to the end of the file because the retry will override the log file */}}
+    local log_file_name="/var/log/nvidia-installer-$(date +%s).log"
+    sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent \
```
mboersma, line 510: Some of these arguments could use their short forms, specifically `--silent|-s`, `--kernel-name|-k`, and `--accept-license|-a`:

```shell
sh $GPU_DEST/nvidia-drivers-$GPU_DV -s \
        -k=${KERNEL_NAME} \
        --log-file-name=${log_file_name} \
        -a --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}"
exit $?
```

Sorry to nitpick, but we have to be careful to keep CSE content as lean as we can.

parts/k8s/cloud-init/artifacts/cse_config.sh
andyliuliming: yes, this is run as a bash sub-command by the retry entrance,
we need to export it.

parts/k8s/cloud-init/artifacts/cse_config.sh
andyliuliming: updating it now.

andyliuliming: @mboersma updated. thanks.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -487,6 +487,33 @@ configAddons() {
...
+
+installGPUDriversRun() {
+    {{- /* there is no file under the module folder, the installation failed, so clean up the dirty directory
+    when you upgrade the GPU driver version, please help check whether the retry installation issue is gone,
+    if yes please help remove the clean up logic here too */}}
+    set -x
+    MODULE_NAME="nvidia"
+    NVIDIA_DKMS_DIR="/var/lib/dkms/${MODULE_NAME}/${GPU_DV}"
+    KERNEL_NAME=$(uname -r)
+    if [ -d "${NVIDIA_DKMS_DIR}" ]; then
```
jackfrancis, line 499: @andyliuliming @xuto2 Why does the `/var/lib/dkms/nvidia/418.40.04` directory exist sometimes but not all the time?

andyliuliming: it's created by the installer.

and if the first try of the installation failed, then the folder is in a dirty state.

and then when the second try of the installation, the installer will think the folder already there, and then fail fast. so we need to do some clean up.

jackfrancis: > and if the first try of the installation failed, then the folder is in a dirty state.

Where does the failure happen?

parts/k8s/cloud-init/artifacts/cse_config.sh
xuto2: oh, I mis-read it, these are two vars not a single one :)

can we pass them in as parameters to the method? global var will some time cause trouble.

parts/k8s/cloud-init/artifacts/cse_config.sh
andyliuliming: from my observing, the failure happens if the nvidia installer choosed the "wrong" /lib/modules/kernel-name, if we have two kernel headers installed.

and then it will fail.
but it's not an always repro.

andyliuliming: and also it will happen if there's some trancient issues happened when the nvidia installer run.

and then it will keep on failing, the retry will not work if we do not do the clean up.

jackfrancis: > if the nvidia installer choosed the "wrong" /lib/modules/kernel-name

When does the above happen sometimes? What line of code?

andyliuliming: that's the "sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent \
This conversation was marked as resolved by mboersma
        --kernel-name=${KERNEL_NAME} \
        --log-file-name=${log_file_name} \
        --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}""


that's the installer command to install multiple things for nvidia.

the installer will choose the lib/modules/kernel-name, and then do the compilation and so on.

that command may fail due to either choosed the wrong lib-module, (there's some guessing logic in the installer), or due to some trancient issue.

and then it will have some dirty directory.

and when we retry to run the installer, then the installer will think there's already one module installed, but that's only a dirty directory, so the retry will fail again and again until reach the max retry times.

andyliuliming: and if we clean up the dirty directory, the retry will succeed.

jackfrancis: O.K., I see what's going on. This PR is making the execution of `/usr/local/nvidia/nvidia-drivers-418.40.04` retryable and idempotent. Is that correct?

jackfrancis: Thanks for the explanation. I see the `-k` option is only meant to be used when you're targeting a different kernel other than the one you're actually installing onto:

```
  -k KERNEL-NAME, --kernel-name=KERNEL-NAME
      Build and install the NVIDIA kernel module for the
      non-running kernel specified by KERNEL-NAME (KERNEL-NAME
      should be the output of `uname -r` when the target kernel
      is actually running).  This option implies
      '--no-precompiled-interface'.  If the options
      '--kernel-install-path' and '--kernel-source-path' are not
      given, then they will be inferred from KERNEL-NAME; eg:
      '/lib/modules/KERNEL-NAME/kernel/drivers/video/' and
      '/lib/modules/KERNEL-NAME/build/', respectively.
```

Does that mean that we don't trust the installer to choose the underlying runtime kernel that's actually on the VM?

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -487,6 +487,33 @@ configAddons() {
...
+        echo "the dkms module folder exists, we need to do the clean up first before retry to install."
+        if [ -x "$(command -v dkms)" ]; then
+          dkms remove -m ${MODULE_NAME} -v ${GPU_DV} -k ${KERNEL_NAME}
+        else
+          echo "the dkms tool is not found, so we just remove the directory to have a try"
```
jackfrancis, line 504: I don't think we can tolerate dkms not being present on the system. The nvidia modproe service needs it:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/nodecustomdata.yml#L273

If there is evidence of dkms not being present on the system, we want to fix that elsewhere (such as when it is installed via apt-get install)

andyliuliming: the "dkms remove" here is just to do the clean up before we retry the installation.

andyliuliming: and then the nvidia-drivers-$GPU_DV will install that for us.

jackfrancis:
See #3031 

Thank you so much for discovering the source of flakes @andyliuliming!

PR #3022, opened 2020-04-03
fix: In case of api server connection error, still complete script

bowang-666:
**Reason for Change**:

In case of validation err(for now api server connection issues), we still want the script to complete. There are edge cases CSE will run before reimaging, we want to short circuit that.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -262,7 +262,7 @@ if [[ $STS != 0 ]]; then
     if [[ $RES == *"168.63.129.16"*  ]]; then
         VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL"}}
     else
-        VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
+        VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_DNS_LOOKUP_FAIL"}}
     fi
 fi
 retrycmd_if_failure 50 1 3 nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -253,17 +253,19 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
 fi
 {{end}}
 
+VALIDATION_ERR=0
+
 {{- if IsHostedMaster }}
 RES=$(retrycmd_if_failure 20 1 3 nslookup ${API_SERVER_NAME})
 STS=$?
 if [[ $STS != 0 ]]; then
-  if [[ $RES == *"168.63.129.16"* ]]; then
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL"}}
-  else
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
-  fi
+    if [[ $RES == *"168.63.129.16"*  ]]; then
+        VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL"}}
+    else
+        VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
+    fi
 fi
-retrycmd_if_failure 50 1 3 nc -vz ${API_SERVER_NAME} 443 || exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -284,4 +286,6 @@ echo $(date),$(hostname), endcustomscript >>/opt/m
 mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
 ps auxfww >/opt/azure/provision-ps.log &
 
+exit VALIDATION_ERR
```
jackfrancis, line 289: missing `$`

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -253,17 +253,19 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
 fi
 {{end}}
 
+VALIDATION_ERR=0
+
 {{- if IsHostedMaster }}
```
jackfrancis, line 258: I think an easier way to complete the script no matter what is to just move this entire block down after the `ps auxfww >/opt/azure/provision-ps.log &` statement, using the original implementation.

bowang-666: The only thing I'm worried about is reboot. If it happens to reboot, then checking connectivity would fail (or the process is gone?).

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -253,17 +253,19 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
 RES=$(retrycmd_if_failure 20 1 3 nslookup ${API_SERVER_NAME})
 STS=$?
 if [[ $STS != 0 ]]; then
-  if [[ $RES == *"168.63.129.16"* ]]; then
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL"}}
-  else
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
-  fi
+    if [[ $RES == *"168.63.129.16"*  ]]; then
```
paulgmiller, line 262: don't understand this magic ip. what is it's significance? 
ANd shouldn't both of the errors below be an azure dns lookup fail? nslookup is never going to connect to the api server?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -284,4 +286,6 @@ echo $(date),$(hostname), endcustomscript >>/opt/m
 mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
```
paulgmiller, line 286: Not blocking but it's wierd that we're racing reboot in 1 minute here. Shoudl we just write out compelete before rebootrequired if/lese block? we're not checking anything there anyways. Guessing there's a reason though I'm just not aware of. But you could move writing out provision.compete to before validation since we're not reallly checking output of anything in the rebootrequiredblock.

paulgmiller: ah you and jack started this conversation up above. good.

parts/k8s/cloud-init/artifacts/cse_main.sh
paulgmiller: think you could leave the reboot block and move the rest up. And even change it to reboot now.

parts/k8s/cloud-init/artifacts/cse_main.sh
bowang-666: the ip is Azure dns ip https://docs.microsoft.com/en-us/azure/virtual-network/virtual-networks-name-resolution-for-vms-and-role-instances#considerations
The main point is to differentiate:
1) if it's reaching out to azure dns but failed it should be an internal server error (there's an issue with private clusters dns right now)
2) if it's custom dns server but not configured correctly, this should be a client error which customer should take actions.

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: O.K. now I understand why we can't move the check down below (because we can't be doing anything serially after a possible shutdown -r command).

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -253,17 +253,19 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
...
+VALIDATION_ERR=0
+
 {{- if IsHostedMaster }}
 RES=$(retrycmd_if_failure 20 1 3 nslookup ${API_SERVER_NAME})
 STS=$?
 if [[ $STS != 0 ]]; then
-  if [[ $RES == *"168.63.129.16"* ]]; then
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL"}}
-  else
-    exit {{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
```
bowang-666, line 263: actually found this error code was somehow wrong

parts/k8s/cloud-init/artifacts/cse_main.sh
paulgmiller: these would be great to have as comments int he error messages... I guess I should just memorize the azure dns ip :)

parts/k8s/cloud-init/artifacts/cse_main.sh
paulgmiller: yeah it's not actualyl the api service connection it's your non azure dns that can't find the api server ip.

PR #3023, opened 2020-04-03
feat: generic cluster-init component

jackfrancis:
**Reason for Change**:

This PR introduces a `"cluster-init"` component as a user-configurable "component" only. The purpose of this is to expose a simple interface for passing base64-encoded yaml data that will work as data to `kubectl apply -f`. E.g.:

```json
"kubernetesConfig": {
    "components": [
        {
            "name": "cluster-init",
            "enabled" : true,
            "data": "YXBpVmVyc2lvbjogdjEKa2luZDogUG9kCm1ldGFkYXRhOgogIG5hbWU6IGFrcy1lbmdpbmUtY2x1c3Rlci1pbml0CnNwZWM6CiAgY29udGFpbmVyczoKICAtIG5hbWU6IGFrcy1lbmdpbmUtY2x1c3Rlci1pbml0CiAgICBpbWFnZTogYnVzeWJveDoxLjMxLjEKICAgIGFyZ3M6CiAgICAtIC9iaW4vc2gKICAgIC0gLWMKICAgIC0gd2hpbGUgdHJ1ZTsgZG8gc2xlZXAgNjAwOyBkb25lCiAgbm9kZVNlbGVjdG9yOgogICAgYmV0YS5rdWJlcm5ldGVzLmlvL29zOiBsaW51eAo="
        }
    ]
}
```

The Kubernetes delivery mechanism is via cloud-init paving a file at `/opt/azure/containers/cluster-init.yaml`, and then CSE executing `kubectl apply- f` against it.

When the `"cluster-init"` component is enabled, the CSE operation will treat the data as a required pre-condition for cluster creation success, and will return exit code `180` if `kubectl apply -f` doesn't return a zero exit code.

// TODO:
- [x] UT
- [x] docs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


Michael-Sinz:
This is exactly what we need.  We had already hacked that into the arm template after creating the template via aks-engine.  This removes that from our list of arm template manipulations.

This also allow our cluster bootstrap to then run as part of the cluster creation process.  We are yet another step closer to the "deploy whole service via arm" with this.

docs/topics/clusterdefinitions.md
```diff
@@ -248,6 +248,7 @@ The reason for the unsightly base64-encoded input type is to optimize delivery p
...
+| cluster-init | false | n/a | This is a "backdoor" interface for delivering manual Kubernetes resource configuration that is appropriate for tight coupling to the ARM deployment specification. For example, if you are using aks-engine in an automated pipeline and you can scale up a cluster more quickly by loading a Kubernetes specification immediately after bootstrapping the control plane, you may use this interface by base64 encoding a yaml representation of that spec. The base64-encoded data string will be decoded at cluster creation time into a yaml file on the first master VM at the `/opt/azure/containers/cluster-init.yaml` filepath, and then loaded into the cluster via `kubectl apply -f`. See[here](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply) for more documentation on how to appropriate specify your source yaml for `kubectl apply` before base64 encoding it. |
```
jackfrancis, line 251: @Michael-Sinz How does this documentation snippet accord with your understanding of this feature?

Michael-Sinz: Looks good - I would mention that this allows us to then produce the arm template via aks-engine that when deployed, deploys a fully configured cluster *with* our full application stack.

Since the yaml is limited in size/scope, it is used to run the rest of our application deployment process as a job in the cluster.  (Maybe not what others want but it is what we do given the size and complexity of what we install, it would not be possible in a yaml that would fit into the arm template.)

docs/topics/clusterdefinitions.md
```diff
@@ -248,6 +248,7 @@ The reason for the unsightly base64-encoded input type is to optimize delivery p
...
+| cluster-init | false | n/a | This is a "backdoor" interface for delivering manual Kubernetes resource configuration that is appropriate for tight coupling to the ARM deployment specification. For example, if you are using aks-engine in an automated pipeline and you can scale up a cluster more quickly by loading a Kubernetes specification immediately after bootstrapping the control plane, you may use this interface by base64 encoding a yaml representation of that spec. The base64-encoded data string will be decoded at cluster creation time into a yaml file on the first master VM at the `/opt/azure/containers/cluster-init.yaml` filepath, and then loaded into the cluster via `kubectl apply -f`. See[here](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply) for more documentation on how to appropriate specify your source yaml for `kubectl apply` before base64 encoding it. Note: the `cluster-init` spec will be ignored during `aks-engine upgrade` operations, nor will it be re-invoked following any node pool scale operations. |
```
mboersma, line 251: I think this could be shortened. Also it should be "base64-encoded".

| cluster-init | false | n/a | This is an interface to deliver Kubernetes resource configuration that may be tightly coupled to an ARM deployment. For example, if you use AKS Engine in an automated pipeline and you can scale up a cluster quickly by loading a Kubernetes specification immediately after bootstrapping the control plane, give the base64-encoded YAML representation of that spec to `cluster-init`. That data is decoded at cluster creation time to the path `/opt/azure/containers/cluster-init.yaml` on the first master VM, and then loaded into the cluster via `kubectl apply -f`. See[`kubectl apply`](https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#apply) for more documentation on how to specify the source YAML. Note: the `cluster-init` spec is ignored during `aks-engine upgrade` and `aks-engine scale`operations.

Issue #2837, opened 2020-03-04
Install nvidia-smi on N-Series node to export gpu metrics


delulu:
**Describe the request**

Install nvidia-smi on N-Series node automatically in aks cluster initialization.

**Explain why AKS Engine needs it**

NVIDIA provides a tool to export GPU metrics into Prometheus, this tool is leveraging driver utility (nvidia-smi) to get gpu metrics and it's a prerequisite that this utility is installed on gpu node.

Also this driver utility seems the only way to get gpu metrics, and there should be a direct way to get gpu metrics beside other metrics (cpu/memory) on gpu node.

**Describe the solution you'd like**

Could you please install nvidia-smi along with GPU driver in aks cluster initialization?

**Describe alternatives you've considered**

**Additional context**
```
PS > .\kubectl.exe --kubeconfig k8s-hc -n prometheus logs prometheus-node-exporter-2t572 nvidia-dcgm-exporter
Starting NVIDIA host engine...
Collecting metrics at /run/prometheus/dcgm.prom every 1000ms...
/usr/local/bin/dcgm-exporter: line 53: nvidia-smi: command not found
```

nvidia-smi: https://developer.nvidia.com/nvidia-system-management-interface
metrics tool: https://github.com/NVIDIA/gpu-monitoring-tools

delulu:
@CecileRobertMichon, could you take a look? or let me know if you need any more information.

CecileRobertMichon:
@delulu added to backlog for triage by the team
/cc @delulu

jackfrancis:
The driver installation you want is probably best delivered as a daemonset. If you can provide a working daemonset spec we can help w/ integration into the existing nvidia addon that is installed in a GPU-enabled VM SKU-backed node.

delulu:
@jackfrancis thanks for your advice. but if it's deployed as a daemonset, this driver utility (CLI utility) will be provided in one serving container hosted on gpu node, and there's no direct access from other pods running on the same node. 

also usually the driver utility is installed along with driver, there should be a way to get metrics of gpu device on gpu node directly.

delulu:
@jackfrancis with a further check, I found you already install this nvidia-smi in in aks cluster initialization.
https://github.com/Azure/aks-engine/blob/3e6b593c7d5e2878be2e536b62a596ad7e530ed7/parts/k8s/cloud-init/artifacts/cse_config.sh#L477

[...]

PR #3029, opened 2020-04-06
fix: make sure nvidia utility and compute bins can be mounted into co…

delulu:
**Reason for Change**:

Make sure nvidia utility and compute bins can be mounted into container as documented in [nvidia-container-runtime readme](https://github.com/NVIDIA/nvidia-container-runtime/blob/2c07b861b81afd4af683a0c1e71a250f1e2e3344/README.md#nvidia_driver_capabilities).

**Issue Fixed**:

NVIDIA utility/compute bins are not mounted into containers when they're required in container env variable as `NVIDIA_DRIVER_CAPABILITIES=compute,utility`

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -519,11 +519,12 @@ configGPUDrivers() {
   retrycmd 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   export -f installNvidiaDrivers
   retrycmd 3 1 600 bash -c installNvidiaDrivers || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  mv ${GPU_DEST}/bin/* /usr/bin
   echo "${GPU_DEST}/lib64" >/etc/ld.so.conf.d/nvidia.conf
   retrycmd 120 5 25 ldconfig || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   umount -l /usr/lib/x86_64-linux-gnu
   retrycmd 120 5 25 nvidia-modprobe -u -c0 || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
-  retrycmd 120 5 25 $GPU_DEST/bin/nvidia-smi || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  retrycmd 120 5 25 nvidia-smi || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   retrycmd 120 5 25 ldconfig || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
 }
 ensureGPUDrivers() {
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -40261,11 +40261,12 @@ configGPUDrivers() {
   retrycmd 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   export -f installNvidiaDrivers
   retrycmd 3 1 600 bash -c installNvidiaDrivers || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  mv ${GPU_DEST}/bin/* /usr/bin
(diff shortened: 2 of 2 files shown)


delulu:
more details can be found from https://github.com/Azure/aks-engine/issues/2837

mboersma:
@delulu could you run `make generate` and then commit and push the code that creates? You can make sure your PR will pass this check by running `make test-style test` locally.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -509,10 +509,11 @@ configGPUDrivers() {
   retrycmd_if_failure 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   retrycmd_if_failure 3 1 600 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   echo "${GPU_DEST}/lib64" >/etc/ld.so.conf.d/nvidia.conf
+  cp "${GPU_DEST}/bin" /usr/bin
```
jackfrancis, line 512: What is special about `/usr/bin` as opposed to `/usr/local/nvidia/bin`?

delulu: `nvidia-container-runtime` and `nvidia-container-cli` bins are located in `/usr/bin`.

It seems it will only search nvidia utility/compute bins in the same folder and then mounted them accordingly into container. Because I've tried to add `/usr/local/nvidia` into system path, and it does not work as expected.

I also opened an issue https://github.com/NVIDIA/nvidia-docker/issues/1226, but there's no response yet. 

So this's a quick fix for this issue.

delulu:
> @delulu could you run `make generate` and then commit and push the code that creates? You can make sure your PR will pass this check by running `make test-style test` locally.

I tried, but there's some unexpected issue as shown below, maybe some more settings are required to setup local box. So could you help run it at your side, thx!

```sh
delzh@MININT-DELU:/mnt/e/git/aks-engine$ make test-style test

==> Running Go linter <==
golangci-lint has version 1.23.7 built from 96964db on 2020-02-28T12:07:44Z
==> Running shell linter <==
ShellCheck - shell script analysis tool
version: 0.3.7
license: GNU Affero General Public License, version 3
website: http://www.shellcheck.net

In ./parts/k8s/cloud-init/artifacts/cse_customcloud.sh line 63:
      jq --arg K8S_CLIENT_CERT_PATH ${K8S_CLIENT_CERT_PATH} '. + {aadClientCertPath:($K8S_CLIENT_CERT_PATH)}' |
                                                            ^-- SC2016: Expressions don't expand in single quotes, use double quotes for that.
...
Makefile:79: recipe for target 'validate-shell' failed
make: *** [validate-shell] Error 1
```

mboersma:
@delulu no problem, I did `make generate` and pushed a commit.

I think the `make test-style` failure was because we assume `shellcheck` 0.7.0 but you have version 0.3.7. You can do `make dev` to run commands from our toolbox container, which has that version, but we don't have a way to enforce tool versions on workstations in general.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -509,10 +509,11 @@ configGPUDrivers() {
   retrycmd_if_failure 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   retrycmd_if_failure 3 1 600 sh $GPU_DEST/nvidia-drivers-$GPU_DV --silent --accept-license --no-drm --dkms --utility-prefix="${GPU_DEST}" --opengl-prefix="${GPU_DEST}" || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   echo "${GPU_DEST}/lib64" >/etc/ld.so.conf.d/nvidia.conf
+  cp "${GPU_DEST}/bin/*" /usr/bin
```
jackfrancis, line 512: This seems to be creating a second copy of the file, should we create a symlink at `/usr/bin/nvidia-smi` to `/usr/local/nvidia/bin/nvidia-smi` instead?

delulu: Actually we can ` mv "${GPU_DEST}/bin/*" /usr/bin ` if you has this concern, since utility libraries have been added to system library path with `echo "${GPU_DEST}/lib64" >/etc/ld.so.conf.d/nvidia.conf`.

It will work with a symlink, but there're also other utility bins and compute bins in `${GPU_DEST}/bin/*`, and I would like to keep the change small to not introduce any potential issue.

jackfrancis: mv works for me, thanks!

delulu:
updated per comments, please have a further check.

jackfrancis:
@yangl900 @andyliuliming FYI

PR #3033, opened 2020-04-06
fix: In case of api server connection error, still complete script (#…

bowang-666:
…3022)

**Reason for Change**:

cherry pick: complete script in case of api server conn error

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -223,17 +223,19 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
 fi
 {{end}}
 
+VALIDATION_ERR=0
+
 {{- if IsHostedMaster }}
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -254,4 +256,6 @@ echo $(date),$(hostname), endcustomscript>>/opt/m
 mkdir -p /opt/azure/containers && touch /opt/azure/containers/provision.complete
 ps auxfww > /opt/azure/provision-ps.log &
 
+exit $VALIDATION_ERR
```
xuto2, line 259: this may create confusion to us since L254 we print out success msg

xuto2: maybe update L254 ?

PR #3034, opened 2020-04-06
chore: reduce CSE by optimizing var/func names

jackfrancis:
**Reason for Change**:

This PR introduces a few targeted optimizations to CSE script with the goal of reducing the payload size by a bit.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-CC_SERVICE_IN_TMP=/opt/azure/containers/cc-proxy.service.in
```
jackfrancis, line 3: These are cruft from (now deprecated) clear-containers container runtime option.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -51,17 +51,6 @@ retrycmd_if_failure() {
   done
   echo Executed \"$@\" $i times
 }
-retrycmd_if_failure_no_stats() {
```
jackfrancis, line 54: It's not clear to me why we still have this option to not report retries in stdout. I think this may be leftover from a time when apt operations were using this func, and stdout/stderr requirements were weird. Dunno.

In any event I think we should just use one, and I think the retry echo statements are valuable for quickly debugging CSE output.

jackfrancis: nevermind it's because apt

PR #3041, opened 2020-04-08
feat: EncryptionAtHost support

andyzhangx:
**Reason for Change**:

This PR is actually a place holder, it's mainly for AKS to support  EncryptionAtHost(Enable Encryption at Host support for VM/VMSS) (JEDI requirement) which is going to be in private preview late this month.

EncryptionAtHost functionality actually does not work by this PR since it depends on
 - go sdk ready and upgrade to supported go sdk version
 - EncryptionAtHost feature on public regions ready
 - subs whitelist

Will make it work when the above 3 requirements are fullfilled.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


jackfrancis:
Thanks @andyzhangx, I've put a hold label on the PR. Let us know when we can consider reviewing/merging the PR.

PR #3043, opened 2020-04-08
fix: the os disk size restriction is changed in azure.

andyliuliming:
**Reason for Change**:

**Issue Fixed**:

the os disk size restriction is changed to 2048 now in Azure.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/vlabs/types.go
+++ b/pkg/api/vlabs/types.go
@@ -413,7 +413,7 @@ type MasterProfile struct {
 	DNSPrefix                 string            `json:"dnsPrefix" validate:"required"`
 	SubjectAltNames           []string          `json:"subjectAltNames"`
 	VMSize                    string            `json:"vmSize" validate:"required"`
-	OSDiskSizeGB              int               `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1023"`
+	OSDiskSizeGB              int               `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1024"`
 	VnetSubnetID              string            `json:"vnetSubnetID,omitempty"`
 	VnetCidr                  string            `json:"vnetCidr,omitempty"`
 	AgentVnetSubnetID         string            `json:"agentVnetSubnetID,omitempty"`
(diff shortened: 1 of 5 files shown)


pkg/api/vlabs/types.go
```diff
@@ -413,7 +413,7 @@ type MasterProfile struct {
 	DNSPrefix                 string            `json:"dnsPrefix" validate:"required"`
 	SubjectAltNames           []string          `json:"subjectAltNames"`
 	VMSize                    string            `json:"vmSize" validate:"required"`
-	OSDiskSizeGB              int               `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1023"`
+	OSDiskSizeGB              int               `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1024"`
```
mboersma, line 416: Shouldn't this also be 2048?

andyliuliming: oh, yes.

andyliuliming: fixed. nice catch.

pkg/api/vlabs/types.go
```diff
@@ -486,7 +486,7 @@ type AgentPoolProfile struct {
 	Name                                string               `json:"name" validate:"required"`
 	Count                               int                  `json:"count" validate:"required,min=1,max=100"`
 	VMSize                              string               `json:"vmSize" validate:"required"`
-	OSDiskSizeGB                        int                  `json:"osDiskSizeGB,omitempty" validate:"min=0,max=1023"`
+	OSDiskSizeGB                        int                  `json:"osDiskSizeGB,omitempty" validate:"min=0,max=2048"`
```
jackfrancis, line 489: Should we update the `DiskSizesGB` property spec as well?

andyliuliming: from this document:
looks like it's 32T now.

will change it.

mboersma: >  Each data disk has a maximum capacity of 32,767 gibibytes (GiB).

PR #3045, opened 2020-04-08
feat: add support for Kubernetes 1.18.1

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#changelog-since-v1180

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This includes a commit to make CI test with 1.18.1, but that could be dropped before merging.

Pipeline build is still in progress, so this won't pass testing yet.

### DIFF
--- a/.pipelines/pr-e2e.yaml
+++ b/.pipelines/pr-e2e.yaml
@@ -60,8 +60,8 @@ jobs:
 
 - template: e2e-job-template.yaml
   parameters:
-    name: 'k8s_1_17_release_e2e'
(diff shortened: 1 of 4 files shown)


mboersma:
Do we want to make 1.18 the default now for PR tests, or should I drop that commit?

jackfrancis:
@mboersma I like this PR as-is :)

PR #3024, opened 2020-04-03
chore: update cluster-autoscaler versions

mboersma:
**Reason for Change**:
Many Azure-specific fixes arrive with these new cluster-autoscaler patch releases.

See https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.18.1
and https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.17.2
and https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.16.5
and https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.15.6
and https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.14.8

**Issue Fixed**:
~~Blocked on #3046~~

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3046, opened 2020-04-08
feat: default to MCR for Kubernetes images

mboersma:
**Reason for Change**:
Switches the default image base for Kubernetes components from k8s.gcr.io to mcr.microsoft.com.

This change was coming, but the switching of base URLs for new cluster-autoscaler releases has made it a priority.

**Issue Fixed**:
Refs #3024

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
For more background on what nudged AKS Engine to make this change, see [Kubernetes vanity domain flip](https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/Vanity-Domain-Flip.md).

PR #3048, opened 2020-04-09
style: mutex is not necessary now that rand uses local object

paulgmiller:
Mutex came in from this comment when the funciton was using rand.seed.
https://github.com/Azure/aks-engine/commit/833fe44015ff6328fa0cfa1fc70f9cd420ffde6e
But it could hav been removed after this commit when we move to using a local source and can't bleed across thread.
Here's where they moved to a local object https://github.com/Azure/aks-engine/commit/a6007e7b7d47dfaf8c2663e24ba592beb05fc9bd

Since this isn't a perf sensitive code path there's nothing gained except a few less lines of code so feel free to reject if it's not worth it. 

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
@paulgmiller our fearless leader @jackfrancis ran some other tests on this which failed, but didn't have time to follow up. Sorry to let this dangle so long. Do you think you could rebase it from current master so we can do fresh testing?

paulgmiller:
Pushed a merge of master No conflicts.... Let me know if there's anything else I can do.

mboersma:
This has passed CI twice (the "upload artifact" error is something unrelated we need to find a solution for).

jackfrancis:
@mboersma did you test upgrade?

mboersma:
@jackfrancis I have tested upgrade with this branch locally, from 1.18 -> 1.19.0-beta.2 and 1.15 -> 1.16. Haven't seen any problems, but I suppose the kind of bug this could potentially introduce wouldn't be deterministic. (Having said that, I think @paulgmiller is correct and this is a safe change.)

jackfrancis:
I ran tests yesterday as well and things looked good.

PR #3052, opened 2020-04-09
feat: Allow specifying Distro and ImageRef at the same time to support air-gapped cloud

sukhong:
**Reason for Change**:
To support a scenario in air-gapped cloud where you need to manually upload an "official" distro image and refer it as ImageRef but STILL be treated as an official distro in aks-engine.

This change is required for treating these manually uploaded "official" distro images exactly same as official Distros VHDs in aks-engine flow, especially passing "{{- if HasVHDDistroNodes}}" variant checks in cse_main.sh to be treated as a "golden image":

https://github.com/Azure/aks-engine/blob/51bde0dbbec2bf90a18ad3fda5bdf3b127c94c3a/parts/k8s/cloud-init/artifacts/cse_main.sh#L71

Also, the semantic makes sense. When you specify Distro and ImageRef at the same time, Distro value is only used as a "hint" in various places. It still outputs ImageRef as an os image in the ARM template.

I believe this change is very safe because it does not affect existing behavior at all (all existing test cases pass).

**Requirements**:
- [ ] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -444,22 +444,24 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			}
 
 			// Distro assignment for masterProfile
-			if cs.Properties.MasterProfile.Distro == "" && cs.Properties.MasterProfile.ImageRef == nil {
-				if cs.Properties.OrchestratorProfile.IsKubernetes() && cs.Properties.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage == "" {
-					cs.Properties.MasterProfile.Distro = AKSUbuntu1604
-				} else {
+			if cs.Properties.MasterProfile.ImageRef == nil {
+				if cs.Properties.MasterProfile.Distro == "" {
+					if cs.Properties.OrchestratorProfile.IsKubernetes() && cs.Properties.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage == "" {
+						cs.Properties.MasterProfile.Distro = AKSUbuntu1604
+					} else {
+						cs.Properties.MasterProfile.Distro = Ubuntu
+					}
+				} else if cs.Properties.OrchestratorProfile.IsKubernetes() && (isUpgrade || isScale) {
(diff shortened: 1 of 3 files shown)


pkg/api/defaults.go
```diff
@@ -444,23 +444,25 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			}
 
 			// Distro assignment for masterProfile
-			if cs.Properties.MasterProfile.Distro == "" && cs.Properties.MasterProfile.ImageRef == nil {
-				if cs.Properties.OrchestratorProfile.IsKubernetes() && cs.Properties.OrchestratorProfile.KubernetesConfig.CustomHyperkubeImage == "" {
-					cs.Properties.MasterProfile.Distro = AKSUbuntu1604
-				} else {
-					cs.Properties.MasterProfile.Distro = Ubuntu
+			if cs.Properties.MasterProfile.ImageRef == nil {
```
sukhong, line 447: I believe previous if-then block was a bug (fall-through to else if wrongly), and this is a correct, desired logic (i.e. only overwrite default Distro when ImageRef is not specified).

sukhong:
@jackfrancis Please review this PR when you have a chance. I also thought about creating separate variables such as IsOfficialVHD or DistroHint. But there are so many places in which it depends on original Distro value (i.e. especially unit test cases such as IsUbuntu(), IsUbuntu1604(), etc...), and this seemed a simpler and more straightforward solution.

jackfrancis:
@CecileRobertMichon FYI, this PR will allow imageRef + distro properties to co-exist in the api model. Because the ARM template composition prefers imageRef over distro, this will work:

- if an imageRef is present, then the VMs are built with it
- which means that the value of distro becomes a higher order configuration declaration: "assume I am a VHD of this type" for CSE and other things that make those assumptions
- so as long as the underlying user-provided image implements the VHD things (e.g., touches the `/opt/azure/containers/provision.complete` file) that the various components that build upon the VHD depend upon, this should be sane

PR #3064, opened 2020-04-13
feat: EncryptionAtHost support(cherry-pick to branch aks-release-v0.47.0-1)

andyzhangx:
**Reason for Change**:

Ported from PR: https://github.com/Azure/aks-engine/pull/3041

This PR is actually a place holder, it's mainly for AKS to support  EncryptionAtHost(Enable Encryption at Host support for VM/VMSS) (JEDI requirement) which is going to be in private preview late this month.

EncryptionAtHost functionality actually does not work by this PR since it depends on
 - go sdk ready and upgrade to supported go sdk version
 - EncryptionAtHost feature on public regions ready
 - subs whitelist

Will make it work when the above 3 requirements are fullfilled.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

/assign @xuto2
cc @jluk

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -529,6 +529,7 @@ type MasterProfile struct {
 	AvailabilityZones         []string          `json:"availabilityZones,omitempty"`
 	SinglePlacementGroup      *bool             `json:"singlePlacementGroup,omitempty"`
 	AuditDEnabled             *bool             `json:"auditDEnabled,omitempty"`
+	EncryptionAtHost          *bool             `json:"encryptionAtHost,omitempty"`
 	CustomVMTags              map[string]string `json:"customVMTags,omitempty"`
 	// Master LB public endpoint/FQDN with port
 	// The format will be FQDN:2376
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -529,6 +529,7 @@ type MasterProfile struct {
 	AvailabilityZones         []string          `json:"availabilityZones,omitempty"`
 	SinglePlacementGroup      *bool             `json:"singlePlacementGroup,omitempty"`
 	AuditDEnabled             *bool             `json:"auditDEnabled,omitempty"`
+	EncryptionAtHost          *bool             `json:"encryptionAtHost,omitempty"`
```
jluk, line 532: Up for discussion pending PRD and design review how this API shapes up

andyzhangx:
Per the discussion in the meeting, we don’t need EncryptionAtHost field any more(same behavior as Accelerated Networking), let me close that PR first, thanks.

andyzhangx:
I was wrong, we still need this field in aks-engine agent pool

PR #3065, opened 2020-04-13
fix: retry more aggressively when install the gpu deb package.

andyliuliming:
when we testing the gpu package installation.
even we have wait_for_apt_locks, the racing can still happen. especially when the unattended upgrade is turned on, and the unattended upgrading may takes 10 min to finish the installation, and we use the 1 second just to try to get the lock more aggressively.


jackfrancis:
Hi @andyliuliming. Did you see this PR?

https://github.com/Azure/aks-engine/pull/1681

The above is meant to disable apt unattended upgrade operations during CSE. Would it make sense to do some tests using a build of aks-engine that includes the above PR (e.g., v0.49.0) and observe if in fact unattended upgrades are still occurring?

PR #3067, opened 2020-04-13
feat: coredns-autoscaler

### LINKED ISSUE #3050: Add coreDNS autoscaling
**Describe the request**
Autoscale the coreDNS pods for availability and scale reasons.

When the cluster size gets on the larger side (~100 nodes, 5k-6k pods) the single coreDNS pod gets CPU bound and dns requests start timing out or becoming a large source of latency.

We fixed this by creating an autoscaling setup for the coredns service using the cluster-proportional-autoscaler package in the kubernetes domain (see yaml below).  This is the solution that runs on our AWS clusters as well, because we use KOPS to deploy them.  This is not the only solution, just what we have done.

```
apiVersion: v1
kind: ServiceAccount
metadata:
  name: coredns-autoscaler
  namespace: kube-system
  labels:
    k8s-addon: coredns.addons.k8s.io

---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    k8s-addon: coredns.addons.k8s.io
  name: coredns-autoscaler
rules:
  - apiGroups: [""]
    resources: ["nodes"]
    verbs: ["list","watch"]
  - apiGroups: [""]
    resources: ["replicationcontrollers/scale"]
    verbs: ["get", "update"]
  - apiGroups: ["extensions", "apps"]
    resources: ["deployments/scale", "replicasets/scale"]
    verbs: ["get", "update"]
# Remove the configmaps rule once below issue is fixed:
# https://github.com/kubernetes-incubator/cluster-proportional-autoscaler/issues/16
  - apiGroups: [""]
    resources: ["configmaps"]
    verbs: ["get", "create"]

---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  labels:
    k8s-addon: coredns.addons.k8s.io
  name: coredns-autoscaler
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: coredns-autoscaler
subjects:
- kind: ServiceAccount
  name: coredns-autoscaler
  namespace: kube-system

---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: coredns-autoscaler
  namespace: kube-system
  labels:
    k8s-app: coredns-autoscaler
spec:
  selector:
    matchLabels:
      k8s-app: coredns-autoscaler
  template:
    metadata:
      labels:
        k8s-app: coredns-autoscaler
    spec:
      containers:
      - name: autoscaler
        image: k8s.gcr.io/cluster-proportional-autoscaler-amd64:1.6.0
        resources:
          requests:
            cpu: 20m
            memory: 10Mi
        command:
        - /cluster-proportional-autoscaler
        - --namespace=kube-system
        - --configmap=coredns-autoscaler
        - --target=Deployment/coredns
        - --default-params={"linear":{"coresPerReplica":256,"nodesPerReplica":16,"min":1}}
        - --logtostderr=true
        - --v=2
      serviceAccount: coredns-autoscaler
      serviceAccountName: coredns-autoscaler
```

jackfrancis:
**Reason for Change**:

This PR introduces a coredns-tuned cluster-proportional-autoscaler pod to the coredns addon.

We also deprecate the never fully baked `dns-autoscaler` addon. Instead of taking the approach of having a discrete addon for this, we are delivering the dns pod autoscaler functionality inside the coredns addon spec itself.

**Issue Fixed**:

Fixes #3050 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/addons/coredns.yaml
```diff
@@ -5,14 +5,14 @@ metadata:
   namespace: kube-system
   labels:
       kubernetes.io/cluster-service: "true"
-      addonmanager.kubernetes.io/mode: Reconcile
+      addonmanager.kubernetes.io/mode: EnsureExists
```
jackfrancis, line 8: In order to better anticipate user-configuration of coredns, this PR will update the addon mode to EnsureExists

vhd/packer/install-dependencies.sh
```diff
@@ -501,6 +501,15 @@ for CSI_SECRETS_STORE_DRIVER_VERSION in ${CSI_SECRETS_STORE_DRIVER_VERSIONS}; do
   echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
+CLUSTER_PROPORTIONAL_AUTOSCALER_VERSIONS="
+1.7.1
+"
+for CLUSTER_PROPORTIONAL_AUTOSCALER_VERSION in ${CLUSTER_PROPORTIONAL_AUTOSCALER_VERSIONS}; do
+  CONTAINER_IMAGE="k8s.gcr.io/cluster-proportional-autoscaler-amd64:${CLUSTER_PROPORTIONAL_AUTOSCALER_VERSION}"
```
jackfrancis, line 508: @jchauncey @sozercan Are we interested in hosting MCR images for the cluster-proportional-autoscaler project?

pkg/api/addons.go
```diff
@@ -680,6 +666,16 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		},
 	}
 
+	if !cs.Properties.IsAzureStackCloud() {
```
mboersma, line 669: Not on Azure Stack?

pkg/api/addons_test.go
```diff
@@ -3749,7 +3768,6 @@ func TestSetAddonsConfig(t *testing.T) {
 				common.IPMASQAgentAddonName,
 				common.AzureCNINetworkMonitorAddonName,
 				common.AzureNetworkPolicyAddonName,
-				common.DNSAutoscalerAddonName,
```
mboersma, line 3752: Do we need to add `common.CoreDNSAutoscalerName` to this list?

jackfrancis: Good question. No, that's the name of the _container_, not the _addon_. This change evolves the existing coredns addon. The above UT is looking for addons enabled by default (no change to that set from this PR).

pkg/api/addons.go
jackfrancis: Right. Because we don't have MCR images of cluster-proportional-autoscaler (that I know of), I think Azure Stack is more complicated.

cc @jadarsie

vhd/packer/install-dependencies.sh
sozercan: we have `mcr.microsoft.com/oss/kubernetes/autoscaler/cluster-proportional-autoscaler` in MCR.

jchauncey: We have the following versions in mcr - 

```
"0.7.0",
"0.7.1",
"0.7.2",
"0.8.0",
"0.8.2",
"0.8.3",
"0.8.4",
"0.8.5",
"1.0.0",
"1.1.0",
"1.1.1",
"1.1.2",
"1.1.2-r2",
"1.1.2-r2_v0.0.5",
"1.2.0",
"1.2.0_v0.0.5",
"1.3.0",
"1.3.0_v0.0.5",
"1.4.0",
"1.5.0",
"1.6.0",
"1.7.0",
"1.7.1",
"1.7.1-hotfix.20200403",
```

jackfrancis: Terrific, thanks!

Anything to share on that April hotfix of 1.7.1? Should we use that instead of the official 1.7.1 release?

jchauncey: I have no idea what was fixed in that version. You would have to contact Tony Xu to find out.

jackfrancis: 10-4, thanks!

pkg/api/addons.go
```diff
@@ -680,6 +666,16 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
...
+		defaultCorednsAddonsConfig.Config["cores-per-replica"] = "256"
```
jackfrancis, line 670: These defaults derived from the official documentation here:

https://kubernetes.io/docs/tasks/administer-cluster/dns-horizontal-autoscaling/

pkg/api/addons.go
jackfrancis: Since we're in fact using MCR, this will be Azure Stack safe, so I'll remove this guard.

examples/e2e-tests/kubernetes/release/default/definition.json
```diff
@@ -39,6 +39,14 @@
           {
             "name": "aad-pod-identity",
             "enabled": true
+          },
+          {
+            "name": "coredns",
+            "enabled": true,
+            "config": {
+              "min-replicas": "1",
+              "nodes-per-replica": "10"
```
jackfrancis, line 48: This configuration in concert with the overall node count in the cluster, so that when we perform node scale out E2E tests we are also able to reliably test coredns-autoscaler behaviors.

docs/topics/clusterdefinitions.md
```diff
@@ -235,6 +235,30 @@ Additionally above, we specified a custom docker image for tiller, let's say we
 
 The reason for the unsightly base64-encoded input type is to optimize delivery payload, and to squash a human-maintainable yaml file representation into something that can be tightly pasted into a JSON string value without the arguably more unsightly carriage returns / whitespace that would be delivered with a literal copy/paste of a Kubernetes manifest.
 
+#### coredns
```
mboersma, line 238: Should we refer to it consistently as `CoreDNS` as the project would prefer? Except when referring to the AKS Engine plugin specifically.

vhd/packer/install-dependencies.sh
```diff
@@ -497,6 +497,15 @@ for CSI_SECRETS_STORE_DRIVER_VERSION in ${CSI_SECRETS_STORE_DRIVER_VERSIONS}; do
   echo "  - ${CONTAINER_IMAGE}" >> ${VHD_LOGS_FILEPATH}
 done
 
+CLUSTER_PROPORTIONAL_AUTOSCALER_VERSIONS="
+1.7.1
```
mboersma, line 501: So we weren't adding this to the VHD at all? I'll try to reconcile calico's version with this one or add it here after this merges.

jackfrancis: Good call

parts/k8s/addons/coredns.yaml
justbert: Hmm... This move also caused `coredns` to never be upgraded to newer configurations on a cluster upgrade. It also introduced #3478 due to the use of the `create` operation instead of the `apply`.

justbert: A quick question here (I attempted to find out but couldn't identify the mechanism, sorry): 

Does the coredns configurations as part of the clusterAPI redeploy the `kube-dns` service if the pod IP changes or does it use addon-manager to deploy the change? If so, during a cluster upgrade, those changes wouldn't be propagated since the service is set to `EnsureExist`.

PR #3072, opened 2020-04-15
feat: modify container runtime data dir

alexeldeib:
**Reason for Change**:
Allow customization of the container runtime data directory.

The location of this directory will be subject to heavy disk IO and offloading it from the main OS disk can significantly stabilize the system. Solutions [already exist](https://github.com/juan-lee/knode) to hack this onto nodes, but we'd like to enable it at provision time. This PR does that.

**Issue Fixed**:
n/a

**Requirements**:
- [x] works, ideally
  
- [x] includes documentation
- [x] adds unit tests

**Notes**:

I haven't made API changes to AKS engine in a while, I wasn't sure what all tests I should add. Do we want an e2e covering this? What's the best way to test the full template generation, cmd/generate_test?


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -282,6 +282,9 @@ installContainerd() {
 ensureContainerd() {
   wait_for_file 1200 1 /etc/systemd/system/containerd.service.d/exec_start.conf || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   wait_for_file 1200 1 /etc/containerd/config.toml || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  {{- if HasContainerDataDir}}
+  echo -e "root = \"{{GetContainerDataDir}}\"\n$(cat /etc/containerd/config.toml)" > /etc/containerd/config.toml
```
alexeldeib, line 286: So, this is ugly. An alternative is doing it in cloud-init for non-vhd installs and doing it in the VHD otherwise. Thoughts?

jackfrancis: So, we don't want to just get the whole generic key/val implementation over with first of all? See the way we use the `GetKubeletConfigKeyVals` go template layer func in cloud-init (`parts/k8s/cloud-init/masternodecustomdata.yml` and `parts/k8s/cloud-init/nodecustomdata.yml`) to accommodate generic key/val injection at runtime.

jackfrancis:
Thank you @alexeldeib!

parts/k8s/cloud-init/artifacts/cse_config.sh
alexeldeib: The part I wasn't sure about was how much to generalize this for docker vs containerd because the config files are totally different.

Are you thinking we could do a func like for kubelet, 1 each for docker and containerd, and have those functions accept the raw `map[string]string` and spit out the raw CRI-specific JSON? Then we can avoid any conditionals in the bash

jackfrancis: Exactly 💯 

Not trying to introduce scope creep, but I think we'll thank ourselves later for doing it the generic way. Thoughts?

alexeldeib: I'd really like to avoid "containerdConfig" and "dockerConfig", but we're kind of hiding it no matter what

alexeldeib: i'm on board with that, will update

jackfrancis: I think we can stick with a single `containerRuntimeConfig`, and then have per-runtime go template helper funcs. As long as we can assume that key/val strings are common across all runtimes (I think we can), then we can use a single interface.

alexeldeib: is there a good way to handle errors in the template funcs? I want to do something like this, but struggling to find a good place to drop the Marshal to catch the error. I could do raw string manipulation, but that feels error prone (pun intended) and I wouldn't be guaranteed to construct valid json anyway.

```go
// GetDockerConfig returns the full docker daemon configuration including user overrides.
func (k *KubernetesConfig) GetDockerConfig() (string, error) {
	config := map[string]interface{}{
		"live-restore": true,
		"log-driver": "json-file",
		"log-opts": {
			"max-size": "50m",
			"max-file": "5",
		},
	}

	dataDir, ok := k.ContainerRuntimeConfig[ContainerDataDirKey]
	if ok {
		config["data-root"] = dataDir
	}

	b, err := json.MarshalIndent(config, "", "    ")
	return string(b), err
}
```

alexeldeib: and containerd uses toml, even more fun

alexeldeib:
@jackfrancis how do you feel about this? Can get some more test coverage if it's agreeable. Curious what you think about adding types like this for containerd/docker config and hiding it internally like this.

pkg/api/vlabs/validate.go
```diff
@@ -1543,6 +1543,32 @@ func (k *KubernetesConfig) Validate(k8sVersion string, hasWindows, ipv6DualStack
...
+		if !strings.HasPrefix(val, "/") {
+			return errors.Errorf("OrchestratorProfile.KubernetesConfig.ContainerRuntimeConfig.DataDir '%s' is invalid: must be absolute path", val)
+		}
+	}
+
+	// Validate base config here, and only allow predefined mutations to ensure invariant.
+	if k.ContainerRuntime == Containerd {
+		_, err := common.GetContainerdConfig(k.ContainerRuntimeConfig, nil)
+		if err != nil {
+			return err
```
alexeldeib, line 1563: I realized the way I'm doing this now, these errors are basically guaranteed not to get hit...they're reducing coverage in the diff but I think keeping them is correct for future sanity

alexeldeib: yay for functional options

pkg/api/common/helper.go
```diff
@@ -351,3 +353,69 @@ func WrapAsParameter(s string) string {
...
+	if ok {
+		config.DataRoot = dataDir
+	}
+
+	b, err := json.MarshalIndent(config, "", "    ")
+	return string(b), err
+}
+
+// GetContainerdConfig transforms the default containerd config with overrides. Overrides may be nil.
+func GetContainerdConfig(opts map[string]string, overrides []func(*ContainerdConfig) error) (string, error) {
```
alexeldeib, line 377: I came up with the functional setters to work around the fact that there are templated values in the docker/containerd config normally, but we don't want that logic in the helper functions (for testing purposes and circular imports).

I realize now that having both map[string]string and []func(*ContainerdConfig) error) is a bit redundant, for example we could have a separate helper with signature `func help(map[string]string) []func(*ContainerdConfig) error)` for containerd/docker each and use only functional setters to construct the final object

pkg/api/common/const.go
```diff
@@ -307,3 +307,55 @@ const (
...
+			MaxSize: "50m",
+			MaxFile: "5",
+		},
+	}
+
+	// DefaultContainerdConfig describes the default configuration of the containerd daemon.
+	DefaultContainerdConfig = ContainerdConfig{
+		Version: 2,
+		// should this be true? seems maybe not https://github.com/containerd/containerd/blob/master/docs/ops.md#base-configuration
+		Subreaper: false,
```
alexeldeib, line 326: I thought recommended docs say to set this to true? Is false a kube thing?

jackfrancis: @cpuguy83 should be able to answer

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -282,6 +282,9 @@ installContainerd() {
 ensureContainerd() {
   wait_for_file 1200 1 /etc/systemd/system/containerd.service.d/exec_start.conf || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   wait_for_file 1200 1 /etc/containerd/config.toml || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  {{- if HasValidContainerdConfig}}
```
jackfrancis, line 285: I think we should do this in the cloud-init file spec directly, rather than as a post-cloud-init shell command on top of the cloud-init-paved file.

alexeldeib: Do you happen to know if apt will avoid overwriting the old daemon config if it already exists? That was my only concern, other than some confusion about VHD vs non-VHD.

alexeldeib: Also do you think it's reasonable to remove the validity check and write whatever we have once we've passed validation, or should we attempt to provide a sane default if we know it's invalid but still got here somehow?

pkg/api/common/helper.go
```diff
@@ -351,3 +353,69 @@ func WrapAsParameter(s string) string {
...
+}
+
+// ContainerdKubenetOverride transforms a containerd config to set details required when using kubenet.
+func ContainerdKubenetOverride(config *ContainerdConfig) error {
+	config.Plugins.IoContainerdGrpcV1Cri.CNI.ConfTemplate = "/etc/containerd/kubenet_template.conf"
+	return nil
+}
+
+// ContainerdSandboxImageOverrider produces a function to transform containerd config by setting the SandboxImage.
+func ContainerdSandboxImageOverrider(image string) func(*ContainerdConfig) error {
```
mboersma, line 403: This is an `Overrider` rather than just an `Override` func so we can use "curried" funcs with an `image` arg baked in elsewhere?

alexeldeib: That was my thinking, the semantics are already fairly clear so I'm happy to keep the naming consistent (`Override`) if you think it makes sense

pkg/api/common/const.go
cpuguy83: Let's remove the setting. It is not part of the config anymore and never really made sense I don't think.
The config was changed to `no_subreaper` in https://github.com/containerd/containerd/pull/1822 and later removed altogether as far as I can tell (can't find where it was removed, but the config doesn't exist).

parts/k8s/cloud-init/artifacts/cse_config.sh
alexeldeib: of course doing that dropped my PR coverage again :P

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -222,26 +208,7 @@ write_files:
...
-          {{end}}
-        [plugins."io.containerd.grpc.v1.cri".containerd]
-          default_runtime_name = "runc"
-         [plugins."io.containerd.grpc.v1.cri".containerd.runtimes]
-            [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.runc]
-              runtime_type = "io.containerd.runc.v2"
-            [plugins."io.containerd.grpc.v1.cri".containerd.runtimes.untrusted]
-              {{/* note: runc really should not be used for untrusted workloads... should we remove this? This is here because it was here before */}}
-              runtime_type = "io.containerd.runc.v2"
-    #EOF
```
jackfrancis, line 244: We need to make sure we keep these #EOF sentinel chars at the end of the data stream, which the CSE scripts use to determine that cloud-init has paved the entire file.

alexeldeib:
felt like it was time for some squash

alexeldeib:
@jackfrancis @mboersma friendly ping for a final review, think this is good to go and squashed

jackfrancis:
@alexeldeib, will do, could you add example usage of this new configuration in one of the agent pools, and the masterProfile, to: `test/e2e/test_cluster_configs/everything.json`. 👈  We use that via jenkins to regularly E2E test cluster config functionality.

Thanks again for this!

pkg/api/common/types.go
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+package common
+
+type ContainerdConfig struct {
+	OomScore  int     `toml:"oom_score,omitempty"`
+	Root      string  `toml:"root,omitempty"`
+	Subreaper bool    `toml:"subreaper"`
```
cpuguy83, line 9: Please remove this, it's not a real config. I removed it from the containerd docs: containerd/containerd#4194

cpuguy83: Same for the actual generated toml.

alexeldeib: gah, thought I pulled this out. will remove

alexeldeib:
@jackfrancis this would be cluster-wide since it's in KubernetesConfig like ContainerRuntime (although it definitely could be per agent pool). Would you like me to add a new file/test case with this, or update the everything cluster to use this by default?

jackfrancis:
@alexeldeib Right, sorry, that's obviously evident in the changeset, thanks for the clarification. :)

So yeah, let's put it into `KubernetesConfig` of the everything cluster (unless that doesn't work w/ a windows pool, which that config includes?). We can consider making this config a per-pool config if we ever get to the larger (and more meaningful) refactor work of actually making the container runtime per-pool configurable.

pkg/engine/template_generator.go
```diff
@@ -692,6 +692,44 @@ func getContainerServiceFuncMap(cs *api.ContainerService) template.FuncMap {
...
+			return err == nil
+		},
+		"HasValidContainerdConfig": func() bool {
+			_, err := common.GetContainerdConfig(cs.Properties.OrchestratorProfile.KubernetesConfig.ContainerRuntimeConfig, nil)
+			return err == nil
+		},
+		"GetDockerConfig": func() string {
+			var overrides []func(*common.DockerConfig) error
+
+			if cs.Properties.HasNSeriesSKU() {
```
alexeldeib, line 706: I already squashed but L704-709 were not present which failed N series VMs with GPUs before this

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -176,21 +176,7 @@ write_files:
...
-      }{{if IsNSeriesSKU .}}
-      ,"default-runtime": "nvidia",
-      "runtimes": {
-         "nvidia": {
-             "path": "/usr/bin/nvidia-container-runtime",
-             "runtimeArgs": []
-        }
-      }{{end}}
-    }
+{{IndentString (GetDockerConfig (IsNSeriesSKU .VMSize)) 4}}
```
jackfrancis, line 179: You're teaching me new things about go template syntax

PR #3073, opened 2020-04-15
feat: disabling windows updates by default

marosset:
**Reason for Change**:

Enabling automatic updates on Windows nodes has caused problems for cluster operators in the past and goes against guidance provided by the aks-engine team so it should not be enabled by default in aks-engine

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/const.go
```diff
@@ -223,7 +223,7 @@ const (
 	// https://docs.microsoft.com/en-us/azure/azure-subscription-service-limits#load-balancer.
 	DefaultMaximumLoadBalancerRuleCount = 250
 	// DefaultEnableAutomaticUpdates determines the aks-engine provided default for enabling automatic updates
-	DefaultEnableAutomaticUpdates = true
+	DefaultEnableAutomaticUpdates = false
```
jackfrancis, line 226: For some reason I thought this already defaulted to false, so lgtm

PS this was being unit tested :)

marosset: ooo - i'll update

marosset: One funny thing is that we disable windows updates in the aks-engine VHDs we produce in the windows registry. I thought this was defaulting to false too.

jackfrancis: Ah, that's why I thought we have been doing this for a long time. I still think this change is good for folks who choose not to use the VHD.

PR #3080, opened 2020-04-16
feat: enable CSI proxy by default when cloud-controller-manager is enabled on Windows cluster

chewong:
**Reason for Change**:

Supports azuredisk-csi-driver and azurefile-csi-driver on Windows clusters via CSI proxy. CSI proxy and CSI drivers will be automatically enabled when out-of-tree cloud-controller-manager is enabled. 

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/windowskubeletfunc.ps1
```diff
@@ -342,6 +341,9 @@ New-NSSMService {
     & "$KubeDir\nssm.exe" set Kubelet AppRotateOnline 1 | RemoveNulls
     & "$KubeDir\nssm.exe" set Kubelet AppRotateSeconds 86400 | RemoveNulls
     & "$KubeDir\nssm.exe" set Kubelet AppRotateBytes 10485760 | RemoveNulls
+    # Do not use & when calling DependOnService since 'docker csi-proxy'
+    # is parsed as a single string instead of two separate strings
+    Invoke-Expression "$KubeDir\nssm.exe set Kubelet DependOnService $kubeletDependOnServices | RemoveNulls"
```
chewong, line 346: cc @marosset

pkg/api/addons.go
```diff
@@ -1131,3 +1171,12 @@ func kubeProxyImageSuffix(cs ContainerService) string {
 	}
 	return ""
 }
+
+func getCSISidecarComponent(csiDriverName, csiSidecarName string, k8sComponents map[string]string) string {
```
jackfrancis, line 1175: I see this is being indirectly tested in the various cluster configs in `addons_test.go`, but would it make sense to add lightweight UT for just this func?

pkg/api/addons.go
```diff
@@ -522,67 +522,83 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		Containers: []KubernetesContainerSpec{
 			{
 				Name:           common.CSIProvisionerContainerName,
-				Image:          specConfig.MCRKubernetesImageBase + k8sComponents[common.CSIProvisionerContainerName],
+				Image:          specConfig.MCRKubernetesImageBase + getCSISidecarComponent(common.AzureDiskCSIDriverAddonName, common.CSIProvisionerContainerName, k8sComponents),
 				CPURequests:    "10m",
 				MemoryRequests: "20Mi",
-				CPULimits:      "200m",
-				MemoryLimits:   "200Mi",
+				CPULimits:      "2",
```
jackfrancis, line 528: The delta between the requests and limits is really broad here, do we feel O.K. about that?

pkg/api/azenvtypes.go
```diff
@@ -99,6 +100,7 @@ var (
 		VnetCNILinuxPluginsDownloadURL:   "https://kubernetesartifacts.azureedge.net/azure-cni/" + AzureCniPluginVerLinux + "/binaries/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
 		VnetCNIWindowsPluginsDownloadURL: "https://kubernetesartifacts.azureedge.net/azure-cni/" + AzureCniPluginVerWindows + "/binaries/azure-vnet-cni-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
 		ContainerdDownloadURLBase:        "https://storage.googleapis.com/cri-containerd-release/",
+		CSIProxyDownloadURL:              "https://k8sreleases.blob.core.windows.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz",
```
jackfrancis, line 103: I assume this is a temporary landing spot for this?

pkg/api/addons.go
chewong: @andyzhangx could you comment on this? We increased the limits to prevent containers from restarting.

pkg/api/azenvtypes.go
chewong: I will be changing that to https://kubernetesartifacts.azureedge.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz

jackfrancis: Is that URL not available in mooncake? If it is, we probably don't need to reference it here and can reference it from the cloud env-context-free maps in `pkg/api/k8s_versions.go`

chewong: The pipeline outputs a mooncake and a non-mooncake artifact so I assume https://kubernetesartifacts.azureedge.net/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz isn't available in mooncake. Let me follow up on that.

pkg/api/addons.go
andyzhangx: 200Mi is not enough, so we set an aggressive limit

pkg/api/azenvtypes.go
feiskyer: @andyzhangx have we mirrored the csi binaries somewhere?

andyzhangx: @fseldow are you maintaining this?
https://kubernetesartifacts.blob.core.chinacloudapi.cn/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz

chewong: https://kubernetesartifacts.blob.core.chinacloudapi.cn/csi-proxy/v0.1.0/binaries/csi-proxy.tar.gz is currently private. We have requested the China team to make it public.

pkg/api/addons.go
jackfrancis: Then should we increase the CPURequests to 1? If we know that we will need up to 2 cores, shouldn't we reserve half to start to prevent the scheduler from thrashing?

chewong:
Testing 1.14 - 1.19 w/ cloud-controller-manager  https://upstream.jenkins.azure-containers.io/blue/organizations/jenkins/k8s-create-cluster-matrix/detail/k8s-create-cluster-matrix/134/pipeline/

pkg/api/addons.go
andyzhangx: actually we don't know now, and usually windows pod would consume more cpu, memory than linux pod. This value is an aggressive value, need to do some perf test in the future. And if we set as half, we may waste resource. Shall we go with this current value now since it's still alpha/beta now?

chewong: I agree with @jackfrancis. @andyzhangx what do you think of the new requests and limits in https://github.com/kubernetes-sigs/azuredisk-csi-driver/pull/363? We should keep the limits for Linux pods the same and slightly decrease resource requests and limits for Windows pods.

chewong: For Linux:
```yaml
resources:
  limits:
    cpu: 200m
    memory: 200Mi
  requests:
    cpu: 10m
    memory: 20Mi
```

For Windows:
```yaml
resources:
  limits:
    cpu: 1
    memory: 1Gi
  requests:
    cpu: 500m
    memory: 512Mi
```

andyzhangx: what about this?
For Linux:
```yaml
resources:
  limits:
    cpu: 500m
    memory: 500Mi
  requests:
    cpu: 10m
    memory: 20Mi
```

For Windows:
```yaml
resources:
  limits:
    cpu: 1
    memory: 1Gi
  requests:
    cpu: 10m
    memory: 20Mi
```

PR #3082, opened 2020-04-16
feat: give metrics-server system-cluster-critical priority

mboersma:
**Reason for Change**:
Adds `priorityClassName: system-cluster-critical` to metrics-server pods so they are less likely to evicted under load.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
cc: @MichaelSinz


Michael-Sinz:
Beautiful - that is exactly what I was looking for.  (And what I had done to the add-on after it was deployed as a post-create fixup)

Thank you!

PR #3085, opened 2020-04-16
feat: enable multiple frontend IPs in Standard LB

### LINKED ISSUE #2377: Be able to configure both the number of IPs or the allocatedOutboundPorts

jackfrancis:
**Reason for Change**:

This PR enables a configurable number of frontend IP addresses to use for Standard LoadBalancer outbound rules, in order to increase available port allocation for large clusters.

The new interface to use this configuration is a property in the `kubernetesConfig` object. E.g.:

```
...
"kubernetesConfig": {
  ...
  "loadBalancerOutboundIPs": 3,
  ...
}
...
```

// TODO

- [x] documentation

Outbound SNAT limitations are described here:

https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-connections#multifesnat

**Issue Fixed**:

Fixes #2377 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/const.go
```diff
@@ -41,6 +41,8 @@ const (
 	DefaultInternalLbStaticIPOffset = 10
 	// DefaultEnableCSIProxyWindows determines if CSI proxy should be enabled by default for Windows nodes
 	DefaultEnableCSIProxyWindows = false
+	// MaxLoadBalancerOutboundIPs is the maximum number of outbound IPs in a Standard LoadBalancer frontend configuration
+	MaxLoadBalancerOutboundIPs = 16
```
jackfrancis, line 45: @dmeytin are you aware of a sensible limit we should put here?

docs/topics/clusterdefinitions.md
```diff
@@ -65,7 +65,8 @@ $ aks-engine get-versions
...
+| loadBalancerSku                  | no                        | Sku of Load Balancer and Public IP. Candidate values are: `basic` and `standard`. If not set, it will be default to "standard". NOTE: Because VMs behind standard SKU load balancer will not be able to access the internet without an outbound rule configured with at least one frontend IP, AKS Engine creates a Load Balancer with an outbound rule and with agent nodes added to the backend pool during cluster creation, as described in the [Outbound NAT for internal Standard Load Balancer scenarios doc](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-outbound-rules-overview#outbound-nat-for-internal-standard-load-balancer-scenarios)                                                                     |
```
jackfrancis, line 68: This documentation was stale, updated here

PR #3086, opened 2020-04-16
fix: Update CSI Secrets Store Provider for Azure to 0.0.4

bdlb77:
**Reason for Change**:
Upgrading from version `0.0.3` to `0.0.4` allows for simplification by removing the requirement for `subscriptionId` and `resourceGroup` properties on the Secrets Store CSI Driver Provider for Azure

/cc @aramase  @paulbouwer


pkg/api/k8s_versions.go
```diff
@@ -53,7 +53,7 @@ const (
 	virtualKubeletImageReference                      string = "virtual-kubelet:latest"
 	azureCNINetworkMonitorImageReference              string = "networkmonitor:v0.0.8"
 	tillerImageReference                              string = "oss/kubernetes/tiller:v2.13.1"
-	csiSecretsStoreProviderAzureImageReference        string = "k8s/csi/secrets-store/provider-azure:0.0.3"
+	csiSecretsStoreProviderAzureImageReference        string = "k8s/csi/secrets-store/provider-azure:0.0.4"
```
aramase, line 56: @bdlb77 Thank you for updating the image. Can you also update the image in the [vhd install script](https://github.com/Azure/aks-engine/blob/master/vhd/packer/install-dependencies.sh#L483)?

bdlb77: updated! thank you @aramase for the guidance on that

PR #3089, opened 2020-04-17
feat: add support for Kubernetes 1.18.2

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.18.md#changelog-since-v1181

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/packer/configure-windows-vhd.ps1
+++ b/vhd/packer/configure-windows-vhd.ps1
@@ -85,7 +85,8 @@ function Get-FilesToCacheOnVHD
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.3/windowszip/v1.17.3-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.4/windowszip/v1.17.4-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.18.0/windowszip/v1.18.0-1int.zip",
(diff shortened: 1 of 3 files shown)


vhd/packer/configure-windows-vhd.ps1
```diff
@@ -85,7 +85,8 @@ function Get-FilesToCacheOnVHD
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.3/windowszip/v1.17.3-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.4/windowszip/v1.17.4-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.18.0/windowszip/v1.18.0-1int.zip",
```
jackfrancis, line 87: get rid of 1.18.0?

mboersma: Let me just double-check with @marosset, because it appeared to me we've been keeping three patches per release in the Windows VHD. If that was unintended, I'll remove this.

mboersma: Since the Windows VHD isn't space-constrained, and this image is shared with AKS, it behooves us to keep more patch versions around. We'll make sure @marrosset checks such changes to the Windows VHD.

marosset: should be fine to keep

PR #3091, opened 2020-04-17
fix: update cluster-proportional-autoscaler path at MCR

### LINKED ISSUE #3079: cluster-proportional-autoscaler-amd64:1.1.2-r2 fails to pull from mcr.microsoft.com ( ImagePullBackOff )
**Describe the bug**
calico-typha-horizontal-autoscaler deployment cannot pull the the image from:

`mcr.microsoft.com/cluster-proportional-autoscaler-amd64:1.1.2-r2`

`calico-typha-horizontal-autoscaler-5757c78465-kbpz5   0/1     ImagePullBackOff`

```bash
41s         Warning   Failed             pod/calico-typha-horizontal-autoscaler-5757c78465-kbpz5    Failed to pull image "mcr.microsoft.com/cluster-proportional-autoscaler-amd64:1.1.2-r2": rpc error: code = Unknown desc = Error response from daemon: manifest for mcr.microsoft.com/cluster-proportional-autoscaler-amd64:1.1.2-r2 not found: manifest unknown: manifest unknown

```
**Steps To Reproduce**
After creating a cluster via passing a prebuild json template ( as we were successfully doing it numerous times with previous versions of aks-engine ) this image fails to download.

**Expected behavior**
Successful pull of the image from mcr.microsoft.com
**AKS Engine version**
v0.49.0
**Kubernetes version**
1.18.1
**Additional context**
If i manualy pass to the `calico-typha-horizontal-autoscaler` deployment the image from another source : `k8s.gcr.io/cluster-proportional-autoscaler-amd64:1.1.2-r2` the image is reachable.

mboersma:
**Reason for Change**:
The cluster-proportional-autoscaler image reference at mcr.microsoft.com was incorrect. Now that MCR is the default, this could cause calico cluster deployments to fail.

**Issue Fixed**:
Fixes #3079

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I did the before and after tests locally to confirm this fixes things. This probably should have been caught somewhere in Jenkins e2e, however. I will follow up.


pkg/api/k8s_versions.go
```diff
@@ -19,7 +19,7 @@ const (
 	calicoCNIImageReference                           string = "cni:v3.8.0"
 	calicoNodeImageReference                          string = "node:v3.8.0"
 	calicoPod2DaemonImageReference                    string = "pod2daemon-flexvol:v3.8.0"
-	calicoClusterProportionalAutoscalerImageReference string = "cluster-proportional-autoscaler-amd64:1.1.2-r2"
+	calicoClusterProportionalAutoscalerImageReference string = "oss/kubernetes/autoscaler/cluster-proportional-autoscaler:1.1.2-r2"
```
jackfrancis, line 22: I think we need to turn this into the full mcr URL, and get rid of the `kubernetesImageBase` prefix dep. Basically we don't need to support both GCR and MCR types here.

Also, not to add scope creep, but is it worth seeing if a newer version of this works w/ the current calico spec?

mboersma: > is it worth seeing if a newer version of this works...?

I think so. I added a commit to put all three versions of C-P-A in the VHD but ideally we would all use 1.7.1 (or the 1.7.1-hotfix?). I'll try that next.

mboersma: Actually, since this is a new regression in v0.49.0, let's defer the version bumps to a separate PR.

andyliuliming:
Hi @jackfrancis  @mboersma  when will this been release? :)

jackfrancis:
@andyliuliming we are in the process of validating master for a v0.50.0 release by end of week

PR #3104, opened 2020-04-20
fix: typos in JSON annotations caught by staticcheck

mboersma:
**Reason for Change**:
Trying to fix #2839, I ran `golangci-lint` with the default linters enabled and couldn't ignore these typos. But I wonder if this could create actual backward-compatibility issues if the previous annotations weren't being honored as expected.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@mboersma I think this should be safe because the back-compat input scenario *would* include the JSON-rendered zero values already present. And those zero values should still be acceptable as `aks-engine` input.

@devigned @CecileRobertMichon for sanity check

PR #3109, opened 2020-04-20
fix: add nodeSelector for coredns-autoscaler

chewong:
**Reason for Change**:

coredns-autoscaler was evicted to a Windows node during an upgrade in https://upstream.jenkins.azure-containers.io/blue/organizations/jenkins/k8s-create-cluster-matrix/detail/k8s-create-cluster-matrix/141/pipeline/271/, causing the test to fail. It looks like it's deployed without a node selector.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -308,6 +308,7 @@ metadata:
   namespace: kube-system
   labels:
     k8s-app: coredns-autoscaler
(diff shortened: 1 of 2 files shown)


jackfrancis:
Thanks!

psss `make generate` and commit

PR #3136, opened 2020-04-24
feat: block imds from VMSS nodes if aad-pod-identity + msi

### LINKED ISSUE #3135: Set iptable rules to block traffic to IMDS
**Describe the request**
When a cluster with pod-identity addon is enabled, there could be scenarios during scaling up where the workload pods are scheduled before pod-identity NMI pods are scheduled on the node. During this period of time, request for token from workload pod will be directly sent to IMDS. This means the workload pod can request token for identity it doesn't have access to.

This is not an issue for VMAS clusters as identity is at the VM level. However, in case of VMSS the identity is assigned at the VMSS level and currently there is no support for identity 
at VMSSVM instance level.

There is currently no solution in kubernetes where node can be set to `NoSchedule` until critical addons are deployed to the cluster. There is a KEP for this which is currently being reviewed - https://github.com/kubernetes/enhancements/pull/1003

This will help ensure VMSS based clusters that use managed identity and pod identity can restrict access to identities at all times during cluster lifetime.

Add a systemd unit file to setup iptable rules during cluster bootstrap. This will ensure there is no access to IMDS by default and access is allowed only through pod-identity.

- Tried to use admission controller to taint nodes as no schedule until NMI is deployed. The problem with this is other critical addons will need to tolerate this taint in order to be deployed.

**Additional context**

iptable rules added by pod-identity -

```
Chain aad-metadata (1 references)
target     prot opt source               destination
DNAT       tcp  -- !localhost            169.254.169.254      tcp dpt:http to:10.240.0.4:2579
RETURN     all  --  anywhere             anywhere
```

```
Chain PREROUTING (policy ACCEPT)
target     prot opt source               destination
aad-metadata  all  --  anywhere             anywhere
```

Placing custom chain - https://github.com/Azure/aad-pod-identity/blob/master/pkg/nmi/iptables/iptables.go#L64-L74
Creating custom chain - https://github.com/Azure/aad-pod-identity/blob/3e2a3dda0e9c3963bc90e08a20f0308a81907149/pkg/nmi/iptables/iptables.go#L76-L102

jackfrancis:
**Reason for Change**:

In an aad-pod-identity cluster configuration, access to IMDS from nodes is proxy'd by the aad-pod-identity nmi daemonset running on each node. This PR introduces an explicit iptables rule that blocks access to IMDS for aad-pod-identity + VMSS configurations when using msi. Because instances in a VMSS share a common msi, we want to prevent a pod container getting a VMSS-scoped msi token between the time that a node comes online and before the nmi daemonset can write its own iptables rules to re-route IMDS to the aad-pod-identity componentry.

**Issue Fixed**:

Fixes #3135 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/nodecustomdata.yml
+++ b/parts/k8s/cloud-init/nodecustomdata.yml
@@ -321,12 +321,13 @@ write_files:
   owner: root
   content: |
     #!/bin/bash
+{{- if and (IsVirtualMachineScaleSets .) IsAADPodIdentityAddonEnabled}}
+    {{- /* Disable TCP access to IMDS endpoint, aad-pod-identity nmi component will provide a complementary iptables rule to re-route this traffic */}}
+    iptables -A OUTPUT -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DROP
+{{end}}
 {{- if not IsIPMasqAgentEnabled}}
     {{if IsAzureCNI}}
     iptables -t nat -A POSTROUTING -m iprange ! --dst-range 168.63.129.16 -m addrtype ! --dst-type local ! -d {{WrapAsParameter "vnetCidr"}} -j MASQUERADE
-    {{end}}
-    {{if BlockIMDS}}
-    iptables -A OUTPUT -s 127.0.0.1/32 -d 169.254.169.254/32 -p tcp -m tcp --dport 80 -j DROP
     {{end}}
 {{end}}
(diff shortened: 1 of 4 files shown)


jackfrancis:
@aramase PTAL

aramase:
Thank you @jackfrancis! changes look good.

jackfrancis:
@yangl900 FYI

PR #3140, opened 2020-04-24
feat: Kubernetes Dashboard addon v2.0.0

mboersma:
**Reason for Change**:
Updates the kubernetes-dashboard addon to use the v2.0.0 release, which includes a metrics-scraper component.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This addon is still disabled by default, which I think is appropriate.

I tested manually with Kubernetes 1.14 and 1.15, as well as 1.18 and 1.19. Graphing data didn't show up on 1.14 (although the metrics-scraper was up) but otherwise it seemed functional everywhere, so it's a step up from the deprecated previous version.

The CPU and Mem requests/limits are basically a stab in the dark.


parts/k8s/addons/kubernetesmasteraddons-kubernetes-dashboard-deployment.yaml
```diff
@@ -1,91 +1,183 @@
+
```
jackfrancis, line 1: nit: yay, we only have one file to maintain! can we rename this to something concise like "kubernetes-dashboard.yml" ?

mboersma: It's not just a deployment, so let's at least drop the `-deployment` bit. But every other addon follows the `kubernetesmasteraddons-component-foo-resource.yaml` convention, so maybe we should drop the prefix in general, but in a separate PR?

jackfrancis: Not every addon! See those elegant `coredns.yaml`, `ip-masq-agent.yaml`, etc...

mboersma: It's true, just the majority... I am persuaded. 👍 

Maybe I'll make a later PR to shorten the names in general.

jackfrancis: Yeah, I started down that road and realized the better optimization is to consolidate the version directory-specific specs into single per-addon specs (using go templateing funcs to do various per-version changes). Then we only have to change the filenames in a single directory!

parts/k8s/addons/kubernetes-dashboard.yaml
```diff
@@ -0,0 +1,319 @@
+
```
jackfrancis, line 1: Is there any difference in the dashboard specs between 1.14 and 1.19? Wondering if we can consolidate these into one file...

docs/topics/monitoring.md
```diff
@@ -6,14 +6,14 @@ There are five main options to monitor your cluster:
...
 1. [Azure Monitor for containers](#azure-monitor-for-containers)
-1. [Kubernetes dashboard](#kubernetes-dashboard)
+1. [Kubernetes Dashboard](#kubernetes-dashboard)
 1. [Monitoring extension](#monitoring-extension)
 1. [Grafana and Influx DB](#grafana-and-influx-db)
 1. [Heapster REST API](#heapster-rest-api)
 
 ## Intro to Heapster
 
-Monitoring your cluster in Kubernetes is powered by a component called [Heapster](https://github.com/kubernetes/Heapster/). Heapster is a pod that is responsible for aggregating monitoring data from across all the nodes and pods in your cluster. Heapster is necessary for viewing monitoring data in the Kubernetes dashboard as well as in Grafana. Heapster comes preinstalled on `aks-engine` deployments. To ensure that Heapster is set up in your cluster and is running:
```
jackfrancis, line 16: blech, what should we do about this defuct heapster documentation?

parts/k8s/addons/kubernetes-dashboard.yaml
mboersma: This manifest works with versions 1.14+ in my testing (at least, the Dashboard is accessible after setting up a role and getting a token). This should be the only manifest we're using for Dashboard for all supported versions.

docs/topics/monitoring.md
mboersma: We should revise it for sure. This whole monitoring doc needs TLC, but I figured I'd just make dashboard-specific changes in this PR.

parts/k8s/addons/kubernetes-dashboard.yaml
mboersma: Here is the compatibility matrix for Kubernetes Dashboard [v2.0.0](https://github.com/kubernetes/dashboard/releases/tag/v2.0.0):

Kubernetes version | 1.15 | 1.16 | 1.17 | 1.18
-- | -- | -- | -- | --
Compatibility | ? | ? | ? | ✓

✓ Fully supported version range.
? Due to breaking changes between Kubernetes API versions, some features might not work correctly in the Dashboard.

jackfrancis: oh sorry jeez

red means deleted

nothing to see here

PR #3143, opened 2020-04-24
chore: aad-pod-identity taint

jackfrancis:
**Reason for Change**:

This PR introduces an aad-pod-identity-specific taint when the aad-pod-identity addon is enabled, and then adds systemd-enforced reconciliation to remove that taint after the `nmi` daemonset has been scheduled successfully onto a node (and thus the node is ready for schedulable, aad-pod-identity-enforced workloads).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/1.19/kubernetesmasteraddons-aad-pod-identity-deployment.yaml
+++ b/parts/k8s/addons/1.19/kubernetesmasteraddons-aad-pod-identity-deployment.yaml
@@ -139,6 +139,11 @@ spec:
             - NET_ADMIN
       nodeSelector:
         kubernetes.io/os: linux
+      tolerations:
+      - key: node.kubernetes.io/aad-pod-identity-not-ready
+        operator: Equal
+        value: "true"
+        effect: NoSchedule
 ---
 apiVersion: v1
 kind: ServiceAccount
--- a/parts/k8s/addons/1.19/kubernetesmasteraddons-aad-pod-identity-deployment.yaml
+++ b/parts/k8s/addons/1.19/kubernetesmasteraddons-aad-pod-identity-deployment.yaml
@@ -140,7 +140,7 @@ spec:
       nodeSelector:
         kubernetes.io/os: linux
       tolerations:
-      - key: node.kubernetes.io/aad-pod-identity-not-ready
+      - key: {{GetAADPodIdentityTaintKey}}
         operator: Equal
         value: "true"
         effect: NoSchedule
--- a/pkg/api/common/const.go
+++ b/pkg/api/common/const.go
@@ -375,3 +375,4 @@ const (
 )
 
 const MasterNodeTaint string = "node-role.kubernetes.io/master=true:NoSchedule"
+const AADPodIdentityTaint string = "node.kubernetes.io/aad-pod-identity-not-ready"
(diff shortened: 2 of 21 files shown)


jackfrancis:
@ericsuhong @DavidParks8 PTAL

DavidParks8:
Won't this not solve the problem for users of AKS? Also, right now, the IAM clusters don't use aks-engine addons because we are trying to separate cluster creation from cluster configuration so we can move to aks eventually.

parts/k8s/addons/1.19/kubernetesmasteraddons-aad-pod-identity-deployment.yaml
```diff
@@ -139,6 +139,11 @@ spec:
             - NET_ADMIN
       nodeSelector:
         kubernetes.io/os: linux
+      tolerations:
+      - key: node.kubernetes.io/aad-pod-identity-not-ready
```
mboersma, line 143: Shouldn't all the manifests be using `key: {{GetAADPodIdentityTaintKey}}` here?

jackfrancis:
@DavidParks8 correct, this will not address the AKS story (other than prototyping a real world end-to-end evolution of the current implementation that includes tainting/untainting as a part of AAD Pod Identity-enabled clusters)

pkg/api/common/const.go
```diff
@@ -371,3 +371,7 @@ const (
 	AntreaInstallCniChainCmd      = "install_cni_chaining"
 	AntreaNetworkPolicyOnlyMode   = "networkPolicyOnly"
 )
+
+// Taints
+// AADPodIdentityTaintKey is the node taint for AAD Pod Identity-enabled clusters before NMI daemonset is ready
+const AADPodIdentityTaintKey string = "node.kubernetes.io/aad-pod-identity-not-ready"
```
jackfrancis, line 377: An alternative to this approach: we could create a programmatically generated (at runtime during cluster provisioning) taint to ensure that workloads aren't able to bypass node scheduling order enforcement by including this well-known taint toleration.

Thoughts @mboersma @DavidParks8 @ritazh @Michael-Sinz @aramase ?

DavidParks8: I feel the random generation method is more vulnerable than a method which temporarily blocks IMDS because anyone who found the random label could bypass it by making a toleration on their deployment. It feels a bit like security through obscurity that gets weaker the longer a cluster is running. Since labels aren't treated like secrets, they may be exposed in logs as well.

jackfrancis: fair enough, it's only really viable during cluster creation, after that the taint is discoverable by scraping the k8s node api

pkg/engine/const.go
```diff
@@ -110,6 +110,8 @@ const (
 	kubernetesDockerMonitorSystemdService    = "k8s/cloud-init/artifacts/docker-monitor.service"
 	labelNodesScript                         = "k8s/cloud-init/artifacts/label-nodes.sh"
 	labelNodesSystemdService                 = "k8s/cloud-init/artifacts/label-nodes.service"
+	untaintNodesScript                       = "k8s/cloud-init/artifacts/untaint-nodes.sh"
```
mboersma, line 113: I think a better antonym for "taint" is "cleanse." I insist that we use this verb everywhere from now on. 😜

PR #3144, opened 2020-04-24
fix: use ctr to remove containerd containers

ksubrmnn:
**Reason for Change**:

Clean up containerd containers on restart so that Kubelet can correctly deploy new containers

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Waiting on test


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -154,6 +156,16 @@ var _ = BeforeSuite(func() {
 		dnsAddonName = common.CoreDNSAddonName
 	}
 	Expect(dnsAddonName).NotTo(Equal(""))
+
+	var client *armhelpers.AzureClient
+	env, err := azure.EnvironmentFromName("AzurePublicCloud") // TODO get this programmatically
```
jackfrancis, line 161: @jadarsie remind me, are you using this test runner in Azure stack? If so we might want to skip this new foo in Azure Stack contexts

parts/k8s/windowsnodereset.ps1
```diff
@@ -45,7 +47,13 @@ if ($global:CsiProxyEnabled) {
 $hnsNetwork = Get-HnsNetwork | Where-Object Name -EQ azure
 if ($hnsNetwork) {
     Write-Log "Cleaning up containers"
-    docker ps -q | ForEach-Object { docker rm $_ -f }
+    if ($UseContainerD -eq $true) {
+        ctr -n k8s.io c ls -q | ForEach-Object { ctr -n k8s.io tasks kill $_ }
```
marosset, line 51: might be safer to use ctr.exe here - powershell does weird things when invoking .exe:s

marosset: and for docker below as well

parts/k8s/windowsnodereset.ps1
```diff
@@ -45,7 +47,13 @@ if ($global:CsiProxyEnabled) {
...
+        ctr -n k8s.io c ls -q | ForEach-Object { ctr -n k8s.io c rm $_ }
```
marosset, line 52: Is there a -f equiv on the ctr.exe commands (like the docker calls below).
Or have you seen the tasks kill operation not complete succesfully - i'm just curious what would happen to the c rm call if the tasks kill command failed

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,101 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				Skip("No windows agent was provisioned for this Cluster Definition")
 			}
 		})
+		It("should be able to deploy containers after restart", func() {
```
marosset, line 2027: nit: reword to something like 'should expect containers to be recreated after restart'

test/e2e/runner.go
```diff
@@ -201,6 +201,23 @@ func main() {
...
+		} else {
+			resourceGroup = cfg.Name
+		}
+
+		if resourceGroup == "" {
+			if cfg.CleanUpIfFail {
+				teardown()
+			}
+			log.Fatalf("Resource Group is empty")
+			os.Exit(1)
```
marosset, line 216: @jackfrancis can you confirm we're ok with an os.Exit() here?
(i generally avoid these calls at all costs)

marosset: maybe returning an error would be cleaner?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,101 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				err = iisService.ValidateWithRetry("(IIS Windows Server)", sleepBetweenRetriesWhenWaitingForPodReady, cfg.Timeout)
+				Expect(err).NotTo(HaveOccurred())
+
+				pods, err := iisDeploy.Pods()
+				Expect(err).NotTo(HaveOccurred())
+				nodeName := pods[0].Spec.NodeName
+				ctx, cancel := context.WithTimeout(context.Background(), 6000*time.Second)
+				defer cancel()
+
+				By("Adding taint to all other Windows nodes")
```
marosset, line 2061: Just confirming - we don't run tests in parallel anyplace do we?

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,101 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				for _, n := range nodeList.Nodes {
+					if n.IsWindows() && n.Metadata.Name != nodeName {
+						n.AddTaint(node.Taint{Key: "key", Value: "value", Effect: "NoSchedule"})
+					}
+				}
+
+				By("Restarting VM " + nodeName + " in resource group " + cfg.ResourceGroup)
+
+				// Getting vmss for the vm
+				vmssPage, err := azureClient.ListVirtualMachineScaleSets(ctx, cfg.ResourceGroup)
```
marosset, line 2072: What happens if the Windows nodes are part of an availability set?
Should we at least add a skip here?

marosset: (well not here - but early in the test case :))

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,101 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				instanceIDs := &compute.VirtualMachineScaleSetVMInstanceIDs{&[]string{instanceID}}
+				err = azureClient.RestartVirtualMachineScaleSets(ctx, cfg.ResourceGroup, vmssName, instanceIDs)
+				Expect(err).NotTo(HaveOccurred())
+
+				//Wait for VM to come up
+				time.Sleep(30 * time.Second)
+
+				By("Verifying that the service is still reachable and returns the default IIS start page")
+				err = iisService.ValidateWithRetry("(IIS Windows Server)", sleepBetweenRetriesWhenWaitingForPodReady, cfg.Timeout)
+				Expect(err).NotTo(HaveOccurred())
```
marosset, line 2109: If this errors out does would the taints stay around?
This seems like it could negatively impact future test cases.

test/e2e/runner.go
ksubrmnn: I found this from earlier in the code so I'll do whatever is recommended

test/e2e/kubernetes/kubernetes_test.go
marosset: in other languages i'd like to see this get put into a finally block.
Can we add a defer block that would run just around this test case in golang?

ksubrmnn: Yeah I guess if it isn't successful, they won't be removed. I don't see a way around this

test/e2e/kubernetes/kubernetes_test.go
ksubrmnn: When does that happen?

parts/k8s/windowsnodereset.ps1
ksubrmnn: No there's no force option :/

test/e2e/kubernetes/kubernetes_test.go
marosset: I think you could add a check right at the beginning of the It("should be able to deploy containers...") and skip on availability sets there.
I'm not sure if there an easy test of this or not but maybe @jackfrancis would know?

test/e2e/kubernetes/kubernetes_test.go
ksubrmnn: Deferred in line 2073

test/e2e/runner.go
ksubrmnn: I removed it

test/e2e/kubernetes/kubernetes_test.go
ksubrmnn: @jackfrancis

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,105 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				Skip("No windows agent was provisioned for this Cluster Definition")
 			}
 		})
+		It("should expect containers to be recreated after node restart", func() {
+			if eng.HasWindowsAgents() {
+				if eng.ExpandedDefinition.Properties.HasAvailabilityZones() {
```
marosset, line 2029: Was this supposed to check for AvailablitySets (agentPoolProfile.availabilityProfile == AvailabilitySet) instead of AvailabilityZones (based on other comments)?

ksubrmnn: Ah yes, I got confused. Will fix

test/e2e/kubernetes/kubernetes_test.go
ksubrmnn: @mboersma and @devigned confirmed that we don't :)

test/e2e/config/config.go
```diff
@@ -51,9 +52,12 @@ type Config struct {
 	BlockSSHPort        bool   `envconfig:"BLOCK_SSH" default:"false"`
 	AddNodePoolInput    string `envconfig:"ADD_NODE_POOL_INPUT" default:""`
 	TestPVC             bool   `envconfig:"TEST_PVC" default:"false"`
+	SubscriptionID      string `envconfig:"SUBSCRIPTION_ID"`
+	ID                  string `envconfig:"CLIENT_ID"`
```
marosset, line 56: nit - ClilentID / ClientSecret?

test/e2e/kubernetes/node/node.go
```diff
@@ -204,6 +204,25 @@ func (n *Node) Describe() error {
...
+}
+
+// Remove taint to node
+func (n *Node) RemoveTaint(taint Taint) error {
+	var commandTimeout time.Duration
+	cmd := exec.Command("k", "taint", "nodes", n.Metadata.Name, taint.Key+":"+taint.Effect+"-")
+	out, err := util.RunAndLogCommand(cmd, commandTimeout)
+	log.Printf("\n%s\n", string(out))
+	return err
+
```
marosset, line 223: nit: extra whitespace

test/e2e/test_cluster_configs/containerd.json
```diff
@@ -33,7 +44,17 @@
                         }
                     ]
                 }
-            }
+            },
+            "windowsProfile": {
+				"adminUsername": "azureuser",
```
marosset, line 49: nit: spacing

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2008,6 +2024,111 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+				}
+
+				// Removing taints
+				defer func(nodeList *node.List, nodeName string) {
+					for _, n := range nodeList.Nodes {
+						if n.IsWindows() && n.Metadata.Name != nodeName {
+							n.RemoveTaint(node.Taint{Key: "key", Value: "value", Effect: "NoSchedule"})
+						}
+					}
+				}(nodeList, nodeName)
```
marosset, line 2085: Do you know if this defer func fires after just this test case.
My concern here is that if this test fails and we don't un-taint the nodes we could skew the resulys of the rest of the test cases.
If would be good to avoid that if we can.

ksubrmnn: I believe it'll execute after the function in line 2028 finishes

ksubrmnn: I verified from the PR tests that the taints are removed after the test

PR #3155, opened 2020-04-28
refactor: enable generic --register-with-taints and --register-node

jackfrancis:
**Reason for Change**:

This PR unwinds some weirdness in the implementation of `--register-with-taints`, especially for masterProfile configuration.

Also addressed some strange `--register-node` configuration.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -361,10 +361,6 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
 {{if IsCustomCloudProfile }}
     AZURE_ENVIRONMENT_FILEPATH=/etc/kubernetes/azurestackcloud.json
 {{end}}
-{{if AnyAgentIsLinux}}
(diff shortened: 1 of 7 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -360,10 +360,6 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
 {{end}}
 {{if IsCustomCloudProfile }}
     AZURE_ENVIRONMENT_FILEPATH=/etc/kubernetes/azurestackcloud.json
-{{end}}
-{{if AnyAgentIsLinux}}
-    KUBELET_REGISTER_NODE=--register-node=true
```
jackfrancis, line 365: As far as I can tell this weird `--register-node=true` config inside this "any agent is Linux" block doesn't make sense. "true" is the default value for this, and we certainly never want to not register a node in the cluster.

Ref:

https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

PR #3158, opened 2020-04-28
fix: don't hardcode pause image for containerD on Windows

marosset:
**Reason for Change**:

This PR replaces updates how the pause image is specified for both containerd and docker on windows (containerd was previously hardcoded) to now use values from kubeclusterconfig.json.

For reference here is how the values are expected in other kubeclusterconfig files:
https://github.com/kubernetes-sigs/sig-windows-tools/blob/master/kubeadm/v1.16.0/Kubeclusterbridge.json

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -41685,11 +41685,14 @@ function Write-KubeClusterConfig {
         [Parameter(Mandatory = $true)][string]	
         $KubeDnsServiceIp
     )
-    $ConfigFile = "c:\k\kubeclusterconfig.json"
+
     $Global:ClusterConfiguration = [PSCustomObject]@{ }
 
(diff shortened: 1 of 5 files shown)


pkg/engine/templates_generated.go
```diff
@@ -43148,13 +43153,21 @@ function Install-Containerd {
     [Parameter(Mandatory = $true)][string]
     $CNIConfDir
   )
+
+  $svc = Get-Service -Name containerd -ErrorAction SilentlyContinue
```
marosset, line 43157: This bit is needed for work containerd VHD work I'm also working on (but i'll highlight it anyway)

PR #3167, opened 2020-04-29
refactor: simplify kubelet systemd service

jackfrancis:
**Reason for Change**:

This PR dusts off the cobwebs from the kubelet systemd implementation. Specifically:

- moves bash block statements out from single line systemd `ExecStartPre` strings into the pre-existing `/opt/azure/containers/kubelet.sh` script (a `ExecStartPre` dependency)
- enables the following containerd kubelet configuration to be overridden via api model:
  - `--container-runtime`
  - `--runtime-request-timeout`
  - `--container-runtime-endpoint`
- removes iptables output from start of kubelet logs

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/kubelet.service
+++ b/parts/k8s/cloud-init/artifacts/kubelet.service
@@ -10,15 +10,6 @@ Restart=always
 EnvironmentFile=/etc/default/kubelet
 SuccessExitStatus=143
 ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
-ExecStartPre=/bin/mkdir -p /var/lib/kubelet
-ExecStartPre=/bin/mkdir -p /var/lib/cni
-ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
-ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
-{{/* This is a partial workaround to this upstream Kubernetes issue: */}}
-{{/* https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731 */}}
-
-ExecStartPre=-/sbin/ebtables -t nat --list
-ExecStartPre=-/sbin/iptables -t nat --numeric --list
 ExecStart=/usr/local/bin/kubelet \
         --enable-server \
         --node-labels="${KUBELET_NODE_LABELS}" \
--- a/parts/k8s/cloud-init/artifacts/kubelet.service
+++ b/parts/k8s/cloud-init/artifacts/kubelet.service
@@ -11,10 +11,7 @@ EnvironmentFile=/etc/default/kubelet
 SuccessExitStatus=143
 ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
 ExecStart=/usr/local/bin/kubelet \
-        --enable-server \
         --node-labels="${KUBELET_NODE_LABELS}" \
-        --v=2 {{if NeedsContainerd}}--container-runtime=remote --runtime-request-timeout=15m --container-runtime-endpoint=unix:///run/containerd/containerd.sock{{end}} \
-        --volume-plugin-dir=/etc/kubernetes/volumeplugins \
         $KUBELET_CONFIG
 [Install]
 WantedBy=multi-user.target
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -382,13 +382,19 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
   content: |
     #!/bin/bash
     set -e
+    MOUNT_DIR=/var/lib/kubelet
+    mkdir -p $MOUNT_DIR /var/lib/cni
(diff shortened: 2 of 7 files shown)


parts/k8s/cloud-init/artifacts/kubelet.service
```diff
@@ -10,20 +10,8 @@ Restart=always
...
 SuccessExitStatus=143
 ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
-ExecStartPre=/bin/mkdir -p /var/lib/kubelet
-ExecStartPre=/bin/mkdir -p /var/lib/cni
-ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
-ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
-{{/* This is a partial workaround to this upstream Kubernetes issue: */}}
-{{/* https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731 */}}
-
-ExecStartPre=-/sbin/ebtables -t nat --list
```
jackfrancis, line 20: I vote we get rid of this extra stdout at the beginning of the kubelet startup. There are better ways to accomplish that, if we want, and `journalctl -u kubelet` reporting that is misleading as the iptables/ebtables rules do not originate from the kubelet runtime.

parts/k8s/cloud-init/artifacts/kubelet.service
```diff
@@ -10,20 +10,8 @@ Restart=always
 EnvironmentFile=/etc/default/kubelet
 SuccessExitStatus=143
 ExecStartPre=/bin/bash /opt/azure/containers/kubelet.sh
-ExecStartPre=/bin/mkdir -p /var/lib/kubelet
-ExecStartPre=/bin/mkdir -p /var/lib/cni
-ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
-ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
-{{/* This is a partial workaround to this upstream Kubernetes issue: */}}
-{{/* https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731 */}}
```
jackfrancis, line 18: I wasn't able to find what in here relates to the above (closed) issue, so I suspect it's an obsolete comment.

parts/k8s/cloud-init/artifacts/kubelet.service
```diff
@@ -10,20 +10,8 @@ Restart=always
...
-
-ExecStartPre=-/sbin/ebtables -t nat --list
-ExecStartPre=-/sbin/iptables -t nat --numeric --list
 ExecStart=/usr/local/bin/kubelet \
-        --enable-server \
         --node-labels="${KUBELET_NODE_LABELS}" \
```
jackfrancis, line 14: As far as I can tell we want to keep the `--node-labels` configuration like this so we can declare the value of `KUBELET_NODE_LABELS` at ARM deployment runtime. The `KUBELET_CONFIG` env var is statically defined in the cloud-init template at generation time, so arbitrary ARM variable/parameters injection isn't possible via that vector.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -382,13 +382,19 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
+    fi
+    mount --make-shared $MOUNT_DIR
     PRIVATE_IP=$(hostname -i | cut -d" " -f1)
 {{- if IsMasterVirtualMachineScaleSets}}
     PRIVATE_IP=$(hostname -i | cut -d" " -f1)
     sed -i "s|<SERVERIP>|https://$PRIVATE_IP:443|g" "/var/lib/kubelet/kubeconfig"
 {{end}}
 {{- if gt .MasterProfile.Count 1}}
-    # Redirect ILB (4443) traffic to port 443 (ELB) in the prerouting chain
+    {{- /* Redirect ILB (4443) traffic to port 443 (ELB) in the prerouting chain */}}
```
jackfrancis, line 397: While I'm in here...

parts/k8s/cloud-init/artifacts/kubelet.service
```diff
@@ -10,20 +10,8 @@ Restart=always
...
-ExecStartPre=/bin/mkdir -p /var/lib/cni
-ExecStartPre=/bin/bash -c "if [ $(mount | grep \"/var/lib/kubelet\" | wc -l) -le 0 ] ; then /bin/mount --bind /var/lib/kubelet /var/lib/kubelet ; fi"
-ExecStartPre=/bin/mount --make-shared /var/lib/kubelet
-{{/* This is a partial workaround to this upstream Kubernetes issue: */}}
-{{/* https://github.com/kubernetes/kubernetes/issues/41916#issuecomment-312428731 */}}
-
-ExecStartPre=-/sbin/ebtables -t nat --list
-ExecStartPre=-/sbin/iptables -t nat --numeric --list
 ExecStart=/usr/local/bin/kubelet \
-        --enable-server \
```
jackfrancis, line 23: `--enable-server` is the default, we don't need to explicitly set it

ref: https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/

PR #3179, opened 2020-05-01
cherry-pick: modify container runtime data dir

alexeldeib:
**Reason for Change**:

cherry-picks #3072 for AKS

**Issue Fixed**:

na

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


alexeldeib:
@xuto2 @bowen5 friendly ping for review if you have a chance

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -246,7 +214,7 @@ write_files:
     {{end}}
 {{end}}
 
-{{if IsNSeriesSKU .}}
+{{if IsNSeriesSKU .VMSize}}
```
xuto2, line 217: just to confirm - is this change in AKS RP?

alexeldeib: No, it is not in RP. Isn’t the point of this PR to vendor AKS engine so we can use it in RP? How would I be able to make the change first there?

Sorry for the confusion :(

pkg/api/common/helper.go
```diff
@@ -337,3 +340,268 @@ func WrapAsParameter(s string) string {
...
+	}
+
+	dataDir, ok := opts[ContainerDataDirKey]
+	if ok {
+		config.DataRoot = dataDir
+	}
+
+	b, err := json.MarshalIndent(config, "", "    ")
+	return string(b), err
+}
```
xuto2, line 361: I don't quite like we move the change from template to go. Why not do it in template?

alexeldeib: The resulting output didn’t have correct indentation when I used go template originally, I changed it to this after.

pkg/api/common/helper.go
```diff
@@ -337,3 +340,268 @@ func WrapAsParameter(s string) string {
...
+// ContainerdSandboxImageOverrider produces a function to transform containerd config by setting the SandboxImage.
+func ContainerdSandboxImageOverrider(image string) func(*ContainerdConfig) error {
+	return func(config *ContainerdConfig) error {
+		config.Plugins.IoContainerdGrpcV1Cri.SandboxImage = image
+		return nil
+	}
+}
+
+// DockerNvidiaOverride transforms a docker config to supply nvidia runtime configuration.
+func DockerNvidiaOverride(config *DockerConfig) error {
```
xuto2, line 398: this can also be a template snippet?

alexeldeib: This applies transformations to the struct, which are serialized all at once. I tried doing everything in the template in my original PR, but it was very difficult to guarantee correct serialization. 

So I converted everything to structs and serialize it once, and unit test against the serialization for both cases.

xuto2: not sure I understand the issue, if it's just template then the unit test just compare the templates. why do we need serialization?

pkg/api/common/helper.go
```diff
@@ -337,3 +340,268 @@ func WrapAsParameter(s string) string {
...
+		for i := 0; i < spaces; i++ {
+			out.WriteString(" ")
+		}
+		out.WriteString(scanner.Text())
+		out.WriteString("\n")
+	}
+	return out.String()
+}
+
+func GetDockerConfigTestCases() map[string]string {
```
xuto2, line 424: why is this not in test files?

pkg/engine/template_generator_test.go
```diff
@@ -1023,6 +1027,151 @@ func TestGetIdentitySystem(t *testing.T) {
...
+		},
+		{
+			name:  "all",
+			vmSKU: "Standard_NC12s_v2",
+			runtimeConfig: map[string]string{
+				"dataDir": "/mnt/docker",
+			},
+		},
+	}
+	expectedOutputs := common.GetDockerConfigTestCases()
```
alexeldeib, line 1061: @xuto2 here it is

pkg/api/common/helper.go
alexeldeib: see other comment

pkg/api/common/helper.go
alexeldeib: Why allow for humans to make errors in whitespace? I made several typos in the original PR and aks-engine team seemed to agree it was less error prone

alexeldeib: I don't see why serialization is worse than templating, if not on par?

alexeldeib: I can change it if you really want, but what does that mean in terms of AKS vs AKS-engine? do we diverge forever? Do I cherry pick that back into AKS-engine? Really want to clarify the process here.

parts/k8s/cloud-init/nodecustomdata.yml
```diff
@@ -168,21 +168,7 @@ write_files:
...
-      }{{if IsNSeriesSKU .}}
-      ,"default-runtime": "nvidia",
-      "runtimes": {
-         "nvidia": {
-             "path": "/usr/bin/nvidia-container-runtime",
-             "runtimeArgs": []
-        }
-      }{{end}}
-    }
+{{IndentString (GetDockerConfig (IsNSeriesSKU .VMSize)) 4}}
```
xuto2, line 171: is there a way to toggle this on/off? What if the changes doesnt work as expect?

alexeldeib: The default change implemented by the PR is a no-op. In RP, the implementation would be as a toggle that sets the config value to e.g. "/mnt/docker". 

I don't know of another way to toggle in yaml directly.

pkg/api/common/helper.go
```diff
@@ -337,3 +340,268 @@ func WrapAsParameter(s string) string {
...
+func GetDockerConfig(opts map[string]string, overrides []func(*DockerConfig) error) (string, error) {
+	config := GetDefaultDockerConfig()
+
+	for i := range overrides {
+		if err := overrides[i](&config); err != nil {
+			return "", err
+		}
+	}
+
+	dataDir, ok := opts[ContainerDataDirKey]
```
xuto2, line 354: If I understand correctly, this is the real change we want to conditionally apply. If so, it seems we can just have another docker daemon template, and choose one of the two templates based on the condition. No?

alexeldeib: This PR is intended make arbitrary configuration changes to containerd/docker easier in the future. That's another reason I opted to model everything as structs. We probably will want to do additional tuning in the future and I'd prefer defaulting/mutating structs to growth of template options.

pkg/api/common/helper.go
juan-lee: @xuto2 why should this be a template snippet instead? I think @alexeldeib makes a good point about reducing potential errors.

parts/k8s/cloud-init/nodecustomdata.yml
palma21: Yeah, suggest a pair review on this one since this one will be tough to ingest on the current setup. I booked some time tomorrow afternoon to discuss open points them and allow some more discussion until then. 

Thanks for the effort folks

pkg/api/common/helper.go
xuto2: my only intention is to be consistent with current template approach and don't want to have mixed implementations. My question here is trying to understand why template is not possible, not saying we cannot use serialization.

parts/k8s/cloud-init/nodecustomdata.yml
xuto2: got you. thanks.

parts/k8s/cloud-init/nodecustomdata.yml
xuto2: nvm. I see the VMSize property in the agent pool.

PR #3184, opened 2020-05-02
chore: check all error returns in main code

### LINKED ISSUE #1032: Enable errcheck linter

mboersma:
**Reason for Change**:
Ignoring an error returned from a func can be a dangerous thing in go, so enabling the errcheck linter is good practice. This contains only the changes necessary to main code. See #3183 for changes to test code.

**Issue Fixed**:
Fixes #1032.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/azureclient.go
+++ b/pkg/armhelpers/azureclient.go
@@ -163,7 +163,9 @@ func NewAzureClientWithDeviceAuth(env azure.Environment, subscriptionID string) 
 	if err != nil {
 		return nil, err
 	}
-	armSpt.Refresh()
+	if err = armSpt.Refresh(); err != nil {
+		return nil, err
+	}
 
 	adRawToken := armSpt.Token()
 	adRawToken.Resource = env.GraphEndpoint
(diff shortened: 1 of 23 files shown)


pkg/armhelpers/converter.go
```diff
@@ -10,11 +10,10 @@ import (
 // DeepCopy dst and src should be the same type in different API version
 // dst should be pointer type
 func DeepCopy(dst, src interface{}) error {
-	defer func() error {
```
jackfrancis, line 13: Where were these deferred `err` responses going prior to this change? Does the caller to `DeepCopy` receive them?

pkg/armhelpers/azureclient.go
```diff
@@ -211,7 +217,6 @@ func NewAzureClientWithClientSecretExternalTenant(env azure.Environment, subscri
 	if err != nil {
 		return nil, err
 	}
-	graphSpt.Refresh()
```
mboersma, line 214: This was returning a consistent error (once we started checking it). Need to investigate more.

pkg/armhelpers/converter.go
mboersma: I think they were being ignored, because the only way to return a value from a `defer` in a func is a named result parameter. I'll try that change instead, because it would preserve the intent here and hopefully also placate the linter.

pkg/armhelpers/azureclient.go
mboersma: I changed this and similar calls to `log.Error(err)` so as not to change existing control flow.

pkg/armhelpers/converter.go
mboersma: I changed this and the other `DeepCopy`...um...copy... as specified above.

PR #3187, opened 2020-05-04
refactor: use simpler file names for addons maintenance

jackfrancis:
**Reason for Change**:

This PR eschews the superfluous `kubernetesmasteraddons-` filename prefix for those addons that have it. The scope of this change is only addons who are singularly maintained in one file under `parts/k8s/addons/`. The remaining addons that use this prefix are decomposed into  per-Kubernetes version child directories. We will consolidate those into single files, and do the renaming then, one PR per addon, to keep the changesets and commit history comprehensible.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/const.go
```diff
@@ -225,8 +225,8 @@ const (
 	reschedulerAddonDestinationFilename            string = "kube-rescheduler-deployment.yaml"
 	nvidiaAddonSourceFilename                      string = "kubernetesmasteraddons-nvidia-device-plugin-daemonset.yaml"
 	nvidiaAddonDestinationFilename                 string = "nvidia-device-plugin.yaml"
-	containerMonitoringAddonSourceFilename         string = "kubernetesmasteraddons-omsagent-daemonset.yaml"
-	containerMonitoringAddonDestinationFilename    string = "omsagent-daemonset.yaml"
+	containerMonitoringAddonSourceFilename         string = "container-monitoring.yaml"
```
jackfrancis, line 228: just taking the opportunity to name this one to match the actual addon name

PR #3191, opened 2020-05-04
ci: enable staticcheck linter

mboersma:
**Reason for Change**:
The `staticcheck` linter is enabled by default in golangci-lint, although AKS Engine has had it disabled. It points out where errors are being assigned but not actually checked, and some other simplifications.

**Issue Fixed**:
Refs #1032 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/armhelpers/azurestack/compute.go
```diff
@@ -181,17 +181,15 @@ func (az *AzureClient) GetAvailabilitySet(ctx context.Context, resourceGroup, av
...
-		ss := strings.Split(id, "/")
-		name := ss[len(ss)-1]
-		vmas, err := az.GetAvailabilitySet(ctx, resourceGroup, name)
-		if err != nil {
-			return 0, err
-		}
-		// Assume that all VMASes in the cluster share a value for platformFaultDomainCount
-		count = int(*vmas.AvailabilitySetProperties.PlatformFaultDomainCount)
-		break
+	id := vmasIDs[0]
```
jackfrancis, line 184: should we check for len(vmasIDs) > 0 before doing this?

PR #3194, opened 2020-05-05
feat: cherry pick changes from master to support kubernetes 1.18.1/1.18.2

bowen5:
**Reason for Change**:

Cherry pick changes from aks-engine master branch to bring the kubernetes 1.18.1 support into this branch.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


vhd/packer/install-dependencies.sh
```diff
@@ -422,6 +423,8 @@ pullContainerImage "docker" "busybox"
 echo "  - busybox" >> ${VHD_LOGS_FILEPATH}
 
 K8S_VERSIONS="
+1.18.2
+1.18.1
```
xuto2, line 427: these needs to go to AKS VHD 
https://github.com/Azure/AgentBaker/blob/master/vhdbuilder/packer/install-dependencies.sh

vhd/packer/configure-windows-vhd.ps1
```diff
@@ -80,7 +80,10 @@ function Get-FilesToCacheOnVHD
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.16.6/windowszip/v1.16.6-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.16.7/windowszip/v1.16.7-1int.zip",
             "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.2/windowszip/v1.17.2-1int.zip",
-            "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.3/windowszip/v1.17.3-1int.zip"
+            "https://kubernetesartifacts.azureedge.net/kubernetes/v1.17.3/windowszip/v1.17.3-1int.zip",
+            "https://kubernetesartifacts.azureedge.net/kubernetes/v1.18.0/windowszip/v1.18.0-1int.zip",
```
xuto2, line 84: @AbelHu do you know if the current windows VHDs contains 1.18 binaries?

xuto2: To support windows 1.18 , 
https://github.com/Azure/aks-engine/blob/c22aad45cabfa797eebba40e4acad567aa5a6355/pkg/api/azenvtypes.go#L94

need to change the above setting from 
"https://acs-mirror.azureedge.net/wink8s/",
to:
"https://acs-mirror.azureedge.net/kubernetes/<k8sversion>/windowszip/",
e.g:
https://acs-mirror.azureedge.net/kubernetes/v1.18.2/windowszip/v1.18.2-1int.zip
https://acs-mirror.azureedge.net/kubernetes/v1.18.1/windowszip/v1.18.1-1int.zip

xuto2: This is because upstream moved to a new storage account and they changed the path. 
The agreement is that they backfilled all the previous versions to the OLD path in the new storage account, but for any new k8s version binaries (1.18 being the first one) , they won't. Instead they ask us to use the new path.

Otherwise the raw windows test egress lockdown test will fail, as we need to download the kubelet from this url, and it will fail for 1.18, if we don't change it.

xuto2: As a reminder, can you also do an AKS RP Raw Ubuntun E2E test for 1.18 and see if it works or not?

bowen5: Thank you @xuto2 !

- As offline discussed, I verified Raw Ubuntu by running "E2Ev2 AKS RP Check-In Test (Default Toggle)" pipeline and enforcing test against 1.18.1. Test of "Test_E2Ev2 Scenario_RawUbuntu_Egress_Lockdown" passed. Corresponding pipeline build id is "30845438" if you want to double check. (I'm not sure whether it is appropriate to paste the DevOps link here, so just put the build id)

- You are right that for 1.18 there is no zip under url "https://acs-mirror.azureedge.net/wink8s/". Thank for sharing the details. I changed the `KubeBinariesSASURLBase` to point to "https://acs-mirror.azureedge.net/kubernetes/", and based on the code https://github.com/Azure/aks-engine/blob/e579e5b3864bb0b9dfb582cbe335b6cac98548d2/pkg/api/k8s_versions.go#L238
the generated value would be e.g. "https://acs-mirror.azureedge.net/kubernetes/v1.18.1/windowszip/v1.18.1-1int.zip", "https://acs-mirror.azureedge.net/kubernetes/v1.17.3/windowszip/v1.17.3-1int.zip" as you said. Note that in my latest iteration I reverted one of your PR https://github.com/Azure/aks-engine/commit/8db438e5ac5f6716a18a4d6dae2fa2aab92d4a5a, to make all versions to follow above pattern instead of just 1.18.x. I verified that for all supported version (1.13.11, 1.13.12, 1.14.7, 1.14.8, 1.15.7, 1.15.9, 1.15.10, 1.16.6, 1.16.7, 1.17.2, 1.17.3, 1.18.1, 1.18.2), zip url with new pattern contains valid result. Please let me know if I should only change the url pattern for 1.18.

- Latest windows VHD used in AKS already contains 1.18 binaries, by checking https://github.com/Azure/AKS/blob/master/CHANGELOG.md and https://github.com/Azure/aks-engine/blob/master/vhd/release-notes/aks-windows/2019-datacenter-core-smalldisk-17763.1158.200421.txt. That also explained why my "Test_E2Ev2 Scenario_Windows" test against 1.18 passed even through my previous `KubeBinariesSASURLBase` for 1.18 is invalid. By checking the code https://github.com/Azure/aks-engine/blob/e579e5b3864bb0b9dfb582cbe335b6cac98548d2/parts/k8s/kuberneteswindowsfunctions.ps1#L27, windows cse will first find from cache before performing real download. Since VHD already contains 1.18 binaries, downloading logic against invalid `KubeBinariesSASURLBase` never performed. @AbelHu please correct if I'm wrong for above statements. Thank you.

PR #3201, opened 2020-05-05
feat: enable alternate kube-reserved cgroups

alexeldeib:
**Reason for Change**:
Kubernetes enforces resource control through cgroups. Kubelet understands several cgroup configuration options like `--kube-reserved-cgroup`, `--system-reserved-cgroup` which specify how to partition a system for resource control. Kubelet subtracts the resource quantities specified through kube-reserved and system-reserved from total node resources to arrive at node allocatable.

Currently AKS-Engine exposes the kubelet flags for configuring enforcement of these options, but there's no way to actually set up the cgroups beforehand. This PR makes that possible by creating a systemd slice and appropriate drop-ins for Kubelet and the container runtime to exist in. That is, when using the features enabled in this PR a cgroup will be created which corresponds to `kube-reserved`, leaving other system daemons in the system slice. The system slice roughly acts as `system-reserved` when a user specifies kubeReservedCgroup.

This PR **does not** align enforcement of the provided value against kubelet -- we could probably do this by checking for a flag that matches `--kube-reserved-cgroup` , but I figured I'd open as is for feedback.

**Issue Fixed**:
consider this PR the feature request 😄 

**Requirements**:
Can place kubelet and docker into arbitrary systemd slices.

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


alexeldeib:
@jackfrancis curious how you feel about this approach vs trying to detect and/or match more tightly what the user might provide as a raw kubelet flag. I found this a bit more palatable, even if it requires more manual jiggering.

alexeldeib:
Cc @xuto2

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -68,6 +68,47 @@ write_files:
...
+  owner: root
+  content: |
+    [Unit]
+    Description=Limited resources slice for Kubernetes services
+    Documentation=man:systemd.special(7)
+    DefaultDependencies=no
+    Before=slices.target
+    Requires=-.slice
+    After=-.slice
+    #EOF
```
jackfrancis, line 83: This `#EOF` sentinel chars pattern is meant to work in concert w/ the `wait_for_file` func in the CSE bootstrap scripts. If the following is true:

- in the event that CSE runs prior to cloud-init being finished paving the filesystem, these missing files will prevent CSE from completing successfully/deterministically

...then, we should keep the `#EOF` and add the appropriate `wait_for_file` invocations against these new files (probably inside the `ensureKubelet` func). Unfortunately, becase the CSE files themselves do not operate against a per-pool or per-master context, we have to pass that context via env vars.

Either that, or we can generalize this approach and pick a "default " cgroup so that we *always* deliver this new cgroup implementation flavor, and allow the value of `kubeReservedCgroup` to be user-configured (we could apply that per-node/per-master user configuration in cloud-init, and skip the CSE env var boilerplate thing).

Hope that makes sense!

alexeldeib: I'm going to always deliver the cgroup and default it to kubereserved, but not run anything in it. That's effectively a no-op unless users manually move stuff into the slice.

alexeldeib: Is it just me, or should https://github.com/Azure/aks-engine/blob/a43a9c4fcc20cee39d0c7cce84af97734d10686f/parts/k8s/cloud-init/artifacts/cse_config.sh#L317 wait for the file to exist?

jackfrancis: Yeah, I'm not sure what's going on there...

alexeldeib: meh, this got a little tricky since the name of the file determines the slice. That doesn't work for the current setup with the system slice, since it already exists (and not via this filepath). I don't really think there's any harm forcing users to pick a slice, but I'm also totally fine with doing this clusterwide and evolving it if we need. That's what I stuck with for now.

alexeldeib:
Creating some monday morning notifications, let me know how this looks to you @jackfrancis?

docs/topics/clusterdefinitions.md
```diff
@@ -65,6 +65,7 @@ $ aks-engine get-versions
...
+| kubeReservedCgroup               | no                        | The name of a systemd slice to create for containtment of both kubelet and the container runtime. This should not point to an existing systemd slice. Defaults to the empty string, which means kubelet and the container runtime will be placed in the system slice.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
jackfrancis, line 68: I think it's more accurate to say:

> Defaults to "kubereserved", which means kubelet and the container runtime will be placed in the system slice.

Is that right?

alexeldeib: No, I think the current phrasing is correct. The default behavior is always to put them in the system slice unless specified. If you specify anything else, that slice will be created and the units will be placed inside it.

Will update for clarity

pkg/api/defaults.go
```diff
@@ -434,6 +434,10 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 			o.KubernetesConfig.ContainerRuntimeConfig = make(map[string]string)
 		}
 
+		if o.KubernetesConfig.KubeReservedCgroup == "" {
```
jackfrancis, line 437: Based on this, it is actually not possible for a fully populated (post-defaults enforcement) api model to have a `""` value of `KubeReservedCgroup`. In which case, the `HasKubeReservedCgroup` func is redundant, right?

docs/topics/clusterdefinitions.md
jackfrancis: I think the "Defaults to the empty string" is what's confusing to me, as the default of that property will actually be "kubereserved".

pkg/api/defaults.go
alexeldeib: whoops, I'll remove this

alexeldeib: This is leftover from playing around with per-nodepool and defaulting, but it didn't work out the way i'd like

pkg/api/common/const.go
```diff
@@ -381,3 +381,7 @@ const (
 	// AADPodIdentityTaintKey is the node taint key for AAD Pod Identity-enabled clusters before NMI daemonset is ready
 	AADPodIdentityTaintKey string = "node.kubernetes.io/aad-pod-identity-not-ready"
 )
+
+const (
+	DefaultKubeReservedCgroup string = "system"
```
alexeldeib, line 386: n.b., unused, will remove

pkg/api/defaults.go
jackfrancis: That explains everything! :)

docs/topics/clusterdefinitions.md
```diff
@@ -65,6 +65,7 @@ $ aks-engine get-versions
...
+| kubeReservedCgroup               | no                        | The name of a systemd slice to create for containtment of both kubelet and the container runtime. When this value is a non-empty string, a file will be dropped at `/etc/systemd/system/$KUBE_RESERVED_CGROUP.slice` creating a systemd slice. Both kubelet and docker will run in this slice. This should not point to an existing systemd slice. If this value is unspecified or specified as the empty string, kubelet and the container runtime will run in the system slice by default.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   |
```
jackfrancis, line 68: s/containtment/containment

jackfrancis: Also, to be clear the default `"kubeReservedCgroup": ""` results in no functional change compared to current aks-engine:master, correct?

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -306,6 +306,7 @@ installContainerd() {
 ensureContainerd() {
   wait_for_file 1200 1 /etc/systemd/system/containerd.service.d/exec_start.conf || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   wait_for_file 1200 1 /etc/containerd/config.toml || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  wait_for_file 1200 1 /etc/systemd/system/containerd.service.d/kubereserved-slice.conf|| exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
```
jackfrancis, line 309: This should be inside a `{{- if HasKubeReservedCgroup}}` block, no?

alexeldeib: Nice catch

My last set of commits...not so clean

docs/topics/clusterdefinitions.md
alexeldeib: > no functional change compared to current aks-engine:master, correct
Correct, that is the intent.

Will fix spelling

test/e2e/test_cluster_configs/everything.json
```diff
@@ -23,9 +23,11 @@
 			"orchestratorProfile": {
 				"orchestratorType": "Kubernetes",
 				"kubernetesConfig": {
+					"kubeReservedCgroup": "kubesystem",
 					"containerRuntimeConfig": {
 						"dataDir": "/mnt/docker"
 					},
+					"kubeReservedCgroup": "kubesystem",
```
jackfrancis, line 30: warning: dups!

alexeldeib: *cries in merge commits*

Issue #3188, opened 2020-05-04
consolidate addons specs in single files


jackfrancis:
Right now we maintain Kubernetes-version-specific addons specs in a discrete, per-Kubernetes-specific directory under `parts/k8s/addons/`. That organization was established prior to processing the addon specs through a go template layer.

Now that we do have access to the Kubernetes version context via go template expressions, we can apply per-version specs in a single file for each addon under `parts/k8s/addons/`. This will make long-term maintenance of the addons much more trivial and easier to understand/document.

mboersma:
Let's do this one addon at a time, to make reviewing easier.

jackfrancis:
Yep, here's the first one: https://github.com/Azure/aks-engine/pull/3189

PR #3211, opened 2020-05-06
refactor: consolidate metrics-server addon spec

jackfrancis:
**Reason for Change**:

This PR consolidates the maintenance of the `metrics-server` addon into a single go template-aware YAML file.

**Issue Fixed**:

Addresses #3188 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/addons.go
```diff
@@ -254,6 +255,10 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		},
 	}
 
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.16.0") {
+		defaultMetricsServerAddonsConfig.Mode = AddonModeReconcile
```
mboersma, line 259: Was there a good reason for `EnsureExists` here before 1.16? It's odd that they differed.

Also if `Reconcile` is the default from 1.16 onward, maybe it should be the default and this block could set `EnsureExists` for earlier versions. Then when 1.16 support drops off, this block could be deleted altogether.

pkg/engine/const.go
```diff
@@ -199,8 +199,8 @@ const (
 const (
 	heapsterAddonSourceFilename                    string = "kubernetesmasteraddons-heapster-deployment.yaml"
 	heapsterAddonDestinationFilename               string = "kube-heapster-deployment.yaml"
-	metricsServerAddonSourceFilename               string = "kubernetesmasteraddons-metrics-server-deployment.yaml"
-	metricsServerAddonDestinationFilename          string = "kube-metrics-server-deployment.yaml"
+	metricsServerAddonSourceFilename               string = "metrics-server.yaml"
```
mboersma, line 202: At some point maybe we should fold source and destination into just a single `metricsServerAddonFilename`, for example.

jackfrancis: yeah, that would probably be easier, just not changing that yet becasue reasons

pkg/api/addons.go
jackfrancis: I can imagine that we decided to prefer `Reconcile` at a time when we were introducing 1.15, and didn't want to change pre-existing versions?

I like your idea.

PR #3218, opened 2020-05-08
docs: Update quickstart.md to add scoop install guide

elemount:
**Reason for Change**:

Scoop is a good tool for install packages. By this PR https://github.com/ScoopInstaller/Main/pull/1010 , scoop can install aks-engine. Scoop will help user to download the aks-engine from Github release and it is secure for users. 

**Issue Fixed**:

None

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/docs/tutorials/quickstart.md
+++ b/docs/tutorials/quickstart.md
@@ -19,7 +19,7 @@ You can also choose to install the `aks-engine` binary using [gofish][gofish-abo
 
 On macOS, you can install the `aks-engine` binary with [Homebrew][homebrew]. Run the command `brew install Azure/aks-engine/aks-engine` to do so. You can install Homebrew following these [instructions][homebrew-install].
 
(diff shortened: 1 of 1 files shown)


mboersma:
Hey, that's great--thanks for doing that @elemount!

So I assume we need to create a new PR at Scoop to match each new release of aks-engine? If so, we should add that to the [release docs](https://github.com/Azure/aks-engine/blob/master/docs/community/release-checklist.md) so we don't forget in the future.

elemount:
> Hey, that's great--thanks for doing that @elemount!
> 
> So I assume we need to create a new PR at Scoop to match each new release of aks-engine? If so, we should add that to the [release docs](https://github.com/Azure/aks-engine/blob/master/docs/community/release-checklist.md) so we don't forget in the future.

Hi @mboersma , it does not needed. Scoop main bucket have a auto check-and-pr tooling and it will daily check the github release page and try to find if there is any version released. So if the bits name pattern is not changed, aks-engine does not need to notify the new version, scoop can automatically pull and update the version.

mboersma:
> scoop can automatically pull and update the version

Excellent! Thanks again for doing this.

PR #3220, opened 2020-05-10
Fixed wrong az group create command

agowdamsft:
Correct command is az group deployment

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/tutorials/quickstart.md
+++ b/docs/tutorials/quickstart.md
@@ -235,7 +235,7 @@ Now we can deploy the files `azuredeploy.json` and `azuredeploy.parameters.json`
 Using the CLI:
 
 ```console
-$ az deployment group create \
+$ az group deployment create \
     --name "contoso-apple-k8s" \
     --resource-group "contoso-apple-5eac6ed8" \
     --template-file "./_output/contoso-apple-5eac6ed8/azuredeploy.json" \


agowdamsft:
assign to @ksubrmnn for approval

mboersma:
@agowdamsft `az group deployment` has been deprecated for several releases of the `az` client now. The preferred syntax is now `az deployment group`. You may have an older version of the CLI that can be updated.

See #3213 for details.

agowdamsft:
I was looking into these docs @mboersma  https://docs.microsoft.com/en-us/cli/azure/group/deployment?view=azure-cli-latest

mboersma:
I guess they haven't updated those docs yet. But a few months ago the `az` client began emitting warnings for `az group deployment`:

```shell
% az group deployment -h  

Group
    az group deployment : Manage Azure Resource Manager deployments.
        This command group has been deprecated and will be removed in a future release. Use
        'deployment group' instead.
...
```

We call this command programmatically in a few places in tests, so I changed those to `az deployment group` to silence the warnings.

I did the same to our docs, but maybe that was a bit premature. If you use an `az` CLI from 2019 or earlier, you may not yet have this change.

PR #3225, opened 2020-05-11
chore: update nvidia device plugin

sozercan:
**Reason for Change**:

updates nvidia-device-plugin version and moves the image to mcr 

Note for versioning change from nvidia's side: https://github.com/NVIDIA/k8s-device-plugin#versioning

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
1.0.0-beta.6 > 1.11 is confusing, but documented as you pointed out.

vhd/packer/install-dependencies.sh
```diff
@@ -296,9 +296,10 @@ done
 
 NVIDIA_DEVICE_PLUGIN_VERSIONS="
 1.11
```
sozercan, line 298: do we still need `1.11`?

mboersma: No, not if 1.0.0-beta.6 replaced it. 

We don't need to keep anything that's not referenced on the VHD (AKS Engine used to do very lazy garbage collection here when the image was shared with AKS, but that's not necessary now).

sozercan: thanks! removed

sozercan:
@mboersma yea nvidia also deprecated nvidia-docker2 but it's still required for k8s 😕  https://github.com/NVIDIA/nvidia-docker#upgrading-with-nvidia-docker2-deprecated

mboersma:
I ran the `cuda-vector-add` test against a new Standard_NC12 cluster off this branch, and it passed:

```shell
% k logs -f cuda-vector-add    
[Vector addition of 50000 elements]
Copy input data from the host memory to the CUDA device
CUDA kernel launch with 196 blocks of 256 threads
Copy output data from the CUDA device to the host memory
Test PASSED
Done
```

The nodes report `nvidia.com/gpu: "2"` and `cpu: "12"` and that the nvidia-device-plugin pods are running
```
    Image:          mcr.microsoft.com/oss/nvidia/k8s-device-plugin:1.0.0-beta6
```

PR #3226, opened 2020-05-11
fix: don't validate Windows version in update scenario

### LINKED ISSUE #3217: Unable to upgrade Windows cluster
**Describe the bug**
aks-engine upgrade outputs following error:

> Error: loading existing cluster: error parsing the api model: Orchestrator Kubernetes version 1.15.7 does not support Windows

I think it is caused by one change in https://github.com/Azure/aks-engine/pull/2930.
https://github.com/Azure/aks-engine/blob/master/pkg/api/vlabs/validate.go#L1176

**Steps To Reproduce**
Build cluster with Windows pool by AKS-Engine 0.45.0, k8s 1.15.7
Upgrade cluster to 1.16 by AKS-Engine 0.50.0

**AKS Engine version**
0.50.0

**Kubernetes version**
1.15.7

jackfrancis:
**Reason for Change**:

This PR doesn't throw a validation error for the Kubernetes Version on a Windows-enabled cluster during scale/upgrade contexts.

**Issue Fixed**:

Fixes #3217

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -173,7 +173,7 @@ func (a *Properties) validate(isUpdate bool) error {
 		return e
 	}
 
-	if e := a.validateWindowsProfile(); e != nil {
+	if e := a.validateWindowsProfile(isUpdate); e != nil {
 		return e
 	}
 	return nil
@@ -1146,7 +1146,7 @@ func validateVMSS(o *OrchestratorProfile, isUpdate bool, storageProfile string) 
 	return nil
 }
 
-func (a *Properties) validateWindowsProfile() error {
+func (a *Properties) validateWindowsProfile(isUpdate bool) error {
 	hasWindowsAgentPools := false
 	for _, profile := range a.AgentPoolProfiles {
 		if profile.OSType == Windows {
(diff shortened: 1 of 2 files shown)


jackfrancis:
@marosset I'll do some manual validation on this, but I think functionally this is what we want to address #3217

jackfrancis:
Validated manually

pkg/api/vlabs/validate.go
```diff
@@ -1176,7 +1176,7 @@ func (a *Properties) validateWindowsProfile() error {
 			false,
```
marosset, line 1176: Should we instead figure out how to plumb in isUpdate here instead of hard-coding the false? 
I think that is the initial cause of the issues.

marosset: (But if that is a lot of work I'm fine with these changes)

jackfrancis: Oh, maybe you're right that that's the way to do this! The current UT in this PR should validate if it's as easy as that.

marosset: Nice - I think I broke this when i was refactoring windows validation in a previous PR :(

PR #3228, opened 2020-05-11
fix: custom cloud Azure CNI config for Windows

jackfrancis:
**Reason for Change**:

This PR only adds the `"environment": "mas"` Azure CNI config if we're in an Azure Stack custom cloud profile context.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -353,7 +353,8 @@ try
                 -KubeClusterCIDR $global:KubeClusterCIDR `
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
-                -VNetCIDR $global:VNetCIDR -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
+                -VNetCIDR $global:VNetCIDR `
+                -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
 
             if ($TargetEnvironment -ieq "AzureStackCloud") {
                 GenerateAzureStackCNIConfig `
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -354,6 +354,7 @@ try
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
                 -VNetCIDR $global:VNetCIDR `
+                {{- /* Azure Stack has discrete Azure CNI config requirements */}}
                 -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
 
             if ($TargetEnvironment -ieq "AzureStackCloud") {
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -353,7 +353,7 @@ try
                 -KubeClusterCIDR $global:KubeClusterCIDR `
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
-                -VNetCIDR $global:VNetCIDR{{if IsAzureStackCloud}} -TargetEnvironment $TargetEnvironment{{end}}
+                -VNetCIDR $global:VNetCIDR -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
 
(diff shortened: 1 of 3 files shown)


parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -354,8 +354,9 @@ try
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
                 -VNetCIDR $global:VNetCIDR `
```
marosset, line 356: I think you need to capture the backtick at the end of this line too

marosset: Can you add a blank line after {{end}} too?

the backtick means the command is continuted on to the next line and in the non-azure stack case this could get interpreted as the if statement geting considered part of the Set-AzureCNIConfig call resulting in everything getting confused

jackfrancis: So the backtick continues the statement on the next line? I could just make the if statement inline, after the `-VNetCIDR` option as well

parts/k8s/windowsazurecnifunc.ps1
```diff
@@ -68,7 +68,7 @@ Set-AzureCNIConfig
         $KubeServiceCIDR,
         [Parameter(Mandatory=$true)][string]
         $VNetCIDR,
-        [Parameter(Mandatory=$true)][string]
+        [Parameter(Mandatory=$false)][string]
         $TargetEnvironment
```
marosset, line 72: If this is no longer required can you set to a default value by turning the last line into
`$TargetEnvironment = ""`

marosset: Or maybe we should always pass this and set the value differently - that would clean up the Set-AzureCNIConfig call too i think

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -353,8 +353,7 @@ try
                 -KubeClusterCIDR $global:KubeClusterCIDR `
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
-                -VNetCIDR $global:VNetCIDR `
-                -TargetEnvironment $TargetEnvironment
+                -VNetCIDR $global:VNetCIDR -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
```
marosset, line 356: Can you move the if/else to a variable assignment before the Set-AzureCNICall and all add a comment explaining what is actually going on?

marosset: and revert back to one param per line (for consistency?)

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -354,7 +354,7 @@ try
                 -MasterSubnet $global:MasterSubnet `
                 -KubeServiceCIDR $global:KubeServiceCIDR `
                 -VNetCIDR $global:VNetCIDR `
-                -TargetEnvironment $TargetEnvironment
+                -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
```
jackfrancis, line 357: @marosset what's the value of creating (and using) a new variable to represent the true/false Azure Stack context? why not just inject the `$true` or `$false` values via go template data model introspection?

marosset: Mostly i wanted a comment near that explaining what was happening.
I'd be ok just adding a comment here with the if/else left inline.

marosset: but i guess this is cleaned up a lot since the initial PR so we don't need a comment :)

jackfrancis: All good I heart comments :)

pkg/engine/templates_generated.go
```diff
@@ -39826,7 +39827,7 @@ Set-AzureCNIConfig
 
     $configJson.plugins.AdditionalArgs[1].Value.DestinationPrefix  = $KubeServiceCIDR
 
-    if ($TargetEnvironment -ieq "AzureStackCloud") {
+    if ($IsAzureStack) {
```
jadarsie, line 39830: is this azure stack specific? 
if this is supposed to apply to all custom clouds, then `IsCustomCloudProfile` may be more appropriate.

jackfrancis:
This change fixed #3245

PR #3231, opened 2020-05-12
fix: update Azure Stack Windows binaries component name

haofan-ms:
**Reason for Change**:
The k8s binary SAS URL for azure stack is inconsistent as the output of the build pipeline (e.g. "v1.15.11-azs/windowszip/v1.15.11-azs-1int.zip"). Update the url to match the output.

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


pkg/api/common/const.go
```diff
@@ -301,6 +301,7 @@ const (
 )
 
 const WindowsArtifactComponentName = "windowszip"
+const WindowsArtifactAzureStackComponentName = ""
```
jadarsie, line 304: this const is the key of the map entry, please use a meaningful name like `windowszip-azs` or similar

PR #3235, opened 2020-05-12
refactor: consolidate nvidia-device-plugin addon spec

mboersma:
**Reason for Change**:
Eases maintenance of the nvidia-device-plugin addon by consolidating manifests into a single YAML file that uses Go templating.

**Issue Fixed**:
Addresses #3188.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
TIL: there never was an nvidia-device-plugin manifest for 1.17. Who knew? 😨

mboersma:
I tested this with a Standard_NC12 cluster on K8s 1.19.0-alpha.3 and it looks good and passes the cuda-vector-add test.

PR #3237, opened 2020-05-12
chore: update zip name for mooncake mirror

tamilmani1989:
**Reason for Change**:

This PR fixes zip name for windows package in mooncake mirror

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


tamilmani1989:
can @sozercan / @jchauncey confirm if this change is good?

sozercan:
@tamilmani1989  I believe we'll need to change the url to `kubernetesartifacts.blob.core.chinacloudapi.cn` (similar to `CSIProxyDownloadURL`) as `mirror.azk8s.cn` doesn't have this artifact.

tamilmani1989:
> @tamilmani1989 I believe we'll need to change the url to `kubernetesartifacts.blob.core.chinacloudapi.cn` (similar to `CSIProxyDownloadURL`) as `mirror.azk8s.cn` doesn't have this artifact.

Ok i see all URLS need to be updated then. I'm not sure about other urls like ContainerdDownloadURLBase, CNIPluginsDownloadURL. do you want me to update for that also?

tamilmani1989:
also why are we not doing for other national clouds but only for mooncake?

sozercan:
I don't think we are hosting those in `kubernetesartifacts.blob.core.chinacloudapi.cn`, just need the change for `VnetCNIWindowsPluginsDownloadURL` and `VnetCNILinuxPluginsDownloadURL`

tamilmani1989:
@sozercan I made the change. Please take a look . 
@mboersma  I dont think e2e is required. I changed for mooncake url and anyway e2e is not going to cover that.

mboersma:
>  I dont think e2e is required

No, but "azp run pr-e2e" is also the way to run the lint checks and unit tests as well.

PR #3259, opened 2020-05-14
fix: resilient etcd data disk mounting

### LINKED ISSUE #3185: Upgrade fails since v.0.50.0 if using availability sets
Upgrade from 1.15.7 to version 1.16.9 fails if using availability sets since v.0.50.0

**Steps To Reproduce**

1. Create a cluster with availability sets, version 1.15.7 and aks-engine v0.48.0
2. Use the aks-engine v0.50.0 to upgrade the cluster to version 1.16.9

**Current behaviour**

Upon upgrading the new master the cse script fails with error code 13
(Hint on linux error code 13 EACCES = Permission denied)

**Expected behavior**

The cluster upgrade should succeed

**AKS Engine version**

v0.48.0 and v0.50.0

**Kubernetes version**

1.15.7 and 1.16.9

**Additional context**

Using availability sets and the basic loadbalancer

### LINKED ISSUE #3246: Do not use device name for etcd mount
**Describe the bug**

Originally reported by @Michael-Sinz. As described in https://docs.microsoft.com/en-us/azure/virtual-machines/troubleshooting/troubleshoot-device-names-problems, device names might change after you restart a Linux VM or reattach the data disks. 

> Normally, the mounts for the OS disk is /dev/sda1 - but this mount it done via a partition GUID and not a device name.
The "temp" disk shows up as /dev/sdb1 and is mounted on /mnt
The etcd disk is /dev/sdc1 and is mounted on /var/lib/etcddisk
However, on the failed boots, the OS disk is something other than /dev/sda1
In fact, the worst case is when the OS disk comes up as /dev/sdc1 - it is still the same disk (data wise) but the device name is not the same.
This lets the OS start but this breaks very quickly when we try to start kubernetes (well, actually, etcd - but that is part of kubernetes).
The reason is that the fstab for the etcd disk specifies mounting /dev/sdc1 to /var/lib/etcddisk (with auto and defaults,nofail)
Doing this ends up making an effective bind mount of the OS disk (/) into /var/lib/etcddisk
This causes etcd to fail as it does not have rights to write to /var/lib/etcddisk since that is the same as (/) and etcd does not have root rights.  
This prevents etcd from starting which then confuses kubelet and the master is out of commission.

We should change the [etcd mount script](https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/mountetcd.sh) to use the UUID or filesystem labels to mount, not the device name `/dev/sda1` which is not guaranteed to stay the same.

See also: https://github.com/kubernetes-sigs/cluster-api/pull/3048/files#diff-f163e4a00e922ec2beac7fb211bba003R221

jackfrancis:
**Reason for Change**:

This PR modifies the etcd data disk mounting implementation thusly:

- `/etc/fstab` does not have a reference to a static `/dev/sd` device, instead it references the filesystem label `etcd_disk`
- we derive the etcd data disk device target by using the lun0 reference under `/dev/disk/azure/scsi1/` 

**Issue Fixed**:

Fixes #3246 
Fixes #3185

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/mountetcd.sh
+++ b/parts/k8s/cloud-init/artifacts/mountetcd.sh
@@ -2,13 +2,15 @@
 # Mounting is done here instead of etcd because of bug https://bugs.launchpad.net/cloud-init/+bug/1692093
 # Once the bug is fixed, replace the below with the cloud init changes replaced in https://github.com/Azure/aks-engine/pull/661.
 set -x
+ETCDDISK=""
 PARTITION=""
 for DISK in $(cat /proc/partitions | grep -o sd[a-z] | uniq); do
   if ! cat /proc/partitions | grep "$DISK"1; then
     if [[ -n $PARTITION ]]; then
       exit 1
     fi
-    PARTITION=/dev/${DISK}1
+    ETCDDISK=/dev/${DISK}
+    PARTITION=${ETCDDISK}1
   fi;
 done
 MOUNTPOINT=/var/lib/etcddisk
@@ -21,7 +23,7 @@ if ! grep "$MOUNTPOINT" /etc/fstab; then
   echo "LABEL=etcd_disk       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
 fi
 if ! ls $PARTITION; then
-  /sbin/sgdisk --new 1 $DISK
+  /sbin/sgdisk --new 1 $ETCDDISK
   /sbin/mkfs.ext4 $PARTITION -L etcd_disk -F -E lazy_itable_init=1,lazy_journal_init=1
 fi
 mount $MOUNTPOINT
--- a/parts/k8s/cloud-init/artifacts/mountetcd.sh
+++ b/parts/k8s/cloud-init/artifacts/mountetcd.sh
@@ -2,31 +2,40 @@
 # Mounting is done here instead of etcd because of bug https://bugs.launchpad.net/cloud-init/+bug/1692093
 # Once the bug is fixed, replace the below with the cloud init changes replaced in https://github.com/Azure/aks-engine/pull/661.
 set -x
+MOUNTPOINT=/var/lib/etcddisk
+LABEL=etcd_disk
 ETCDDISK=""
 PARTITION=""
+ETCDDISK_PERSIST=/opt/azure/containers/etcd_disk_initial
 udevadm settle
-for DISK in $(grep -o -G "sd[a-z]" /proc/partitions | uniq); do
-  if ! grep "$DISK"1 /proc/partitions; then
-    if [[ -n $PARTITION ]]; then
-      exit 1
-    fi
-    ETCDDISK=/dev/${DISK}
-    PARTITION=${ETCDDISK}1
-  fi;
-done
-if [[ -z $ETCDDISK ]]; then
-  exit 1
+if [ ! -f $ETCDDISK_PERSIST ]; then
+  for DISK in $(grep -o -G "sd[a-z]" /proc/partitions | uniq); do
+    if ! grep "$DISK"1 /proc/partitions; then
+      if [[ -n $ETCDDISK ]]; then
+        exit 1
+      fi
+      ETCDDISK=/dev/${DISK}
+      echo "${ETCDDISK}" > /opt/azure/containers/etcd_disk_initial
+    fi;
+  done
+  if [[ -z $ETCDDISK ]]; then
+    exit 1
+  fi
+else
+  ETCDDISK=$(cat ${ETCDDISK_PERSIST})
 fi
-MOUNTPOINT=/var/lib/etcddisk
-mkdir -p $MOUNTPOINT
-umount $MOUNTPOINT
-if ! grep "$MOUNTPOINT" /etc/fstab; then
-  echo "LABEL=etcd_disk       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
-fi
+PARTITION=${ETCDDISK}1
 if ! ls $PARTITION; then
   /sbin/sgdisk --new 1 $ETCDDISK
-  /sbin/mkfs.ext4 $PARTITION -L etcd_disk -F -E lazy_itable_init=1,lazy_journal_init=1
 fi
+if ! blkid -L $LABEL; then
+  /sbin/mkfs.ext4 $PARTITION -L $LABEL -F -E lazy_itable_init=1,lazy_journal_init=1
+fi
+mkdir -p $MOUNTPOINT
+if ! grep "$MOUNTPOINT" /etc/fstab; then
+  echo "LABEL=${LABEL}       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
+fi
+umount $MOUNTPOINT
 mount $MOUNTPOINT
 /bin/chown -R etcd:etcd /var/lib/etcddisk
 #EOF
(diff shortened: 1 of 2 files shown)


CecileRobertMichon:
> we derive the target /dev/sd device programmatically, by inspecting /proc/partitions for disk devices that have no existing filesystem mounts

This won't work if there is more than one data disk, which is possible if the customer specifies disks in AgentPoolProfile.DiskSizesGB.

For now, let's do what @Michael-Sinz suggested and just keep [the label change](https://github.com/Azure/aks-engine/pull/3259/files#diff-68a375c3d9c31b5f27994806a7f564cfR23) so we can fix 18.04 reboots ASAP and then focus on the second problem which is how to partition deterministically when there is more than one data disk as a follow up.

parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -2,20 +2,26 @@
...
 # Once the bug is fixed, replace the below with the cloud init changes replaced in https://github.com/Azure/aks-engine/pull/661.
 set -x
-DISK=/dev/sdc
-PARTITION=${DISK}1
-MOUNTPOINT=/var/lib/etcddisk
+ETCDDISK=""
+PARTITION=""
+for DISK in $(grep -o -G "sd[a-z]" /proc/partitions | uniq); do
+  if ! grep "$DISK"1 /proc/partitions; then
+    if [[ -n $PARTITION ]]; then
```
jackfrancis, line 9: This check is essentially a panic, as the script assumes only a single unpartitioned disk will be found in /proc/partitions on master vms: we deliver a static disk device config:

1. OS disk
2. temp disk
3. 1 data disk (for etcd)

parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -2,20 +2,26 @@
...
 udevadm settle
+MOUNTPOINT=/var/lib/etcddisk
 mkdir -p $MOUNTPOINT
-if mount | grep $MOUNTPOINT; then
-  echo "disk is already mounted"
-  umount /dev/sdc1
-fi
-if ! grep "/dev/sdc1" /etc/fstab; then
-  echo "$PARTITION       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
+umount $MOUNTPOINT
```
jackfrancis, line 19: We can simply always run the `umount` here. If the script only has to be run once, then this will complain:

```
+ umount /var/lib/etcddisk
umount: /var/lib/etcddisk: not mounted.
```

And we just ignore that error and continue along.

parts/k8s/cloud-init/artifacts/mountetcd.sh
Michael-Sinz: Would you want to log some message just so there is more than just a blank exit?
I know this should not get upset but it is a "emergency out" in case of confusion.

`echo "Found more than 1 unpartitioned disk - confused!"`

pkg/engine/templates_generated.go
```diff
@@ -35797,13 +35797,15 @@ var _k8sCloudInitArtifactsMountetcdSh = []byte(`#!/bin/bash
...
   if ! cat /proc/partitions | grep "$DISK"1; then
     if [[ -n $PARTITION ]]; then
       exit 1
     fi
-    PARTITION=/dev/${DISK}1
+    ETCDDISK=/dev/${DISK}
+    PARTITION=${ETCDDISK}1
   fi;
 done
 MOUNTPOINT=/var/lib/etcddisk
```
Michael-Sinz, line 35811: You should check that ETCDDISK is set by the time you get out of the loop.  Just in case there was no disk...
(That would be bad, but it would be worse to use "" as the disk...)

parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -2,20 +2,26 @@
...
+for DISK in $(grep -o -G "sd[a-z]" /proc/partitions | uniq); do
+  if ! grep "$DISK"1 /proc/partitions; then
+    if [[ -n $PARTITION ]]; then
+      exit 1
+    fi
+    ETCDDISK=/dev/${DISK}
+    PARTITION=${ETCDDISK}1
+  fi;
+done
 udevadm settle
```
Michael-Sinz, line 16: You should have this before you grep through partitions

pkg/engine/templates_generated.go
```diff
@@ -35797,20 +35797,26 @@ var _k8sCloudInitArtifactsMountetcdSh = []byte(`#!/bin/bash
...
+for DISK in $(grep -o -G "sd[a-z]" /proc/partitions | uniq); do
+  if ! grep "$DISK"1 /proc/partitions; then
+    if [[ -n $PARTITION ]]; then
+      exit 1
+    fi
+    ETCDDISK=/dev/${DISK}
+    PARTITION=${ETCDDISK}1
+  fi;
+done
 udevadm settle
```
Michael-Sinz, line 35811: This udevadm settle should go before you search for partitions.

parts/k8s/cloud-init/artifacts/mountetcd.sh
jackfrancis: I know it sounds lame but we are trying to be parsimonious with the data payload we send to the Azure cloud-init interface, as there are hard limits. So we can introspect what the heck went wrong here from /var/log/azure/cluster-provision.log

Michael-Sinz: Make sense...

parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -4,27 +4,9 @@
...
-  done
-  if [[ -z $ETCDDISK ]]; then
-    exit 1
-  fi
-else
-  ETCDDISK=$(cat ${ETCDDISK_PERSIST})
-fi
+ETCDDISK=$(readlink -f /dev/disk/azure/scsi1/lun0)
 PARTITION=${ETCDDISK}1
+udevadm settle
```
Michael-Sinz, line 9: I would make sure this is up earlier (before the readlink) - just in case.

pkg/engine/templates_generated.go
```diff
@@ -35799,27 +35799,9 @@ var _k8sCloudInitArtifactsMountetcdSh = []byte(`#!/bin/bash
...
-  done
-  if [[ -z $ETCDDISK ]]; then
-    exit 1
-  fi
-else
-  ETCDDISK=$(cat ${ETCDDISK_PERSIST})
-fi
+ETCDDISK=$(readlink -f /dev/disk/azure/scsi1/lun0)
 PARTITION=${ETCDDISK}1
+udevadm settle
```
Michael-Sinz, line 35804: I would move this up above the readlink - just in case.  (Basically, before accessing anything in /dev)

Michael-Sinz:
Much nicer - it only assumes the azure disk ordering now.

CecileRobertMichon:
> Much nicer - it only assumes the azure disk ordering now.

@Michael-Sinz not even, there is no assumption on ordering, it only relies on the LUN which is set deterministically in the template. For etcd it's always 0: https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachines.go#L172-L173

Michael-Sinz:
> > Much nicer - it only assumes the azure disk ordering now.
> 
> @Michael-Sinz not even, there is no assumption on ordering, it only relies on the LUN which is set deterministically in the template. For etcd it's always 0: https://github.com/Azure/aks-engine/blob/master/pkg/engine/virtualmachines.go#L172-L173

Yes, I saw that email - so as long as we get the ordering in the arm template consistent then we can use the lun ordering to pick off the correct volumes.

mboersma:
Maybe this also fixes #3185?

jackfrancis:
@mboersma Yes, I suspect so, good call

PR #3261, opened 2020-05-14
chore: add legal notice to Windows VHD

### LINKED ISSUE #2084: discussion: vhd legal notice

jadarsie:
**Reason for Change**:
Fixes #2084

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/packer/windows-vhd-builder.json
+++ b/vhd/packer/windows-vhd-builder.json
@@ -83,6 +83,12 @@
             "destination": "c:\\akse-cache\\collect-windows-logs.ps1"
         },
         {
+            "type": "file",
+            "direction": "upload",
+            "source": "vhd/notice/notice.txt",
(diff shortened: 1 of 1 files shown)


marosset:
That file doesn't look relevant to Windows...
Most of the files/components mentioned in there are specific to Linux.

PR #3262, opened 2020-05-14
fix: mount etcd data disk with label

### LINKED ISSUE #3246: Do not use device name for etcd mount

jackfrancis:
**Reason for Change**:

This PR changes the `/etc/fstab` mount definition for the etcd data disk so that it references the disk partition label instead of the device path, as the device path may not be consistent across reboots.

**Issue Fixed**:

Fixes #3246

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/mountetcd.sh
+++ b/parts/k8s/cloud-init/artifacts/mountetcd.sh
@@ -12,7 +12,7 @@ if mount | grep $MOUNTPOINT; then
   umount /dev/sdc1
 fi
 if ! grep "/dev/sdc1" /etc/fstab; then
-  echo "$PARTITION       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
+  echo "LABEL=etcd_disk       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
 fi
 if ! ls $PARTITION; then
   /sbin/sgdisk --new 1 $DISK
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -35807,7 +35807,7 @@ if mount | grep $MOUNTPOINT; then
   umount /dev/sdc1
 fi
 if ! grep "/dev/sdc1" /etc/fstab; then
-  echo "$PARTITION       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
+  echo "LABEL=etcd_disk       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
 fi
 if ! ls $PARTITION; then
   /sbin/sgdisk --new 1 $DISK


parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -12,7 +12,7 @@ if mount | grep $MOUNTPOINT; then
   umount /dev/sdc1
 fi
 if ! grep "/dev/sdc1" /etc/fstab; then
```
CecileRobertMichon, line 14:
```suggestion
if ! grep "etcd_disk" /etc/fstab; then
```

parts/k8s/cloud-init/artifacts/mountetcd.sh
```diff
@@ -12,7 +12,7 @@ if mount | grep $MOUNTPOINT; then
...
-  echo "$PARTITION       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
+  echo "LABEL=etcd_disk       $MOUNTPOINT       auto    defaults,nofail       0       2" >>/etc/fstab
```
CecileRobertMichon, line 15: nit: we should make `etcd_disk` a var like `PARTITION` and `MOUNTPOINT`

jackfrancis: All of these changes are happening here: https://github.com/Azure/aks-engine/pull/3259

The purpose of this PR is to change as few things as possible to deliver the fix for reboots getting different OS devices mapped to underlying disks

CecileRobertMichon: I'm fine with that, moving to a var is just better for code safety so we don't accidentally change one reference to it but not all. My other comment is required for this to work though, with the change as-is a new line will be added every time you reboot.

parts/k8s/cloud-init/artifacts/mountetcd.sh
jackfrancis: This suggestion (again, see #3259) will have no effect upon reboot scenarios, as this script only runs once (during cluster provisioning)

CecileRobertMichon: why are we even doing that check then? if the script only runs once, seems like that check shouldn't be needed. What am I missing?

jackfrancis: It was added recently to support the script being re-run *during provisioning*. See:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_config.sh#L24

Michael-Sinz: I would actually check for the mount point being in the fstab since that would be a variable and it does not depend on which way it may be in the fstab.

I fully agree with the "smallest change to get clusters stable" first so it likely does not matter but correctness means the fstab line and the if/grep line need to be changed.

CecileRobertMichon:
/hold

should we close this in favor of #3259

jackfrancis:
@CecileRobertMichon yes, I'll close this once that merges

PR #3270, opened 2020-05-15
[WIP] fix: use DisableOutboundSnat configuration for SLB

### LINKED ISSUE #3257: Standard LB SKU configuration doesn't use explicit outbound rules
**Describe the bug**
The LB with Standard SKU is created with flag "Create implicit outbound rules" set to true

As a result the LB doesn't use the explicit outbound rules  that are also configured but not used

It could be seen by checking outbound IPs on pods running on different nodes: some are going out using inbound IP and some using outbound IP, as opposite to rules definition

**Steps To Reproduce**

Deploy cluster with standard LB using aks-engine 0.49
**Expected behavior**
Create implicit outbound rules flag is set to false and explicit outbound rules been used
**AKS Engine version**
0.49
**Kubernetes version**
1.16.8
**Additional context**

jackfrancis:
**Reason for Change**:

This PR sets the proper `DisableOutboundSnat` SLB configuration

**Issue Fixed**:

Fixes #3257

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/loadbalancers.go
+++ b/pkg/engine/loadbalancers.go
@@ -132,6 +132,9 @@ func CreateMasterLoadBalancer(prop *api.Properties, isVMSS bool) LoadBalancerARM
 				},
 			},
 		}
+		if to.Bool(prop.OrchestratorProfile.KubernetesConfig.CloudProviderDisableOutboundSNAT) {
+			loadBalancingRules.LoadBalancingRulePropertiesFormat.DisableOutboundSnat = to.Bool(true)
+		}
 		probes := &[]network.Probe{
 			{
 				Name: to.StringPtr("tcpHTTPSProbe"),


pkg/engine/loadbalancers.go
```diff
@@ -132,6 +132,9 @@ func CreateMasterLoadBalancer(prop *api.Properties, isVMSS bool) LoadBalancerARM
 				},
 			},
 		}
+		if to.Bool(prop.OrchestratorProfile.KubernetesConfig.CloudProviderDisableOutboundSNAT) {
+			loadBalancingRules.LoadBalancingRulePropertiesFormat.DisableOutboundSnat = to.Bool(true)
```
CecileRobertMichon, line 136: You also need to explicitly add outbound rules for TCP and UDP otherwise the cluster won't have outbound which is problematic (see https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/536/files#diff-b3b16dd828cf19adbdc5be2d4e28d2d5R139)

Issue #3269, opened 2020-05-15
Seeing nvidia installation errors in 5.3.0-1020-azure-backed GPU nodes


jackfrancis:
Example:

```
Error from deployment for kubernetes-westeurope-4290 in resource group kubernetes-westeurope-4290:exit status 1
 2020/05/15 15:26:32 Command Output: Deployment failed. Correlation ID: 7ce64e49-5230-4ea1-85ff-6243e9899bc9. {
   "status": "Failed",
   "error": {
     "code": "ResourceDeploymentFailure",
     "message": "The resource operation completed with terminal provisioning state 'Failed'.",
     "details": [
       {
         "code": "VMExtensionProvisioningError",
         "message": "VM has reported a failure when processing extension 'vmssCSE'. Error message: \"Enable failed: failed to execute command: command terminated with exit status=84\n[stdout]\nFri May 15 15:19:48 UTC 2020,k8s-poolgpu-10230551-vmss000000\n\n[stderr]\n\"\r\n\r\nMore information on troubleshooting is available at https://aka.ms/VMExtensionCSELinuxTroubleshoot "
       }
     ]
   }
 }
```

jackfrancis:
```
ERROR: Failed to run `/usr/sbin/dkms build -m nvidia -v 418.40.04 -k
       5.3.0-1020-azure`: 
       Kernel preparation unnecessary for this kernel.  Skipping...
       
       Building module:
       cleaning build area...
       'make' -j6 NV_EXCLUDE_BUILD_MODULES='nvidia-drm '
       KERNEL_UNAME=5.3.0-1020-azure IGNORE_CC_MISMATCH='1' modules....(bad
       exit status: 2)
       ERROR (dkms apport): binary package for nvidia: 418.40.04 not found
       Error! Bad return status for module build on kernel:
       5.3.0-1020-azure (x86_64)
       Consult /var/lib/dkms/nvidia/418.40.04/build/make.log for more
       information.
```

jackfrancis:
Here's the full log output from the failed install:

```
cat /var/log/nvidia-installer-1589561704.log 
nvidia-installer log file '/var/log/nvidia-installer-1589561704.log'
creation time: Fri May 15 16:55:20 2020
installer version: 418.40.04

PATH: /usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin

nvidia-installer command line:
    ./nvidia-installer
    -s
[...]

PR #3275, opened 2020-05-15
chore: disable 18.04 + N series VM node pool

jackfrancis:
**Reason for Change**:

This PR adds fail fast validation to prevent errors installing nvidia drivers on 18.04-LTS N series node pools.

#3269 describes failures installing nvidia drivers on 18.04-LTS.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon:
would it make sense to default the distro to 16.04 if GPU is enabled? otherwise a user might get a validation error for using 18.04 when not explicitly specifying 18.04.

CecileRobertMichon:
/hold

see comment above. feel free to hold cancel once resolved.

jackfrancis:
I'm of two minds about that. On the one hand, it's more convenient, on the other hand, it has the likely outcome of going beyond a user's notice, and then only later being confused as to why the node pools are running a hetergeneous OS configuration.

Validating introduces annoyance that doesn't exist, with the benefit that the user is more likely to know exactly the type of cluster she is building.

@mboersma @rbitia @marosset thoughts?

marosset:
I'm not a big fan of changing the default SKU based on if gpu is enabled or not (unless there was a way to may it clear the defaults where getting changed with a confirm dialog or something) since it could be confusing if folks where expecting 18.04 nodes.

CecileRobertMichon:
We could print a warning... Agreed it's not ideal but I'm concerned about people having working apimodels in CI, etc. that suddenly stop working. If we don't change the default we have to call this out as a breaking change in release notes.

marosset:
I'm trying to decide if I think it would be better to break users right away so they are aware of the issue and call this out as a breaking change like you suggested or allow users to continue along without seeing breaks in CI but potentially not be aware of their cluster configs...

Right now I'm leaning toward the former but am curious what other folks think.

jackfrancis:
I actually intentionally broke upgrade scenarios (there's a UT) in this PR because we *don't* want to allow `aks-engine upgrade` to proceed against a working cluster w/ 18.04-backed GPU nodes. Essentially, the new 5.3 kernel broke our current nvidia implementation, so it's important that we don't allow folks to do that.

So actually, I realized this PR is actually functionally incomplete, as @CecileRobertMichon points out. If we want to permit a distro value of `""` during vlabs api model validation, we have to change the defaults. If we don't want to permit that, then we can essentially add the `""` (no user preference) to the not allowed list of distros here:

https://github.com/Azure/aks-engine/pull/3275/files#diff-f45af0b14c127459636440a5981a6934R585

Let's put the choice to a vote...

jackfrancis:
Disallow a non-user-opinionated distro for N series SKUs, and require the user choose between the two available 16.04 options.

Add thumbs-up reaction to choose this one.

jackfrancis:
Update the defaults flow to set distro defaults for N series SKUs distinctly as compared to other pools.

Add thumbs-up reaction to choose this one.

jackfrancis:
@CecileRobertMichon @marosset @mboersma @ksubrmnn @devigned @ritazh could you please vote if you have a preference?

jackfrancis:
Based on the votes, I've implemented a "permit no user opinion when specifying N series VM SKU, and default to 16.04" solution.

PR #3276, opened 2020-05-15
chore: don't force create etcd_disk filesystem

jackfrancis:
**Reason for Change**:

This PR removes the "force create" option when we create the etcd_disk filesystem via cloud-init.

If for some reason the device partition we target for etcd_disk already exists as a filesystem, we want cloud-init to throw an error, not forcibly create a new filesystem and ignore the previous state.

Ref:

```
-F     Force mke2fs to create a filesystem, even if the specified device is not a partition on a block special device, or if other parameters do not make sense.  In order to force mke2fs to create a filesystem even if the filesystem appears to be in use or is mounted (a truly dangerous thing to do), this option must be specified twice.
```

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -539,7 +539,6 @@ fs_setup:
     filesystem: ext4
     device: /dev/disk/azure/scsi1/lun0
     extra_opts:
-      - -F
       - -E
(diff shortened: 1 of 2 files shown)


johnsonshi:
LGTM

PR #3283, opened 2020-05-18
refactor: consolidate azure-network-policy addon spec

mboersma:
**Reason for Change**:
Eases maintenance of the (newly renamed) azure-network-policy addon by consolidating manifests into a single YAML file that uses Go templating.

**Issue Fixed**:
Addresses #3188.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Just to be pedantic we aren't renaming the addon here, just the source yaml file that we use to maintain it.

parts/k8s/addons/azure-network-policy.yaml
```diff
@@ -65,15 +65,20 @@ spec:
     metadata:
       labels:
         k8s-app: azure-npm
+{{- if IsKubernetesVersionGe "1.16.0"}}
       annotations:
         scheduler.alpha.kubernetes.io/critical-pod: ''
+  {{- if IsKubernetesVersionGe "1.17.0"}}
+        cluster-autoscaler.kubernetes.io/daemonset-pod: "true"
+  {{- end}}
```
jackfrancis, line 73: We need the newlines here in both negative cases, so should be `{{end}}`

jackfrancis: disregard this comment

PR #3284, opened 2020-05-18
refactor: consolidate pod-security-policy addon spec

jackfrancis:
**Reason for Change**:

Eases maintenance of the pod-security-policy addon by consolidating manifests into a single YAML file that uses Go templating.

**Issue Fixed**:

Addresses #3188.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/const.go
```diff
@@ -246,8 +246,8 @@ const (
 	corednsAddonDestinationFilename                string = "coredns.yaml"
 	kubeProxyAddonSourceFilename                   string = "kubernetesmasteraddons-kube-proxy-daemonset.yaml"
 	kubeProxyAddonDestinationFilename              string = "kube-proxy-daemonset.yaml"
-	podSecurityPolicyAddonSourceFilename           string = "kubernetesmasteraddons-pod-security-policy.yaml"
-	podSecurityPolicyAddonDestinationFilename      string = "pod-security-policy.yaml"
+	podSecurityPolicyAddonSourceFilename           string = "pod-security-policy.yaml"
+	podSecurityPolicyAddonDestinationFilename      string = "0-pod-security-policy.yaml"
```
jackfrancis, line 250: @aramase This should influence the kube-addon-manager to load this spec first (as it loads the files alphanumerically), right? It won't guarantee anything other than that (e.g., it won't guarantee that all resources in this file be loaded before resources in other specs), but it should reduce the rate of flakes compared to having the spec have a "higher" alphanumeric filename.

aramase: @jackfrancis That's right. Having a "higher" alphanumeric filename will ensure its the first file that gets applied, but still won't guarantee the resources in the file have been loaded.

PR #3293, opened 2020-05-19
fix: apiserver connection timeout for private apiserver

feiskyer:
**Reason for Change**:

Increase apiserver connection timeout for private apiserver.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


feiskyer:
tests failed because of the following reasons:

```
Error response from daemon: Get https://quay.io/v2/deis/go-dev/manifests/v1.27.0: received unexpected HTTP status: 500 Internal Server Error
##[error]Docker pull failed with exit code 1
```

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -276,7 +276,11 @@ if [[ $STS != 0 ]]; then
         VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_DNS_LOOKUP_FAIL"}}
     fi
 fi
-retrycmd 50 1 3 nc -vz ${API_SERVER_NAME} 443 || VALIDATION_ERR={{GetCSEErrorCode "ERR_K8S_API_SERVER_CONN_FAIL"}}
+API_SERVER_CONN_RETRIES=50
+if [[ $API_SERVER_NAME == *.privatelink.* ]]; then
```
jackfrancis, line 280: is `== *.privatelink.*` equivalent to `== *"privatelink"*` ?

@mboersma @alexeldeib thoughts? the existing shell globbing equality comparisons in the codebase use the latter pattern

mboersma: I don't think they are equivalent. The proposed pattern matches ".privatelink." with dots included:

```shell
% export API_SERVER_NAME=myprivatelinkdealie                            
% if [[ $API_SERVER_NAME == *.privatelink.* ]]; then echo Equivalent; fi
% if [[ $API_SERVER_NAME == *"privatelink"* ]]; then echo Equivalent; fi
Equivalent
% export API_SERVER_NAME=my.privatelink.dealie                          
% if [[ $API_SERVER_NAME == *.privatelink.* ]]; then echo Equivalent; fi
Equivalent
% if [[ $API_SERVER_NAME == *"privatelink"* ]]; then echo Equivalent; fi
Equivalent

```

jackfrancis: It looks like `*.privatelink.*` is equivalent to `*".privatelink."*`:

```
$ export API_SERVER_NAME=my.privatelink.dealie
$ if [[ $API_SERVER_NAME == *"privatelink"* ]]; then echo Match; fi
Match
$ if [[ $API_SERVER_NAME == *".privatelink."* ]]; then echo Match; fi
Match
```

So I think this means that the `"` chars in our existing shell globbing comparisons are superfluous.

feiskyer: yep, '.privatelink.' means dots are included

PR #3294, opened 2020-05-19
fix: increase apiserver connection timeout for private apiserver

feiskyer:
**Reason for Change**:

Cherry-pick of https://github.com/Azure/aks-engine/pull/3293 and #3294.

Fix: increase apiserver connection timeout for private apiserver.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -229,7 +229,11 @@ fi
 VALIDATION_ERR=0
 
 {{- if IsHostedMaster }}
-RES=$(retrycmd_if_failure 20 1 3 nslookup ${API_SERVER_NAME})
+API_SERVER_DNS_RETRIES=20
+if [[ $API_SERVER_NAME == *.privatelink.* ]]; then
+  API_SERVER_DNS_RETRIES=200
+fi
+RES=$(retrycmd_if_failure ${API_SERVER_DNS_RETRIES} 1 3 nslookup ${API_SERVER_NAME})
 STS=$?
 if [[ $STS != 0 ]]; then
     if [[ $RES == *"168.63.129.16"*  ]]; then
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -229,16 +229,25 @@ fi
 VALIDATION_ERR=0
 
 {{- if IsHostedMaster }}
-RES=$(retrycmd_if_failure 20 1 3 nslookup ${API_SERVER_NAME})
+API_SERVER_DNS_RETRIES=20
+if [[ $API_SERVER_NAME == *.privatelink.* ]]; then
+  API_SERVER_DNS_RETRIES=100
```
xuto2, line 234: is this enough? I read the email discussion that the dns resolution failed for 10minutes, is that not the cae?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -229,16 +229,25 @@ fi
...
+fi
+RES=$(retrycmd_if_failure ${API_SERVER_DNS_RETRIES} 1 3 nslookup ${API_SERVER_NAME})
 STS=$?
 if [[ $STS != 0 ]]; then
     if [[ $RES == *"168.63.129.16"*  ]]; then
         VALIDATION_ERR=$ERR_K8S_API_SERVER_AZURE_DNS_LOOKUP_FAIL
     else
         VALIDATION_ERR=$ERR_K8S_API_SERVER_DNS_LOOKUP_FAIL
     fi
+    exit $VALIDATION_ERR
```
xuto2, line 244: nit: instead of exit here, we can skip the subsequent api server connection check if $STS != 0, this way we can still see the last cse message with code.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -229,16 +229,25 @@ fi
...
+fi
+API_SERVER_CONN_RETRIES=50
+if [[ $API_SERVER_NAME == *.privatelink.* ]]; then
+  API_SERVER_CONN_RETRIES=200
```
xuto2, line 248: question: so private cluster node max timeout is (100+200) * (1+3) = 1200 second that is 20min, do we plan to have the same timeout in ccp readiness check in RP?

parts/k8s/cloud-init/artifacts/cse_main.sh
feiskyer: Good catch, 200 should be here and 100 in the later check.

parts/k8s/cloud-init/artifacts/cse_main.sh
feiskyer: yep, would move the following check into else block

parts/k8s/cloud-init/artifacts/cse_main.sh
feiskyer: I think we already have that (longer than 20m by default)

PR #3273, opened 2020-05-15
refactor: move etcd data disk mounting to cloud init 

CecileRobertMichon:
**Reason for Change**:

PR https://github.com/Azure/acs-engine/pull/661 added a temporary fix for mounting disks until issue https://bugs.launchpad.net/cloud-init/+bug/1692093 was resolved. The issue has now been resolved in the latest Ubuntu image and the workaround script can be removed in favor of more (hopefully) a more stable implementation leveraging cloud init.

**Issue Fixed**:

Fixes #3125 without needing retries & fixes https://github.com/Azure/acs-engine/issues/1643 

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3297, opened 2020-05-19
fix: ensure /mnt is ext4

jackfrancis:
**Reason for Change**:

This PR follows on from #3273 to ensure that `/mnt` is mounted as ext4 filesystem type via cloud-init.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


anhvoms:
You might want to add a comment that this is due to cloud-init bug:
https://bugs.launchpad.net/cloud-init/+bug/1879552

The resource disk should get partitioned, formatted, mounted correctly regardless of the presence of datadisks configuration. However, it looks like it only got partitioned and mounted, not formatted correctly

PR #3300, opened 2020-05-20
fix: fix issue provisioning windows nodes with kubenet

### LINKED ISSUE #3292: Windows + kubenet node doesn't come online
When building a Windows node in a kubenet cluster that node never reports as ready.

marosset:
**Reason for Change**:

**Issue Fixed**:

Fixes #3292 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubeletstart.ps1
+++ b/parts/k8s/kubeletstart.ps1
@@ -22,7 +22,7 @@ $global:AzureCNIBinDir = [Io.path]::Combine("$global:AzureCNIDir", "bin")
 $global:AzureCNIConfDir = [Io.path]::Combine("$global:AzureCNIDir", "netconf")
 
 $global:CNIPath = [Io.path]::Combine("$global:KubeDir", "cni")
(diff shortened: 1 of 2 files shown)


marosset:
@jackfrancis @ksubrmnn @AbelHu PTAL

This is a quick fix to address the issue - I think we should do a more thorough review the powershell as a separate work item/PR

parts/k8s/kubeletstart.ps1
```diff
@@ -246,7 +246,7 @@ if (($global:NetworkPlugin -eq "kubenet") -and ($global:ContainerRuntime -eq "do
         if (-not $podCidrDiscovered) {
             $argList = $KubeletArgListStr
```
marosset, line 247: StartProcess is expecting -ArgumentList to be <string[]> but line 74 sets $KubeletArtListStr (and $argList by proxy) to a string

PR #3311, opened 2020-05-20
feat: add support for Kubernetes 1.17.6

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.17.md#changelog-since-v1175

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

This PR has a temporary commit for testing with Kubernetes 1.17. It should be removed before merging.

mboersma:
This [passed CI](https://github.com/Azure/aks-engine/pull/3311/checks?check_run_id=694263424), so I'll strip off the testing commit and it should be ready for review.

vhd/packer/install-dependencies.sh
```diff
@@ -351,10 +351,9 @@ K8S_VERSIONS="
 1.19.0-alpha.3
 1.18.2
 1.18.1
+1.17.6
 1.17.5
 1.17.5-azs
-1.17.4
-1.17.4-azs
```
jackfrancis, line 357: @jadarsie Are there any in-code Azure Stack docs we want to update now that we're removing support for 1.17.4?

jadarsie: I cannot think of anything besides the azure-stack doc/topic, I can update that page in the next few days.

PR #3322, opened 2020-05-22
chore: Add EnableAHUB in WindowsProfile

AbelHu:
**Reason for Change**:

Allow users to use Azure hybrid benefit for Windows server.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


jackfrancis:
@AbelHu How can we add E2E test coverage for this? Can the current Windows VHD image be associated with a LicenseType?

AbelHu:
@jackfrancis From my understanding and the test result, `LicenseType` is only a property of VMSS for Windows and there is no additional change for the Windows VHD.
I manually ran `aks-engine deploy` with and without `LicenseType` and both passed. How about to add `"LicenseType": "Windows_Server"` in `windowsProfile` of the file test\e2e\test_cluster_configs\windows\cse_benchmark_aks_vhds.json?

AbelHu:
One question for the type `WindowsLicenseType` I added in common/types.go: Is this correct? Or should I add this separately in vlabs and api, and then convert them to *string in `pkg/api/converterfromapi.go` and ` pkg/api/convertertoapi.go `?

jackfrancis:
@AbelHu I see the `WindowsLicenseType` type as being defined in both the api and vlabs packages, and then converted. Strictly speaking you could define this thinly wrapped string type in a common library that both api and vlabs imported (like common), to avoid the unnecessary conversion.

The only consideration would be whether or not the evolution of that type definition might want to evolve discretely for the api and vlabs data structures. In this case I don't think it will.

AbelHu:
@jackfrancis Agree with your point. That is why I defined it in common in the original PR. But later I found that common/types.go is never introduced in AKS and it seems like to introduce many conflicts for namespace in AKS unit tests so I define them separately for vlabs and api.

pkg/api/types.go
```diff
@@ -205,23 +205,34 @@ type CustomNodesDNS struct {
...
+	WindowsImageSourceURL     string              `json:"windowsImageSourceURL"`
+	WindowsPublisher          string              `json:"windowsPublisher"`
+	WindowsOffer              string              `json:"windowsOffer"`
+	WindowsSku                string              `json:"windowsSku"`
+	WindowsDockerVersion      string              `json:"windowsDockerVersion"`
+	Secrets                   []KeyVaultSecrets   `json:"secrets,omitempty"`
+	SSHEnabled                *bool               `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates    *bool               `json:"enableAutomaticUpdates,omitempty"`
+	IsCredentialAutoGenerated *bool               `json:"isCredentialAutoGenerated,omitempty"`
+	LicenseType               *WindowsLicenseType `json:"licenseType,omitempty"`
```
marosset, line 235: Can this get added in alphabetically?
This really help reduce merge conflicts in code

marosset:
@AbelHu Does AKS ever plan to use liscenseType of Windows_Client here?
I don't think a lot of scenarios are testing or would work on client editions (Docker EE, network setup etc) and I'm wondering if we should just hardcode this to Windows_Server for now.

docs/topics/clusterdefinitions.md
```diff
@@ -895,6 +895,7 @@ https://{keyvaultname}.vault.azure.net:443/secrets/{secretName}/{version}
...
 | sshEnabled                    | no       | If set to `true`, OpenSSH will be installed on windows nodes to allow for ssh remoting. **Only for Windows version 1809/2019 or later**. The same SSH authorized public key(s) will be added from [linuxProfile.ssh.publicKeys](#linuxProfile). Default: `true` |
+| licenseType                   | no       | Specifies that the image or disk that is being used was licensed on-premises. This element is only used for images that contain the Windows Server operating system. <br> Possible values are: Windows_Client, Windows_Server <br> If this element is included in a request for an update, the value must match the initial value. This value cannot be updated. <br> For more information, see [Azure Hybrid Use Benefit for Windows Server](https://docs.microsoft.com/azure/virtual-machines/virtual-machines-windows-hybrid-use-benefit-licensing?toc=%2fazure%2fvirtual-machines%2fwindows%2ftoc.json) <br> Minimum api-version: 2015-06-15 |
```
marosset, line 898: Can you note the default value too? - it looks like that will leave the licensing option empty correct?

marosset:
Can you either add comments in docs/validation about this only working for scale sets or set this up for availability sets too?

In the portal I see on option to 'use Azure Hybrid Benefit' when creating non-scaleset VMs which get added to availability sets?

docs/topics/clusterdefinitions.md
AbelHu: This is a pointer so the default value is nil and no licenseType will be generated in the template.

pkg/api/types.go
AbelHu: We just follow the current implementation. It seems like that we always add new property in the end. I think that the conflict is not caused by the sequence but by the different indent.

AbelHu:
> Can you either add comments in docs/validation about this only working for scale sets or set this up for availability sets too?
> 
> In the portal I see on option to 'use Azure Hybrid Benefit' when creating non-scaleset VMs which get added to availability sets?

I will add the comment. AKS only supports Windows with VMSS. If later we need to support this in VMAS, we can add the support later.

AbelHu:
> @AbelHu Does AKS ever plan to use liscenseType of Windows_Client here? If not it might be good to add validation to fail if licenseType is set to Windows_Client.

AKS only uses `Windows_Server` and will validate this in AKS side. It seems like that SDK and API allows users to set it to `Windows_Client` so I think that it may be better to keep this option.

pkg/api/types.go
marosset: > > @AbelHu Does AKS ever plan to use liscenseType of Windows_Client here? If not it might be good to add validation to fail if licenseType is set to Windows_Client.
> 
> AKS only uses `Windows_Server` and will validate this in AKS side. It seems like that SDK and API allows users to set it to `Windows_Client` so I think that it may be better to keep this option.

We don't support configuring Windows client nodes in aks-engine so I think it would be appropriate to add validation in aks-engine to dis-allow this even if it is possible in arm.

marosset: > We just follow the current implementation. It seems like that we always add new property in the end. I think that the conflict is not caused by the sequence but by the different indent.

I know things aren't currently alphabetized everywhere but my ask is to start inserted new fields in alphabetically order.
It does help keep things neat and tidy.
I have been doing this whenever possible.

(also we have linting rules that make sure consistent indentation is caught in PRs)

pkg/api/convertertoapi_test.go
```diff
@@ -946,9 +950,9 @@ func TestConvertVLabsWindowsProfile(t *testing.T) {
 	}
 
 	for _, c := range cases {
+		c := c
```
AbelHu, line 953: @marosset FYI. This must be set to capture range variable. Otherwise, only one case will be run multiple times.
Reference: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

AbelHu:
@marosset @jackfrancis @mboersma I will update this PR with below changes soon:

1. Support VMAS by adding code in pkg\engine\virtualmachines.go
1. Change LicenseType (*string) to EnableAHUB (*bool)

marosset:
Sounds good!

AbelHu:
@keikhara helped to confirm that AKS only needs to set `LicenseType` to `Windows_Server`. Now I have updated this PR. I also test VMAS with `examples/windows/kubernetes-wincni.json` and it works.

Please help to review it. @marosset @jackfrancis @mboersma

AbelHu:
@marosset Could you help to take a look at the updated PR? Thanks.

pkg/api/types.go
```diff
@@ -1805,6 +1811,14 @@ func (w *WindowsProfile) GetIsCredentialAutoGenerated() bool {
 	return false
 }
 
+// GetEnableAHUB gets it AHUB should be enabled for Windows nodes
```
marosset, line 1814: type: if

marosset:
Can we add one e2e test config that exercises this behavior?

you can add a new json file at https://github.com/Azure/aks-engine/tree/master/test/e2e/test_cluster_configs/windows with in enabled and also set
 
	"env": {
		"SKIP_TESTS": "true",
	},
	"options": {
		"allowedOrchestratorVersions": ["latestReleasedVersion"]
	},

in the top (like in https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/windows/vhd_url.json)

marosset:
Looks good after we get one e2e test config with this set to true.

AbelHu:
> Can we add one e2e test config that exercises this behavior?
> 
> you can add a new json file at https://github.com/Azure/aks-engine/tree/master/test/e2e/test_cluster_configs/windows with in enabled and also set
> 
> ```
> "env": {
> 	"SKIP_TESTS": "true",
> },
> "options": {
> 	"allowedOrchestratorVersions": ["latestReleasedVersion"]
> },
> ```
> 
> in the top (like in https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/windows/vhd_url.json)

Added now. Why I did not add it is because it is difficult to verify it after enabling AHUB. currently it only can verify whether the cluster can be created successfully with enabling AHUB.

marosset:
> > Can we add one e2e test config that exercises this behavior?
> > you can add a new json file at https://github.com/Azure/aks-engine/tree/master/test/e2e/test_cluster_configs/windows with in enabled and also set
> > ```
> > "env": {
> > 	"SKIP_TESTS": "true",
> > },
> > "options": {
> > 	"allowedOrchestratorVersions": ["latestReleasedVersion"]
> > },
> > ```
> > 
> > 
> > in the top (like in https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/windows/vhd_url.json)
> 
> Added now. Why I did not add it is because it is difficult to verify it after enabling AHUB. currently it only can verify whether the cluster can be created successfully with enabling AHUB.

That is alright, the SKIP_TESTS equals true will cause the e2e test to only validate deployment and will skip running the actually e2e tests.
Thanks or adding this!

PR #3333, opened 2020-05-23
fix: Add latest DC-Series VM SKUs

joekottke:
**Reason for Change**:
VM SKUs used to detect DC-Series VMs are outdated, and the code doesn't check for newer SKUs (DCv2 VMs in particular)

**Issue Fixed**:

**Requirements**:

- [X] includes documentation
- [X] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/helper.go
+++ b/pkg/api/common/helper.go
@@ -245,11 +245,27 @@ func GetDCSeriesVMCasesForTesting() []struct {
 		Expected bool
 	}{
 		{
+			"Standard_DC1s_v2",
+			true,
+		},
+		{
 			"Standard_DC2s",
 			true,
 		},
 		{
+			"Standard_DC2s_v2",
+			true,
+		},
+		{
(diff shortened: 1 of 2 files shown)


jackfrancis:
@mboersma to comment if these updates will be resilient to the VM SKU code generators

mboersma:
The preferred way to update SKUs is by running ./scripts/azure-const.sh with appropriate environment variables set. But I ran this and didn't see any changes, so these DC-series VMs must not be visible to my subscription.

In that case, we can add the SKUs to the `VMSkus` data structure in `pkg/helpers/azure_skus_const.go` and then to the appropriate unit tests. That data structure is treated as append-only and is the source of truth for determining storage account and accelerated networking support.

@joekottke let me know if I can clarify anything or if you'd like me to update the PR.

joekottke:
I understand and will have a crack at it. Thanks!

PR #3340, opened 2020-05-26
ci: enable multiple static resource group exclusions in cleanup script

jackfrancis:
**Reason for Change**:

This PR enables us to configure a set of "never delete" resource group substrings in the cleanup script.

In this PR we add the "aksimages" reserved exclusion, to ensure that we never accidentally delete these VHD CI-produced image resource group(s).

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
Ref: https://stackoverflow.com/a/44704837/3423859

jackfrancis:
/hold

This doesn't do what we think it does

PR #3343, opened 2020-05-26
ci: use ginkgo 1.12, run go mod vendor during make coverage

jackfrancis:
**Reason for Change**:

This PR fixes up the CI configuration thusly:

- install all binary tools in the `hack/tools` go module
- go mod tidy in `hack/tools` and `test/e2e` as well as root.
- update to the latest ginkgo (v1.12.2)

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


scripts/ginkgo.coverage.sh
```diff
@@ -25,6 +25,8 @@ hash goveralls 2>/dev/null || go get github.com/mattn/goveralls
 hash godir 2>/dev/null || go get github.com/Masterminds/godir
 
 generate_cover_data() {
+  ginkgo version
+  git diff --exit-code --quiet go.mod go.sum vendor/modules.txt || go mod vendor
```
jackfrancis, line 29: This is the key change to unblock CI, we want to figure out why `make coverage` gets itself into a place where the above files are dirty.

PR #3352, opened 2020-05-28
feat: add support for GPU VHD

bowang-666:
Add image config for GPU VHD, also make NvidiaEnabledSKUs accessible publicly so RP can use it to generate error message.

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -179,7 +179,7 @@ var (
 		ImageOffer:     "aks",
 		ImageSku:       "aks-gpu-1804-202005",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "2020.05.31",
+		ImageVersion:   "2020.05.28",
 	}
 
 	// AKSWindowsServer2019OSImageConfig is the AKS image based on Windows Server 2019
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -173,9 +173,9 @@ var (
 		ImageVersion:   "2020.05.13",
 	}
 
-	// AKSGPU1804OSImageConfig is the AKS image with GPU drivers and device plugin based on Ubuntu 18.04-LTS.
+	// AKSUbuntuGPU1804OSImageConfig is the AKS image with GPU drivers and device plugin based on Ubuntu 18.04-LTS.
 	// TODO: Update sku and version after GPU image is actually published
-	AKSGPU1804OSImageConfig = AzureOSImageConfig{
+	AKSUbuntuGPU1804OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks",
 		ImageSku:       "aks-ubuntu-gpu-1804-202005",
 		ImagePublisher: "microsoft-aks",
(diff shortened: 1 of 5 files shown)


pkg/api/azenvtypes.go
```diff
@@ -173,6 +173,15 @@ var (
 		ImageVersion:   "2020.05.13",
 	}
 
+	// AKSGPU1804OSImageConfig is the AKS image with GPU drivers and device plugin based on Ubuntu 18.04-LTS.
+	// TODO: Update sku and version after GPU image is actually published
+	AKSGPU1804OSImageConfig = AzureOSImageConfig{
+		ImageOffer:     "aks",
+		ImageSku:       "aks-gpu-1804-202005",
```
xuto2, line 180: can we follow the pattern aks-ubuntu-1804-202005, maybe call it aks-ubuntu-1804-gpu-202005?

pkg/api/azenvtypes.go
```diff
@@ -221,6 +230,7 @@ var (
 			AKSUbuntu1804:     AKSUbuntu1804OSImageConfig,
 			AKS1804Deprecated: AKSUbuntu1804OSImageConfig, // for back-compat
 			ACC1604:           ACC1604OSImageConfig,
+			AKSGPU1804:        AKSGPU1804OSImageConfig,
```
xuto2, line 233: We should also update
AzureUSGovernmentCloudSpec 
AzureChinaCloudSpec

pkg/api/const.go
```diff
@@ -36,6 +36,7 @@ const (
 	AKSUbuntu1604     Distro = "aks-ubuntu-16.04"
 	AKSUbuntu1804     Distro = "aks-ubuntu-18.04"
 	ACC1604           Distro = "acc-16.04"
+	AKSGPU1804        Distro = "aks-gpu-18.04"
```
xuto2, line 39: nit: AKSUbuntu1804GPU

pkg/api/azenvtypes.go
bowang-666: made it aks-ubuntu-gpu-1804 so numbers go in the end

pkg/api/azenvtypes.go
bowang-666: added. btw I don't see ACC1604 in USGov and China Cloud, are they not enabled?

pkg/api/const.go
bowang-666: made them AKSUbuntuGPU1804

pkg/api/azenvtypes.go
xuto2: we don't plan to support 1604 for gen2, since we want to move to 1804

pkg/api/common/helper.go
```diff
@@ -14,6 +14,44 @@ import (
 	validator "gopkg.in/go-playground/validator.v9"
 )
 
+var (
+	/* If a new GPU sku becomes available, add a key to this map, but only if you have a confirmation
+	   that we have an agreement with NVIDIA for this specific gpu.
+	*/
+	NvidiaEnabledSKUs = map[string]bool{
```
xuto2, line 21: For future reference - we don't want to diverge from aks-e code when absolutely necessary.

PR #3363, opened 2020-05-28
refactor: consolidate cloud-node-manager addon spec

jackfrancis:
**Reason for Change**:

Eases maintenance of the cloud-node-manager addon by consolidating manifests into a single YAML file that uses Go templating.
**Issue Fixed**:

Addresses #3188.

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/addons/cloud-node-manager.yaml
```diff
@@ -93,7 +93,7 @@ spec:
           limits:
             cpu: 2000m
             memory: 512Mi
-{{- if HasWindows}}
+{{- if and HasWindows (IsKubernetesVersionGe "1.18.0")}}
```
jackfrancis, line 96: cc @chewong to sanity check that this is the only version-specific aspect of this spec

chewong: That's right.

PR #3371, opened 2020-05-29
fix: bypass AzurePublicCloud assumption in e2e suite

jadarsie:
**Reason for Change**:
E2E should not hard-code public Azure cloud as it breaks custom clouds running E2E.
This PR is not a proper fix, it just makes things work again for Azure Stack.

Introduced here PR #3144.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -165,13 +165,15 @@ var _ = BeforeSuite(func() {
 	}
 	Expect(dnsAddonName).NotTo(Equal(""))
 
-	env, err = azure.EnvironmentFromName("AzurePublicCloud") // TODO get this programmatically
-	if err != nil {
-		Expect(err).NotTo(HaveOccurred())
-	}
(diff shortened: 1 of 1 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2146,7 +2148,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 						break
 					}
 				}
-
+				// TODO refactor to remove the "compute" usage so the test can be run on Azure Stack
```
jackfrancis, line 2151: We're O.K. with leaving this TODO as is? This means these tests will fail, right?

jackfrancis: (fail in Azure Stack, that is)

jadarsie: OK, for now, it wont get to this point as we do not yet support VMSS

jadarsie: https://github.com/Azure/aks-engine/pull/3371/files/7910d7b76b9f89603208235636059dc2deb76d68#diff-3509723ee1508f842e543bf659b43e20R2067

PR #3376, opened 2020-06-01
chore: Bump moby version to 3.0.12

cpuguy83:
**Reason for Change**:

New moby patch release

**Issue Fixed**:

N/A

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -222,7 +222,7 @@
       }
     },
     "mobyVersion": {
-      "defaultValue": "3.0.11",
+      "defaultValue": "3.0.12",
       "metadata": {
         "description": "The Azure Moby build version"
       },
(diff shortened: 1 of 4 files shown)


mboersma:
@cpuguy83 we should also bump the `DefaultMobyVersion` const in [`pkg/api/const.go`](https://github.com/Azure/aks-engine/blob/master/pkg/api/const.go#L467).

cpuguy83:
@mboersma Fixed.

parts/k8s/kubernetesparams.t
```diff
@@ -237,6 +237,7 @@
          "3.0.8",
          "3.0.10",
          "3.0.11"
```
jackfrancis, line 239: missing comma at the end of "3.0.11"

cpuguy83: :rage1:

cpuguy83: Fixed

PR #3267, opened 2020-05-15
feat: add support for Flatcar Container Linux

invidian:
**Reason for Change**:

This PR adds support for [Flatcar Container Linux](https://www.flatcar-linux.org/), based on previously supported CoreOS, which support has been removed because of upcoming EoL.

**Issue Fixed**:

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3380, opened 2020-06-01
feat: flatcar nodes

jackfrancis:
**Reason for Change**:

This PR introduces flatcar-backed nodes by specifying the `"flatcar"` distro. E.g., in an agentPoolProfile definition:

`"distro": "flatcar"`

Derived from @invidian's original changeset in #3267.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -529,6 +546,85 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
+      drop-ins:
+        - name: "10-flatcar.conf"
+          content: |
+            [Service]
+            ExecStart=
+            ExecStart=/opt/bin/etcd $DAEMON_ARGS
+    - name: etcd-member.service
+      mask: true
+{{else}}
 disk_setup:
```
jackfrancis, line 628: @johnsonshi it looks like flatcar doesn't support these cloud-init interfaces? How should we be mounting `/var/lib/etcddisk` on flatcar-enabled master VMs?

Alternatively, we can only add flatcar support as a worker node, and require Ubuntu for master nodes.

cc @rbitia @ritazh

jackfrancis: @invidian can you confirm that it makes sense to scope this work to worker nodes only? I can fix up the PR so that flatcar is enabled as a node distro, but not available for the master (control plane) VMs.

johnsonshi: The other alternative would be for Flatcar to include those cloud-init udev rules so that the symlinks are created. However that may exceed this PR's scope.

johnsonshi: Waagent also has the udev rules that creates symlinks to data disks at /dev/disk/azure/scsi/lun#.

I deployed a Flatcar VM. I confirmed that the udev rules for data disks are present. I then used `udevadm monitor` to monitor for udev events, then used a separate console to attach a data disk. From the monitor I see that the udev rule is fired. After that, I see that the symlink has been created.

https://paste.ubuntu.com/p/rqCqh7BPcT/

invidian: From my tests, Flatcar was working just fine for master nodes. I didn't test the reboots though, which were problematic in the past :thinking:

invidian: Are there any known issues for Flatcar for master nodes? I can investigate that, if you point me to them.

We could only merge support for worker nodes, but I'd rather avoid doing more changes to this, already large changeset if possible, I'd rather prefer incremental improvements and fixes. The whole reverting was a lot of work and resolving conflicts while rebasing on `master` is also not trivial.

johnsonshi: Hey @invidian I think @jackfrancis 's concern is whether Flatcar can correctly mount the etcd data disk. Being able to mount the etcd data disk is a prerequisite for a distro image to become a master node.

Well from my validations (and from a cloud-init provisioning perspective), Flatcar can correctly mount the etcd data disk. Here's why:

The WALinuxAgent package comes with the required [udev rules](https://github.com/Azure/WALinuxAgent/blob/develop/config/66-azure-storage.rules) for handling attached VM data disks. These udev rules are required on Azure VMs to create publicly-documented symlinks from `/dev/disk/azure/scsi1/lun#` to the attached VM data disks. From [this specific line in the WALinuxAgent udev rules](https://github.com/Azure/WALinuxAgent/blob/78c660062f2ef86976cf4d03c8d99b70fe0a7255/config/66-azure-storage.rules#L12), we can see that the udev rule creates symlinks from `/dev/disk/azure/scsi1/lun#` to VM data disks (which will always have `f8b3781b-1e82-4818-a1c3-63d806ec15bb` as the device ID`). Since the WALinuxAgent package comes with these rules and since the package is installed by default in Flatcar, these udev rules are guaranteed to always be in the system.

Well I tested to see if Flatcar is able to handle a VM data disk (such as the etcd data disk). I did my validation over [here](https://paste.ubuntu.com/p/rqCqh7BPcT/). From that link (my manual validation), you can see that `/lib/udev/rules.d/66-azure-storage.rules` contains the WALinuxAgent udev rules. Furthermore, I used `udevadm monitor` to monitor for kernel udev events. I then attached a VM data disk to the Flatcar VM from another console, and as expected, the udev rules were fired, and a symlink from `/dev/disk/azure/scsi1/lun0` is created by the fired udev rule. This symlink allows Flatcar to have a reliable mechanism to refer to the etcd data disk.

Here is the reason why we need the udev rule and symlink to exist on Azure VMs for data disks. On Azure, there is no guarantee after VM reboot that `/dev/sdc` is the etcd data disk. Now, because the udev rule is always present for all Flatcar VMs on Azure, we have a safe guarantee that `/etc/fstab` can have a reliable path (symlink) to refer to the etcd data disk.

If you are keen on reading up on more reasons, here we gooo... :)

Take a look at the [master node cloud-init custom-data YAML at these lines](https://github.com/Azure/aks-engine/blob/2062d0793f58fda7c6ceeb61a3f1dd5004e94ee4/parts/k8s/cloud-init/masternodecustomdata.yml#L538-L543). You can see that the cloud-init `fs_setup` module uses `/dev/disk/azure/scsi1/lun0` to refer to the etcd data disk. Those highlighted lines of the cloud-init `fs_setup` module indicate that the etcd data disk should be formatted to `ext4` and most importantly should have the `etcd_disk` label. Cloud-init happily formats and applies the `etcd_disk` label as instructed.

Now, take a look at [the next few lines](https://github.com/Azure/aks-engine/blob/2062d0793f58fda7c6ceeb61a3f1dd5004e94ee4/parts/k8s/cloud-init/masternodecustomdata.yml#L550-L552). From the [docs on cloud-init's `mounts` module](https://cloudinit.readthedocs.io/en/latest/topics/modules.html?highlight=label#mounts), we see that it says the `mounts` module writes the mount directives to `/etc/fstab`, which ensures that these directives are persisted across reboots. The mount directive (which is written to fstab) instructs the system to "please mount the filesystem with the `etcd_disk` label at `/var/lib/etcddisk`.

Now because the etcd file system has been mounted at `/var/lib/etcddisk`, this triggers the systemd unit [etcd.service](https://github.com/Azure/aks-engine/blob/2062d0793f58fda7c6ceeb61a3f1dd5004e94ee4/parts/k8s/cloud-init/artifacts/etcd.service#L7), which starts etcd. Notice that this systemd unit `RequiresMountsFor=/var/lib/etcddisk`.

johnsonshi: @invidian @jackfrancis Now the only validations I have run is from a Linux system provisioning perspective, which is specifically validating whether the udev rule is fired and the data disk symlink is created (which both did). Because the symlink is created, the rest of the pieces (cloud-init fs setup and label application, mounting the disk at the expected location, and triggering the etcd systemd service) should logically follow. From a Linux system provisioning perspective, this is a definite pass and LGTM.

However, @invidian mentioned that Flatcar reboots were problematic in the past. I don't know what these problematic issues were, so maybe you can clarify what they were?

Before this thing is merged in, can one or both of you test that (1) Flatcar works without any issues when provisioned as master nodes using this PR's code, and (2) the Flatcar master nodes can be restated with no issues, and (3) after restart the etcd service is up and alive because the etcd disk was mounted at the expected location to trigger the etcd systemd unit.

jackfrancis:
Pro tip for flatcar cloud-init debugging: `journalctl -u oem-cloudinit`

parts/k8s/cloud-init/masternodecustomdata.yml
jackfrancis: This is why the mounting is not working, from oem-cloudinit logs:

```
Jun 04 23:25:23 k8s-master-36217359-0 coreos-cloudinit[871]: 2020/06/04 23:25:23 line 422: warning: unrecognized key "disk_setup"
Jun 04 23:25:23 k8s-master-36217359-0 coreos-cloudinit[871]: 2020/06/04 23:25:23 line 427: warning: unrecognized key "fs_setup"
Jun 04 23:25:23 k8s-master-36217359-0 coreos-cloudinit[871]: 2020/06/04 23:25:23 line 438: warning: unrecognized key "mounts"
Jun 04 23:25:23 k8s-master-36217359-0 coreos-cloudinit[871]: 2020/06/04 23:25:23 line 362: warning: unrecognized key "groups"
```

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -529,6 +546,85 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
+    retrycmd_if_failure 5 5 10 curl --retry 5 --retry-delay 10 --retry-max-time 10 --max-time 60 https://127.0.0.1:2379/v2/machines
+    {{- if EnableAggregatedAPIs}}
+    sudo bash /etc/kubernetes/generate-proxy-certs.sh
+    {{- end}}
+    touch /opt/azure/containers/runcmd.complete
+
+- path: "/etc/kubernetes/manifests/.keep"
+
+  {{- if .OrchestratorProfile.KubernetesConfig.RequiresDocker}}
+groups:
```
jackfrancis, line 566: @invidian, oem-cloudinit is complaining that `groups` is not recognized:

```
Jun 04 23:25:23 k8s-master-36217359-0 coreos-cloudinit[871]: 2020/06/04 23:25:23 line 362: warning: unrecognized key "groups"
```

parts/k8s/cloud-init/masternodecustomdata.yml
johnsonshi: @jackfrancis and I got together and diagnosed the issue preventing Flatcar from being provisioned correctly for master nodes.

When the Flatcar VM is provisioned as a master node VM with the etcd data disk, the udev rule is indeed fired and the symlink from `/dev/disk/azure/scsi1/lun0` is created. However, as seen in the coreos-cloudinit logs, the issue is that coreos-cloudinit does not recognize [these cloud-init directives](https://github.com/Azure/aks-engine/blob/94cfb92ccc122048f910b61cb5d1c6b2652a9537/parts/k8s/cloud-init/masternodecustomdata.yml#L532-L552).

The coreos-cloudinit codebase makes no mention of these upstream cloud-init directives. Lack of support and significant divergence between coreos-cloudinit and upstream cloud-init is preventing the etcd data disk from being set up correctly.

From a code maintainability point of view, it will be difficult to maintain aks-engine code to support VM provisioning using both upstream cloud-init and coreos-cloudinit. Perhaps Flatcar for worker nodes only is the way to go.

invidian: Oh, it worked by the time I submitted the PR, but febebda6a6883cf8a1ad4a91ef0f3fe6d9a1a20b changed the way etcddisk is mounted :disappointed:

invidian: Thanks for explanation @jackfrancis. I understand the reasoning behind this change and it seem indeed reasonable. It's just unfortunate that it happen and this time and that Flatcar does not support required options. 

I think it's fine if we add support only for workers for now and I will collaborate with Flatcar team to maybe add those missing options to Flatcar itself.

Who should revert the changes done in this PR for master nodes? Also, is there a way to keep this revert as a separate commit? Once Flatcar has support for missing options, it would be easier to add support for master nodes then.

jackfrancis: I'm happy to fix up this PR to make it worker node-only.

I'll file a backlog item that documents that flatcar isn't yet implemented for control plane VMs, and will reference your original PR, which is probably the cleanest changeset representation of how we may want to revisit that work in the future.

How does that sound?

invidian: Sounds good to me, thank you for your help!

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -429,7 +432,11 @@ ensureJournal() {
   systemctlEnableAndStart systemd-journald || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
 }
 installKubeletAndKubectl() {
-  if [[ ! -f "/usr/local/bin/kubectl-${KUBERNETES_VERSION}" ]]; then
+  path=/usr/local/bin
+  if [[ $OS == $FLATCAR_OS_NAME ]]; then
+    path=/opt
```
jackfrancis, line 437: @invidian is there a reason we're installing kubelet and kubectl in /opt/ instead of /opt/bin? Should we change this to /opt/bin?

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -175,12 +188,16 @@ installImg() {
 extractHyperkube() {
   CLI_TOOL=$1
   path="/home/hyperkube-downloads/${KUBERNETES_VERSION}"
+  targetpath="/usr/local/bin"
+  if [[ $OS == $FLATCAR_OS_NAME ]]; then
+    targetpath="/opt"
```
jackfrancis, line 193: sorry @invidian , earlier comment was for kublet only, same question for kubectl here, should we install in /opt/bin?

invidian: Yeah, it should be OK to install it in `/opt/bin`. I used `/opt`, as it was used for some other binaries IIRC.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -189,16 +206,26 @@ extractHyperkube() {
...
+  fi
 }
 extractKubeBinaries() {
   KUBE_BINARY_URL=${KUBE_BINARY_URL:-"https://kubernetesartifacts.azureedge.net/kubernetes/v${KUBERNETES_VERSION}/binaries/kubernetes-node-linux-amd64.tar.gz"}
   K8S_TGZ_TMP=${KUBE_BINARY_URL##*/}
   mkdir -p "${K8S_DOWNLOADS_DIR}"
   retrycmd_get_tarball 120 5 "$K8S_DOWNLOADS_DIR/${K8S_TGZ_TMP}" ${KUBE_BINARY_URL} || exit 31
+  path=/usr/local/bin
+  if [[ $OS == $FLATCAR_OS_NAME ]]; then
+    path=/opt
```
jackfrancis, line 225: @invidian and same comment here

parts/k8s/addons/blobfuse-flexvolume.yaml
```diff
@@ -21,6 +21,15 @@ spec:
 {{- end}}
     spec:
       priorityClassName: system-cluster-critical
+      affinity:
```
jackfrancis, line 24: @invidian adding these declarations instead of not allowing the addon to be installed in a flatcar scenario, as we support heterogenous node pool configurations

test/e2e/test_cluster_configs/flatcar/OWNERS
```diff
@@ -0,0 +1,2 @@
+maintainers:
+- invidian
```
jackfrancis, line 2: @invidian is there anyone else at kinvolk that would like to maintain this going forward?

invidian: I'll ask around, perhaps it would be good to have more than one person around. Maybe @iaguis?

test/e2e/test_cluster_configs/flatcar/flatcar.json
```diff
@@ -0,0 +1,59 @@
...
+						}
+					]
+				}
+			},
+			"masterProfile": {
+				"count": 1,
+				"dnsPrefix": "",
+				"vmSize": "Standard_D2_v3"
+			},
+			"agentPoolProfiles": [
```
johnsonshi, line 35: @jackfrancis Just reviewed these + the options defined in https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#agentPoolProfiles

LGTM for `agentPoolProfiles` test configs.

test/e2e/test_cluster_configs/flatcar/flatcar.json
```diff
@@ -0,0 +1,59 @@
...
+					"vmSize": "Standard_D2_v3",
+					"distro": "flatcar"
+				}
+			],
+			"linuxProfile": {
+				"adminUsername": "azureuser",
+				"ssh": {
+					"publicKeys": [
+						{
+							"keyData": ""
```
johnsonshi, line 48: @jackfrancis Is the doc at https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#notes-on-ssh-public-keys out of date? That doc suggests that at least 1 ssh pubkey is required for `linuxProfile`.

jackfrancis: Good question. The test runner that we ship as part of the AKS Engine project interprets these empty strings as "I'll create ssh keys for you", so that we don't have to publish public keys in public.

Something similar is happening for the empty string service principal values below.

test/e2e/test_cluster_configs/flatcar/OWNERS
iaguis: Thanks for volunteering me :D

Happy to be added here.

jackfrancis:
Observed consistent failures in HPA functionality in 1.19-beta.1, and filed an issue here:

https://github.com/Azure/aks-engine/issues/3446

Tests look good across 1.15-1.18, so this is ready to merge.

PR #3393, opened 2020-06-03
feat: re-enable container runtime dir relocation

alexeldeib:
This reverts commit 5af21fe837f7043d17afb2f8b807c851b4759573.

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
not sure if there have been material changes in between that should affect this, the revert was fairly clean. doing some manual testing.


alexeldeib:
Manually tested with docker + containerd since it's been a while since this was originally applied, still looks good though.

I would like to avoid special case logic for handling copy to temp disk at provision time. If our target is ephemeral OS disk or using data disk for docker, both of those will use a cached VHD. This avoids increasing provisioning latency. For temp disk, we will never make it the default and so will not increase latency by default. We will only allow it as an opt-in toggle in the short/medium-term, especially for targeted SKUs like N series.

When these SKUs support ephemeral OS disk or AKS supports using a data disk for docker root, we would de-prioritize temp disk and ask those same customers to use ephemeral OS + data disk, which will not have the same latency issues. In short I don't think it's worth it to detect if the user placed it on "/mnt" vs a data disk because we would remove that logic in the future anyway, and prefer for it not to be used.

alexeldeib:
I did some testing around provisioning time and using this option with a VHD (the AKS scenario). I tested provisioning time with docker on temp disk (this should trigger network pull at provision time)

Standard_D8s_v3 for all tests

100gb, default
```
real	3m37.317s
user	0m11.176s
sys	0m0.102s
```

100gb, reroot
```
real	3m16.865s
user	0m12.421s
sys	0m0.118s
```

1024gb, default
```
real	2m51.709s
user	0m12.575s
sys	0m0.222s
```

1024gb, reroot
```
real	2m53.774s
user	0m6.842s
sys	0m0.127s
```

copy from 100 GB OS disk:
```
time cp -R /var/lib/docker /mnt/container

real	9m24.323s
user	0m1.495s
sys	0m27.279s
```

Seems we have an enormous cache, but pulling a few images to temp disk on fast VMs is basically negligible, and we even gain some improvements by offloading from the OS disk. For the N-series SKUs we are primarily targeting with this change, temp disk will be even better. 

I'm waiting on a longer run with more data, but figured I'd share this in favor of letting the network pull happen.

PR #3401, opened 2020-06-03
fix: Set spg only when spg is not set 

zqingqing1:
**Reason for Change**:

We need to set SinglePlacementGroup default value only when it is not set even node count > 100.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -344,11 +344,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		if a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
-			if a.HasAvailabilityZones() {
-				a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = StandardLoadBalancerSku
-			} else {
-				a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = DefaultLoadBalancerSku
-			}
+			a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = a.GetLoadBalancerDefault()
 		}
 
 		if strings.ToLower(a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku) == strings.ToLower(BasicLoadBalancerSku) {
@@ -651,7 +647,10 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			}
 
 			if profile.SinglePlacementGroup == nil {
-				if p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == StandardLoadBalancerSku {
+				if p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
+					p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = p.GetLoadBalancerDefault()
+				}
+				if strings.EqualFold(p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku, StandardLoadBalancerSku) {
 					profile.SinglePlacementGroup = to.BoolPtr(false)
 				} else {
 					profile.SinglePlacementGroup = to.BoolPtr(DefaultSinglePlacementGroup)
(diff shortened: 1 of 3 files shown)


pkg/api/defaults.go
```diff
@@ -649,11 +649,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			if profile.VMSSOverProvisioningEnabled == nil {
 				profile.VMSSOverProvisioningEnabled = to.BoolPtr(DefaultVMSSOverProvisioningEnabled && !isUpgrade && !isScale)
 			}
-			if profile.Count > 100 {
-				profile.SinglePlacementGroup = to.BoolPtr(false)
-			}
+
 			if profile.SinglePlacementGroup == nil {
 				profile.SinglePlacementGroup = to.BoolPtr(DefaultSinglePlacementGroup)
```
xuto2, line 654: move this to the else part?

pkg/api/defaults.go
```diff
@@ -649,11 +649,12 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
...
+				if profile.Count > 100 {
+					profile.SinglePlacementGroup = to.BoolPtr(false)
```
xuto2, line 656: do we want to set to false if the cluster is BLB?

xuto2: the master branch behavior seems more accurate:
https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults.go#L546

pkg/api/defaults.go
zqingqing1: read the master branch, seems like the "profile.Count > 100" condition is removed. It only checks if it is SLB.

pkg/api/defaults.go
zqingqing1: I think it is right to consistent with master branch. removing count> 100 logic, instead of adding SLB checking logic.

pkg/api/defaults.go
```diff
@@ -649,11 +649,13 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			if profile.VMSSOverProvisioningEnabled == nil {
 				profile.VMSSOverProvisioningEnabled = to.BoolPtr(DefaultVMSSOverProvisioningEnabled && !isUpgrade && !isScale)
 			}
-			if profile.Count > 100 {
-				profile.SinglePlacementGroup = to.BoolPtr(false)
-			}
+
 			if profile.SinglePlacementGroup == nil {
-				profile.SinglePlacementGroup = to.BoolPtr(DefaultSinglePlacementGroup)
+				if p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == StandardLoadBalancerSku {
```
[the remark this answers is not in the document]
zqingqing1: Ok, sure

pkg/api/defaults.go
```diff
@@ -651,7 +647,10 @@ func (p *Properties) setAgentProfileDefaults(isUpgrade, isScale bool) {
 			}
 
 			if profile.SinglePlacementGroup == nil {
-				if p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == StandardLoadBalancerSku {
+				if p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
+					p.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = p.GetLoadBalancerDefault()
```
xuto2, line 651: The LB SKU is cluster default, should not be set in setAgentProfileDefaults
It think it's safe to assume LB sku is not "" at this step, otherwise we should fix in the orchestrator default

pkg/api/defaults.go
```diff
@@ -344,11 +344,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		}
 
 		if a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku == "" {
-			if a.HasAvailabilityZones() {
-				a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = StandardLoadBalancerSku
-			} else {
-				a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = DefaultLoadBalancerSku
-			}
+			a.OrchestratorProfile.KubernetesConfig.LoadBalancerSku = a.GetLoadBalancerDefault()
```
xuto2, line 347: nit: I won't try to refactor the aks-e code.

pkg/api/defaults.go
xuto2: actually it's set at L347 right?

zqingqing1: no, it is not like that. setAgentPoolDefaults() called before setOrchestratorDefaults(), where LB sku default sets. It is possible that LB sku is empty at setAgentPoolDefault().

Issue #3392, opened 2020-06-02
Multiple nginx ingress containers add large, inconsistent amounts of latency


gaieges:
We bumped into an odd issue the other day with a few different versions of k8s (1.16 and 1.17), with a new aks-engine library (versions below).  This previously did not happen when we were using earlier versions of aks-engine (`v0.44.2`)

We use the helm version of the nginx controller via the below install method:

```
helm upgrade --install nginx-ingress stable/nginx-ingress \
    --version 1.33.5 \ # also broken in 1.39.0
    --set rbac.create=true \
    --set controller.replicaCount=3 \
    --namespace kube-system \
    --wait \
    --timeout 6000
```

When we enable this configuration, setting to 3 replicas as defined above causes very inconsistent amounts of latency to the most basic of ingress configurations (ie the default 404 fallback ingress).  Hitting the IP defined by the below command with a simple `GET /` request will take up to 10s sometimes, will sometimes take 1s, and often somewhere in between.  

The trick to solving this was to reduce the replica count to 1 rather than > 1. 

This however is a problem, as now we have a central point of failure and have no way to add redundancy to this if a node dies (and we use spot nodes which often get pre-empted and take down service for a few seconds)

### Getting IP of the ingress
`kubectl get service nginx-ingress-controller -n kube-system -o jsonpath='{.status.loadBalancer.ingress[0].ip}'`

## Versions
- k8s: 
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-11-13T23:50:10Z", GoVersion:"go1.12.13", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.9", GitCommit:"a17149e1a189050796ced469dbd78d380f2ed5ef", GitTreeState:"clean", BuildDate:"2020-04-16T23:15:50Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
- aks-engine: `v0.51.0`
- helm: `v2.16.1`

cc: @durch @posix4e

posix4e:
It would also be nice to know what the load balancer configuration should look like

jackfrancis:
Hi folks, what's the cluster configuration (api model) you're using to create your cluster(s)? (reminder to scrub any secrets)

[...]

PR #3402, opened 2020-06-04
[WIP] fix: repro ingress-nginx replica latency

jackfrancis:
**Reason for Change**:

This PR introduces additional E2E coverage to attempt to repro https://github.com/Azure/aks-engine/issues/3392

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/kubernetes/deployment/deployment.go
+++ b/test/e2e/kubernetes/deployment/deployment.go
@@ -123,7 +123,7 @@ metadata:
     role: %s
   name: %s
 spec:
-  replicas: 1
+  replicas: 9
   selector:
     matchLabels:
       app: %s
--- a/test/e2e/kubernetes/kubernetes_test.go
+++ b/test/e2e/kubernetes/kubernetes_test.go
@@ -1463,7 +1463,7 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				By("Ensuring we can create a curl pod to connect to the service")
 				Expect(err).NotTo(HaveOccurred())
 				for i := 0; i < 100; i++ {
-					ilbCurlPod, err := pod.RunLinuxWithRetry("byrnedo/alpine-curl", "curl-to-ilb", "default", fmt.Sprintf("curl %s", sILB.Status.LoadBalancer.Ingress[0]["ip"]), false, 1*time.Minute, cfg.Timeout)
+					ilbCurlPod, err := pod.RunLinuxWithRetry("byrnedo/alpine-curl", "curl-to-ilb", "default", fmt.Sprintf("curl %s", sILB.Status.LoadBalancer.Ingress[0]["ip"]), true, 1*time.Minute, 1*time.Second, 30*time.Second)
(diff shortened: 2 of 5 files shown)


test/e2e/kubernetes/deployment/deployment.go
```diff
@@ -123,7 +123,7 @@ metadata:
     role: %s
   name: %s
 spec:
-  replicas: 1
+  replicas: 9
```
jackfrancis, line 126: artificially setting replica count high here

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -1422,8 +1422,12 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 				Expect(err).NotTo(HaveOccurred())
 
 				By("Ensuring we can create a curl pod to connect to the service")
-				ilbCurlPod, err := pod.RunLinuxWithRetry("byrnedo/alpine-curl", "curl-to-ilb", "default", fmt.Sprintf("curl %s", sILB.Status.LoadBalancer.Ingress[0]["ip"]), false, 1*time.Minute, cfg.Timeout)
-				Expect(err).NotTo(HaveOccurred())
+				for i := 0; i < 100; i++ {
+					ilbCurlPod, err := pod.RunLinuxWithRetry("byrnedo/alpine-curl", "curl-to-ilb", "default", fmt.Sprintf("curl %s", sILB.Status.LoadBalancer.Ingress[0]["ip"]), false, 1*time.Minute, 1*time.Second, 30*time.Second)
```
jackfrancis, line 1426: this test (which we run 100 times in succession) will fail if it takes longer than one second to curl to the LB IP

test/e2e/kubernetes/deployment/deployment.go
gaieges: Is this what gets used to set up the nginx pod count?  I presume our issue was something to do with the azure LB not being aware of what nodes the nginx pods are living on.

jackfrancis: Correct. These are the LB specs we test against the ingress-nginx deployment (w/ 9 replicas in this test):

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/ingress-nginx-ilb.yaml
https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/ingress-nginx-elb.yaml

As you can see, the LB routes to the pod (and eventually the node the pod is scheduled onto) via the app selector.

What's different about your scenario?

posix4e:
Neat;

test/e2e/kubernetes/deployment/deployment.go
posix4e: can we try this with k8s 1.16?

jackfrancis: Updated CI config, the next run will test 1.15 and 1.16

PR #3403, opened 2020-06-04
feat: Update container monitoring addon for may omsagent release 

rashmichandrashekar:
Updated api versions for cluster role and cluster role binding
Adding permissions for nodes/proxy
Updating agent image version to ciprod05262020

### DIFF
--- a/parts/k8s/addons/container-monitoring.yaml
+++ b/parts/k8s/addons/container-monitoring.yaml
@@ -22,7 +22,7 @@ metadata:
     addonmanager.kubernetes.io/mode: Reconcile
 ---
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: omsagent-reader
   labels:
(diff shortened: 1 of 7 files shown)


parts/k8s/addons/container-monitoring.yaml
```diff
@@ -22,15 +22,15 @@ metadata:
     addonmanager.kubernetes.io/mode: Reconcile
 ---
 kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
```
jackfrancis, line 25: Does this change work across all currently supported Kubernetes versions? (1.15-1.19)

pkg/api/addons_test.go
```diff
@@ -1564,7 +1564,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Enabled: to.BoolPtr(true),
 					Config: map[string]string{
 						"omsAgentVersion":       "1.10.0.1",
-						"dockerProviderVersion": "8.0.0-3",
+						"dockerProviderVersion": "10.0.0-1",
```
jackfrancis, line 1567: This changed in the unit test, but we didn't change the actual addon default, so the UT is failing.

rashmichandrashekar: good catch.. fixed it

parts/k8s/addons/container-monitoring.yaml
rashmichandrashekar: Yes this works.

PR #3415, opened 2020-06-05
feat: Dualstack support for Windows containers

tamilmani1989:
**Reason for Change**:

This PR adds dual stack support for azure cni windows cluster as like linux cluster. V4 Ips will be in vnet and v6 IPs will be in non-vnet. Non lifting validation check for windows dualstack until windows team release OS support. This PR also fixes creating multiple v6 configs per nic as its not supported.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubeproxystart.ps1
+++ b/parts/k8s/kubeproxystart.ps1
@@ -24,4 +24,9 @@ Import-Module $global:HNSModule
 # and https://github.com/kubernetes/kubernetes/pull/78612 for <= 1.15
 Get-HnsPolicyList | Remove-HnsPolicyList
 
-.$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+if (("--feature-gates=IPv6DualStack=true" | ? { $Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs -match $_ }) -ne $null) {
+    .$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --feature-gates=IPv6DualStack=true --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+}
+else {
+    .$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+}
--- a/parts/k8s/kubeproxystart.ps1
+++ b/parts/k8s/kubeproxystart.ps1
@@ -18,7 +18,7 @@ while (!$hnsNetwork) {
 }
 
 # add dualstack feature gate if dualstack enabled
-$isDualStackEnabled = ("--feature-gates=IPv6DualStack=true" | ? { $Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs -match $_ }) -ne $null
+$isDualStackEnabled = ("IPv6DualStack=true" | ? { $Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs -match $_ }) -ne $null
 if ($isDualStackEnabled) {
     $global:KubeproxyArgList += @("--feature-gates=IPv6DualStack=true")
 }
--- a/parts/k8s/kubeproxystart.ps1
+++ b/parts/k8s/kubeproxystart.ps1
@@ -7,13 +7,20 @@ if ($Global:ClusterConfiguration.Cni.Name -eq "kubenet") {
 
 $env:KUBE_NETWORK = $KubeNetwork
 $global:HNSModule = "c:\k\hns.psm1"
-$KubeDir = $Global:ClusterConfiguration.Install.Destination
+$global:KubeDir = $Global:ClusterConfiguration.Install.Destination
+$global:KubeproxyArgList = @("--v=3", "--proxy-mode=kernelspace", "--hostname-override=$env:computername", "--kubeconfig=$KubeDir\config")
 
 $hnsNetwork = Get-HnsNetwork | ? Name -EQ $KubeNetwork
 while (!$hnsNetwork) {
     Write-Host "$(Get-Date -Format o) Waiting for Network [$KubeNetwork] to be created . . ."
     Start-Sleep 10
     $hnsNetwork = Get-HnsNetwork | ? Name -EQ $KubeNetwork
+}
+
+# add dualstack feature gate if dualstack enabled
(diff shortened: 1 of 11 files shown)


tamilmani1989:
@mboersma fixed nit issue. can you please start pipeline again?

parts/k8s/kubeproxystart.ps1
```diff
@@ -24,4 +24,9 @@ Import-Module $global:HNSModule
 # and https://github.com/kubernetes/kubernetes/pull/78612 for <= 1.15
 Get-HnsPolicyList | Remove-HnsPolicyList
 
-.$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+if (("--feature-gates=IPv6DualStack=true" | ? { $Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs -match $_ }) -ne $null) {
```
marosset, line 27: Would it be possible to refactor this so that there is a single call to kube-proxy.exe at the end and conditionals are used when building an argument list?
That would make it easier to support other cases in the future and also more clear what is going on.

ksubrmnn: +1. see parts/k8s/kubeletstart.ps1 for an example

parts/k8s/kubeproxystart.ps1
```diff
@@ -24,4 +24,9 @@ Import-Module $global:HNSModule
...
+    .$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --feature-gates=IPv6DualStack=true --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+}
+else {
+    .$KubeDir\kube-proxy.exe --v=3 --proxy-mode=kernelspace --hostname-override=$env:computername --kubeconfig=$KubeDir\config
+}
```
ksubrmnn, line 32: add newline

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -357,7 +357,8 @@ try
                 -KubeServiceCIDR $global:KubeServiceCIDR `
                 -VNetCIDR $global:VNetCIDR `
                 {{- /* Azure Stack has discrete Azure CNI config requirements */}}
-                -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}}
+                -IsAzureStack {{if IsAzureStackCloud}}$true{{else}}$false{{end}} `
+                -IsDualStackEnabled {{if IsIPv6DualStackFeatureEnabled}}$true{{else}}$false{{end}}
```
ksubrmnn, line 361: Can we make a global var for isDualStackEnabled?

parts/k8s/kubeproxystart.ps1
tamilmani1989: sure .will do that

tamilmani1989:
@ksubrmnn @marosset please take a look, Addressed your comments.

parts/k8s/kubeproxystart.ps1
```diff
@@ -16,6 +17,12 @@ while (!$hnsNetwork) {
     $hnsNetwork = Get-HnsNetwork | ? Name -EQ $KubeNetwork
 }
 
+# add dualstack feature gate if dualstack enabled
+$isDualStackEnabled = ("--feature-gates=IPv6DualStack=true" | ? { $Global:ClusterConfiguration.Kubernetes.Kubelet.ConfigArgs -match $_ }) -ne $null
```
aramase, line 21: Would this check fail if `IPv6DualStack` feature gate was not first in the list `--feature-gates=CSIInlineVolume=true,IPv6DualStack=true`?

parts/k8s/windowsazurecnifunc.ps1
```diff
@@ -93,7 +101,25 @@ Set-AzureCNIConfig
...
     }
 
-    $configJson.plugins.AdditionalArgs[1].Value.DestinationPrefix  = $KubeServiceCIDR
+    if ($IsDualStackEnabled){
+        $configJson.plugins[0]|Add-Member -Name "ipv6Mode" -Value "ipv6nat" -MemberType NoteProperty
+        $serviceCidr = $KubeServiceCIDR -split ","
+        $configJson.plugins[0].AdditionalArgs[1].Value.DestinationPrefix = $serviceCidr[0]
+        $valueObj = [PSCustomObject]@{
+            Type = 'ROUTE'
+            DestinationPrefix = $serviceCidr[1]
```
aramase, line 110: What about the case where cluster is dual stack, but services are just single stack? The user could just have a single family CIDR for service.

parts/k8s/kubeproxystart.ps1
tamilmani1989: you're right, it will fail. let me change the match to "IPv6DualStack=true"

parts/k8s/windowsazurecnifunc.ps1
tamilmani1989: so if user didn't specify ipv6 service cidr, won't aks-e assign default service v6 cidr? May be im interpreting your ask wrongly

aramase: aks-e default v6 CIDR only for `ClusterCIDR`. Having v4 and v6 CIDR for services isn't mandatory.

tamilmani1989: i believe we should add default for v6 service cidr also here if user opted for dualstack. 
https://github.com/Azure/aks-engine/blob/24e9029e83d7349dcc9d1fab5e33ec0f3729e6e9/pkg/api/defaults.go#L278

aramase: @tamilmani1989 Having dual stack CIDRs for services is not mandatory unlike `ClusterCIDRs`. The user can just define single stack v4 or v6 service CIDR in dual stack cluster. Thats the reason for not appending v6 service CIDR unless the user explicitly requests it.

tamilmani1989: @aramase Yes you are right. Its not mandatory to create ipv6 svc but  if user decided want to have in future he has to recreate cluster. What's the user going to lose if we assign default svc cidr if its not explicitly specified?

tamilmani1989:
@aramase  as per our discussion offline, addressed those changes

jackfrancis:
Thank you @tamilmani1989!

marosset:
yay!

tamilmani1989:
Thank you guys!

PR #3416, opened 2020-06-05
fix: Correct CNI plugin download URL for Mooncake

AbelHu:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This commit is not cherry-pick from master. I will create a PR for master soon.

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -310,9 +310,9 @@ var (
 			EtcdDownloadURLBase:              "mcr.microsoft.com/oss/etcd-io/",
 			KubeBinariesSASURLBase:           DefaultKubernetesSpecConfig.KubeBinariesSASURLBase,
 			WindowsTelemetryGUID:             DefaultKubernetesSpecConfig.WindowsTelemetryGUID,
-			CNIPluginsDownloadURL:            "https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
-			VnetCNILinuxPluginsDownloadURL:   "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
-			VnetCNIWindowsPluginsDownloadURL: "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
+			CNIPluginsDownloadURL:            "https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/cni-plugins-linux-amd64-" + CNIPluginVer + ".tgz",
+			VnetCNILinuxPluginsDownloadURL:   "https://mirror.azk8s.cn/azure-cni/" + AzureCniPluginVerLinux + "/binaries/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
+			VnetCNIWindowsPluginsDownloadURL: "https://mirror.azk8s.cn/azure-cni/" + AzureCniPluginVerWindows + "/binaries/azure-vnet-cni-singletenancy-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
 			ContainerdDownloadURLBase:        "https://mirror.azk8s.cn/kubernetes/containerd/",
 		},
 		DCOSSpecConfig: DCOSSpecConfig{


andyzhangx:
would you also fix it in master branch? @AbelHu

pkg/api/azenvtypes.go
```diff
@@ -310,9 +310,9 @@ var (
 			EtcdDownloadURLBase:              "mcr.microsoft.com/oss/etcd-io/",
 			KubeBinariesSASURLBase:           DefaultKubernetesSpecConfig.KubeBinariesSASURLBase,
```
AbelHu, line 311: This may also needs to be updated.

AbelHu: @jackfrancis @marosset Could you help to confirm whether `KubeBinariesSASURLBase` is not used now? If so, we do not need to update this URL for Mooncake and update our mirror settings to sync the binaries. BTW: This PR will only be merged for AKS. Later I will create a new one to master branch.

https://github.com/Azure/aks-engine/blob/6c1891c93de325ccfc4949c2e696707de3642b23/pkg/engine/params_k8s.go#L89-L101

pkg/api/azenvtypes.go
```diff
@@ -310,9 +310,9 @@ var (
...
 			WindowsTelemetryGUID:             DefaultKubernetesSpecConfig.WindowsTelemetryGUID,
-			CNIPluginsDownloadURL:            "https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/cni-plugins-amd64-" + CNIPluginVer + ".tgz",
-			VnetCNILinuxPluginsDownloadURL:   "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
-			VnetCNIWindowsPluginsDownloadURL: "https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-windows-amd64-" + AzureCniPluginVerWindows + ".zip",
+			CNIPluginsDownloadURL:            "https://mirror.azk8s.cn/kubernetes/containernetworking-plugins/cni-plugins-linux-amd64-" + CNIPluginVer + ".tgz",
```
jackfrancis, line 313: This change already exists in master.

pkg/api/azenvtypes.go
```diff
@@ -310,9 +310,9 @@ var (
...
+			VnetCNILinuxPluginsDownloadURL:   "https://mirror.azk8s.cn/azure-cni/" + AzureCniPluginVerLinux + "/binaries/azure-vnet-cni-linux-amd64-" + AzureCniPluginVerLinux + ".tgz",
```
jackfrancis, line 314: These changes will need to be applied to master, if they are correct generally. See:

https://github.com/Azure/aks-engine/blob/master/pkg/api/azenvtypes.go#L285

pkg/api/azenvtypes.go
jackfrancis: `KubeBinariesSASURLBase` is definitely used if `kubernetesConfig.CustomWindowsPackageURL` is not specified. Looking through the code, it seems like `kubernetesConfig.CustomWindowsPackageURL` is a user-provided configuration, so the normal (default) cluster configuration will exclude it.

So, yes, this line will depend upon `KubeBinariesSASURLBase`:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/kuberneteswindowssetup.ps1#L68

pkg/api/azenvtypes.go
paulgmiller: I didn't think they'd changed the azure mirror in china's lay out yet.
https://mirror.azk8s.cn/kubernetes/azure-container-networking/azure-vnet-cni-linux-amd64-v1.1.3.tgz
still works
but so does
https://mirror.azk8s.cn/azure-cni/v1.1.3/binaries/azure-vnet-cni-linux-amd64-v1.1.3.tgz 

So that's good.

paulgmiller: Ah but there's also an update to the filename to add singletenancy

pkg/api/azenvtypes.go
xuto2: This file is not in AKS VHD. I would rather we hold it

pkg/api/azenvtypes.go
AbelHu: Thanks to confirm it, @jackfrancis. We should also update the `KubeBinariesSASURLBase` and make sure our proxy for Mooncake can provide these binaries. @andyzhangx

pkg/api/azenvtypes.go
AbelHu: Yes. We are preparing a PR for the master. https://github.com/Azure/aks-engine/compare/master...AbelHu:dev?expand=1

AbelHu:
Abandon it since Paul has changed two URLs for Mooncake. I will follow `KubeBinariesSASURLBase `

PR #3127, opened 2020-04-22
fix: don't hardcode csi enableproxy in kubeclusterconfig.json

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3434, opened 2020-06-09
chore: Add EnableAHUB in WindowsProfile (#3322)

AbelHu:
* Add LicenseType in WindowsProfile

* fix run tests in parallel with same variable

* Change LicenseType to EnableAHUB

* Support LicenseType in VMAS

* Update doc

* Add E2E test config for Windows AHUB

**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:
Cherry-pick from master

- chore: Add EnableAHUB in WindowsProfile (#3322)

- fix: don't hardcode csi enableproxy in kubeclusterconfig.json (#3127)


pkg/api/convertertoapi_test.go
```diff
@@ -918,9 +920,9 @@ func TestConvertVLabsWindowsProfile(t *testing.T) {
 	}
 
 	for _, c := range cases {
+		c := c
```
andyliuliming, line 923: what's this?

pkg/api/vlabs/types.go
```diff
@@ -158,22 +158,32 @@ type CustomNodesDNS struct {
...
+	ImageVersion              string            `json:"imageVersion,omitempty"`
+	WindowsImageSourceURL     string            `json:"WindowsImageSourceUrl"`
+	WindowsPublisher          string            `json:"WindowsPublisher"`
+	WindowsOffer              string            `json:"WindowsOffer"`
+	WindowsSku                string            `json:"WindowsSku"`
+	WindowsDockerVersion      string            `json:"windowsDockerVersion"`
+	Secrets                   []KeyVaultSecrets `json:"secrets,omitempty"`
+	SSHEnabled                *bool             `json:"sshEnabled,omitempty"`
+	EnableAutomaticUpdates    *bool             `json:"enableAutomaticUpdates,omitempty"`
+	IsCredentialAutoGenerated *bool             `json:"isCredentialAutoGenerated,omitempty"`
```
andyliuliming, line 185: no coverage for this yet?

andyliuliming: "IsCredentialAutoGenerated"

pkg/engine/virtualmachinescalesets.go
```diff
@@ -611,6 +611,11 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 			CustomData: to.StringPtr(customDataStr),
 		}
 		vmssVMProfile.OsProfile = &windowsOsProfile
+
+		if cs.Properties.WindowsProfile.GetEnableAHUB() {
```
andyliuliming, line 615: may WindowsProfile be nil here?

pkg/api/convertertoapi_test.go
AbelHu: This must be set to capture range variable. Otherwise, only one case will be run multiple times. This will cause code coverage drop.
Reference: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks

pkg/api/vlabs/types.go
AbelHu: Normally we do not add more test cases for this file. This will only be used in aks-engine.

pkg/engine/virtualmachinescalesets.go
AbelHu: No. Since above code has checked the agent pool is Windows.

pkg/api/convertertoapi_test.go
andyliuliming, line 923: I would suggest to give it a new variable name.
say localCase.

andyliuliming: localCase := c

and then use localCase inner the t.Runblabla

pkg/api/vlabs/types.go
andyliuliming: if the "IsCredentialAutoGenerated" is only used in aks-engine. then I would suggest remove this from the PR.

pkg/api/convertertoapi_test.go
AbelHu: Good suggestion. We could change it later in master. Since this is a cherry-pick, let us keep it in this PR.

pkg/api/vlabs/types.go
AbelHu: no. it is for aks. When users do not provide username and password to create a k8s cluster with VMSS and azure cni, AKS will generate username and password automatically and set this to true.

pkg/api/convertertoapi_test.go
AbelHu: And you also can see the example in the reference: https://golang.org/pkg/testing/#hdr-Subtests_and_Sub_benchmarks. The name should be ok.

pkg/api/vlabs/types.go
AbelHu: But we need to add `enableAHUB` in WindowsProfile into aks-engine. That is why you see `IsCredentialAutoGenerated ` in this PR. When I added `IsCredentialAutoGenerated `, I did not add it in this file.

andyliuliming:
@xuto2 I will merge this PR first for Abel, please ping me if any concerns.

Issue #3435, opened 2020-06-09
DNS fails intermitentnly with Win-bridge (kubenet) on Windows


jsturtevant:
**Describe the bug**
Seeing intermittent failures with kubenet on windows nodes in the e2e tests:

```
020/06/08 20:48:17 #### $ k logs windows-2-linux-kubernetes-westus2-53936-18444 -c windows-2-linux-kubernetes-westus2-53936-18444 -n default completed in 250.590741ms
2020/06/08 20:48:17 
iwr : The remote name could not be resolved: 
'nginx-dns-kubernetes-westus2-53936-8014'
At line:1 char:1
+ iwr -UseBasicParsing -TimeoutSec 60 nginx-dns-kubernetes-westus2-5393 ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (System.Net.HttpWebRequest:Htt 
   pWebRequest) [Invoke-WebRequest], WebException
    + FullyQualifiedErrorId : WebCmdletWebResponseException,Microsoft.PowerShe 
   ll.Commands.InvokeWebRequestCommand
 

2020/06/08 20:48:17 $ k describe pod windows-2-linux-kubernetes-westus2-53936-18444 -n default
2020/06/08 20:48:17 #### $ k describe pod windows-2-linux-kubernetes-westus2-53936-18444 -n default completed in 197.26936ms
2020/06/08 20:48:17 
Name:         windows-2-linux-kubernetes-westus2-53936-18444
Namespace:    default
Priority:     0
Node:         3641k8s00000001/10.240.0.5
Start Time:   Mon, 08 Jun 2020 20:28:13 -0700
Labels:       run=windows-2-linux-kubernetes-westus2-53936-18444
Annotations:  kubernetes.io/psp: privileged
Status:       Failed
IP:           10.244.1.115
IPs:
  IP:  10.244.1.115
Containers:
  windows-2-linux-kubernetes-westus2-53936-18444:
    Container ID:  docker://e9e0a75eb09a497e1882a60ca76edf3cad8d535ee0cb57a93c2d8402fbca067e
    Image:         mcr.microsoft.com/windows/servercore:ltsc2019
[...]

Issue #3429, opened 2020-06-08
Windows pod never recovered after VM restart


jackfrancis:
Observed an E2E failure now that we're testing restarting VMs and validating that scheduled pods are able to recover.

```
Events:
  Type    Reason          Age                    From                      Message
  ----    ------          ----                   ----                      -------
  Normal  Scheduled       30m                    default-scheduler         Successfully assigned default/iis-kubernetes-eastus-692-91403-674d68f9ff-w4ntq to 1514k8s00000002
  Normal  Pulled          30m                    kubelet, 1514k8s00000002  Container image "mcr.microsoft.com/windows/servercore/iis:20191112-windowsservercore-ltsc2019" already present on machine
  Normal  Created         30m                    kubelet, 1514k8s00000002  Created container iis-kubernetes-eastus-692-91403
  Normal  Started         30m                    kubelet, 1514k8s00000002  Started container iis-kubernetes-eastus-692-91403
  Normal  SandboxChanged  3m10s (x117 over 28m)  kubelet, 1514k8s00000002  Pod sandbox changed, it will be killed and re-created.
```

Essentially, the pod was in a terminal Error state after the restart. A manual delete of the pod worked (the deployment created a new one that quickly went into a Running state).

jackfrancis:
This may be a kubenet-only error symptom (i.e., not reproduced if Azure CNI-backed)

PR #3436, opened 2020-06-09
test: Disable failing kubnet tests

jsturtevant:
**Reason for Change**:

Disables failing tests in where kubenet does not work with node restarts.   Disabling the tests as kubenet is used for Debugging/test scenarios.  

**Issue Fixed**:

Addresses #3429 and #3435

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Currently Windows + Kubenet is for development and testing only, not recommended for production

### DIFF
--- a/pkg/api/vlabs/validate.go
+++ b/pkg/api/vlabs/validate.go
@@ -1667,6 +1667,10 @@ func (k *KubernetesConfig) validateNetworkPlugin(hasWindows bool) error {
 		return errors.Errorf("networkPlugin '%s' is not supporting windows agents", networkPlugin)
 	}
 
+	if networkPlugin == NetworkPluginKubenet && hasWindows {
+		log.Warnf("Windows + Kubenet is for development and testing only, not recommended for production")
+	}
+
 	return nil
 }
 
--- a/test/e2e/test_cluster_configs/windows/network_plugin/kubenet.json
+++ b/test/e2e/test_cluster_configs/windows/network_plugin/kubenet.json
@@ -1,5 +1,8 @@
 {
-	"env": {},
+	"env": {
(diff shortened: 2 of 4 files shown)


pkg/api/vlabs/validate.go
```diff
@@ -1667,6 +1667,10 @@ func (k *KubernetesConfig) validateNetworkPlugin(hasWindows bool) error {
 		return errors.Errorf("networkPlugin '%s' is not supporting windows agents", networkPlugin)
 	}
 
+	if networkPlugin == NetworkPluginKubenet && hasWindows {
```
jackfrancis, line 1670: Let's unit test this using the super cool `Example`-style UT. See `ExampleProperties_validateLocation` as an example in `validate_test.go`.

test/e2e/test_cluster_configs/windows/network_plugin/kubenet.json
```diff
@@ -1,5 +1,8 @@
 {
-	"env": {},
+	"env": {
+		"GINKGO_SKIP": "resolve DNS across windows and linux deployments|should expect containers to be recreated after node restart",
```
marosset, line 3: Should we disable this with a ginkgo.Skip() in code if we detect Windows + kubenet?
That seems to be the more common pattern.
@jackfrancis what do you think?

jackfrancis: Right, we can add an additional condition here (AND NOT is kubenet):

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L1934

And then modify the skip message to say something like "we only run this test on Windows-enabled clusters using Azure CNI".

And yeah, then we can get rid of the updates to this test config file, as the test runner itself will always skip those non-working tests in a kubenet context.

PR #3447, opened 2020-06-11
fix: Do not mutate OrchestratorVersion when set default

gtxistxgao:
**Reason for Change**:

When provision or update VMAS cluster, AKS RP will first call deployTemplateWithoutAgentPool.
When call deployTemplateWithoutAgentPool, it will follow the below calling trace and AKS-Engine changes the managed cluster's OrchestratorVersion silently.

deployTemplateWithoutAgentPool =>
GenerateTemplate() => 
SetPropertiesDefaults() =>
setOrchestratorDefaults() =>
    o.OrchestratorVersion = common.GetValidPatchVersion(
        o.OrchestratorType,
        o.OrchestratorVersion, isUpdate, a.HasWindows())

This is out of the expectation of AKS RP behavior, we should not adjust the MC's OrchestratorVersion if already given one.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -117,9 +117,11 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		return
 	}
 	o := a.OrchestratorProfile
-	o.OrchestratorVersion = common.GetValidPatchVersion(
-		o.OrchestratorType,
-		o.OrchestratorVersion, isUpdate, a.HasWindows())
+	if o.OrchestratorVersion == "" {
+		o.OrchestratorVersion = common.GetValidPatchVersion(
+			o.OrchestratorType,
+			o.OrchestratorVersion, isUpdate, a.HasWindows())
+	}
 
 	switch o.OrchestratorType {
 	case Kubernetes:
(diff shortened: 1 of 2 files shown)


pkg/api/defaults.go
```diff
@@ -117,9 +117,11 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 		return
 	}
 	o := a.OrchestratorProfile
-	o.OrchestratorVersion = common.GetValidPatchVersion(
-		o.OrchestratorType,
-		o.OrchestratorVersion, isUpdate, a.HasWindows())
+	if o.OrchestratorVersion == "" {
```
xuto2, line 120: would this cause issue later if the existing version is not supported?

Shouldn't we error out before calling this?

gtxistxgao: if we want to do that, it's better to be done in RP. That's a behavior change. That means when the version is not supported, even scale is not allowed for VMAS cluster.

gtxistxgao: For creating cluster scenario, RP already block it if the k8s version is not supported.
For updating, this change will also let AKS-E honor the input k8s version when generating template.

PR #3448, opened 2020-06-11
chore: add csi windows related images

ZeroMagic:
**Reason for Change**:

These images will be used by CSI Drivers.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

/cc @andyzhangx @xuto2

### DIFF
--- a/vhd/packer/configure-windows-vhd.ps1
+++ b/vhd/packer/configure-windows-vhd.ps1
@@ -47,7 +47,9 @@ function Get-ContainerImages
     $imagesToPull = @(
         "mcr.microsoft.com/windows/servercore:ltsc2019",
         "mcr.microsoft.com/windows/nanoserver:1809",
-        "mcr.microsoft.com/oss/kubernetes/pause:1.3.2")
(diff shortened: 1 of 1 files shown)


andyzhangx:
@marosset could you approve? thanks. We will make csi-proxy enabled by default on AKS windows agent node and by caching these images, the csi driver on windows could be loaded quite faster.

ZeroMagic:
Rebase to the master to solve the latest conflict.

marosset:
We don't currently have logic to remove un-used container images for Windows nodes and I'm a little hesitant to include container images for a feature still in alpha.

@MikkelHegn @AbelHu @keikhara  do you have any concerns here from AKS?

Should we add logic to provisioning to remove Windows related CSI images if CSI proxy isn't enabled for Windows?

/cc @jsturtevant

marosset:
Sorry, i missed the comment above where AKS is trying to enable CSI proxy by default.
If that is the case I don't have concerns with this.

/LGTM

PR #3454, opened 2020-06-11
fix: make security context configs more restrictive

vtikoo:
**Reason for Change**:
Updated the security context related configs of the sgx device plugin to be more restrictive.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/sgx/device-plugin-before-k8s-1-17.yaml
+++ b/docs/topics/sgx/device-plugin-before-k8s-1-17.yaml
@@ -42,7 +42,10 @@ spec:
         - name: dev-sgx
           mountPath: /dev/sgx
         securityContext:
-          privileged: true
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+              - ALL
       volumes:
       - name: device-plugin
         hostPath:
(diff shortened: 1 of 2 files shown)


jackfrancis:
We should definitely keep the E2E test up-to-date w/ doc recommendations!

vtikoo:
@salsal97 @jackfrancis the e2e test is a slightly different scenario, as its not using a device plugin and directly accessing /dev/sgx on the host.
We can however update the e2e test to use a device plugin in a future PR, as thats the workflow we are recommending users. Any thoughts on that?

jackfrancis:
@vtikoo Makes sense. So it sounds like we're not really testing/validating what we're recommending?

What would be involved to update our tests? Is it as simple as updating the spec we kubectl apply onto the cluster?

vtikoo:
It would involve installing the sgx device plugin before the actual test workload runs. 

For installing the device plugin, we could either kubectl apply it and wait for it to be ready, or create an addon/extension. The latter would be more effort, but might be useful to users as well.

jackfrancis:
Right, thanks for clarifying, yeah, let's add a follow-up task to do the kubectl apply + run updated spec in the E2E tests.

jackfrancis:
Added this backlog issue: https://github.com/Azure/aks-engine/issues/3462

PR #3468, opened 2020-06-12
chore: Add June Windows Patches

jsturtevant:
**Reason for Change**:

Adds release notes and updates default windows image to use June's patches

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
CC: @AbelHu

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -182,7 +182,7 @@ var (
 		ImageOffer:     "aks-windows",
 		ImageSku:       "2019-datacenter-core-smalldisk-2006",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.1217.200603",
+		ImageVersion:   "17763.1282.200610",
 	}
 
 	// WindowsServer2019OSImageConfig is the 'vanilla' Windows Server 2019 image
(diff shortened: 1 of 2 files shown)


pkg/api/azenvtypes.go
```diff
@@ -182,7 +182,7 @@ var (
 		ImageOffer:     "aks-windows",
 		ImageSku:       "2019-datacenter-core-smalldisk-2006",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.1217.200603",
+		ImageVersion:   "17763.1282.200610",
 	}
 
 	// WindowsServer2019OSImageConfig is the 'vanilla' Windows Server 2019 image
```
marosset, line 188: Can you update this one to use the same ImageVersion the aks-windows VHD was based off of?

jsturtevant: For the vanilla windows server the June patches haven't been released as an image yet.  So I believe does match: https://github.com/Azure/aks-engine/pull/3439/files#diff-d61a27407ad48d7a4154938c47584c43

marosset: Wow - yea.
I've never seen the -with-containers image not get released at the same time as the vanilla WS2019 images..

PR #3488, opened 2020-06-16
feat: ensure all AKS required ports on api-server are exposed

gtracer:
**Reason for Change**:

AKS requires two ports to be exposed beyond 443. These ports are used for tunnelfront to connect to tunnelend and when customers block them (firewall etc.) it causes clusters to fail

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Looks to me like the apiserver connection check is sectioned to aks only since it is for hosted master but if that is wrong and aks-engine clusters could also trigger this code then please let me know as these ports are only needed for aks clusters


jackfrancis:
Hi @gtracer, if you run `make generate` and then commit the resultant generated code we can kick off E2E

jveski:
lgtm for the tunnel ports 👍

PR #3490, opened 2020-06-16
fix: scale + AvailabilitySet errata

### LINKED ISSUE #3489: Intermittent Windows + AvailabilitySet scale error
Seeing this error intermittently in our AvailabilitySet scale tests:

```
aks-engine scale --azure-env AzurePublicCloud --subscription-id **** --api-model _output/kubernetes-southeastasia-67768/apimodel.json --location southeastasia --resource-group kubernetes-southeastasia-67768 --apiserver kubernetes-southeastasia-67768.southeastasia.cloudapp.azure.com --node-pool poolwin --new-node-count 1 --auth-method client_secret --client-id **** --client-secret ****
 Error: None of the VMs in the provided resource group contain any nodes
```

The resource group *does* have VMs, so that error indicates that the resource group didn't return any VMs that matched what we'd expect given the agentPoolProfile name value.

jackfrancis:
**Reason for Change**:

This PR addresses three errors observed:

- `printScaleTargetEqualsExisting` convenience func doesn't return at the right time when it can't print
- retry calling the VM API when looking for nodes in the case where the call produces no matching VMs
- don't rely on tags being present on Windows VMs to correlate VMs w/ node pools, because in practice tags just sometimes aren't there :(

**Issue Fixed**:

Fixes #3489 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -15,9 +15,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/aks-engine/pkg/armhelpers/utils"
+
 	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/aks-engine/pkg/armhelpers"
-	"github.com/Azure/aks-engine/pkg/armhelpers/utils"
 	"github.com/Azure/aks-engine/pkg/engine"
 	"github.com/Azure/aks-engine/pkg/engine/transform"
 	"github.com/Azure/aks-engine/pkg/helpers"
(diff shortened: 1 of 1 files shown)


cmd/scale.go
```diff
@@ -560,11 +570,14 @@ func (sc *scaleCmd) vmInAgentPool(vmName string, tags map[string]*string) bool {
 		}
 	}
 
-	// For Windows, we rely upon the tags
-	if sc.agentPool.OSType == api.Windows {
+	// Fall back to checking the VM name to see if it fits the naming pattern
+	if sc.agentPool.OSType == api.Windows && strings.HasPrefix(vmName, sc.containerService.Properties.GetClusterID()[:4]) {
```
jackfrancis, line 574: Finally implementing a fallback for Windows VMs if the tag isn't present.

PR #3513, opened 2020-06-20
fix: Make sure nvidia utility and compute bins can be mounted into container

bowang-666:
**Reason for Change**:
Make sure nvidia utility and compute bins can be mounted into container as documented in nvidia-container-runtime readme.
https://github.com/Azure/aks-engine/pull/3029

**Issue Fixed**:
NVIDIA utility/compute bins are not mounted into containers when they're required in container env variable as NVIDIA_DRIVER_CAPABILITIES=compute,utility

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


xuto2:
LGTM

PR #3530, opened 2020-06-24
feat: enable alternate kube-reserved cgroups

alexeldeib:
cherry-pick #3201 to AKS

this is a somewhat old commit, and contains a lot of stuff that is irrelevant with agentbaker. mostly the field "kubeReservedCgroup" is important in KubernetesConfig.


pkg/engine/template_generator.go
```diff
@@ -295,6 +295,18 @@ func getContainerServiceFuncMap(cs *api.ContainerService) template.FuncMap {
 			}
 			return kc.GetOrderedKubeletConfigStringForPowershell()
 		},
+		"HasKubeReservedCgroup": func() bool {
+			kc := cs.Properties.OrchestratorProfile.KubernetesConfig
```
xuto2, line 299: May need some tuning when really enable this in AKS due containerd in agentpool level

PR #3532, opened 2020-06-24
fix: use kube-proxy image instead of hyperkube for 1.16

### LINKED ISSUE #3529: 1.16.11 hyperkube assumes nf_tables for kube-proxy
Our current 1.16.11 implementation uses the hyperkube image for kube-proxy. The 1.16.11 hyperkube image is based on Debian buster, and assumes nf_tables instead of legacy iptables. Because our Ubuntu images assume legacy iptables, kube-proxy breaks.

The 1.16.11 kube-proxy image (not hyperkube) includes "which iptables should I use?" wrapping, which supports either nf_tables or legacy iptables depending on which one is present on the underlying OS that it's scheduled onto. If we can convert our 1.16.11 kube-proxy implementation to use that image we should be able to deliver a working 1.16.11.

chewong:
Use kube-proxy image instead of hyperkube for 1.16.

**Reason for Change**:

**Issue Fixed**:

Fixes #3529.

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/kube-proxy.yaml
+++ b/parts/k8s/addons/kube-proxy.yaml
@@ -82,7 +82,7 @@ spec:
         operator: Exists
       containers:
       - command:
-{{- if not (IsKubernetesVersionGe "1.17.0")}}
+{{- if not (IsKubernetesVersionGe "1.16.0")}}
         - /hyperkube
 {{- end}}
         - kube-proxy
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -768,7 +768,7 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		Containers: []KubernetesContainerSpec{
 			{
 				Name:  common.KubeProxyAddonName,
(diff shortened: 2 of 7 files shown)


parts/k8s/addons/kube-proxy.yaml
```diff
@@ -82,9 +82,6 @@ spec:
         operator: Exists
       containers:
       - command:
-{{- if not (IsKubernetesVersionGe "1.17.0")}}
```
jackfrancis, line 85: Removing this entirely breaks 1.15. If we have a 1.15 kube-proxy image, we can test that, too, otherwise maybe easiest to just change "1.17.0" to "1.16.0" in the original conditional

jadarsie:
testing on my side as well

chewong:
/hold 
@jadarsie will add an additional commit to make it work on Azure Stack.

jadarsie:
thx @chewong, so far so good, waiting for e2e to finish

PR #3536, opened 2020-06-24
fix: always use /var/run/reboot-required

### LINKED ISSUE #3533: improve reboot detection in CSE
Let's check for /var/run/reboot-required at the end of CSE instead of at the beginning to account for that requirement being influenced by CSE operations.

jackfrancis:
**Reason for Change**:

This PR standardizes the CSE bootstrapping script to always evaluate `[ -f /var/run/reboot-required ]` (does the file exist?) to determine whether or not we need to reboot.

In addition, in the case where we *want* to reboot, we touch that file, which as far as I can tell is a safe tactic as `/var/run` is tmpfs and files therein will be reliably removed upon reboot. Cf:

```
azureuser@k8s-master-31453872-1:~$ ls -la /var/run
lrwxrwxrwx 1 root root 4 Nov 13  2019 /var/run -> /run
azureuser@k8s-master-31453872-1:~$ df -h /run
Filesystem      Size  Used Avail Use% Mounted on
tmpfs           796M  1.2M  795M   1% /run
```

**Issue Fixed**:

Fixes #3533

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -259,7 +259,7 @@ configureCNI() {
   {{if HasCiliumNetworkPlugin}}
   systemctl enable sys-fs-bpf.mount
   systemctl restart sys-fs-bpf.mount
-  REBOOTREQUIRED=true
+  touch /var/run/reboot-required
   {{end}}
 {{- if IsAzureStackCloud}}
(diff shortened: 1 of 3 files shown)


jackfrancis:
cc @MichaelSinz

Michael-Sinz:
Looks good to me.

PR #3540, opened 2020-06-25
fix: downgrade blobfuse to 1.1.1 since there is breaking change on 1.2.3

andyzhangx:
**Reason for Change**:

fix: downgrade blobfuse to 1.1.1 since there is breaking change on 1.2.3
more details could be found here: https://github.com/Azure/kubernetes-volume-drivers/issues/66#issuecomment-649188681

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
`make generate` and commit generated code to pass UT

PR #3549, opened 2020-06-26
chore: Use moby 19.03.x packages

cpuguy83:
**Reason for Change**:

Update moby verison.

**Issue Fixed**:

N/A

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

This is the same as 3.0.13, except the new versioning to match upstream
and things are packaged differently (e.g. containerd is now managed with
systemd instead of by dockerd).
The packaging/build changes are fairly significant, but the underlying
moby commits are the same.

### DIFF
--- a/vhd/packer/install-dependencies.sh
+++ b/vhd/packer/install-dependencies.sh
@@ -70,6 +70,7 @@ echo "  - bpftrace" >> ${VHD_LOGS_FILEPATH}
 
 MOBY_VERSION="19.03.12"
 installMoby
+systemctl start docker
 echo "  - moby v${MOBY_VERSION}" >> ${VHD_LOGS_FILEPATH}
 downloadGPUDrivers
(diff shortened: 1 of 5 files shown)


vhd/packer/install-dependencies.sh
```diff
@@ -68,7 +68,7 @@ echo "  - apmz $apmz_version" >> ${VHD_LOGS_FILEPATH}
 installBpftrace
 echo "  - bpftrace" >> ${VHD_LOGS_FILEPATH}
 
-MOBY_VERSION="3.0.13"
+MOBY_VERSION="19.03.12"
 installMoby
```
jackfrancis, line 72: @cpuguy83 add the working "systemctl start docker" command here below `installMoby`

cpuguy83: Adding this here so I can directly commit. :)

```suggestion
installMoby
systemctl start docker
```

PR #3551, opened 2020-06-26
feat: add support for Kubernetes v1.16.12

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#changelog-since-v11611

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

This has a temporary commit for testing that should be discarded before merging.

mboersma:
This [passed CI](https://github.com/Azure/aks-engine/pull/3551/checks?check_run_id=812796438), so I'll strip off the testing commit and should be ready for review.

mboersma:
>  Could you add [kube-proxy] 1.16.12...

Good catch! Added.

chewong:
/hold
I think there are plans to drop 1.16.11 support after v0.53.0 is released and resume using hyperkube for kube-proxy image.

jackfrancis:
@chewong @mboersma we don't need to worry about dropping v1.16.11 with this change, we can continue to include 1.16 kube-proxy artifacts in the VHD spec as @chewong pointed out earlier in this PR thread. We're good on that path!

PR #3560, opened 2020-06-29
test: testing the recommended flow of sgx-device-plugin

### LINKED ISSUE #3462: Add SGX device plugin E2E test case
Currently we cover the SGX node pool scenario by relying upon CSE to install the drivers here:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/cse_config.sh#L610

And then validate that we can run this workload to completion:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/sgx-test.yaml

Let's continue to do that, but also add a test where we kubectl apply sgx device drivers and run a test that matches the recommended SGX workflow documented in the repo.

salsal97:
**Reason for Change**:

This PR adds a test in the pipeline that deploys the sgx device plugin, and then a test workload in order to check the proper exposure of EPC memory by the plugin & also test the recommended way to use DC series SKUs on aks-engine

**Issue Fixed**:

Fixes #3462

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/kubernetes/workloads/sgx-device-plugin.yaml
```diff
@@ -0,0 +1,56 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: sgx-device-plugin
+  namespace: kube-system
```
jackfrancis, line 5: The namespace in the spec doesn't match the namespace we're looking for in the test ("default"). In general, we should put these out of kube-system namespace, unless there's a reason to. (In which case we'd change the test to validate the pod creation in the "kube-system" namespace).

salsal97: I feel stupid, thanks for pointing that out

jackfrancis: I recommend feeling silly or something positive instead!

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -1620,6 +1620,37 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
...
+			if eng.ExpandedDefinition.Properties.HasDCSeriesSKU() {
+
+				sgx_device_plugin := ""
+				if common.IsKubernetesVersionGe(eng.ExpandedDefinition.Properties.OrchestratorProfile.OrchestratorVersion, "1.17.0") {
+					sgx_device_plugin = "sgx-device-plugin.yaml"
+				} else {
+					sgx_device_plugin = "sgx-device-plugin-before-k8s-1-17.yaml"
+				}
+
+				p, err := pod.CreatePodFromFileIfNotExist(filepath.Join(WorkloadDir, sgx_device_plugin), "sgx-device-plugin", "default", 1*time.Second, cfg.Timeout)
```
jackfrancis, line 1634: So, the `CreatePodFromFileIfNotExist` function was designed to `kubectl apply` a `Pod` spec (see https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/dns-liveness.yaml).

The reason a DaemonSet spec won't work as coded here is that the actual Pod resource rendered by the DaemonSet will have a unique suffix appended to its name. So we can't expect a pod to be present in the cluster literally named "sgx-device-plugin". We *can*, however use that func to kubectl apply the spec (just drop the first response object, e.g.: `_, err := pod.CreatePodFromFileIfNotExist(`...

And then we can use a pattern like this to ensure that all pod resources with a given prefix ("sgx-device-plugin" in this case), are in a Running state:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L1038

salsal97: Gotcha, I've made the required changes - thank you for your guidance

jackfrancis:
Thanks so much @salsal97!

PR #3570, opened 2020-07-01
fix: kube-proxy upgrade from v1.15 to v1.16

### LINKED ISSUE #3557: kube-proxy daemonset not reconciled after upgrading to v1.16 or greater
**Describe the bug**

Addon-manager fails to reconcile the kube-proxy manifest after an upgrade from v1.15 (or lower) to v1.16 (or greater).

The reason is that the new kube-proxy manifest contains updates to immutable fields (`selector.matchLabels`).

**Steps To Reproduce**

Deploy a v1.15 cluster, upgrade to v1.16, check kube-proxy daemonset still references the v1.15 image 

**Additional context**

Addon-manager error message:
 
```
Error from server (Invalid): error when applying patch:
Resource: "apps/v1, Resource=daemonsets", GroupVersionKind: "apps/v1, Kind=DaemonSet"
Name: "kube-proxy", Namespace: "kube-system"
for: "/etc/kubernetes/addons/kube-proxy.yaml": DaemonSet.apps "kube-proxy" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{...}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```
Source of the issue:
https://github.com/Azure/aks-engine/blob/master/parts/k8s/addons/kube-proxy.yaml#L51-L54

Deleting the daemonset recreates the resource with the expected state.

jadarsie:
**Reason for Change**:
Added an extra step in between control-plane and agent upgrades to delete the kube-proxy daemon if the upgrade version is 1.16.

**Issue Fixed**:
Fixes #3557

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This extra step will fail silently otherwise it would break test "Should not fail if a Kubernetes client cannot be created" (upgradecluster_test.go)

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -14,6 +14,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/Azure/aks-engine/pkg/api"
+	"github.com/Azure/aks-engine/pkg/api/common"
 	"github.com/Azure/aks-engine/pkg/armhelpers"
 	"github.com/Azure/aks-engine/pkg/armhelpers/utils"
 	"github.com/Azure/aks-engine/pkg/engine"
@@ -35,6 +36,7 @@ type Upgrader struct {
 	stepTimeout        *time.Duration
 	cordonDrainTimeout *time.Duration
 	AKSEngineVersion   string
+	CurrentVersion     string
 	ControlPlaneOnly   bool
 }
 
@@ -44,6 +46,7 @@ const (
 	defaultTimeout                     = time.Minute * 20
 	defaultCordonDrainTimeout          = time.Minute * 20
 	nodePropertiesCopyTimeout          = time.Minute * 5
+	getResourceTimeout                 = time.Minute * 1
 	clusterUpgradeTimeout              = time.Minute * 180
 	vmStatusUpgraded          vmStatus = iota
 	vmStatusNotUpgraded
@@ -76,6 +79,8 @@ func (ku *Upgrader) RunUpgrade() error {
 		return err
 	}
 
+	ku.handleUnreconcilableAddons(ctx)
+
 	if ku.ControlPlaneOnly {
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -702,6 +703,16 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			}
 		})
 
+		It("should have core kube-system addons running the correct version", func() {
```
jackfrancis, line 706: This is nice, we should probably guard this in a `if eng.ExpandedDefinition.Properties.OrchestratorProfile.KubernetesConfig.CustomKubeProxyImage == "" {` block

jadarsie: good call

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -88,6 +93,33 @@ func (ku *Upgrader) RunUpgrade() error {
...
 }
 
+// handleUnreconcilableAddons ensures addon upgrades that addon-manager cannot handle by itself.
+// This method fails silently otherwide it would break test "Should not fail if a Kubernetes client cannot be created" (upgradecluster_test.go)
+func (ku *Upgrader) handleUnreconcilableAddons() {
+	// kube-proxy upgrade fails from v1.15 to 1.16: https://github.com/Azure/aks-engine/issues/3557
+	// deleting daemonset so addon-manager recreates instead of patching
+	upgradeVersion := ku.DataModel.Properties.OrchestratorProfile.OrchestratorVersion
+	if !common.IsKubernetesVersionGe(ku.CurrentVersion, "1.16.0") && common.IsKubernetesVersionGe(upgradeVersion, "1.16.0") {
+		ku.logger.Errorf("Attempting to delete kube-proxy daemonset.")
```
jadarsie, line 103: Will change this from Error to Warning

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -88,6 +93,33 @@ func (ku *Upgrader) RunUpgrade() error {
...
+	// deleting daemonset so addon-manager recreates instead of patching
+	upgradeVersion := ku.DataModel.Properties.OrchestratorProfile.OrchestratorVersion
+	if !common.IsKubernetesVersionGe(ku.CurrentVersion, "1.16.0") && common.IsKubernetesVersionGe(upgradeVersion, "1.16.0") {
+		ku.logger.Infof("Attempting to delete kube-proxy daemonset.")
+		client, err := ku.getKubernetesClient(getResourceTimeout)
+		if err != nil {
+			ku.logger.Errorf("Error getting Kubernetes client: %v", err)
+			return
+		}
+		ds, err := client.GetDaemonSet("kube-system", common.KubeProxyAddonName)
```
mboersma, line 109: Do we need to bother with `GetDaemonSet`, or can we just call `DeleteDaemonSet` and handle the possible error if it didn't exist?

jadarsie: can update

PR #3556, opened 2020-06-28
feat: reconcile private hosts for AKS BYO DNS

levimm:
**Reason for Change**:

This pr is required by AKS private cluster BYO DNS scenario. Since this pr also changes the api model, the pr need to be merged into master.
For AKS BYO DNS, we'll config the /etc/hosts file on all agent nodes so we can resolve the private dns without azure private dns zone.

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3441, opened 2020-06-10
feat: Consume signed powershell scripts

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [x] includes documentation
- [x] adds unit tests
- [x] tested upgrade from previous version

**Notes**:

PR #3572, opened 2020-07-02
feat: add hosts config agent support for Windows nodes

feiskyer:
**Reason for Change**:

Follow up of #3556: add hosts config agent support for Windows nodes

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


feiskyer:
the first commit is coming from #3556, needs a rebase after it gets merged.

parts/k8s/hostsconfigagent.ps1
```diff
@@ -0,0 +1,59 @@
+$Global:ClusterConfiguration = ConvertFrom-Json ((Get-Content "c:\k\kubeclusterconfig.json" -ErrorAction Stop) | out-string)
+$clusterFQDN = $Global:ClusterConfiguration.Kubernetes.ControlPlane.IpAddress
+$hostsFile="C:\Windows\System32\drivers\etc\hosts"
+$retryDelaySeconds = 15
+
+# TODO: test only, remove it after test
+$clusterFQDN = "$clusterFQDN.local"
```
jsturtevant, line 7: looks like this needs to be removed

parts/k8s/hostsconfigagent.ps1
```diff
@@ -0,0 +1,59 @@
...
+        if ($values[0] -eq "aksAPIServerIPAddress")
+        {
+            return $values[1]
+        }
+    }
+
+    return ""
+}
+
+while ($true)
```
jsturtevant, line 38: this looks like an indefinite loop.  Is that the intent?

jsturtevant: I see below this file is installed as a service to continually poll

parts/k8s/hostsconfigagent.ps1
```diff
@@ -0,0 +1,59 @@
...
+{
+    $clusterIP = Get-APIServer-IPAddress
+    Write-Log "Get current APIServer IP address: $clusterIP"
+
+    if ($clusterIP -eq "") {
+        Write-Log "Doesn't find clusterIP from aksAPIServerIPAddress tag, skipping"
```
jsturtevant, line 44: We should not write to disk every loop. In the past we have filled up the log file and it caused unnecessary disk consumption

parts/k8s/hostsconfigagent.ps1
```diff
@@ -0,0 +1,59 @@
...
+        Start-Sleep $retryDelaySeconds
+        continue
+    }
+
+    $hostsContent=Get-Content -Path $hostsFile -Encoding UTF8
+    if ($hostsContent -match "$clusterIP $clusterFQDN") {
+        Write-Log "$clusterFQDN has already been set to $clusterIP"
```
jsturtevant, line 51: Same here

jsturtevant:
Look like there was at least one debugging TODO left. Also want to minimize the noise in the logs so we don't fill them up since the call runs regularly.

A rebase will be need for recent changes to the powershell.  We moved the powershell scripts due to signing requirements: https://github.com/Azure/aks-engine/pull/3441.  There are instructions in the PR that demonstrate where the PS should live, how to test and how to make updates but the functionality itself should stay similar.  Let me know if you have questions.

parts/k8s/hostsconfigagent.ps1
feiskyer: exactly

parts/k8s/hostsconfigagent.ps1
feiskyer: good catch, need to remove this line

parts/k8s/hostsconfigagent.ps1
feiskyer: simplified logs and it would generate at most 100m one year, so it should be ok

feiskyer:
@jsturtevant could you help to have a look again?

staging/provisioning/windows/hostsconfigagent.ps1
```diff
@@ -0,0 +1,54 @@
+$Global:ClusterConfiguration = ConvertFrom-Json ((Get-Content "c:\k\kubeclusterconfig.json" -ErrorAction Stop) | out-string)
+$clusterFQDN = $Global:ClusterConfiguration.Kubernetes.ControlPlane.IpAddress
+$hostsFile="C:\Windows\System32\drivers\etc\hosts"
+$retryDelaySeconds = 15
+
+# Dot-source scripts with functions that are called in this script
+. c:\AzureData\k8s\kuberneteswindowsfunctions.ps1
```
jsturtevant, line 7: I am not sure if since these files are not signed we might run into issues.  @ksubrmnn do you know if we can use the non signed powershell here?

staging/provisioning/windows/hostsconfigagent.ps1
```diff
@@ -0,0 +1,54 @@
...
+
+function Get-APIServer-IPAddress
+{
+    $uri = "http://169.254.169.254/metadata/instance/compute/tags?api-version=2019-03-11&format=text"
+    $response = Retry-Command -Command "Invoke-RestMethod" -Args @{Uri=$uri; Method="Get"; ContentType="application/json"; Headers=@{"Metadata"="true"}} -Retries 3 -RetryDelaySeconds 5
```
jsturtevant, line 12: How often does the ip address change? can this be cached so there are not continuous calls?

staging/provisioning/windows/hostsconfigagent.ps1
```diff
@@ -0,0 +1,54 @@
...
+        Start-Sleep $retryDelaySeconds
+        continue
+    }
+
+    $hostsContent=Get-Content -Path $hostsFile -Encoding UTF8
+    if ($hostsContent -match "$clusterIP $clusterFQDN") {
+        Write-Log "APIServer FQDN has already been set to $clusterIP in hosts file"
+    } else {
+        $hostsContent -notmatch "$clusterFQDN" | Out-File $hostsFile -Encoding UTF8
+        Add-Content -Path $hostsFile -Value "$clusterIP $clusterFQDN" -Encoding UTF8
```
jsturtevant, line 49: This appends to the host file. If the value changes do you need to clean up the old one?

feiskyer: it is clean up in the above line

staging/provisioning/windows/hostsconfigagent.ps1
feiskyer: it depends. But when it changed, the script would catch it as soon as possible. So I think depending on intervals should be better. Do you insist on the caching, or do you think we should increase the interval?

staging/provisioning/windows/hostsconfigagent.ps1
feiskyer: if so, should I move it to `staging/provisioning/windows`?

staging/provisioning/windows/hostsconfigagent.ps1
jsturtevant: I don't have enough context. It looks like the [linux solution](https://github.com/Azure/aks-engine/pull/3556/files) is doing the same so I say we keep similar

staging/provisioning/windows/hostsconfigagent.ps1
jsturtevant: I don't have enough context. It looks like the [linux solution](https://github.com/Azure/aks-engine/pull/3556/files) is doing the same so I say we keep similar

staging/provisioning/windows/hostsconfigagent.ps1
ksubrmnn: I think it depends on when the code is run. During provisioning vs after provisioning. We've only been signing code that runs after provisioning.

feiskyer: removed this by copying the required functions.

feiskyer:
@jsturtevant could you help to have a look again?

PR #3574, opened 2020-07-03
fix: always extract kubelet and kubectl from a custom URL if defined

chewong:
**Reason for Change**:

We should extract kubelet and kubectl from a custom URL (`customHyperkubeImage` / `customKubeBinaryURL`) when it's defined. Also, we should allow the use of VHD when `customHyperkubeImage` is define to prevent a reboot problem in some of our Windows testgrid jobs.

**Issue Fixed**:

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


jsturtevant:
lgtm but I don't have a ton of context on Linux side so would refer to @jackfrancis or @mboersma for approval

We should update the docs at https://github.com/Azure/aks-engine/blob/master/docs/topics/kubernetes-developers.md#aks-engine-api-model

jackfrancis:
It would be nice to have a cluster config that uses `customHyperkubeImage` under `test/e2e/test_cluster_configs`. Do we have a "canonical" 1.16-vintage custom hyperkube we can test?

test/e2e/test_cluster_configs/custom_hyperkube.json
```diff
@@ -0,0 +1,46 @@
...
+    "options": {
+        "allowedOrchestratorVersions": [
+            "1.16"
+        ]
+    },
+    "apiModel": {
+        "apiVersion": "vlabs",
+        "properties": {
+            "orchestratorProfile": {
+                "orchestratorType": "Kubernetes",
```
jackfrancis, line 11: I wonder if the right pattern is to include `"orchestratorVersion": "1.16.0"` here to ensure that any foo that operates on < or > per patch version does the right thing?

test/e2e/kubernetes/scripts/custom-hyperkube-validate.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -x
+
+kubelet_version=$(docker run ${CUSTOM_HYPERKUBE_IMAGE} kubelet --version)
```
jackfrancis, line 5: ❤️

test/e2e/test_cluster_configs/custom_hyperkube.json
chewong: SGTM

chewong: The best I can do is setting "orchestratorRelease": "1.16" due to the following error

```
02:44:19  2020/07/09 21:44:19 Output:Error: loading API model in generateCmd: error parsing the api model: the following OrchestratorProfile configuration is not supported: OrchestratorType: "Kubernetes", OrchestratorRelease: "", OrchestratorVersion: "1.16.0". Please use one of the following versions: [1.6.9 1.15.11 1.15.12 1.16.11 1.16.12 1.17.7 1.17.8 1.18.4 1.18.5 1.19.0-alpha.1 1.19.0-alpha.2 1.19.0-alpha.3 1.19.0-beta.0 1.19.0-beta.1 1.19.0-beta.2]

```

jackfrancis: Yep, I think you're right!

jackfrancis: So let's roll that back and then lgtm, thanks so much!

test/e2e/test_cluster_configs/custom_hyperkube.json
```diff
@@ -0,0 +1,47 @@
...
+                "orchestratorRelease": "1.16",
```
jackfrancis, line 12: nit: this is actually superfluous because the Jenkinsfile will interpret the `allowedOrchestratorVersions` array before generating api models, and inject the `orchestratorRelease` value that correlates with the value in the array.

PR #3583, opened 2020-07-07
feat: etcdStorageLimitGB for configurable etcd storage limit

### LINKED ISSUE #3581: aks-engine support for setting larger kubernetes etcd maximum size
**Describe the request**
The etcd that is set up for kubernetes by aks-engine is running a default maximum size of 2GB.  We need a way to change this limit.

The 2GB limit is usually fine for clusters that are smaller or not very dynamic but we have run into a case where so much change happened within a short period of time that even with the pruning of history by kubernetes every 5 minutes (default) we overflowed etcd and it ended up completely failing (and thus causing the cluster to fail).

Either to have the default be 8GB (which should be significant overkill) or at least the min(8GB, etcd_disk_size/2)  (16GB disk gives you an 8GB etcd - larger disks are just for higher performance and not more space - we already run with the largest premium SSD storage available for best I/O performance)  (Do not have etcd limit the same as the disk size as there are overheads that need to be taken into account for etcd control structures and the filesystem)

If changing the default to 8GB is not reasonable to change, being able to express this value in the config to aks-engine would be the next best thing.  

NOTE:  Current etcd documentation suggests not setting this higher than 8GB which is why I have been suggesting that value:  https://etcd.io/docs/v3.3.13/dev-guide/limit/#storage-size-limit 

There are potentially other advanced etcd configuration values that may be useful for advanced/high scale clusters to control.  So making a way to provide options to etcd would allow for setting the size (which you currently don't do) and possibly other values.

**Additional context**
The speech services clusters are very dynamic and relatively large and as such stress kubernetes relatively hard.  We have not validated that larger etcd storage limits would be safe even for smaller clusters but I would not expect it to be a problem.  The reason is that even at the default 2GB limit, most clusters use a very small percentage of that and setting the value lower did not help them.  However, at such small sizes, there is no where near the stress on the clusters so I can not say that blanket change to 8GB maximum would not have a negative impact.  (My gut feeling is that it would not be a problem)

jackfrancis:
**Reason for Change**:

This PR adds a configurable `etcdStorageLimitGB` property to the `kubernetesConfig` to enable configurable etcd storage limits. We default to 2, which is what the etcd runtime defaults to if we aren't passing this configuration in. i.e., there is no functional change for the default scenario, but this enables folks to use up to 8 GB of storage for the etcd data store.

**Issue Fixed**:

Fixes #3581 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @CecileRobertMichon does this sound like something we want to track as a configuration/feature gap for capz?

pkg/api/vlabs/validate.go
```diff
@@ -371,6 +371,15 @@ func (a *Properties) ValidateOrchestratorProfile(isUpdate bool) error {
 						}
 					}
 				}
+
+				if o.KubernetesConfig.EtcdStorageLimitGB != 0 {
+					if o.KubernetesConfig.EtcdStorageLimitGB > 8 {
```
Michael-Sinz, line 376: While I somewhat agree here, I would not put a limit.  etcd recommends not going about 8GB but if someone really needs to play with this value, they should at least look at the limit etcd has.

You could output a warning about that (if you wish) but I would not make it a failure and thus block setting it to, say, 10GB.

NOTE:  I would hope that we never need to go above, say, 4GB as that is rather huge already but the reality is that we don't know where or when this limit may be hit under dynamic cluster conditions.

Michael-Sinz: PS - I would not block this PR due to this as it is potentially important to us to get a larger (say 4GB) etcd to test our busy cluster environment.  Just thought it would be nice to not artificially limit the value at this level.

jackfrancis: Understood, so we want to wrap this behavior:

> 8GB is a suggested maximum size for normal environments and etcd warns at startup if the configured value exceeds it.

Sounds like there are known de-optimizations to the operationalization of the key/val store above 8GB, but there isn't really a hard limit?

jackfrancis: Done

PR #3594, opened 2020-07-14
feat: Support configurable Windows PauseImage

### LINKED ISSUE #3586: Support to configure endpoint for Windows MCR images.
**Describe the request**
Support to configure endpoint for Windows MCR images.

Windows: pkg\engine\templates_generated.go
```
function Write-KubeClusterConfig {
    param(		
        $MasterIP,
        $KubeDnsServiceIp
    )

    $Global:ClusterConfiguration = [PSCustomObject]@{ }

    $Global:ClusterConfiguration | Add-Member -MemberType NoteProperty -Name Cri -Value @{
        Name   = $global:ContainerRuntime;
        Images = @{
            "Pause" = "mcr.microsoft.com/oss/kubernetes/pause:1.4.0"
        }
    }
```
Need for custom cloud environment.

Can use a variable to specify.

**Additional context**

AbelHu:
**Reason for Change**:

Support configurable Windows PauseImage:
1. Allow to configure endpoint of Windows PauseImage
2. Allow to always pull Windows PauseImage in provisioning.

**Issue Fixed**:

Fixes #3586

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


AbelHu:
The coverage drop in cmd/scale.go is caused by @jackfrancis's [PR](https://github.com/Azure/aks-engine/pull/3590)

devigned:
I agree with you, @AbelHu. It looks like the coverage report is including Jack's earlier commit. Perhaps, squash your two commits and see if that makes a difference.

The PR looks good to me and if squashing doesn't make a difference, I vote to ignore coverage.

@jsturtevant, what do you think?

docs/topics/clusterdefinitions.md
```diff
@@ -910,6 +910,8 @@ https://{keyvaultname}.vault.azure.net:443/secrets/{secretName}/{version}
 | enableAutomaticUpdates        | no       | If set to `true` Windows Update will be configured to automatically apply updates on Windows nodes. Default: `false`                                                                                                                                            |
 | enableCSIProxy                | no       | If set to `true` the csi-proxy specified by `windowsProfile.csiProxyURL` will get installed during node provisioning. See [Windows Csi Proxy](csi-proxy-windows.md) for more details.                                                                           |
 | provisioningScriptsPackageURL | no       | Path to a package containing provisioning scripts found in [/staging/provisioning/windows](/staging/provisioning/windows). Default values will be maintained by aks-engine team and contain script content signed by Microsoft.                                 |
+| windowsPauseImageURL          | no       | Path to a Windows pause image. Default values will be maintained by aks-engine team.                                 |
```
jsturtevant, line 913: can we add an example to the docs:  # e.g. "mcr.microsoft.com/oss/kubernetes/pause:1.4.0"

```suggestion
| windowsPauseImageURL          | no       | Url to a Windows pause image. Default values will be maintained by aks-engine team.  e.g. "mcr.microsoft.com/oss/kubernetes/pause:1.4.0"                                |
```

docs/topics/clusterdefinitions.md
```diff
@@ -910,6 +910,8 @@ https://{keyvaultname}.vault.azure.net:443/secrets/{secretName}/{version}
...
+| alwaysPullWindowsPauseImage   | no       | If set to `true`, it will always pull the Windows pause image from MCR repo. Default: `false`                                   |
```
jsturtevant, line 914: it will always pull but not nessarilly from the MCR repo.  It would always pull from the repo specified.

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -204,14 +204,14 @@ New-InfraContainer {
 
     if ($pauseImageVersions -icontains $windowsVersion) {
```
jsturtevant, line 205: I think this whole section would be skipped and the custom pause image wouldn't be pulled.  The `$global:AlwaysPullWindowsPauseImage` needs be added up here for the image to be pulled at provisioning time

AbelHu: I think that it will not be skipped since we only added PauseImage for those supported kernel versions.

AbelHu: I tested this PR with aks-engine and confirmed that the PauseImage will be pulled when `$global:AlwaysPullWindowsPauseImage` is true.

jsturtevant: Yes I had it backwards.  Though I think in the case of a insider base image where you provide a custom pause image, it would not get pulled which would lead to some confusion I think.

PR #3599, opened 2020-07-16
fix: updating hotfix packages used for windows inlcuding CVE and metric

AbelHu:
fixes

**Reason for Change**:

updating hotfix packages used for windows inlcuding CVE and metric

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jsturtevant:
code coverage is down a bit but I am not sure there is much we can do on this?  @mboersma any thoughts?

mboersma:
@jsturtevant codecov is complaining about `k8s_versions.go` which is actually tested. I don't think there's anything reasonable we can do here, so let's ignore it and merge anyway. (Unfortunately, code coverage here is a blunt instrument that can be buggy, so we have to ignore its signal sometimes.)

jsturtevant:
thanks!

/approve

PR #3602, opened 2020-07-16
feat: add support for Kubernetes 1.16.13

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.16.md#changelog-since-v11612

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
/hold

This has a temporary commit for testing which needs to be discarded after it passes CI.

pkg/api/vlabs/validate_test.go
```diff
@@ -5272,7 +5272,7 @@ func TestValidateAzureStackSupport(t *testing.T) {
 			cs.Properties.OrchestratorProfile.OrchestratorVersion = "1.16.11"
 			if err := cs.Validate(false); !helpers.EqualError(err, test.expectedErr) {
 				t.Logf("scenario %q", test.name)
-				t.Errorf("expected error: %v, got: %v", test.expectedErr, err)
+				t.Logf("FIXME: expected error: %v, got: %v", test.expectedErr, err)
```
mboersma, line 5275: @jadarsie if I follow our usual process, this leaves Azure Stack without a supported 1.16.x version, so I disabled this test. But let me know if you prefer a different approach--we can be flexible about version support if need be.

mboersma:
This [passed CI](https://github.com/Azure/aks-engine/pull/3602/checks?check_run_id=879503582), so I'll strip off the testing commit and it should be ready for review.

pkg/api/vlabs/validate_test.go
jadarsie: ASH support for v1.16 here #3613, usual process OK

jackfrancis: @jadarsie should I merge this as-is, and then #3613 can rebase and undo the error log changes?

jadarsie: @JesusAlvarezTorres FYI

JesusAlvarezTorres: @jackfrancis go ahead and merge this in, I'll update #3613 when this is in

PR #3607, opened 2020-07-16
[wip] feat: arc agents integration

jadarsie:
**Reason for Change**:
Proposal to add `connectedClusterProfile` to integrate arc onboarding with cluster deployment.

**Requirements**:
- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


docs/design/arc-onboarding.md
```diff
@@ -0,0 +1,27 @@
+# Azure Arc for Kubernetes integration
+
+## Motivation
+
+Simplify Arc agents deploymet by making it an optional last step of the cluster deployment process.
```
jackfrancis, line 5: s/deploymet/deployment

jackfrancis: Also, should be `agents' deployment` if we're talking about multiple agents, or `agent's deployment` if we're talking about a single agent

docs/design/arc-onboarding.md
```diff
@@ -0,0 +1,27 @@
...
+      "resourceGroup": "connected-cluster-rg-azure",
+      "clusterName": "friendly-name",
+      "clientID": "azure-spn",
+      "clientSecret": "azure-spn-secret",
+      "region": "azure-region"
+    }
+}
+```
+
+- AKSe replaces the placeholders in [arc-onboarding.yml](https://github.com/Azure/azure-arc-kubernetes-onboarding/blob/master/arc-onboarding.yml), base64 encodes the yaml and save the onboarding manifest in the control plane nodes.
```
jackfrancis, line 25: Let's always refer to AKS Engine as "AKS Engine"

jadarsie: This file will go away now that I am actually working on the feature

pkg/api/defaults.go
```diff
@@ -917,6 +921,25 @@ func (p *Properties) setTelemetryProfileDefaults() {
...
+	if p.ConnectedClusterProfile.Location == "" {
+		p.ConnectedClusterProfile.Location = cs.Location
+	}
+	if p.ConnectedClusterProfile.ClientID == "" {
+		p.ConnectedClusterProfile.ClientID = p.ServicePrincipalProfile.ClientID
+	}
+	if p.ConnectedClusterProfile.ClientSecret == "" {
+		p.ConnectedClusterProfile.ClientSecret = p.ServicePrincipalProfile.Secret
+	}
+	// p.ConnectedClusterProfile.ClusterName = default to resource group name?
```
jadarsie, line 938: @jackfrancis is there a way for `defaults.go` to access `resource-group` and `subscriptionID` other than passing them to method `SetPropertiesDefaults`? would it be acceptable to add these two fields to `PropertiesDefaultsParams`?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -245,7 +245,7 @@ if [[ -n ${MASTER_NODE} ]]; then
   time_metric "EnsureK8sControlPlane" ensureK8sControlPlane
   {{- if HasClusterInitComponent}}
   if [[ $NODE_INDEX == 0 ]]; then
-    retrycmd 120 5 30 $KUBECTL apply -f /opt/azure/containers/cluster-init.yaml || exit {{GetCSEErrorCode "ERR_CLUSTER_INIT_FAIL"}}
+    retrycmd 120 5 30 $KUBECTL apply -f /opt/azure/containers/init || exit {{GetCSEErrorCode "ERR_CLUSTER_INIT_FAIL"}}
```
jadarsie, line 248: should this be `apply -f /opt/azure/containers/cluster-init.yaml -f /opt/azure/containers/init` instead?

parts/k8s/cloud-init/arc-onboarding.yaml
```diff
@@ -0,0 +1,92 @@
+---
```
jackfrancis, line 1: So, as I read through this, I think this is probably O.K. to implement as an addon. I don't think there are any mutations to any of the resources once the onboarding finishes (it's a one-time process).

The only thing to consider is: what would happen if a user manually deletes the job object? According to the official docs, Job resources (and their resultant Pod resources) will stick around forever:

https://kubernetes.io/docs/concepts/workloads/controllers/job/#job-termination-and-cleanup

Is it O.K. if a user deletes the Job object, and then the kube-addon-manager notices and re-runs it? Is the azure-arc-onboarding operation idempotent?

jadarsie: Just checked, at this point the onboarding job is idempotent.
I can move to the addon approach.

PR #3627, opened 2020-07-21
test: run pod schedule tests first

jackfrancis:
**Reason for Change**:

In response to a test flake that resulted from a Windows node not being able to schedule pods (but the test results not showing that explicitly), this PR adds a discrete "should be able to schedule a pod to a Windows node" test, and does all pod scheduling tests before running other tests that in fact depend upon pod scheduling.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @jsturtevant

PR #3647, opened 2020-07-29
chore: Only use Windows telemetry client when `enableTelemetry` is true

AbelHu:
**Reason for Change**:

Only handle Windows telemetry when enableTelemetry is true

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/windowsazurecnifunc.ps1
+++ b/parts/k8s/windowsazurecnifunc.ps1
@@ -76,10 +76,14 @@ Set-AzureCNIConfig
     # Fill in DNS information for kubernetes.
     if ($IsDualStackEnabled){
         $subnetToPass = $KubeClusterCIDR -split ","
-        $exceptionAddresses = @($subnetToPass[0], $MasterSubnet, $VNetCIDR)
+        $exceptionAddresses = @($subnetToPass[0], $MasterSubnet)
     }
     else {
-        $exceptionAddresses = @($KubeClusterCIDR, $MasterSubnet, $VNetCIDR)
+        $exceptionAddresses = @($KubeClusterCIDR, $MasterSubnet)
+    }
+    $vnetCIDRs = $VNetCIDR -split ","
+    foreach ($cidr in $vnetCIDRs) {
+        $exceptionAddresses += $cidr
     }
 
     $fileName  = [Io.path]::Combine("$AzureCNIConfDir", "10-azure.conflist")
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -24479,10 +24479,14 @@ Set-AzureCNIConfig
     # Fill in DNS information for kubernetes.
     if ($IsDualStackEnabled){
         $subnetToPass = $KubeClusterCIDR -split ","
-        $exceptionAddresses = @($subnetToPass[0], $MasterSubnet, $VNetCIDR)
+        $exceptionAddresses = @($subnetToPass[0], $MasterSubnet)
     }
     else {
-        $exceptionAddresses = @($KubeClusterCIDR, $MasterSubnet, $VNetCIDR)
+        $exceptionAddresses = @($KubeClusterCIDR, $MasterSubnet)
+    }
+    $vnetCIDRs = $VNetCIDR -split ","
+    foreach ($cidr in $vnetCIDRs) {
+        $exceptionAddresses += $cidr
     }
 
     $fileName  = [Io.path]::Combine("$AzureCNIConfDir", "10-azure.conflist")


AbelHu:
The failed case is not related to this change. Please help to confirm it.
```
2020/07/29 14:34:20 Command Output: Deployment failed. Correlation ID: 611c249a-8e4c-4f4d-add9-49e5b540a7d1. {
  "status": "Failed",
  "error": {
    "code": "ResourceDeploymentFailure",
    "message": "The resource operation completed with terminal provisioning state 'Failed'.",
    "details": [
      {
        "code": "DiskServiceInternalError",
        "message": "An unexpected error occurred while processing disk. Please retry later.  Target: '/subscriptions/aa3d3369-e814-4495-899d-d31e8d7d09ce/resourceGroups/KUBERNETES-SOUTHEASTASIA-76715/providers/Microsoft.Compute/disks/4214k8s06_4214k8s06_0_OsDisk_1_792a998a89414e2c9a9854fa32df9475'."
      }
    ]
  }
}

```

jackfrancis:
This lgtm, @jsturtevant to review as well

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -182,67 +182,77 @@ try
...
-
-        $configAppInsightsClientTimer = [System.Diagnostics.Stopwatch]::StartNew()
-        # Get app insights binaries and set up app insights client
-        mkdir c:\k\appinsights
-        DownloadFileOverHttp -Url "https://globalcdn.nuget.org/packages/microsoft.applicationinsights.2.11.0.nupkg" -DestinationPath "c:\k\appinsights\microsoft.applicationinsights.2.11.0.zip"
-        Expand-Archive -Path "c:\k\appinsights\microsoft.applicationinsights.2.11.0.zip" -DestinationPath "c:\k\appinsights"
-        $appInsightsDll = "c:\k\appinsights\lib\net46\Microsoft.ApplicationInsights.dll"
-        [Reflection.Assembly]::LoadFile($appInsightsDll)
-        $conf = New-Object "Microsoft.ApplicationInsights.Extensibility.TelemetryConfiguration"
-        $conf.DisableTelemetry = -not $global:enableTelemetry
```
jsturtevant, line 195: I believe this line disables the sending of telemetry.  https://docs.microsoft.com/en-us/dotnet/api/microsoft.applicationinsights.extensibility.telemetryconfiguration.disabletelemetry?view=azure-dotnet

If that is the case it makes the resulting code much simpler to reason about as the telemetry is not sent even if the calls are made.  What is the idea for removing the use all together?

AbelHu: In some custom clouds, it cannot download the application insight client so it will fail even if it is disabled.

jsturtevant: We should add the application client to the VHD cache then.  The `DownloadFileOverHttp` checks the cache first https://github.com/Azure/aks-engine/blob/a1870f9a3ec6668515065df1bd41c6d37bff132b/parts/k8s/kuberneteswindowsfunctions.ps1#L27-L30

jsturtevant: We have a no outbound tests that we run before releases: https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/no_outbound.json#L32 

It might be because telemetry isn't enabled it doesn't catch it.

jsturtevant: It is in the vhd cache: https://github.com/Azure/aks-engine/blob/f321785bd81b178dcff02736bf3c4799eeaf1c54/vhd/packer/configure-windows-vhd.ps1#L82  

Maybe there is something else causing the outbound traffic?

AbelHu: 1. Skipping installation and configuration can reduce the provisioning time when telemetry is disabled.
2. Later if we remove hardcoded versions from this script, it also can avoid that mismatched version causes unexpected egress.

jsturtevant: 1.  Do you see a huge provision time improvement? I wouldn't think that the configuration has a significant time difference but I might be wrong.

My biggest trouble with this is the added if statements make the code a bit more difficult to reason and brittle, and if it is disabled we are not seeing this as the egress issue unless I am missing something.  We run [no-out-bound](https://github.com/Azure/aks-engine/blob/master/test/e2e/test_cluster_configs/no_outbound.json) tests as part of the release process which has telemetry disabled and we don't see outbound traffic.

As a way forward it be better if all this logic was in a separate module that hides the if statements?

AbelHu: I think that the issue is 
```$global:EnableTelemetry = "{{WrapAsVariable "enableTelemetry" }}";```
=>
```$global:EnableTelemetry = [System.Convert]::ToBoolean("{{WrapAsVariable "enableTelemetry" }}");```

marosset: I did test that setting $conf.DisableTelemetry will prevent sending telemetry so I think @AbelHu is correct that the issue is the system.convert missing.

If that fixes the issue can we remove all of the added if statements from this PR?

craiglpeters:
@AbelHu Is this still active?

PR #3654, opened 2020-07-30
fix: only remove CRI if re-installing

jackfrancis:
**Reason for Change**:

This PR changes the "install moby|containerd" CSE implementation so that we only do the "remove all other CRI" operations if we actually need to install either moby or containerd. In the event that we detect that the desired version is already installed, we can return immediately.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -296,7 +296,6 @@ configureAzureCNI() {
 }
 {{- if NeedsContainerd}}
 installContainerd() {
-  removeMoby
   CURRENT_VERSION=$(containerd -version | cut -d " " -f 3 | sed 's|v||')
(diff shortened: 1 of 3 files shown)


jackfrancis:
cc @cpuguy83

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -678,10 +678,4 @@ cleanUpContainerd() {
 removeEtcd() {
   rm -rf /usr/bin/etcd
 }
-removeMoby() {
```
jackfrancis, line 681: Moving these funcs to cse_install because that's the only shell script file that VHD CI sources, so we always have these (in fact innocuous) errors during VHD builds:

`/home/packer/provision_installs.sh: line 93: removeMoby: command not found`

`/home/packer/provision_installs.sh: line 90: removeContainerd: command not found`

PR #3658, opened 2020-08-03
feat: cherry-pick #2960 into AKS official branch

norshtein:
**Reason for Change**:

This PR cherry-picks #2960 into AKS official branch. The commit enables BYO identity for kubelet, so after this commit cherry-picked, kubelet's identity can be created by AKS RP via Go client and then provide this identity to AKS-Engine. AKS RP don't need to rely on AKS-Engine generated ARM template to create the identity. 

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


norshtein:
@jackfrancis could you please help merge the PR? @yangl900 @xuto2 FYI.

pkg/api/types.go
```diff
@@ -1970,25 +1970,8 @@ func (k *KubernetesConfig) SystemAssignedIDEnabled() bool {
...
-}
-
-// GetUserAssignedClientID returns the user assigned client ID if it is enabled.
-func (k *KubernetesConfig) GetUserAssignedClientID() string {
-	if k.UserAssignedClientIDEnabled() {
-		return k.UserAssignedClientID
-	}
-	return ""
+func (k *KubernetesConfig) ShouldCreateNewUserAssignedIdentity() bool {
+	return !(k.UserAssignedIDEnabled() && strings.Contains(k.UserAssignedID, "/"))
```
xuto2, line 1974: why do we need this strings.Contains(k.UserAssignedID, "/") check here but not inside UserAssignedIDEnabled()?

pkg/engine/armresources.go
```diff
@@ -27,24 +27,30 @@ func GenerateARMResources(cs *api.ContainerService) []interface{} {
...
 		useManagedIdentity = kubernetesConfig.UseManagedIdentity
-		userAssignedIDEnabled = useManagedIdentity && kubernetesConfig.UserAssignedID != ""
+		userAssignedIDEnabled = kubernetesConfig.UserAssignedIDEnabled()
+		createNewUserAssignedIdentity = kubernetesConfig.ShouldCreateNewUserAssignedIdentity()
 	}
 
 	isHostedMaster := cs.Properties.IsHostedMasterProfile()
 	if userAssignedIDEnabled {
-		userAssignedID := createUserAssignedIdentities()
+		if createNewUserAssignedIdentity {
```
xuto2, line 41: I'm curious, in which case do we need to create a new identity when there is already one asssigned by user?

pkg/api/types.go
xuto2: ok. so this means it already exists in arm

pkg/engine/armvariables.go
```diff
@@ -67,15 +67,22 @@ func getK8sMasterVars(cs *api.ContainerService) (map[string]interface{}, error)
...
 		enableEncryptionWithExternalKms = to.Bool(kubernetesConfig.EnableEncryptionWithExternalKms)
 		useInstanceMetadata = kubernetesConfig.UseInstanceMetadata
 		excludeMasterFromStandardLB = to.Bool(kubernetesConfig.ExcludeMasterFromStandardLB)
 		maxLoadBalancerCount = kubernetesConfig.MaximumLoadBalancerRuleCount
 		provisionJumpbox = kubernetesConfig.PrivateJumpboxProvision()
+
+		if kubernetesConfig.ShouldCreateNewUserAssignedIdentity() {
+			userAssignedIDReference = "[resourceId('Microsoft.ManagedIdentity/userAssignedIdentities/', variables('userAssignedID'))]"
+		} else {
+			userAssignedIDReference = "[variables('userAssignedID')]"
```
xuto2, line 84: The userAssignedID here is a full ARM resource ID, but not something that starts with "Microsoft.ManagedIdentity/userAssignedIdentities", I assume arm can handle both cases?

PR #3660, opened 2020-08-03
test: update kubernetes e2e to use GINKGO_FAIL_FAST parameter value

haofan-ms:
**Reason for Change**:

Update kubernetes e2e to use GINKGO_FAIL_FAST parameter value

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/cluster.sh
```diff
@@ -114,7 +114,7 @@ docker run --rm \
 -e CUSTOM_KUBE_PROXY_IMAGE="${CUSTOM_KUBE_PROXY_IMAGE}" \
 -e IS_JENKINS="${IS_JENKINS}" \
 -e SKIP_TEST="${SKIP_TESTS}" \
--e GINKGO_FAIL_FAST=true \
+-e GINKGO_FAIL_FAST="${GINKGO_FAIL_FAST}" \
```
jackfrancis, line 117: The problem with using any other value than true is that the "focus" and "skip" options stop working. So if you want to use fail fase = false, beware that you won't be able to use focus/skip.

Also, if we make this change, let's make sure to set the default to true at the top of the script, e.g.:

`GINKGO_FAIL_FAST ="${GINKGO_FAIL_FAST:-true}"`

haofan-ms: Thanks Jack, I am a little confused here. I see on the GINKGO doc it says "When using the command line flags you can specify one or both of --focus and --skip. If both are specified the constraints will be ANDed together." But I am not sure why the GINKGO_FAIL_FAST is also affecting the other two.

jackfrancis: Feel free to test and see if my recollection is mistaken!

haofan-ms: I think you are right, because for now we are suffering the other problem you mentioned, which is GINKGO_SKIP not working (Sorry I failed to see your comments earlier). But it's just confusing me why the design of GINKGO features is like this and it's not documented.

PR #3666, opened 2020-08-04
feat: add support for Kubernetes 1.19.0-rc.4

mboersma:
**Reason for Change**:
See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#changelog-since-v1190-rc3

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
The Azure k8s artifacts are still building, so this will fail at first.

### DIFF
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -18,7 +18,7 @@ pr: none
 
 jobs:
 - job: build_vhd_windows
-  timeoutInMinutes: 120
+  timeoutInMinutes: 150
   strategy:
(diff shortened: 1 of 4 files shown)


mboersma:
Still working out a `kind`-related issue with the build pipeline, and then we will re-run CI once artifacts are published.

mboersma:
It looks like the **AKS-Engine WS2019 VHD - docker** job is taking just over two hours in this case (and 1:57 on some other PRs where it passed). So I added a commit that gives that job 2.5 hours. 🤞

PR #3685, opened 2020-08-10
chore: force delete addon manager when addon manager pods get stuck on terminating state

JesusAlvarezTorres:
**Reason for Change**:

Workaround an issue seen on 1.15 to 1.16 upgrade on Azure Stack disconnected scenario where the CSE will timeout on deleting the addon manager pods; the pods go into terminating state and are not deleted unless force delete is invoked.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -395,7 +395,7 @@ ensureAddons() {
   wait_for_file 1200 1 $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   sed -i "s|${ADDONS_DIR}/init|${ADDONS_DIR}|g" $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
   {{/* Force re-load all addons because we have changed the source location for addon specs */}}
-  retrycmd 120 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
+  retrycmd 120 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system --force --grace-period 0 || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
   {{if HasCiliumNetworkPolicy}}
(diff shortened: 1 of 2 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -395,7 +395,7 @@ ensureAddons() {
   wait_for_file 1200 1 $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   sed -i "s|${ADDONS_DIR}/init|${ADDONS_DIR}|g" $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
   {{/* Force re-load all addons because we have changed the source location for addon specs */}}
-  retrycmd 120 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
+  retrycmd 120 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system --force --grace-period 0 || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
```
mboersma, line 398: Should we first try a normal (polite) `kubectl delete pods` before adding `--force -grace-period 0`?

JesusAlvarezTorres: change made

jackfrancis:
@JesusAlvarezTorres did we root cause *why* the pod delete operations are stalling indefinitely?

JesusAlvarezTorres:
> @JesusAlvarezTorres did we root cause _why_ the pod delete operations are stalling indefinitely?

@jackfrancis no, we didn't. The issue seems to be intermittent, while validating this change we noticed on some clusters the graceful delete works while others had to fall back to the force delete, will need some time to see why this is happening

jackfrancis:
Also @JesusAlvarezTorres can you change the title of the PR, this change doesn't actually care about 1.15--1.16, it's enabling the --force short-circuit foo for every CSE run (create/scale/upgrade).

JesusAlvarezTorres:
> Also @JesusAlvarezTorres can you change the title of the PR, this change doesn't actually care about 1.15--1.16, it's enabling the --force short-circuit foo for every CSE run (create/scale/upgrade).

@jackfrancis Sure, updated title.

Issue #3668, opened 2020-08-05
Creating a new cluster leaves some masters in need of reboot (still)


Michael-Sinz:
**Describe the bug**
After creating a new cluster with 5 masters, 4 of the 5 masters are tagged with needing an OS reboot.

**Steps To Reproduce**
We create a v7.0.2 skyman cluster (aks-engine 0.54.0, kubernetes 1.18.6) with 5 masters and at least one VMSS node pool.

The input aks-engine looks like this:
```
 {
   "apiVersion": "vlabs",
   "location": "westus2",
   "properties": {
     "agentPoolProfiles": [
       {
         "OSDiskSizeGB": 128,
         "availabilityProfile": "VirtualMachineScaleSets",
         "count": 5,
         "ipAddressCount": 64,
         "kubernetesConfig": {
           "kubeletConfig": {
             "--image-pull-progress-deadline": "30m",
             "--max-pods": "63"
           }
         },
         "name": "agentpool1",
         "singlePlacementGroup": false,
         "storageProfile": "ManagedDisks",
         "vmSize": "Standard_DS13_v2"
       }
     ],
     "linuxProfile": {
       "adminUsername": "azureuser",
       "ssh": {
         "publicKeys": [
           {
             "keyData": "********"
           }
         ]
       }
     },
     "masterProfile": {
       "OSDiskSizeGB": 128,
       "count": 5,
       "dnsPrefix": "speech1-ppe7",
       "vmSize": "Standard_D8s_v3"
     },
     "orchestratorProfile": {
       "kubernetesConfig": {
         "ExcludeMasterFromStandardLB": true,
         "addons": [
           {
             "enabled": false,
             "name": "blobfuse-flexvolume"
           },
           {
             "enabled": false,
             "name": "scheduled-maintenance"
           },
           {
             "enabled": false,
             "name": "smb-flexvolume"
           },
           {
             "enabled": false,
             "name": "keyvault-flexvolume"
           },
           {
             "enabled": true,
             "name": "nvidia-device-plugin"
           },
           {
             "config": {
               "cores-per-replica": "320",
               "min-replicas": "2",
               "nodes-per-replica": "60"
             },
             "enabled": true,
             "name": "coredns"
           }
         ],
         "apiServerConfig": {
           "--feature-gates": "TTLAfterFinished=true,EphemeralContainers=true",
           "--profiling": "True",
[...]

PR #3721, opened 2020-08-17
fix: do not restart addon-manager if a reboot is required

CecileRobertMichon:
**Reason for Change**:

Related to https://github.com/Azure/aks-engine/issues/3668#issuecomment-674305778
Seeing some clusters fail with exit 36 recently where CSE failed to get the addon-manager pod as part of ensureAddons().

```
+ for i in $(seq 1 $retries)
+ timeout 30 /usr/local/bin/kubectl --kubeconfig=/home/azureuser/.kube/config get pods -l app=kube-addon-manager -n kube-system
+ '[' 120 -eq 120 ']'
+ echo Executed '"/usr/local/bin/kubectl' --kubeconfig=/home/azureuser/.kube/config get pods -l app=kube-addon-manager -n 'kube-system"' 120 times
Executed "/usr/local/bin/kubectl --kubeconfig=/home/azureuser/.kube/config get pods -l app=kube-addon-manager -n kube-system" 120 times
```

This PR might not fix the underlying root cause but we should not be getting/restarting pods when the vm is marked for reboot so this PR addresses that.

/cc @mboersma @Michael-Sinz

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -392,8 +392,7 @@ ensureAddons() {
   retrycmd 120 5 30 $KUBECTL get podsecuritypolicy privileged restricted || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
   rm -Rf ${ADDONS_DIR}/init
 {{- end}}
-  wait_for_file 1200 1 $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  sed -i "s|${ADDONS_DIR}/init|${ADDONS_DIR}|g" $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
+  replaceAddonsInit
   {{/* Force re-load all addons because we have changed the source location for addon specs */}}
   retrycmd 10 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system || \
   retrycmd 120 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system --force --grace-period 0 || \
@@ -419,6 +418,10 @@ ensureAddons() {
     sleep 3
   done
   {{end}}
+}
+replaceAddonsInit() {
+  wait_for_file 1200 1 $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  sed -i "s|${ADDONS_DIR}/init|${ADDONS_DIR}|g" $ADDON_MANAGER_SPEC || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
 }
 ensureLabelNodes() {
   LABEL_NODES_SCRIPT_FILE=/opt/azure/containers/label-nodes.sh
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,17 +387,21 @@ ensureKubelet() {
...
 
 ensureAddons() {
-  retrycmd 120 5 30 $KUBECTL get pods -l app=kube-addon-manager -n kube-system || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
+  if ! { [ -f /var/run/reboot-required ] || [ "$NO_OUTBOUND" = "true" ]; }; then
+    retrycmd 120 5 30 $KUBECTL get pods -l app=kube-addon-manager -n kube-system || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
 {{- if not HasCustomPodSecurityPolicy}}
-  retrycmd 120 5 30 $KUBECTL get podsecuritypolicy privileged restricted || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
-  rm -Rf ${ADDONS_DIR}/init
+    retrycmd 120 5 30 $KUBECTL get podsecuritypolicy privileged restricted || exit {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}}
+    rm -Rf ${ADDONS_DIR}/init
```
CecileRobertMichon, line 394: I'm not too sure what this does and whether we want to run it every time or also skip it

Michael-Sinz: I think your change is correct - we only want to clean up after the deployment.

Michael-Sinz: Only question I have is what should happen if we are in "need of reboot" - what additional work below is still acceptable?

(we should think/model this as a state machine)

jackfrancis: So what is happening here:

1) Cloud-init populates a subset of "must load first" addons (at present pod-security-policy) to `/etc/kubernetes/addons/init`
2) The remaining addons are in `/etc/kubernetes/addons/`
3) kube-addon-manager configuration says "source addons from the init directory"
4) we wait for kube-addon-manager pod to be Running (which is an indirect way of confirming that the "init" addons have been loaded)
5) we then move all addons into the main `/etc/kubernetes/addons` directory, update the kube-addon-manager configuration so that it sources from there (and no longer init) and then restart kube-addon-manager to load the remaining addons

So, if we're in a node reboot situation, we should ensure that we do step 5, but we can ignore the others.

Does that make sense?

CecileRobertMichon: which step is `rm -Rf ${ADDONS_DIR}/init` part of?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -226,7 +226,11 @@ if [[ -n ${MASTER_NODE} ]]; then
 {{if IsAzurePolicyAddonEnabled}}
   time_metric "EnsureLabelExclusionForAzurePolicyAddon" ensureLabelExclusionForAzurePolicyAddon
 {{end}}
-  time_metric "EnsureAddons" ensureAddons
+  if [ -f /var/run/reboot-required ] || [ "$NO_OUTBOUND" = "true" ]; then
```
CecileRobertMichon, line 229: looking at https://github.com/Azure/aks-engine/pull/3410/files, we might actually not need this for NO_OUTBOUND @jackfrancis

CecileRobertMichon: removed the no_outbound condition

PR #3722, opened 2020-08-17
test: add E2E config for "control plane reboot" scenario

jackfrancis:
**Reason for Change**:

This PR adds an E2E test config for validating the health of cluster deployments when control plane VMs are rebooted immediately after provisioning.

In practice, we re-use the "ubuntu-18.04" distro to implement "requires reboot", as that distro includes known cloud-init-related reboot behavior.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/test_cluster_configs/control_plane_reboot.json
+++ b/test/e2e/test_cluster_configs/control_plane_reboot.json
@@ -1,5 +1,7 @@
 {
-	"env": {},
+	"env": {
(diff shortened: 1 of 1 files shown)


jackfrancis:
No longer needed, using the `REBOOT_CONTROL_PLANE_NODES` config option instead

PR #3723, opened 2020-08-17
chore: specify containerd version to install with moby-engine package

cpuguy83:
**Reason for Change**:

Since moby packages now depend on a separate containerd package, we should specify which version of the containerd package should be picked up so that we don't accidentally introduce regressions for new installs/upgrades due to `apt-get` picking the latest containerd (which may include major version bumps).

This also enables testing aks-engine with different versions of containerd (such as the newly released 1.4) since the api model now allows specifying a containerd version when the moby version is >= 19.03.

**Issue Fixed**:

N/A

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -92,21 +92,34 @@ removeMoby() {
 removeContainerd() {
   apt_get_purge moby-containerd || exit 27
 }
+mobyPkgVersion() {
+  dpkg -s "${1}" | grep "Version:" | awk '{ print $2 }' | cut -d '+' -f 1
+}
 installMoby() {
-  CURRENT_VERSION=$(dockerd --version | grep "Docker version" | cut -d "," -f 1 | cut -d " " -f 3 | cut -d "+" -f 1)
-  if [[ $CURRENT_VERSION != "${MOBY_VERSION}" ]]; then
-    removeContainerd
+  install_pkgs=""
+  CURRENT_CONTAINERD_VERSION="$(mobyPkgVersion moby-containerd)"
+  if [ -n "${CONTAINERD_VERSION}" ] && [ ! "${CURRENT_CONTAINERD_VERSION}" = "${CONTAINERD_VERSION}" ]; then
+    install_pkgs+=" moby-containerd=${CONTAINERD_VERSION}*"
     removeMoby
+    removeContainerd
+  fi
+  CURRENT_ENGINE_VERSION="$(mobyPkgVersion moby-engine)"
+  if [ ! "${CURRENT_ENGINE_VERSION}" = "${MOBY_VERSION}" ]; then
+    install_pkgs+=" moby-engine=${MOBY_VERSION}*"
+    MOBY_CLI="${MOBY_VERSION}"
+    if [ "${MOBY_CLI}" = "3.0.4" ]; then
+      MOBY_CLI="3.0.3"
+    fi
+    install_pkgs+=" moby-cli=${MOBY_CLI}*"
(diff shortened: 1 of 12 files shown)


pkg/api/components_test.go
```diff
@@ -780,7 +780,6 @@ func TestSynthesizeComponentsConfig(t *testing.T) {
 	for _, c := range cases {
 		c := c
 		t.Run(c.name, func(t *testing.T) {
-			t.Parallel()
```
mboersma, line 783: Can these not run in parallel now, or was this for debugging?

cpuguy83: I have this split up in a separate commit.
Tests are modifying `c.components`, is shared between a couple of cases.

I got a concurrent read/write panic here.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -92,21 +92,44 @@ removeMoby() {
 removeContainerd() {
   apt_get_purge moby-containerd || exit 27
 }
+mobyPkgVersion() {
+  # This gets the version of the package and trims extra bits that the moby packages have (e.g. `+azure`)
```
mboersma, line 96: Since this script is part of the cloud-init payload, every byte is precious. So I think the comments here and below need to be omitted, sadly.

pkg/versions/compare.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
mboersma, line 1: AKS Engine already uses https://github.com/blang/semver--could it be used here instead of new code?

cpuguy83: No, since that package is quite strictly semver whereas the versions we are comparing are calver.
https://play.golang.org/p/aClagUgV2Fk

parts/k8s/cloud-init/artifacts/cse_install.sh
cpuguy83: Ah, interesting.

cpuguy83: I bet we could do something to strip comments in the generated script.

cpuguy83: Took comments and extra line breaks out.

pkg/api/vlabs/validate.go
```diff
@@ -37,7 +38,7 @@ var (
 		"3.1.0", "3.1.1", "3.1.2", "3.1.2", "3.1.3", "3.1.4", "3.1.5", "3.1.6", "3.1.7", "3.1.8", "3.1.9", "3.1.10",
 		"3.2.0", "3.2.1", "3.2.2", "3.2.3", "3.2.4", "3.2.5", "3.2.6", "3.2.7", "3.2.8", "3.2.9", "3.2.11", "3.2.12",
 		"3.2.13", "3.2.14", "3.2.15", "3.2.16", "3.2.23", "3.2.24", "3.2.25", "3.2.26", "3.3.0", "3.3.1", "3.3.8", "3.3.9", "3.3.10", "3.3.13", "3.3.15", "3.3.18", "3.3.19", "3.3.22"}
-	containerdValidVersions              = [...]string{"1.3.2"}
+	containerdValidVersions              = [...]string{"1.3.2", "1.3.3", "1.3.4", "1.3.5", "1.3.6", "1.3.7"}
```
amrmahdi, line 41: Can 1.4.0 also be allowed? If I understand correctly, even with this change I won't be able to use containerRuntime as `containerd` and version `1.4.0`

PR #3738, opened 2020-08-21
test: fix ginkgo failFast

jadarsie:
**Reason for Change**:

The way the E2E runner was building the `ginkgo` command was not gracefully handling empty parameters (in this case `failFast`)

Related to #3660

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@jadarsie @haofan-ms thanks for root causing this!

Let's go ahead and change this default to `false` so we can start running the complete set of tests on all test runs by default:

https://github.com/Azure/aks-engine/blob/master/test/e2e/config/config.go#L50

(Doing so in this PR will also act as a simple smoke test for this fix.)

PR #3739, opened 2020-08-22
chore: Update moby version to 19.03.12

bowang-666:


mboersma:
@bowang-666 could you `git commit --amend` and force push this so the commit message is [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) style? Something like you have as the PR title:

```
chore: update moby version to 19.03.11
```

pkg/api/const.go
```diff
@@ -457,7 +457,7 @@ const (
 	// DefaultKubernetesDNSServiceIPv6 specifies the IPv6 address that kube-dns listens on by default. must by in the default Service CIDR range.
 	DefaultKubernetesDNSServiceIPv6 = "fd00::10"
 	// DefaultMobyVersion specifies the default Azure build version of Moby to install.
-	DefaultMobyVersion = "3.0.10"
+	DefaultMobyVersion = "19.03.11"
```
cpuguy83, line 460: 19.03.12 is latest.

Can't believe we missed this one.

bowang-666: seems on github it says latest release is 19.03.11: https://github.com/Azure/moby/releases
also 11 was released on Jun.26 but 12 was released on Jun.25 which looks weird

cpuguy83: That's just tag date vs when the release was created in GH.

cpuguy83: One is automatic when a tag is created, the other is manual after the fact.

bowang-666: OK sounds good. I'll update to .12

PR #3747, opened 2020-08-25
test: add private key input to e2e suite + keep all junit result files

haofan-ms:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/config/config.go
```diff
@@ -44,6 +44,8 @@ type Config struct {
 	ResourceGroup       string `envconfig:"RESOURCE_GROUP" default:""`
 	SoakClusterName     string `envconfig:"SOAK_CLUSTER_NAME" default:""`
 	ForceDeploy         bool   `envconfig:"FORCE_DEPLOY" default:"false"`
+	PublicSSHKey        string `envconfig:"PUBLIC_SSH_KEY" default:""`
+	PrivateSSHKeyPath   string `envconfig:"PRIVATE_SSH_KEY_FILE" default:""`
```
jadarsie, line 48: It would be nice to update the field name or add a comment stating that this is a relative path

jadarsie:
Taking a look

test/e2e/runner/cli_provisioner.go
```diff
@@ -113,13 +113,15 @@ func (cli *CLIProvisioner) provision() error {
 	}
 	os.Setenv("NAME", cli.Config.Name)
 
-	outputPath := filepath.Join(cli.Config.CurrentWorkingDir, "_output")
-	if !cli.Config.UseDeployCommand {
-		publicSSHKey, err := createSaveSSH(outputPath, cli.Config.Name+"-ssh")
-		if err != nil {
-			return errors.Wrap(err, "Error while generating ssh keys")
+	if cli.Config.PublicSSHKey == "" || cli.Config.PrivateSSHKeyPath == "" {
```
jadarsie, line 116: Users may see unexpected hard-to-troubleshoot behavior if they set one out of the two variables.

I would validate that either none or both are set.

Also, this code should also set `PUBLIC_SSH_KEY`, probably not good to assume that this is always going to be called from `cluster.sh`

test/e2e/config/config.go
```diff
@@ -187,11 +189,23 @@ func (ccc *CustomCloudConfig) SetEnvironment() error {
 	// as azure-cli complains otherwise
 	azsSelfSignedCaPath := "/aks-engine/Certificates.pem"
 	if _, err = os.Stat(azsSelfSignedCaPath); err == nil {
+		// include cacert.pem from python2.7 path for upgrade scenario
```
jadarsie, line 192: the only difference seems to be the python version

instead of duplicating code the entire block, you may be able to solve it with an extra local variable

test/e2e/cluster.sh
```diff
@@ -42,6 +42,26 @@ ${ADD_NODE_POOL_INPUT}
...
+fi
+
+function tryExit {
+  if [ "${AZURE_ENV}" != "AzureStackCloud" ]; then
+    exit 1
+  fi
+}
+
+function moveStageJunit {
+  if [ "${AZURE_ENV}" == "AzureStackCloud" ]; then
```
jadarsie, line 60: It is probably ok to do this for all clouds

test/e2e/cluster.sh
```diff
@@ -42,6 +42,26 @@ ${ADD_NODE_POOL_INPUT}
...
+  PRIVATE_SSH_KEY_FILE=$(realpath --relative-to=$(pwd) ${PRIVATE_SSH_KEY_FILE})
+fi
+
+function tryExit {
+  if [ "${AZURE_ENV}" != "AzureStackCloud" ]; then
+    exit 1
+  fi
+}
+
+function moveStageJunit {
```
jadarsie, line 59: nit: `renameResultsFile` may be more specific

test/e2e/cluster.sh
```diff
@@ -42,6 +42,22 @@ ${ADD_NODE_POOL_INPUT}
...
+fi
+
+function tryExit {
+  if [ "${AZURE_ENV}" != "AzureStackCloud" ]; then
+    exit 1
+  fi
+}
+
+function renameResultsFile {
+  if [ "${AZURE_ENV}" == "AzureStackCloud" ] && [ "${SKIP_TEST}" == "false" ]; then
```
jadarsie, line 56: What about variables `SKIP_TESTS_AFTER_*`? I may be simpler to check if the results file exists

jadarsie (changes requested):

test/e2e/cluster.sh
```diff
@@ -42,6 +42,23 @@ ${ADD_NODE_POOL_INPUT}
...
+function tryExit {
+  if [ "${AZURE_ENV}" != "AzureStackCloud" ]; then
+    exit 1
+  fi
+}
+
+function renameResultsFile {
+  JUNIT_PATH=$(pwd)/test/e2e/kubernetes/junit.xml
+  if [ "${AZURE_ENV}" == "AzureStackCloud" ] && [ -f ${JUNIT_PATH} ]; then
+    sudo mv ${JUNIT_PATH} $(pwd)/test/e2e/kubernetes/${1}-junit.xml
```
jadarsie, line 58: is `sudo` needed?

test/e2e/runner/cli_provisioner.go
```diff
@@ -87,15 +87,30 @@ func (cli *CLIProvisioner) Run() error {
...
+			return "", errors.Wrapf(err, "Error while trying to move private ssh key")
+		}
+		cmd = exec.Command("ssh-keygen", "-y", "-f", keyPath)
+	}
+
 	util.PrintCommand(cmd)
-	out, err := cmd.CombinedOutput()
-	if err != nil {
-		return "", errors.Wrapf(err, "Error while trying to generate ssh key\nOutput:%s", out)
+	out, err2 := cmd.CombinedOutput()
```
jadarsie, line 104: Instead of declaring a few `err#` variables, it would be more idiomatic to declare `var err error` at the beginning of the func and reuse it.

haofan-ms: Updated

test/e2e/cluster.sh
haofan-ms: From the test runs this is needed otherwise there will be permission issue, since we are running the cluster.sh as "azureuser".

jackfrancis: Why is sudo needed? Is it the directory permissions?

jadarsie: it may be better to `chown` the aks-engine directory before `cluster.sh` is executed.

haofan-ms: Tested again and it works without "sudo". Sorry for the false statement, I have removed the "sudo" part.

jadarsie (approved):

PR #3748, opened 2020-08-25
chore: rev default etcd version to 3.4.13

### LINKED ISSUE #3663: add etcd 3.4 support
Let's add support for etcd 3.4 as it is sufficiently mature.

https://github.com/etcd-io/etcd/releases

jackfrancis:
**Reason for Change**:

This PR updates the default etcd version to v3.4.25 or control plane nodes.

Ref:

https://github.com/etcd-io/etcd/releases/tag/v3.4.13

**Issue Fixed**:

Fixes #3663 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
So, after going through this prototype exercise to make etcd 3.4 work, I strongly suspect that maintaining a common configuration surface area that will work across all the currently supported etcd versions is not tenable. We should strongly consider introducing 3.4 as a "you can't go back to 3.3 or prior" change.

cc @Michael-Sinz @devigned

Michael-Sinz:
Interesting - we should look at a what this buys us.  We have a lot of miles on 3.3.x but we know that there are demons in there.

We should ask the aks team what they saw with 3.4 vs 3.3 in their setup.  However, there have been critical fixes that went into 3.4 first and then backported to 3.3.  For example, you really don't want to be [at or before 3.3.13 due to at least one critical issue](https://github.com/kubernetes/kubernetes/issues/72102) that was first fixed in 3.4 and then finally backported to 3.3.14

mboersma:
Just fyi, as of Kubernetes 1.19.1, etcd server [3.4.13 is the default](https://github.com/kubernetes/kubernetes/pull/94536).

mboersma (approved):
/lgtm

jackfrancis:
This PR is not compatible w/ etcd 3.3 and below

bridgetkromhout:
Closing per @jackfrancis, given AKS Engine deprecation.

PR #3752, opened 2020-08-26
feat: variable upgrade timeout based on num nodes

### LINKED ISSUE #3746: Upgrade has a Timeout of 3 hrs
**Describe the bug**
Aks Engine upgrade comes to a sudden stop  on reaching 3 hrs mark.

**Steps To Reproduce**
Run Upgrade for around 3 hrs. For me it was a cluster with 3 Masters and 20 Nodes

**Expected behavior**
The timeout should be configurable or adjusted a bit higher, as because of another issue (https://github.com/Azure/aks-engine/issues/3520) we are hitting the 3 hr limit for 20 Nodes. 

**AKS Engine version**
v0.54.0

**Kubernetes version**
15
**Additional context**
If possible look into #3520 bug as that is whats contributing to the Timeout. Simply restarting Upgrade will not work for our usecase as we are doing Force Upgrades.

jackfrancis:
**Reason for Change**:

This PR replaces the static 3 hour upgrade timeout with a "per-node" timeout. We calculate a a sum 20 mins per node for every node-to-be upgraded, and allow the aks-engine binary that much total time before throwing a general `aks-engine upgrade` operation timeout error.

For small clusters (less than 9 nodes including control plane nodes), this has the practical effect of reducing the total timeout tolerance for `aks-engine upgrade` operations. For larger clusters (10 or more total nodes), this has the practical effect of increasing the timeout tolerance. For really large clusters, that timeout tolerance becomes quite large.

In summary, the observed ~5 mins per-node upgrade time means that clusters with > 30 nodes cannot be upgraded in one operation successfully: they will consistently run into the 3 hour `aks-engine upgrade` timeout. This change unblocks such large cluster `aks-engine upgrade` scenarios, with the tradeoff of waiting a really long time in the event of a "terminally stuck somewhere in Azure" encounter.

**Issue Fixed**:

Fixes #3746 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -49,7 +49,7 @@ const (
 	defaultCordonDrainTimeout          = time.Minute * 20
 	nodePropertiesCopyTimeout          = time.Minute * 5
 	getResourceTimeout                 = time.Minute * 1
-	clusterUpgradeTimeout              = time.Minute * 180
+	perNodeUpgradeTimeout              = time.Minute * 20
 	vmStatusUpgraded          vmStatus = iota
 	vmStatusNotUpgraded
 	vmStatusIgnored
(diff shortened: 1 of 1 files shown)


devigned:
👏 much better than what I suggested by having a configurable parameter. You are right. A user should not need to know the details. The software should have a good idea of how long it should take.

mboersma:
> For small clusters (less than 9 nodes including control plane nodes), this has the practical effect of reducing the total timeout tolerance

I wonder if there are clusters in this range that could be impacted. Perhaps some VM type and/or region that is particularly slow to provision might currently be expecting more of the three-hour limit for `aks-engine upgrade` than 20min/node will give them?

But no, that sounds pathological anyway once I've typed it out, and this is a more generous timeout for clusters where it's likely to matter.

PR #3754, opened 2020-08-26
feat: add support for Kubernetes v1.19.0

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.19.md#v1190

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

~The official release is in progress, and then we will kick off the pipeline to build signed artifacts hosted with Azure. So this will fail testing for now.~ *Edit*: all done!

### DIFF
--- a/parts/k8s/addons/1.15/calico.yaml
+++ b/parts/k8s/addons/1.15/calico.yaml
@@ -41,7 +41,7 @@ data:
 
 ---
 {{- /* Source: calico/templates/kdd-crds.yaml */}}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1{{- if not (IsKubernetesVersionGe "1.19.0")}}beta1{{end}}
 kind: CustomResourceDefinition
 metadata:
   name: felixconfigurations.crd.projectcalico.org
@@ -57,7 +57,7 @@ spec:
     singular: felixconfiguration
 ---
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1{{- if not (IsKubernetesVersionGe "1.19.0")}}beta1{{end}}
 kind: CustomResourceDefinition
 metadata:
   name: bgpconfigurations.crd.projectcalico.org
@@ -73,7 +73,7 @@ spec:
     singular: bgpconfiguration
 ---
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1{{- if not (IsKubernetesVersionGe "1.19.0")}}beta1{{end}}
 kind: CustomResourceDefinition
 metadata:
   name: ippools.crd.projectcalico.org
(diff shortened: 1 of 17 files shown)


parts/k8s/addons/1.15/calico.yaml
```diff
@@ -41,7 +41,7 @@ data:
 
 ---
 {{- /* Source: calico/templates/kdd-crds.yaml */}}
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1{{- if not (IsKubernetesVersionGe "1.19.0")}}beta1{{end}}
```
mboersma, line 44: We can omit these conditionals if `apiextensions.k8s.io/v1` was supported on Kubernetes 1.15.

jackfrancis: This calico spec will only be delivered in a Kubernetes v1.15 cluster creation context, do we need to do anything here?

mboersma: Oops...bad example. I reverted changes to this file, but there are many similar changes in other manifests.

jackfrancis:
The changes to the apiextensions.k8s.io/v1beta1 apiVersion are breaking the secrets-store-csi spec. From kube-addon-manager logs:

```
error validating "/etc/kubernetes/addons/secrets-store-csi-driver.yaml": error validating data: [ValidationError(CustomResourceDefinition.spec): unknown field "validation" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec, ValidationError(CustomResourceDefinition.spec): unknown field "version" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1.CustomResourceDefinitionSpec]; if you choose to ignore these errors, turn validation off with --validate=false
```

This seems to be complaining about these lines:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/addons/secrets-store-csi-driver.yaml#L203
https://github.com/Azure/aks-engine/blob/master/parts/k8s/addons/secrets-store-csi-driver.yaml#L266

@chewong @aramase Do we need to undo the api version change and revert back to:

```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
```

Or do we want to delete that `version` property from the spec using the new `apiVersion: apiextensions.k8s.io/v1` declaration?

aramase:
> Do we need to undo the api version change and revert back to:

@jackfrancis Can we keep the version for now as `apiextensions.k8s.io/v1beta1`. I'll open a follow up PR later to update the api version after some testing upstream.

`apiextensions.k8s.io/v1beta1` was deprecated in 1.19 and is planned for removal in `1.22` (https://github.com/kubernetes/kubernetes/issues/82022). I'll setup some tests upstream to run with 1.19 before including that change in aks-engine. WDYT?

jackfrancis:
@aramase Great plan, will do!

PR #3758, opened 2020-08-27
fix: Restart=always docker systemd service

jackfrancis:
**Reason for Change**:

This PR ensures that Restart=always is configured for the docker systemd service, and that the included monitor script for both docker and kubelet has a fail-safe "start the systemd service" if the systemd job spec doesn't do so itself.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/health-monitor.sh
+++ b/parts/k8s/cloud-init/artifacts/health-monitor.sh
@@ -32,6 +32,9 @@ container_runtime_monitoring() {
       fi
       systemctl kill --kill-who=main "${container_runtime_name}"
       sleep 120
+      if ! systemctl is-active ${container_runtime_name}; then
+        systemctl start ${container_runtime_name}
+      fi
     else
       sleep "${SLEEP_SECONDS}"
     fi
@@ -49,6 +52,9 @@ kubelet_monitoring() {
       echo "Kubelet is unhealthy!"
       systemctl kill kubelet
       sleep 60
+      if ! systemctl is-active kubelet; then
+        systemctl start kubelet
+      fi
     else
       sleep "${SLEEP_SECONDS}"
     fi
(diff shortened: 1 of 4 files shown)


jackfrancis:
@xuto2 @yangl900 FYI, this pathology may exist for AKS node systemd definitions as well

xuto2:
/lgtm.

Thanks @jackfrancis for sharing!

jackfrancis:
After more investigation, it appears that Restart=always is already present for the docker systemd service:

$ grep Restart /lib/systemd/system/docker.service
RestartSec=2
Restart=always

(Though not via aks-engine-delivered config).

So, we should still not just rely upon that being there, and explicitly declare that config (like we're doing in this PR). However, the fact that we're still seeing terminally stopped docker systemd services in the wild suggests that we aren't as resilient as we want to in terms of guaranteeing the availability of the docker systemd service. These docs suggest that Restart=always in fact has limits:

https://www.freedesktop.org/software/systemd/man/systemd.service.html#Restart=

The key statement is:

> As exceptions to the setting above, the service will not be restarted if the exit code or signal is specified in RestartPreventExitStatus= (see below) or the service is stopped with systemctl stop or an equivalent operation.

We aren't specifying any non-qualifying exit codes for docker:

```
$ sudo grep -R "RestartPreventExitStatus" /etc/systemd/
/etc/systemd/system/sshd.service:RestartPreventExitStatus=255
/etc/systemd/system/multi-user.target.wants/snapd.service:RestartPreventExitStatus=42
/etc/systemd/system/multi-user.target.wants/ssh.service:RestartPreventExitStatus=255
```

So there must be some systemctl stop-equivalent events that occasionally happen. In order to restart docker after that, I've added `systemctl start` logic to the script that the pre-existing docker-monitor service uses. I think with this combination, we'll be in better shape to deal with docker systemd service events.

PR #3767, opened 2020-08-31
fix: availabilityZones value in template to read from parameter

### LINKED ISSUE #3762: Availibility zone of virtualMachineScaleSets is specified directly in template file instead of reading from parameter file
**Describe the bug**

The value of the availability zone in the generated template file is directly specified and does not use parameters.

**Steps To Reproduce**

Create a Kubernetes cluster specifying the availability zone. 

e.g 

```json
"agentPoolProfiles": [
        {
          "name": "",
          "count": 3,
          "vmSize": "Standard_D2_v3",
          "vnetSubnetId": "",
          "availabilityZones": ["1","2", "3"],
        }
```

Generate template using 
```
aks-engine generate kubeconfig.json
```

Check the template resource section you and search for vmss. Check the availability zine value it will be fixed and not being read from the parameter.

```json
"zones": [
        "1",
        "2",
        "3"
      ]
```

**Expected behavior**

zones value should be read from parameter. 

```json
"zones": "[parameters('registryAvailabilityZones')]"
```

**AKS Engine version**

Version: v0.53.0
GitCommit: b1655d6cc
GitTreeState: clean

**Kubernetes version**

1.18

**Additional context**

manishram:
fix: Fix availabilityZones value in template to read from parameter

**Reason for Change**:
The availability zone of virtualMachineScaleSets is specified directly in the template file instead of reading from the parameter file.

**Issue Fixed**:
Fixes #3762 

fix: Bug Fixes 🐞

**Earlier:**
grege4r

**Now:**
reef

### DIFF
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -85,7 +85,7 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 	addCustomTagsToVMScaleSets(cs.Properties.MasterProfile.CustomVMTags, &virtualMachine)
 
 	if hasAvailabilityZones {
-		virtualMachine.Zones = to.StringPtr("[parameters('availabilityZones')]")
+		virtualMachine.Zones = &[]string{"[parameters('availabilityZones')]"}
 	}
 
 	if userAssignedIDEnabled {
@@ -417,7 +417,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	addCustomTagsToVMScaleSets(profile.CustomVMTags, &virtualMachineScaleSet)
 
 	if profile.HasAvailabilityZones() {
-		virtualMachineScaleSet.Zones = to.StringPtr(fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name))
+		virtualMachineScaleSet.Zones = &[]string{fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name)}
 	}
 
 	var useManagedIdentity bool
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -417,7 +417,11 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	addCustomTagsToVMScaleSets(profile.CustomVMTags, &virtualMachineScaleSet)
 
 	if profile.HasAvailabilityZones() {
-		virtualMachineScaleSet.Zones = &[]string{fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name)}
+		zones := []string{}
+		for i := range profile.AvailabilityZones {
+			zones = append(zones, fmt.Sprintf("[parameters('%sAvailabilityZones')[%d]]", profile.Name, i))
+		}
+		virtualMachineScaleSet.Zones = &zones
 	}
 
 	var useManagedIdentity bool
--- a/pkg/engine/virtualmachinescalesets.go
+++ b/pkg/engine/virtualmachinescalesets.go
@@ -85,7 +85,7 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 	addCustomTagsToVMScaleSets(cs.Properties.MasterProfile.CustomVMTags, &virtualMachine)
 
 	if hasAvailabilityZones {
-		virtualMachine.Zones = &masterProfile.AvailabilityZones
+		virtualMachine.Zones = to.StringPtr("[parameters('availabilityZones')]")
 	}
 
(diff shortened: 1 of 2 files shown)


jackfrancis:
@manishram Thanks! can you get rid of the change in the vendor directory (may have been a local go mod artifact)?

manishram:
@jackfrancis should I revert the change in line number 2488 and do that for line number 10636 or completely revert changes from that file?

manishram:
@jackfrancis why tests are failing? Any more changes I need to do?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -85,7 +85,7 @@ func CreateMasterVMSS(cs *api.ContainerService) VirtualMachineScaleSetARM {
 	addCustomTagsToVMScaleSets(cs.Properties.MasterProfile.CustomVMTags, &virtualMachine)
 
 	if hasAvailabilityZones {
-		virtualMachine.Zones = &masterProfile.AvailabilityZones
+		virtualMachine.Zones = to.StringPtr("[parameters('availabilityZones')]")
```
jackfrancis, line 88: the `Zones` property is a `*[]string` (pointer to a string slice), so this would be the syntax for expressing the above change as a literal value:

`&[]string{"[parameters('availabilityZones')]"}`

pkg/engine/virtualmachinescalesets.go
```diff
@@ -417,7 +417,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	addCustomTagsToVMScaleSets(profile.CustomVMTags, &virtualMachineScaleSet)
 
 	if profile.HasAvailabilityZones() {
-		virtualMachineScaleSet.Zones = &profile.AvailabilityZones
+		virtualMachineScaleSet.Zones = to.StringPtr(fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name))
```
jackfrancis, line 420: ditto above comment:

`&[]string{fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name)}

pkg/engine/virtualmachinescalesets.go
manishram: fixed

pkg/engine/virtualmachinescalesets.go
manishram: fixed

pkg/engine/virtualmachinescalesets.go
manishram: Do you think `availabilityZones` should be `AvailabilityZones`?

manishram: do you think this can be the fix? ` &[]string{"[string(parameters('availabilityZones')]"}`

manishram: do you think this can be the fix? &[]string{"[string(parameters('availabilityZones'))]"}

manishram: do you think this can be the fix? `&[]string{"[string(parameters('availabilityZones'))]"}`

pkg/engine/virtualmachinescalesets.go
manishram: And for this one this can be the fix? `&[]string{fmt.Sprintf("[string(parameters('%sAvailabilityZones'))]", profile.Name)}`

manishram:
@jackfrancis what is the cause for failing?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -417,7 +417,7 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	addCustomTagsToVMScaleSets(profile.CustomVMTags, &virtualMachineScaleSet)
 
 	if profile.HasAvailabilityZones() {
-		virtualMachineScaleSet.Zones = &profile.AvailabilityZones
+		virtualMachineScaleSet.Zones = &[]string{fmt.Sprintf("[parameters('%sAvailabilityZones')]", profile.Name)}
```
jackfrancis, line 420: @manishram, I didn't quite give you the right direction here. Because the profile zones param resolves to an array under the hood, which means we're essentially composing an array inside an array. Here's the error:

```
InvalidTemplate - Deployment template parse failed: 'Error reading string. Unexpected token: StartArray. Path '[0]'.'.
```

Instead, we need to manually compose the array one index item at time.

This is a working solution for this entire if block:

```
	if profile.HasAvailabilityZones() {
		zones := []string{}
		for i := range profile.AvailabilityZones {
			zones = append(zones, fmt.Sprintf("[parameters('%sAvailabilityZones')[%d]]", profile.Name, i))
		}
		virtualMachineScaleSet.Zones = &zones
	}
```

manishram: Okay

manishram:
@jackfrancis  still some problem with codecov

jackfrancis:
@manishram don't worry about the codecov check, thanks for this contribution!

PR #3773, opened 2020-09-01
chore: simplify addons config for calico and flannel

jackfrancis:
**Reason for Change**:

This PR removes some of the sed tomfoolery for calico and flannel addons specs and instead uses go templating for more easier maintenance.

Also modified the E2E test cluster configs so that the "calico" test uses kubenet for IPAM, and the "calico_azure" test config uses Azure CNI.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma (approved):
/lgtm

PR #3774, opened 2020-09-01
fix: Enable Windows VHD version update in aks-engine upgrade scenario

haofan-ms:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -52,6 +52,7 @@ type upgradeCmd struct {
 	force                                    bool
 	controlPlaneOnly                         bool
 	disableClusterInitComponentDuringUpgrade bool
+	upgradeWindowsVHD                        bool
 
 	// derived
 	containerService    *api.ContainerService
@@ -87,6 +88,7 @@ func newUpgradeCmd() *cobra.Command {
 	f.IntVar(&uc.cordonDrainTimeoutInMinutes, "cordon-drain-timeout", -1, "how long to wait for each vm to be cordoned in minutes")
 	f.BoolVarP(&uc.force, "force", "f", false, "force upgrading the cluster to desired version. Allows same version upgrades and downgrades.")
 	f.BoolVarP(&uc.controlPlaneOnly, "control-plane-only", "", false, "upgrade control plane VMs only, do not upgrade node pools")
+	f.BoolVarP(&uc.upgradeWindowsVHD, "upgrade-windows-vhd", "", true, "upgrade image reference of Windows VHD")
 	addAuthFlags(uc.getAuthArgs(), f)
 
 	_ = f.MarkDeprecated("deployment-dir", "deployment-dir is no longer required for scale or upgrade. Please use --api-model.")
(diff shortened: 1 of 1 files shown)


cmd/upgrade.go
```diff
@@ -87,6 +88,7 @@ func newUpgradeCmd() *cobra.Command {
 	f.IntVar(&uc.cordonDrainTimeoutInMinutes, "cordon-drain-timeout", -1, "how long to wait for each vm to be cordoned in minutes")
 	f.BoolVarP(&uc.force, "force", "f", false, "force upgrading the cluster to desired version. Allows same version upgrades and downgrades.")
 	f.BoolVarP(&uc.controlPlaneOnly, "control-plane-only", "", false, "upgrade control plane VMs only, do not upgrade node pools")
+	f.BoolVarP(&uc.upgradeWindowsVHD, "upgrade-windows-vhd", "", true, "upgrade image reference of Windows VHD")
```
jadarsie, line 91: `of the Windows nodes` maybe

cmd/upgrade.go
```diff
@@ -168,6 +170,17 @@ func (uc *upgradeCmd) loadCluster() error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	// Use the Windows VHD associated with the aks-engine version if upgradeWindowsVHD is set to "true"
+	if uc.upgradeWindowsVHD {
+		if uc.containerService.Properties.WindowsProfile.WindowsPublisher == api.AKSWindowsServer2019OSImageConfig.ImagePublisher && uc.containerService.Properties.WindowsProfile.WindowsOffer == api.AKSWindowsServer2019OSImageConfig.ImageOffer {
```
jadarsie, line 175: this maybe more readable if `uc.containerService.Properties.WindowsProfile` is stored in a variable.

cmd/upgrade.go
haofan-ms: updated

cmd/upgrade.go
haofan-ms: updated

cmd/upgrade.go
```diff
@@ -168,6 +170,18 @@ func (uc *upgradeCmd) loadCluster() error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	// Use the Windows VHD associated with the aks-engine version if upgradeWindowsVHD is set to "true"
+	if uc.upgradeWindowsVHD {
```
jackfrancis, line 174: @marosset to review if this logic is what we want:

- if `--upgrade-windows-vhd=true`, then
  - set the appropriate Windows VHD defaults for this particular aks-engine build

jadarsie:
lgtm

cmd/upgrade.go
marosset: This same logic is *mostly* already duplicated twice

- https://github.com/Azure/aks-engine/blob/2569b075ebe7d7b54d93f2fc351b0290116ebb91/pkg/api/defaults.go#L840-L860
- https://github.com/Azure/aks-engine/blob/2569b075ebe7d7b54d93f2fc351b0290116ebb91/pkg/api/defaults.go#L895-L909

Can we maybe refactor that logic out to a reusable method?

I'm worried that if/when we update the logic it will get out of sync.

haofan-ms: My consideration for this is that defaults.go and upgrade.go handles different behaviors. One is to set the default value for the Windows VHD image reference, while the other one is to update/not update the Windows VHD image reference at a specific step. When we set the flag "upgrade-windows-vhd", it's not an indicator for default value. Does this make sense?

marosset: Yes and I agree with you.

marosset:
@haofan-ms Is there anything you can do to increase code coverage to pass the codecov automated check here?
Otherwise the PR looks good.

jackfrancis:
I'm O.K. w/ this merging as-is. Manual upgrade tests checked out.

Thanks @haofan-ms!

jackfrancis (approved):
/lgtm

PR #3787, opened 2020-09-03
feat: secrets store addon can pull from custom env

### LINKED ISSUE #3786: Secrets store addon pulling from custom clouds
**Describe the request**

The `secrets-store-csi-driver` addon needs access to the custom cloud environment json in order to pull Key Vault secrets stored on custom clouds.

This file is currently unreachable to the addon as AKS Engine drops the json file in path `/etc/kubernetes/azurestackcloud.json` (see [current addon volumes](https://github.com/Azure/aks-engine/blob/v0.55.0/parts/k8s/addons/secrets-store-csi-driver.yaml#L395-L411))

So, I am thinking we could either (1, preferred IMO) mount `/etc/kubernetes` or (2) copy the environment json file to `/etc/kubernetes/secrets-store-csi-providers`.

@jackfrancis  / @aramase, what's your take?

**Additional context**

Related work: https://github.com/Azure/secrets-store-csi-driver-provider-azure/pull/220

jadarsie:
**Reason for Change**:

Make sure that the secrets store addon is able to pull Key Vault entries from air-gapped clouds. 

**Issue Fixed**:

Fixes #3786

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/secrets-store-csi-driver.yaml
+++ b/parts/k8s/addons/secrets-store-csi-driver.yaml
@@ -365,8 +365,8 @@ spec:
             - name: mountpoint-dir
               mountPath: /var/lib/kubelet/pods
               mountPropagation: Bidirectional
-            - name: providers-dir
-              mountPath: /etc/kubernetes/secrets-store-csi-providers
+            - name: etc-kubernetes
+              mountPath: /etc/kubernetes
           resources:
             limits:
               cpu: {{ContainerCPULimits "secrets-store"}}
@@ -405,9 +405,9 @@ spec:
           hostPath:
             path: /var/lib/kubelet/plugins/csi-secrets-store/
             type: DirectoryOrCreate
-        - name: providers-dir
+        - name: etc-kubernetes
           hostPath:
-            path: /etc/kubernetes/secrets-store-csi-providers
+            path: /etc/kubernetes
             type: DirectoryOrCreate
       nodeSelector:
         kubernetes.io/os: linux
--- a/parts/k8s/addons/secrets-store-csi-driver.yaml
+++ b/parts/k8s/addons/secrets-store-csi-driver.yaml
@@ -367,9 +367,11 @@ spec:
               mountPropagation: Bidirectional
             - name: providers-dir
               mountPath: /etc/kubernetes/secrets-store-csi-providers
+{{- if IsCustomCloudProfile}}
             - name: custom-environment
               mountPath: /etc/kubernetes/azurestackcloud.json
               readOnly: true
+{{end}}
           resources:
             limits:
               cpu: {{ContainerCPULimits "secrets-store"}}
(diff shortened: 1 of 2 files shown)


parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -365,8 +365,8 @@ spec:
             - name: mountpoint-dir
               mountPath: /var/lib/kubelet/pods
               mountPropagation: Bidirectional
-            - name: providers-dir
-              mountPath: /etc/kubernetes/secrets-store-csi-providers
```
jackfrancis, line 369: @aramase @chewong can you comment on the safety of this change?

aramase: @jackfrancis Thank you for asking about the safety. That makes me think more about this and on second thought, mounting `/etc/kubernetes` also mounts all the Kubernetes certs into the pod. If the single driver pod is compromised it could result in exposing all those certs. Keeping in accordance with best practices, it might be better to mount the least possible content required to make this work which in this case would be `/etc/kubernetes/environment.json`. For the standalone deployments outside of aks-engine addons, we could provider an option to custom `volumeMounts` and `volumes` in the `secrets-store-csi-driver` helm charts that users can leverage to mount this custom environment file. WDYT?

parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -367,6 +367,9 @@ spec:
               mountPropagation: Bidirectional
             - name: providers-dir
               mountPath: /etc/kubernetes/secrets-store-csi-providers
+            - name: custom-environment
```
jackfrancis, line 370: Let's wrap this in a `{{if IsAzureStack}}` condition.

jackfrancis: Sorry, looks like `{{if IsAzureStackCloud}}` is what we'll need. (`IsAzureStackCloud` is the existent func in the `getAddonFuncMap` that the addons files are go-template-generated w/

jadarsie: `{{if IsCustomCloudProfile}}` probably better?

jackfrancis: so I assume `azurestackcloud.json` is what all customCloudProfile-enabled clouds will get?

jadarsie: correct

jadarsie: https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/masternodecustomdata.yml#L537-L543

jackfrancis (approved):
/lgtm

PR #3788, opened 2020-09-04
fix: working systemd monitor jobs

jackfrancis:
**Reason for Change**:

This PR updates the implementation of the various systemd monitor jobs so that the following critical services are monitored for failure, and restarted:

- docker
- containerd
- etcd
- kubelet

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


examples/e2e-tests/kubernetes/release/default/definition.json
```diff
@@ -8,7 +8,7 @@
         "addons": [
           {
             "name": "kubernetes-dashboard",
-            "enabled": true
+            "enabled": false
```
jackfrancis, line 11: these are temporary changes while we work on reducing customData size

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 NODE_INDEX=$(hostname | tail -c 2)
 NODE_NAME=$(hostname)
-PRIVATE_IP=$(hostname -I | cut -d' ' -f1)
+PRIVATE_IP=$(hostname -i | cut -d' ' -f1)
```
jackfrancis, line 4: It's unclear to me why we're using `-I` (get me all interfaces) vs `-i` get me my primary interface...

Michael-Sinz: Note that the ip address here (-i) is a resolved IP address based on the host name.  Is that what you want?

The hostname -I returns all ip addresses without DNS resolution.  This means you get things like loopback, etc.

Michael-Sinz: Note that I would look into another way to get the right IP address for the master (like, it must be known somewhere already)

parts/k8s/cloud-init/artifacts/docker-monitor.timer
```diff
@@ -1,7 +1,7 @@
 [Unit]
```
jackfrancis, line 1: I would advocate we eliminate this timer spec. It adds complexity to the overall implementation, and the way we're implementing the docker health check (docker ps) should not be "racy" given that the health check script runs *only after* the docker systemd service has started (see `After=docker.service` above)

@Michael-Sinz @mboersma  thoughts?

jackfrancis: I think what we're implicitly saying by delaying things for 30 mins is that we are tolerant of `docker ps` repeatedly failing during the first 30 mins of boot time, which I don't think is defensible.

mboersma: I think providing a warmup period for Docker to get its act together was basically what Azure/acs-engine#4050 was about. I agree simpler is way better when it comes to systemd units especially, so l'm ok with removing it to see if it's unneeded now.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -372,6 +374,8 @@ ensureKubelet() {
   wait_for_file 1200 1 /etc/systemd/system/kubelet.service.d/kubereserved-slice.conf || exit {{GetCSEErrorCode "ERR_KUBELET_SLICE_SETUP_FAIL"}}
   {{- end}}
   systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  wait_for_file 1200 1 /etc/systemd/system/kubelet-monitor.service || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
```
jackfrancis, line 378: Also (continuing along the systemd relationship conversation), we are as a rule blocking the startup of the monitor jobs on the jobs themselves starting successfully, as exemplified by this kubelet-monitor start operation

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -22,6 +22,14 @@ NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
 NVIDIA_CONTAINER_RUNTIME_VER=2.0.0
 NVIDIA_DOCKER_SUFFIX=docker18.09.2-1
+PRIVATE_IP=$(ip -4 addr show eth0 | grep -Po '(?<=inet )[\d.]+')
```
jackfrancis, line 25: This is a new, common "what is my private IP address?" implementation to be shared across the various places where that runtime determination is needed.

eth0 always has this information at first boot; in an Azure CNI configuration, an "azure0" bridge interface is set up with the primary NIC IP address (and the eth0 interface no longer containes it)

In the edge case scenario where none of those exists, we fall back to the IP address that the hostname resolves to. The reason we don't want to do that primarily is because we don't want to rely upon DNS lookups.

cc @Michael-Sinz

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 NODE_INDEX=$(hostname | tail -c 2)
 NODE_NAME=$(hostname)
-PRIVATE_IP=$(hostname -I | cut -d' ' -f1)
```
jackfrancis, line 4: This var assignment has been moved into the `cse_helpers.sh` instead for more general purpose usage, and the etcd vars are moved down to local vars inside the funcs that use them

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -366,12 +365,22 @@ ensureKubelet() {
   sysctl_reload 10 5 120 || exit {{GetCSEErrorCode "ERR_SYSCTL_RELOAD"}}
   wait_for_file 1200 1 /etc/default/kubelet || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
   wait_for_file 1200 1 /var/lib/kubelet/kubeconfig || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  if [[ -n ${MASTER_NODE} ]]; then
```
jackfrancis, line 368: This one-time VMSS master-specific foo has been moved into the bootstrap script, and out of the shell script that runs every time that the kubelet systemd service starts.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -22,6 +22,11 @@ NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
 NVIDIA_CONTAINER_RUNTIME_VER=2.0.0
 NVIDIA_DOCKER_SUFFIX=docker18.09.2-1
+PRIVATE_IP=$( (ip -br -4 addr show eth0 || ip -br -4 addr show azure0) | grep -Po '\d+\.\d+\.\d+\.\d+')
```
jackfrancis, line 25: This code does the following:

1) Get me the IP addresses on eth0
2) Or, if there aren't any, get me the IP addresses on azure0
3) If there isn't exactly 1 IP address from trying those, then just get the IP address that the hostname DNS entry returns

#3 is undesirable as it relies upon functional DNS, so we only do it in a fallback scenario

We consider the possibility that eth0 won't have the desired IP address in order to make this solution resilient during the lifecycle of the VM: in Azure CNI configuration scenarios the IP address will be attached to the "azure0" interface when the first (non-hostNetwork) pod is scheduled onto the node that the VM represents (that's how Azure CNI routes container traffic out of the VM).

parts/k8s/cloud-init/artifacts/docker-monitor.service
```diff
@@ -5,5 +5,8 @@ After=docker.service
 Restart=always
 RestartSec=10
 RemainAfterExit=yes
+Environment=CONTAINER_RUNTIME={{GetContainerRuntime}}
```
jackfrancis, line 8: This service is now overloaded to support both moby or containerd

parts/k8s/cloud-init/artifacts/docker-monitor.timer
jackfrancis: After further investigation, it *is* a bit tricky to tell a systemd service "wait until this service starts, and *also* until it is fully activated; and so I've moved the delay into the health script itself. Arguably that is less complicated than maintaining systemd overhead to do that.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -488,11 +484,11 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
         echo "" >> /etc/environment
     fi
   {{- if IsMasterVirtualMachineScaleSets}}
+    source {{GetCSEHelpersScriptFilepath}}
```
jackfrancis, line 487: This source statement gets us the `$PRIVATE_IP` var we need

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -78,7 +78,7 @@ var (
 	stabilityCommandTimeout         time.Duration
 	env                             azure.Environment
 	azureClient                     *armhelpers.AzureClient
-	firstMasterRegexStr             = fmt.Sprintf("^%s-", common.LegacyControlPlaneVMPrefix)
+	firstMasterRegexStr             = fmt.Sprintf("^%s-.*-0", common.LegacyControlPlaneVMPrefix)
```
jackfrancis, line 81: If you look hard enough you find interesting things

parts/k8s/cloud-init/artifacts/cse_helpers.sh
Michael-Sinz: Which one is priority?  The azure0 or the eth0?
If azure0 is there then eth0 should not be, then would the order be better first check azure0 and then eth0?

(Unclear from your comment)

jackfrancis: When we define the VM in the ARM template, we declare a primary IP address on a single NIC, so we expect it to be reflected in eth0. Because Azure CNI creates a bridge "azure0" interface and "takes over" the IP address from eth0, that's where we have to look (if we want to look locally and not rely upon DNS) for the primary IP address of the host.

An additional consideration is that Ubuntu 18.04-LTS's cloud-init implementation has a bug in that it doesn't respect the ARM template-declared primary IP address when the spec also includes secondary IP addresses, and so eth0 will have more than one IP address (those secondary IP addresses are for Azure CNI to use in the container networking layer, not the host layer, so eth0 should only ever have 1 IP address). To deal with *that* edge case (e.g., 18.04-LTS on the first boot, before the cloud-init network config override has been applied) we evaluate the number of IP addresses returned (the `| grep -c '^'` part), and if it's not exactly 1, then we just fallback to relying upon DNS.

parts/k8s/cloud-init/artifacts/health-monitor.sh
```diff
@@ -66,20 +87,16 @@ if [[ $# -ne 1 ]]; then
...
-SLEEP_SECONDS=10
+SLEEP_TIME=10
 component=$1
 echo "Start kubernetes health monitoring for ${component}"
 
 if [[ ${component} == "container-runtime" ]]; then
   container_runtime_monitoring
 elif [[ ${component} == "kubelet" ]]; then
   kubelet_monitoring
+elif [[ ${component} == "etcd" ]]; then
```
mboersma, line 98: We should change the `Usage:` statement a few lines above to say:

```
echo "Usage: health-monitor.sh <container-runtime|kubelet|echo>"
```

mboersma (approved):
/lgtm

PR #3792, opened 2020-09-04
fix: Azure Stack CNI network interfaces file creation fix

JesusAlvarezTorres:
**Reason for Change**:

Fixing an issue seen in Azure Stack during the creation of the network interfaces file for the Azure CNI plugin

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


scripts/collect-logs.sh
```diff
@@ -18,6 +18,12 @@ collectCloudProviderJson() {
     if [ -f /etc/kubernetes/azurestackcloud.json ]; then
         jq . /etc/kubernetes/azurestackcloud.json > ${DIR}/azurestackcloud.json
     fi
+    if [ -f /etc/kubernetes/network_interfaces.json ]; then
+        jq . /etc/kubernetes/network_interfaces.json > ${DIR}/network_interfaces.json
+    fi
```
vikasnav, line 23: This seems to copy files on Linux. What about Windows nodes. We should copy this file even on windows nodes if there is a failure

parts/k8s/cloud-init/artifacts/cse_customcloud.sh
```diff
@@ -210,7 +215,7 @@ configureAzureStackInterfaces() {
 
     if [[ -z $SUBNET_PREFIX ]]; then
       echo "Error fetching the subnet address prefix for a subnet ID"
-      exit 122
+      exit 123
```
rjaini, line 218: Dont we need to specify what is the error code mapping of 123  in https://github.com/Azure/aks-engine/blob/058b1dc164206de2b0641e73f34150ed447f300a/pkg/engine/cse.go ?.

JesusAlvarezTorres: yes, @jadarsie pointed this out, I will include this on the next commit.

JesusAlvarezTorres: change made

scripts/collect-logs.sh
JesusAlvarezTorres: change made

scripts/collect-windows-logs.ps1
```diff
@@ -33,6 +33,22 @@ if (Test-Path $containerd) {
...
 
+$networkInterfaces = "c:\k\network-interfaces.json"
+if (Test-Path $networkInterfaces) {
+  $tempFile = Copy-Item $networkInterfaces $lockedTemp -PassThru -ErrorAction Ignore
+  if ($tempFile) {
+    $paths += $tempFile
+  }
+}
+
+$interfaces = "c:\k\interfaces.json.json"
```
jadarsie, line 44: extra `.json`

jadarsie (dismissed):

scripts/collect-windows-logs.ps1
JesusAlvarezTorres: good catch, fixing

JesusAlvarezTorres: change made

jadarsie (dismissed):

parts/k8s/cloud-init/artifacts/cse_customcloud.sh
```diff
@@ -195,7 +195,12 @@ configureAzureStackInterfaces() {
 
   mapfile -t local_interfaces < <(cat /sys/class/net/*/address | tr -d : | sed 's/.*/\U&/g')
 
-  SDN_INTERFACES=$(jq ".value | map(select(.properties.macAddress | inside(\"${local_interfaces[*]}\"))) | map(select((.properties.ipConfigurations | length) > 0))" ${NETWORK_INTERFACES_FILE})
+  SDN_INTERFACES=$(jq ".value | map(select(.properties != null) | select(.properties.macAddress != null) | select(.properties.macAddress | inside(\"${local_interfaces[*]}\"))) | map(select((.properties.ipConfigurations | length) > 0))" ${NETWORK_INTERFACES_FILE})
```
jackfrancis, line 198: How is this additional null protection actually fixing anything?

JesusAlvarezTorres: there is an underlying issue on upgrade where the upgrade for Windows nodes is creating an extra nic. The nic is not attached to a VM thus it doesnt have a macAddress property. Without this null check for macAddress jq throws an error which prevents the configuration file to be created

jackfrancis: So basically this change filters out all interfaces that have a null properties or properties.macAddress value, and in practice the right interface is selected?

How are we addressing the underlying issue of an unexpected additional NIC?

JesusAlvarezTorres: This change is a defensive fix to unblock the CNI scenario for Azure Stack on upgrade, I am still investigating the issue for the additional NIC, will send a separate PR for it once I've figured it out.

jackfrancis (dismissed):
lgtm for linux fixes

@marosset and @jsturtevant to review the powershell additions

jackfrancis:
ping again @marosset @jsturtevant

scripts/collect-windows-logs.ps1
```diff
@@ -33,6 +33,22 @@ if (Test-Path $containerd) {
...
+}
+
+$interfaces = "c:\k\interfaces.json"
+if (Test-Path $interfaces) {
+  $tempFile = Copy-Item $interfaces $lockedTemp -PassThru -ErrorAction Ignore
+  if ($tempFile) {
+    $paths += $tempFile
+  }
+}
+
```
marosset, line 51: Could we do something like
$files = @( 
    "c:\"C:\ProgramData\containerd\root\panic.log",
     "c:\k\network-interfaces.json", 
      "c:\k\interfaces.json")

$files | Foreach-Object {
if (Test-Path $_) {
    $tempFile = Copy-Item $_ $lockedTemp -PassThru -ErrorAction Ignore
    if ($tempFile) {
      $paths += $tempFile
    }
  }
}

To reduce code duplication?

marosset: or just add interfaces.json and network-interfaces.json to $lockedFiles above...
the panic.log was special cased b/c it was outside of c:\k

jsturtevant: adding the files to $lockedfiles would be the best, they will be picked up automatically

JesusAlvarezTorres: thanks for the feedback, change made

PR #3799, opened 2020-09-08
chore: check VHD media name length before being published to Marketplace

mainred:
**Reason for Change**:

Maketplace has a VHD media name length, which should be under 63 characters.
This change check the media name length before the vhd is published to Marketplace.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/publishing/add-image-version-to-sku.sh
+++ b/vhd/publishing/add-image-version-to-sku.sh
@@ -52,7 +52,7 @@ image_version="${vhd_version}.${version_date}"
 # generate media name
 # Media name must be under 63 characters
 sku_prefix=$(< $SKU_INFO jq -r ".sku_prefix")
-media_name="${sku_prefix}-${image_version}"
+media_name="aks-windows-${sku_prefix}-${image_version}"
 if [ "${#media_name}" -ge 63 ]; then
 	echo "$media_name should be under 63 characters"
 	exit 1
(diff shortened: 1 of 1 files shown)


vhd/publishing/add-image-version-to-sku.sh
```diff
@@ -50,8 +50,13 @@ version_date=$(date +"%y%m%d")
 image_version="${vhd_version}.${version_date}"
 
 # generate media name
+# Media name must be under 63 characters
 sku_prefix=$(< $SKU_INFO jq -r ".sku_prefix")
-media_name="aks-windows-${sku_prefix}-${image_version}"
```
jsturtevant, line 54: Is dropping the `aks-windows` from the media name intentional?  To confirm, this doesn't affect sku name?

mainred: > Is dropping the aks-windows from the media name intentional?

I restored the media name prefix. That is used for AKS now, because we have extended sku-prefix, and we have plan id and sku id as another level of scopes, so we don't have to  try the limit of media name length.

>  To confirm, this doesn't affect sku name?

sku name is not affected

jsturtevant:
ltgm 

/assign @marosset 
for VHD script changes

marosset (approved):

PR #3801, opened 2020-09-08
chore: targeting sept updates for Windows 2019 VHD

marosset:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/packer/configure-windows-vhd.ps1
+++ b/vhd/packer/configure-windows-vhd.ps1
@@ -186,9 +186,8 @@ function Install-WindowsPatches {
     # Windows Server 2019 update history can be found at https://support.microsoft.com/en-us/help/4464619
     # then you can get download links by searching for specific KBs at http://www.catalog.update.microsoft.com/home.aspx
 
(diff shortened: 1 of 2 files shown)


vhd/packer/configure-windows-vhd.ps1
```diff
@@ -188,7 +188,7 @@ function Install-WindowsPatches {
 
     # KB4558998 contains August 11, 2020 cumulative updates for Windows Server 2019
     # https://www.catalog.update.microsoft.com/Search.aspx?q=KB4565349
```
mainred, line 190: Maybe we may just change this comment into a template instead for each specific patch.

mainred (dismissed):
The patch URL and image version look good to me. But we need to change the patch comment, approve this PR as I think you'll fix it.

vhd/packer/configure-windows-vhd.ps1
marosset: Will update

marosset: Updated comment, thanks!

Issue #3753, opened 2020-08-26
Migrate all container images to MCR


jackfrancis:
AKS Engine currently refers to these dockerhub-hosted images in VHD CI:

- microsoft/virtual-kubelet
- registry

And currently delivers these dockerhub-hosted images via addons:

- calico/typha:v3.8.0
- calico/cni:v3.8.0
- calico/node:v3.8.0
- calico/pod2daemon-flexvol:v3.8.0
- docker.io/cilium/cilium:v1.4
- docker.io/cilium/cilium-init:2018-10-16
- docker.io/cilium/operator:v1.4
- docker.io/cilium/cilium-etcd-operator:v2.0.5
- antrea/antrea-ubuntu:v0.6.0
- microsoft/virtual-kubelet

sozercan:
These should be available in MCR today:

```
calico/typha:v3.8.0
calico/cni:v3.8.0
calico/node:v3.8.0
calico/pod2daemon-flexvol:v3.8.0
microsoft/virtual-kubelet
```

jackfrancis:
@jadarsie does Azure Stack still need the registry container image? As long as Azure Stack users always use the curated VHD they should be calling to dockerhub.

But if the need is still there, is there a plan to put it into MCR?

jackfrancis:
Closing this for now, as all blessed configurations now have dependent container images hosted in MCR.

jadarsie:
fyi @rjaini

rjaini:
@jackfrancis : As of now we still need the registry image in the VHD for disconnected case. Let me check and discuss once again with Azure Stack PM and get back to you. After then we can decide if we need to move the image to MCR or not.

PR #3803, opened 2020-09-08
chore: get calico and v-k from mcr.microsoft.com

mboersma:
**Reason for Change**:

Moves references to the calico and virtual-kubelet components to mcr.microsoft.com, where they are already published.

**Issue Fixed**:

Refs #3753

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

I `docker pull`ed all these images to make sure they're actually published to MCR.

In Mooncake, AKS Engine uses `dockerhub.azk8s.cn/calico/` as the base for Calico images and `dockerhub.azk8s.cn/microsoft/` as the base for virtual-kubelet. Are there parallel replacements at `mcr.azk8s.cn`?


jackfrancis (approved):
/lgtm

thanks @mboersma!

vhd/packer/install-dependencies.sh
```diff
@@ -285,7 +285,7 @@ done
 
 VIRTUAL_KUBELET_VERSIONS="latest"
 for VIRTUAL_KUBELET_VERSION in ${VIRTUAL_KUBELET_VERSIONS}; do
-    CONTAINER_IMAGE="microsoft/virtual-kubelet:${VIRTUAL_KUBELET_VERSION}"
+    CONTAINER_IMAGE="mcr.microsoft.com/oss/virtual-kubelet/virtual-kubelet:${VIRTUAL_KUBELET_VERSION}"
```
sozercan, line 288: can we pin this? cc @cpuguy83

mboersma: It's always floated at `:latest`, but we could use `:0.9.1`. They point to the [same image](https://hub.docker.com/r/microsoft/virtual-kubelet/tags) and haven't been updated in a year.

cpuguy83: The hub image is not able to be updated anymore (per policy), which is why it was moved to ACR.
Honestly, I think the virtual-node add-on is using 'latest' as well 😞

sozercan: Looks like `microsoft/virtual-kubelet` registry is deprecated. published `v1.2.1.2` to MCR.

mboersma: I think pinning virtual-kubelet to a recent version is logically separate from this PR, so I'm going to merge this and we can follow up on #3810.

sozercan (approved):

PR #3806, opened 2020-09-09
feat: Generate ARM template for Windows agent pool for different VHD images

AbelHu:
**Reason for Change**:

feat: Generate ARM template for Windows agent pool for different VHD images

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/virtualmachines.go
+++ b/pkg/engine/virtualmachines.go
@@ -498,7 +498,25 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 	storageProfile := compute.StorageProfile{}
 
 	if profile.IsWindows() {
-		storageProfile.ImageReference = createWindowsImageReference(profile.Name, cs.Properties.WindowsProfile)
+		imageRef := profile.ImageRef
+		if profile.HasImageRef() {
+			if profile.HasImageGallery() {
+				storageProfile.ImageReference = &compute.ImageReference{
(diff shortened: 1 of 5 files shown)


pkg/engine/virtualmachines.go
```diff
@@ -498,7 +498,25 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 	storageProfile := compute.StorageProfile{}
 
 	if profile.IsWindows() {
-		storageProfile.ImageReference = createWindowsImageReference(profile.Name, cs.Properties.WindowsProfile)
```
mainred, line 501: Should we add these into `createWindowsImageReference` instead?

pkg/engine/virtualmachinescalesets.go
```diff
@@ -665,7 +665,27 @@ func CreateAgentVMSS(cs *api.ContainerService, profile *api.AgentPoolProfile) Vi
 	vmssStorageProfile := compute.VirtualMachineScaleSetStorageProfile{}
 
 	if profile.IsWindows() {
-		vmssStorageProfile.ImageReference = createWindowsImageReference(profile.Name, cs.Properties.WindowsProfile)
+		if profile.HasImageRef() {
```
mainred, line 668: ditto

pkg/engine/virtualmachines.go
AbelHu: I planned to remove createWindowsImageReference  later. I think that keeping the logic here as Linux does makes it more readable.

mainred: But it leaves `createWindowsImageReference` in a wrong logic, I prefer we leave a TODO comment to fix it later.  It depends on you.

mainred (approved):

PR #3818, opened 2020-09-11
feat: Add T4 skus for samsung

xuto2:
**Reason for Change**:

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma:
@xuto2 should this change have also gone into AKS Engine master?

xuto2:
@mboersma This is for T4 which is still in preview. we just need them for Samsung for a urgent ask. Not sure if AKS-E wants that

mboersma:
Ah...probably not if the SKUs aren't GA, although we should keep an eye on them. Thanks for the clarification!

PR #3822, opened 2020-09-14
fix: clean up stale CNI data regardless of HNS state

ashvindeodhar:
fix: cleanup stale CNI data regardless of HNS state
Currently CNI stale data is cleaned up only when a stale HNS network is found.
This change removes that unnecessary dependency and cleans up the stale CNI data
regardless of the HNS state. This is needed when HNS network is not persisted across the reboots.

**Reason for Change**:

**Issue Fixed**:

Fixes https://github.com/Azure/azure-container-networking/issues/642

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -216,6 +216,7 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.19.0-rc.3":    false,
 	"1.19.0-rc.4":    false,
 	"1.19.0":         true,
+	"1.19.1":         true,
 }
 
 // AllKubernetesSupportedVersionsAzureStack is a hash table of all supported Kubernetes version strings on Azure Stack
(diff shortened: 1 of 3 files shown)


jackfrancis:
Thanks @ashvindeodhar, is this meant to be a Windows-only change? There is no equivalent Linux change to benefit those scenarios?

ashvindeodhar:
@jackfrancis yes this is for windows-only. There is no change for linux scenario since linux removes all the CNI stale data upon reboot.

marosset:
@AbelHu @mainred as FYI

ashvindeodhar:
@marosset / @jackfrancis Do you know which AKS release will this get into? I am looking to get some idea as to when (ETA) will this is available to the customers.

jackfrancis:
@xuto2 are you able to answer @ashvindeodhar's question?

xuto2:
@AbelHu Can you comment on the above question? It's an issue for windows only.

AbelHu:
@ashvindeodhar I think that aks-e will sign this script and publish a new Windows signed binary. After that, AKS can pick up it. Please let me know when the new signed binary is ready. @marosset. Thanks.

mainred:
Want to confirm from @ashvindeodhar, we can resolve https://github.com/Azure/azure-container-networking/issues/642 by only this PR, right?
If so, Windows signed binary for CNI is required, but we need to include changes in this PR to AKS used branch.

zhiweiv:
Just FYI, the issue in AKS is https://github.com/Azure/AKS/issues/1772, it is great to update the ETA there.

PR #3830, opened 2020-09-15
feat: update VMSS node pools

jackfrancis:
**Reason for Change**:

This PR introduces the new CLI feature `aks-engine update --node-pool <node pool name>` to update VMSS node pools on existing cluster via AKS Engine ARM template generation.

tl;dr With a new version of AKS Engine compared to the version used to build your cluster, and/or with changes to the agentPoolProfiles definition in the generated `apimodel.json` output, you use `aks-engine update` to essentially push a new VMSS model to your node pool. Subsequent scale out operations against that VMSS will therefore use the updated VMSS model.

Some real-world examples:

- change the VM SKU
- change the OS distro
- change other AKS Engine-exposed OS configuration (e.g., sysctld)

In practice the implementation essentially re-uses `aks-engine scale`, giving that preëxisting command the right inputs to do an ARM template deployment against an existing VMSS with no change to the `Count` value.

**Issue Fixed**:

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/update.go
+++ b/cmd/update.go
@@ -1,0 +1,284 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strconv"
+
+	"github.com/Azure/aks-engine/pkg/api"
+	"github.com/Azure/aks-engine/pkg/armhelpers"
+	"github.com/Azure/aks-engine/pkg/engine"
+	"github.com/Azure/aks-engine/pkg/helpers"
+	"github.com/Azure/aks-engine/pkg/i18n"
+	"github.com/leonelquinteros/gotext"
+	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	prefixed "github.com/x-cray/logrus-prefixed-formatter"
+	v1 "k8s.io/api/core/v1"
+)
+
+type updateCmd struct {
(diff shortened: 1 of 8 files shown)


jackfrancis:
Worth mentioning: detailed documentation will follow this PR (as part of a larger set of improved documentation around all CLI functionality)

cmd/update.go
```diff
@@ -0,0 +1,260 @@
...
+				if !sc.vmInAgentPool(vmssName, vmss.Tags) {
+					continue
+				}
+			}
+
+			if vmss.Sku != nil {
+				sc.newDesiredAgentCount = int(*vmss.Sku.Capacity)
+				uc.agentPool.Count = sc.newDesiredAgentCount
+			} else {
+				return errors.Wrap(err, fmt.Sprintf("failed to detect find VMSS matching node pool %s in resource group %s", sc.agentPoolToScale, sc.resourceGroupName))
```
mboersma, line 249:
```suggestion
				return errors.Wrap(err, fmt.Sprintf("failed to find VMSS matching node pool %s in resource group %s", sc.agentPoolToScale, sc.resourceGroupName))
```

mboersma (commented):
This makes sense to me and should help users quite a bit.

We should update the `aks-engine help` example output in docs/community/developer-guide.md to include this command.

jackfrancis:
> This makes sense to me and should help users quite a bit.
> 
> We should update the `aks-engine help` example output in docs/community/developer-guide.md to include this command.

Done

mboersma (approved):
/lgtm

PR #3835, opened 2020-09-16
docs: remove mention about library for AKS

jackfrancis:
**Reason for Change**:

This PR removes the mention in the landing page README.md about AKS Engine being a library for AKS.

Historically, this was true, but more and more any AKS Engine usage by AKS is more or less piecemeal, and not obvious to the public, so we should stop suggesting otherwise.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ AKS Engine is the easiest way to provision a self-managed Kubernetes cluster on 
 
 ## Overview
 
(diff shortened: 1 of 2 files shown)


jackfrancis:
cc @craiglpeters @palma21

mboersma:
In the FAQ we also say:

> AKS uses AKS Engine internally, but they are not the same.

I assume that language is vague and correct enough to remain intact.

jackfrancis:
@mboersma good find! No, I don't think we should say that either, updated the FAQ language.

mboersma (approved):
/lgtm

PR #3847, opened 2020-09-18
feat: Updating Windows VHD build files to support building for multiple OS versions

marosset:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder-windows.yaml
+++ b/.pipelines/vhd-builder-windows.yaml
@@ -40,6 +40,7 @@ jobs:
       -e BUILD_ID=$(Build.BuildId) \
       -e BUILD_NUMBER=$(Build.BuildNumber) \
       -e CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) \
+      -e WINDOWS-SERVER-VERSION=2019 \
       ${DEIS_GO_DEV_IMAGE} make run-packer-windows
     displayName: Building windows VHD
 
@@ -49,7 +50,7 @@ jobs:
       -w /go/src/github.com/Azure/aks-engine \
       ${DEIS_GO_DEV_IMAGE} make all
     displayName: build aks-engine-test - make all
-    condition: eq(variables.COPY_VHD, 'False')
+    condition: and(succeeded(), eq(variables.COPY_VHD, 'False'))
 
   - script: |
       OS_DISK_URI="$(cat packer-output | grep "OSDiskUri:" | cut -d " " -f 2)" && \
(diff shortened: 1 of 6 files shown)


marosset:
I decided to not convert the Windows vhd packer file to use the HashiCorp Config Langauge becasue this is a smaller set of changes that accomplishes the same goals.

vhd/packer/windows-vhd-builder.json
```diff
@@ -23,12 +28,12 @@
...
-            "image_publisher": "MicrosoftWindowsServer",
-            "image_offer": "WindowsServer",
-            "image_sku": "2019-Datacenter-Core-smalldisk",
-            "image_version": "17763.1339.2007101755",
+            "image_publisher": "{{user `image_publisher`}}",
+            "image_offer": "{{user `image_offer`}}",
+            "image_sku": "{{user `image_sku`}}",
+            "image_version": "{{user `image_version`}}",
             "resource_group_name": "{{user `resource_group_name`}}",
-            "capture_container_name": "aksengine-vhds-windows-ws2019",
```
CecileRobertMichon, line 31: this will change from `ws2019` to `2019` by default, is that okay?

marosset: I think it should be ok. This is only used in the paths in the storage account the VHD gets captured into correct?

marosset:
Let me revert chagnes tot he scale.go file.. not sure how those got picked up...

CecileRobertMichon:
it's a github beta feature `Unchanged files with check annotations`, no changes were pushed, it's just showing them to reviewers

marosset:
First time I saw that. Thanks.

vhd/packer/windows-vhd-builder.json
CecileRobertMichon: yeah should be fine, we get the URI directly from the output so we don't expect it to be in a specific place 👍

CecileRobertMichon (approved):
the packer/pipeline parts lgtm

.pipelines/vhd-builder-windows.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
       -e BUILD_ID=$(Build.BuildId) \
       -e BUILD_NUMBER=$(Build.BuildNumber) \
       -e CONTAINER_RUNTIME=$(CONTAINER_RUNTIME) \
+      -e WINDOWS-SERVER-VERSION=2019 \
```
jsturtevant, line 43: should 2019 be an environment variable so it can be configured on the job?

marosset: Possibly. We still need to have a discussion around what images aks-engine should actually be producing. If we want to produce more than just Windows Server 2019 + Docker I think I would rather just update the pipepline to build all the images as different stages and avoid the need to trigger multiple pipelines.
I'm hoping we can resolve that with a different PR tho.

PR #2583, opened 2020-01-15
fix: create role assignment for master VMs

dennis-benzinger-hybris:
Fixes #2373 
Fixes #2367

PR #3837, opened 2020-09-17
docs: CLI operations

jackfrancis:
**Reason for Change**:

This PR thoroughly documents the CLI operations provided by AKS Engine.

**Issue Fixed**:

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [x] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3857, opened 2020-09-23
fix: de-dup vnet roleAssignment deployments

### LINKED ISSUE #3863: system-assigned identity broken in multiple node pools w/ common VNET scenario
**Describe the bug**

If you're running a cluster configuration that includes the following:

1) multiple node pools
2) bring your own VNET
3) at least 2 of the node pools share the same VNET

Then you can't use system-assigned identity for authenticating to Azure, because there is foo in the code base that creates a per-master VM --> VNET role assignment *per node pool*. That foo assumes that each pool is using a discrete VNET, in which case you *do* need the per node pool role assignment to ensure that your control plane and node pool VMs are in a common security domain in the event that they are all in discrete VNETs.

In essence, the current implementation allows control plane VMs and node pool VMs to be flexibly assigned to discrete VNETs, and to use system-assigned identity properly. But the code *does not* allow you to do the same with your control plane and node pools residing in the *same* VNET. This PR changeset demonstrates the bug:

https://github.com/Azure/aks-engine/pull/3861

**Steps To Reproduce**

**Expected behavior**

**AKS Engine version**

**Kubernetes version**

**Additional context**

jackfrancis:
**Reason for Change**:

This PR addresses and edge case resulting from the "master VM roleAssignment" change introduced in #2583. Specifically, we want to create a VNET roleAssignment  per master VM *for each* VNET referenced by any node pool. In the event that we are sharing a common VNET among several node pools, we want to ensure that we only create one Deployment resource per VNET.

The CI jobs in this repro PR demonstrate that system-assigned identity is currently broken w/ multiple node pools + custom VNET:

https://github.com/Azure/aks-engine/pull/3861

In addition, this PR *only* includes the above roll assignment `Deployment` resources if the master VNET is in fact distinct as compared to any one node pool VNET.

Also, we standardize to the `2018-09-01-preview` API version for defining system-assigned ID-derived role assignments, so that we can declare the `"principalType": "ServicePrincipal"` configuration.

Also includes changes to the cluster configuration used to test PRs so that we're testing system-assigned identity-enabled clusters regularly.

**Issue Fixed**:

Fixes #3863 

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @dennis-benzinger-hybris

pkg/engine/systemroleassignments.go
```diff
@@ -88,13 +90,24 @@ func createKubernetesMasterRoleAssignmentForAgentPools(masterProfile *api.Master
...
-	for agentPoolIdx, agentPool := range agentPoolProfiles {
+	found := []string{}
+	for _, agentPool := range agentPoolProfiles {
 		var roleAssignments = make([]interface{}, masterProfile.Count)
+		subnetElements := strings.Split(agentPool.VnetSubnetID, "/")
+		if len(subnetElements) < 9 {
+			continue
+		}
+		vnet := strings.Join(subnetElements[:9], "/")
+		if helpers.IsStringInStringSlice(vnet, found) {
```
CecileRobertMichon, line 106: instead of having to loop through the string slice everytime you could use a `map[string]struct{}` which is more performant and doesn't require the addionnal helper func, ie. https://www.davidkaya.com/sets-in-golang/

jackfrancis: Thanks for the optimization! Done

test/e2e/azure/cli.go
```diff
@@ -278,16 +279,27 @@ func (a *Account) CreateDeployment(name string, e *engine.Engine) error {
...
+		if bytes.Contains(out, []byte("PrincipalNotFound")) {
+			for err != nil {
+				cmd = exec.Command("az", azArgsStringSlice...)
+				util.PrintCommand(cmd)
+				out, err = cmd.CombinedOutput()
+				if err != nil {
+					log.Printf("Command Output: %s\n", out)
+				}
+			}
+		}
```
jackfrancis, line 302: Here's an example of the result of this change:

```
$ az deployment group create --name kubernetes-eastus-37551 --resource-group kubernetes-eastus-37551 --template-file /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-eastus-37551/azuredeploy.json --parameters /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-eastus-37551/azuredeploy.parameters.json
.......2020/09/23 17:26:56 
Error from deployment for kubernetes-eastus-37551 in resource group kubernetes-eastus-37551:exit status 1
2020/09/23 17:26:56 Command Output: Deployment failed. Correlation ID: 0934f995-914a-47ec-a60c-cbf5774220ba. {
  "error": {
    "code": "PrincipalNotFound",
    "message": "Principal 4aaa67de8ac644c5bf234a1d069a995e does not exist in the directory 72f988bf-86f1-41af-91ab-2d7cd011db47."
  }
}


$ az deployment group create --name kubernetes-eastus-37551 --resource-group kubernetes-eastus-37551 --template-file /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-eastus-37551/azuredeploy.json --parameters /Users/jackfrancis/work/src/github.com/Azure/aks-engine/_output/kubernetes-eastus-37551/azuredeploy.parameters.json
.
```

And then the E2E execution proceeded successfully. Basically, this change enables us to functionally test system-assigned identity reliably, as we need to assume that such a config will regularly produce an error response from the initial ARM template deployment based on observed behaviors.

Additionally, documentation outlining this "redeploy your ARM template" guidance will be added to this PR:

https://github.com/Azure/aks-engine/pull/3837

pkg/engine/systemroleassignments.go
```diff
@@ -88,13 +92,35 @@ func createKubernetesMasterRoleAssignmentForAgentPools(masterProfile *api.Master
...
+	// We're gonna keep track of distinct VNETs in use across all our node pools;
+	//   we only want to define master VM --> VNET role assignments once per VNET.
+	// If our cluster configuration includes more than one pool sharing a common VNET,
+	//   we define the master VM --> VNET role assignments (one per master VM) just once for those pools
+	var vnetInCluster = struct{}{}
+	vnets := make(map[string]struct{})
+	for _, agentPool := range agentPoolProfiles {
 		var roleAssignments = make([]interface{}, masterProfile.Count)
+		subnetElements := strings.Split(agentPool.VnetSubnetID, "/")
+		// We expect a very specific string format for the VnetSubnetID property;
```
mboersma, line 110: We could instead use a capturing regex here (and the other places we're splitting into elements by `/`). Could be safer and more self-documenting. 

Happy to help with that here or in a follow-on PR if you agree.

mboersma (approved):
/lgtm

pkg/engine/systemroleassignments.go
jackfrancis: What's that regex look like? I think it'll definitely be more concise, less convinced about more readable/self-documenting, but maybe that's a lack of imagination on my part.

mboersma: Something like:

```regex
(?P<BASE>\/subscriptions\/.*\/resourceGroups\/.*\/providers\/Microsoft\.Network\/virtualNetworks\/.*\/subnets)\/.*
```

But wow I forgot about forward slashes needing escaping...ick. And go requires a few lines of code to extract a named capturing expression. So maybe never mind.

jackfrancis: Right, I actually do think that's maybe slightly better, but let's merge and see how annoyed we get with maintaining it in this more brute force way.

PR #3862, opened 2020-09-23
fix: add #EOF sentinel chars to custom search file

jackfrancis:
**Reason for Change**:

This PR adds the needed `#EOF` sentinel chars to the custom search domains script so that the CSE file wait func will properly detect its arrival on disk via cloud-init.

**Issue Fixed**:

Fixes #3859 

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [ ] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #3867, opened 2020-09-24
fix: adds an apt update to custom search file #3862

### LINKED ISSUE #3859: The customSearchDomain is not functional
**Describe the bug**
I have the problem that I would like to use a custom DNS, so I enter the variables as desired

```bash
            "customSearchDomain": {
                "name": "REDACTED.tld",
                "realmUser": "svc.K8sDNS",
                "realmPassword": "REDACTED"
            },
            "customNodesDNS": {
                "dnsServer": "$IPofcustomNodesDNS"
            }
```

I get the fallowing error from the log:

+ grep -Fq '#EOF' /opt/azure/containers/setup-custom-search-domains.sh
+ '[' 3600 -eq 3600 ']'
+ return 1
+ exit 6

This is caused by the ./provision_source.sh:wait_for_file()
```bash
wait_for_file() {
  retries=$1; wait_sleep=$2; filepath=$3
  paved=/opt/azure/cloud-init-files.paved
  grep -Fq "${filepath}" $paved && return 0
    grep -Fq '#EOF' $filepath && break
    if [ $i -eq $retries ]; then
      return 1
    else
      sleep $wait_sleep
    fi
  done
  sed -i "/#EOF/d" $filepath
  echo $filepath >>$paved
}
```

The script assumes that the script [parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh](https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh) contains a #EOF. But this is not the case here, so the script is not included in /opt/azure/cloud-init-files.paved on the server and the corresponding abort occurs. If i run setup-custom-search-domains.sh manually, the server are joining the realm.

**Steps To Reproduce**
```json
{
    "apiVersion": "vlabs",
    "location": "",
    "properties": {
        "orchestratorProfile": {
            "orchestratorRelease": "1.17",
            "orchestratorVersion": "1.17.5",
            "kubernetesConfig": {
                "cloudProviderBackoff": true,
                "cloudProviderBackoffRetries": 1,
                "cloudProviderBackoffDuration": 30,
                "cloudProviderRateLimit": true,
                "cloudProviderRateLimitQPSWrite": 3,
                "cloudProviderRateLimitBucketWrite": 10,
                "kubernetesImageBase": "mcr.microsoft.com/k8s/azurestack/core/",
                "useInstanceMetadata": false,
                "networkPlugin": "kubenet",
                "kubeletConfig": {
                    "--node-status-update-frequency": "1m",
                    "--kube-reserved": "memory=500Mi",
                    "--system-reserved": "memory=500Mi",
                    "--eviction-hard": "memory.available<500Mi"
                },
                "controllerManagerConfig": {
                    "--node-monitor-grace-period": "5m",
                    "--pod-eviction-timeout": "5m",
                    "--route-reconciliation-period": "1m"
                },
                "privateCluster": {
                    "enabled": true
                },
                "etcdDiskSizeGB": "32"
            }
        },
        "customCloudProfile": {
            "portalURL": "https://azurestackhub/"
        },
        "featureFlags": {
            "enableTelemetry": true
        },
        "masterProfile": {
            "dnsPrefix": "mas",
            "distro": "aks-ubuntu-16.04",
            "count": 3,
            "vmSize": "Standard_DS2_v2",
            "vnetSubnetId": "/subscriptions/...REDACTED.../subnets/kubernetes-master-Subnet",
            "firstConsecutiveStaticIP": "..REDACTED...",
            "customVMTags": {
                "Application": "k8s Test"
            }
        },
        "agentPoolProfiles": [
            {
                "name": "worker",
                "count": 3,
                "vmSize": "Standard_DS2_v2",
                "distro": "aks-ubuntu-16.04",
                "availabilityProfile": "AvailabilitySet",
                "AcceleratedNetworkingEnabled": false,
                "vnetSubnetId": "/subscriptions/...REDACTED.../subnets/kubernetes-agent-Subnet",
                "customVMTags": {
                  "Application": "k8s Test"
                }
            }
        ],
        "linuxProfile": {
            "adminUsername": "...REDACTED...",
            "ssh": {
                "publicKeys": [
                    {
                        "keyData": "...REDACTED..."
                    }
                ]
            },
            "customSearchDomain": {
                "name": "...REDACTED...",
                "realmUser": "...REDACTED...",
                "realmPassword": "...REDACTED..."
            },
            "customNodesDNS": {
                "dnsServer": "...REDACTED..."
            }
        },
        "servicePrincipalProfile": {
            "clientId": "...REDACTED...",
            "secret": "...REDACTED..."
        }
    }
}
```
aks-engine-v0.51.1 deploy -f --azure-env AzureStackCloud --api-model kubernetes-azurestack.json --location _REDACTED_ --resource-group RG-k8s-Test --client-id _REDACTED_ --client-secret _REDACTED_ --subscription-id _REDACTED_ --output-directory RG-k8s-Test

**Expected behavior**
The script should be functional so that the nodes register themselves in the AD.

**AKS Engine version**
aks-engine-v0.51.1
**Kubernetes version**
1.17.5

**Additional context**
Another thing is, thet the Names of the Nodes are much to long, for the Active-Directory Computer section. There is a limit, so the nodes are shorten to k8s-master-3500 and get all the same name, is there an way to get shorter names?

0hlov3:
**Reason for Change**:
This PR adds the needed apt update to the custom search domains script so that the packages the file expects to install, can be installed.

**Issue Fixed**:
Fixes #3859

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
+++ b/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
@@ -5,7 +5,9 @@ source {{GetCSEHelpersScriptFilepath}}
 echo "  dns-search {{GetSearchDomainName}}" | tee -a /etc/network/interfaces.d/50-cloud-init.cfg
 systemctl_restart 20 5 10 networking
 wait_for_apt_locks
-retrycmd 10 5 120 apt-get update && apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
+retrycmd 10 5 120 apt-get update
+wait_for_apt_locks
+retrycmd 10 5 120 apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
 wait_for_apt_locks
 echo "{{GetSearchDomainRealmPassword}}" | realm join -U {{GetSearchDomainRealmUser}}@$(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/) $(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/)
 #EOF
--- a/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
+++ b/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
@@ -5,9 +5,7 @@ source {{GetCSEHelpersScriptFilepath}}
 echo "  dns-search {{GetSearchDomainName}}" | tee -a /etc/network/interfaces.d/50-cloud-init.cfg
 systemctl_restart 20 5 10 networking
 wait_for_apt_locks
-retrycmd 10 5 120 apt-get update
-wait_for_apt_locks
-retrycmd 10 5 120 apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
+retrycmd 10 5 120 apt-get update && apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
 wait_for_apt_locks
 echo "{{GetSearchDomainRealmPassword}}" | realm join -U {{GetSearchDomainRealmUser}}@$(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/) $(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/)
 #EOF
--- a/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
+++ b/parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
@@ -5,6 +5,8 @@ source {{GetCSEHelpersScriptFilepath}}
 echo "  dns-search {{GetSearchDomainName}}" | tee -a /etc/network/interfaces.d/50-cloud-init.cfg
 systemctl_restart 20 5 10 networking
 wait_for_apt_locks
+retrycmd 10 5 120 apt-get update
+wait_for_apt_locks
 retrycmd 10 5 120 apt-get -y install realmd sssd sssd-tools samba-common samba samba-common python2.7 samba-libs packagekit
 wait_for_apt_locks
 echo "{{GetSearchDomainRealmPassword}}" | realm join -U {{GetSearchDomainRealmUser}}@$(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/) $(echo "{{GetSearchDomainName}}" | tr /a-z/ /A-Z/)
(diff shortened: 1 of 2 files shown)


0hlov3:
I have to correct an issue first. Its not functional

0hlov3:
It should work with that changes. I have another issue and opened a BUG report.
But the script is working as apected with that apt-get update.

parts/k8s/cloud-init/artifacts/setup-custom-search-domains.sh
```diff
@@ -5,7 +5,9 @@ source {{GetCSEHelpersScriptFilepath}}
 echo "  dns-search {{GetSearchDomainName}}" | tee -a /etc/network/interfaces.d/50-cloud-init.cfg
 systemctl_restart 20 5 10 networking
 wait_for_apt_locks
+retrycmd 10 5 120 apt-get update
```
jackfrancis, line 8: Hi @0hlov3, why do we need to run `apt-get update` before installing the packages below? This script doesn't modify anything in the system apt configuration, so we shouldn't have any need to run `apt-get update`.

If there *is* a reason to do so, we can actually use the convenience func `apt_get_update` instead of passing in `apt-get update` to `retrycmd`. Doing so will do the proper lock handling for us.

Also, while we're here, we should probably change the `apt-get install` implementation to follow the usages in `installDeps`. So, basically, enumerate through an array of packages and call `apt_get_install` against each of them.

This is slightly slower, but more resilient, including lock handling. See:

https://github.com/Azure/aks-engine/blob/v0.56.0/parts/k8s/cloud-init/artifacts/cse_install.sh#L36
https://github.com/Azure/aks-engine/blob/v0.56.0/parts/k8s/cloud-init/artifacts/cse_install.sh#L53

0hlov3: Hi @jackfrancis,

I was surprised that I have to do an apt-get update, because there are no changes made to the apt config, but if apt tries to install the packages, single packages can't be found by apt, so apt aborts with the message that packages were not found. 

Unfortunately I had to get my POC running today and couldn't go into deeper debugging, but tomorrow I can take a closer look at, why the package installation doesn't work if I don't do an apt update. 
Then I'll copy the error messages out again so that we can have a look at it.

Yes we should use the convenience func apt_get_update and yes we can change it to follow the usages in installDeps, I think that could work better.

If you want, you can close this Pull request and i create a new one after i took a closer look, or we can let this open and i'll check it tomorrow and update the pull request.

Thanks a lot. =)

jackfrancis: Feel free to update this pull request when you have more confidence in the right approach, no rush! And thank you!

PR #3870, opened 2020-09-25
fix: Support ImageRef in WindowsProfile

AbelHu:
**Reason for Change**:

Support ImageRef and CustomImage in WindowsProfile.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/engine/params.go
+++ b/pkg/engine/params.go
@@ -216,9 +216,16 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
 			// Set ImageRef if it is not nil and always set the Windows VHD information in WindowsProfile.
 			// ImageRef will be used to generate ARM template for the agent pool if it is set.
 			// Otherwise, the Windows VHD information in WindowsProfile will be used to generate ARM template.
+			// Priority:
+			//   1. ImageRef in agent pool
+			//   2. ImageRef in WindowsProfile
+			//   3. PIR image in WindowsProfile
 			if agentProfile.ImageRef != nil {
 				addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), agentProfile.ImageRef.Name)
 				addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), agentProfile.ImageRef.ResourceGroup)
+			} else if properties.WindowsProfile.HasImageRef() {
+				addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), properties.WindowsProfile.ImageRef.Name)
+				addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), properties.WindowsProfile.ImageRef.ResourceGroup)
 			}
 			addValue(parametersMap, fmt.Sprintf("%sosImageOffer", agentProfile.Name), properties.WindowsProfile.WindowsOffer)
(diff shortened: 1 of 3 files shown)


mainred (approved):
LGTM, but 
1. we need @xuto2 ' approval.
2. Considering the complicated logic and string concatenation, I'd suggest we have some test to verify this change and for future protection.
3. We have similar logic for vmss and vm, we can scope the common operations into a function. But considering the future support of only vmss(Please correct me if I am wrong), we can keep the code like this.

pkg/engine/virtualmachines.go
```diff
@@ -498,8 +498,14 @@ func createAgentAvailabilitySetVM(cs *api.ContainerService, profile *api.AgentPo
 	storageProfile := compute.StorageProfile{}
 
 	if profile.IsWindows() {
-		imageRef := profile.ImageRef
+		// Priority:
+		//   1. ImageRef in agent pool
+		//   2. Custom image in WindowsProfile. AKS does not use this for now but it may use this later for testing.
+		//   3. ImageRef in WindowsProfile
+		//   4. PIR image in WindowsProfile
+		windowsProfile := cs.Properties.WindowsProfile
```
xuto2, line 506: the rule is a bit complicated, why do we need to use both agent pool and windows profile? can we not combine them into one?

pkg/engine/params.go
```diff
@@ -216,9 +216,16 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
...
+			//   1. ImageRef in agent pool
+			//   2. ImageRef in WindowsProfile
+			//   3. PIR image in WindowsProfile
 			if agentProfile.ImageRef != nil {
 				addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), agentProfile.ImageRef.Name)
 				addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), agentProfile.ImageRef.ResourceGroup)
+			} else if properties.WindowsProfile.HasImageRef() {
+				addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), properties.WindowsProfile.ImageRef.Name)
+				addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), properties.WindowsProfile.ImageRef.ResourceGroup)
 			}
```
xuto2, line 229: why do we have to use this? Is agentProfile.ImageRef not enough?

pkg/engine/virtualmachines.go
AbelHu: Current code is using ImageRef in WindowsProfile. To support more Windows image, for example, 2004, we need to support multiple VHD images in the same cluster so we need to use ImageRef (SIG uses) in agent pool. Later the VHD info in WindowsProfile will be deprecated.

pkg/engine/params.go
AbelHu: I found that current code in AKS is using WindowsProfile.ImageRef. If some regions are using this, it will fail. I need the confirmation from @xizhamsft

AbelHu: @xuto2 WindowsProfile.ImageRef is used by CustomizedWindowsOSImage.
After syncing with @xizhamsft, WindowsProfile.ImageRef is not used by those regions for now. I will hold this PR.

AbelHu:
@mainred I used the SIG you provided to test customized sig and the test passed.

PR #3872, opened 2020-09-25
feat: enable flexible SSD configuration

### LINKED ISSUE #2243: I'm not able to create a cluster using Ultra SSD
**Describe the bug**

I tried to create a cluster using [Ultra SSD](https://azure.microsoft.com/en-us/blog/announcing-ultra-ssd-the-next-generation-of-azure-disks-technology-preview/) and I failed miserably.

**Steps To Reproduce**

1. Tried to spawn a cluster with **Standard LB**, masters with **AvailabilitySet**, minion with **VirtualMachineScaleSets**, with the right **availabilityZones** in a custom **VNET**. UltraSSD works only with **VirtualMachineScaleSets** 😞 
2. Tried to change masters to **VirtualMachineScaleSets** but I received this message back ```Error: loading API model in generateCmd: error parsing the api model: standard loadBalancerSku should exclude master nodes. Please set KubernetesConfig "ExcludeMasterFromStandardLB" to "true"```
3. Tried to add "**ExcludeMasterFromStandardLB**" to "true". I get ```Error: validating API model after populating values: when master profile is using VirtualMachineScaleSets and is custom vnet, set "vnetsubnetid" and "agentVnetSubnetID" for master profile```.
4. Added "**agentVnetSubnetID**". I get ```WARN[0000] Clusters with VMSS masters are not yet upgradable! You will not be able to upgrade your cluster until a future version of aks-engine!```

**AKS Engine version**

v0.43.0

**Kubernetes version**

1.15.5

jackfrancis:
**Reason for Change**:

This PR addresses remaining gaps in enabling SSD flavors across OS and data disks.

Will fix #2243

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/virtualmachines.go
```diff
@@ -229,6 +245,58 @@ func CreateMasterVM(cs *api.ContainerService) VirtualMachineARM {
...
+func createEtcdDisk(cs *api.ContainerService) DiskARM {
+	var ssdType compute.DiskStorageAccountTypes
+	if cs.Properties.MasterProfile != nil {
+		switch cs.Properties.MasterProfile.DataDiskSSDType {
+		case api.Ultra:
+			ssdType = compute.UltraSSDLRS
+		case api.Premium:
+			ssdType = compute.PremiumLRS
+		case api.Standard:
+			ssdType = compute.StandardLRS
```
CecileRobertMichon, line 257: Standard SSD should actually be `StandardSSDLRS `. `StandardLRS ` is standard HDD:

https://github.com/Azure/aks-engine/blob/b3c4f9eb7e0bd91f7b5977f8534ed96c21a1355c/vendor/github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2019-12-01/compute/models.go#L378

examples/e2e-tests/kubernetes/release/default/definition.json
```diff
@@ -79,7 +81,9 @@
         "availabilityProfile": "VirtualMachineScaleSets",
         "vnetSubnetId": "/subscriptions/SUB_ID/resourceGroups/RG_NAME/providers/Microsoft.Network/virtualNetworks/VNET_NAME/subnets/SUBNET_NAME",
         "osDiskCachingType": "ReadOnly",
-        "dataDiskCachingType": "ReadWrite"
+        "dataDiskCachingType": "ReadWrite",
+        "osDiskSSDType": "Standard",
```
CecileRobertMichon, line 85: instead of SSD type, have you considered calling this `osDiskStorageAccountType`? I would 1) be consistent with other usages across Azure and 2) allow for Standard_HDD and any other non-SSD types that may be added in the future

pkg/api/const.go
```diff
@@ -554,3 +554,10 @@ const TLSStrongCipherSuitesAPIServer = "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS
 
 // TLSStrongCipherSuitesKubelet is a kube-bench-recommended allowed cipher suites for kubelet
 const TLSStrongCipherSuitesKubelet = "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_GCM_SHA256"
+
+// SSD Types
```
CecileRobertMichon, line 558: same comment here, this could be StorageAccountTypes and we should consider using the same syntax as Azure (`Standard_LRS`, `StandardSSD_LRS`, `Premium_LRS ` etc)

pkg/api/defaults.go
```diff
@@ -732,6 +749,20 @@ func (p *Properties) setMasterProfileDefaults(isUpgrade bool) {
 	if p.MasterProfile.OSDiskCachingType == "" {
 		p.MasterProfile.OSDiskCachingType = string(compute.CachingTypesReadWrite)
 	}
+
+	if !isUpgrade {
+		if p.MasterProfile.DataDiskSSDType == "" {
+			p.MasterProfile.DataDiskSSDType = Premium
+		}
+		if p.MasterProfile.DataDiskSSDType == Ultra {
+			if p.MasterProfile.UltraSSDEnabled == nil {
```
CecileRobertMichon, line 758: what happens if `UltraSSDEnabled` is false and `DataDiskSSDType` is Ultra?

test/e2e/test_cluster_configs/everything.json
```diff
@@ -97,14 +98,16 @@
 				"availabilityZones": [
 					"1",
 					"2"
-				]
+				],
+				"ultraSSDEnabled": true
```
CecileRobertMichon, line 102: what about regions that don't support this yet? is it guaranteed that e2e won't run in those regions?

examples/e2e-tests/kubernetes/release/default/definition.json
```diff
@@ -93,10 +97,13 @@
...
         "count": 1,
-        "vmSize": "Standard_D2_v3",
+        "vmSize": "Standard_D4s_v3",
         "distro": "aks-ubuntu-16.04",
         "availabilityProfile": "VirtualMachineScaleSets",
-        "vnetSubnetId": "/subscriptions/SUB_ID/resourceGroups/RG_NAME/providers/Microsoft.Network/virtualNetworks/VNET_NAME/subnets/SUBNET_NAME"
+        "vnetSubnetId": "/subscriptions/SUB_ID/resourceGroups/RG_NAME/providers/Microsoft.Network/virtualNetworks/VNET_NAME/subnets/SUBNET_NAME",
+        "availabilityZones": ["1", "2"],
+        "osDiskSSDType": "Standard",
+        "ultraSSDEnabled": true
```
CecileRobertMichon, line 106: I think we should consider deprecating this altogether since it's redundant information and simply assume that if dataDisk type is ultraSSD it should be enabled on the VM. Is there a use case for enabling it but not using it for any disks?

pkg/api/defaults.go
jackfrancis: That will be disallowed in validation (// TODO)

test/e2e/test_cluster_configs/everything.json
jackfrancis: The test config interface supports specific regions, so yeah, we'll need to make sure it is set to only include regions that support ultra ssd for cluster configs that are testing ultra ssd (like this one)

examples/e2e-tests/kubernetes/release/default/definition.json
jackfrancis: Well, it's complicated, as it is in fact a discrete configuration for the VM (arguably, unfortunately). This configuration tells Azure to "get me a VM that is capable of attaching to an Ultra SSD disk resource". There's a different configuration to actually connect an Ultra SSD disk.

Is there a use-case for getting a ultra ssd-enabled VM, but not actually using ultra ssd for the k8s IaaS bootstrapped by AKS Engine? I don't know... I think not knowing definitively that that will never make sense combined with the fact that that config property already exists... so yeah, just working with it, so to speak.

examples/e2e-tests/kubernetes/release/default/definition.json
jackfrancis: How about just `osDiskType` and `dataDiskType`?

Just reading this doc, that would be an accurate concise way to represent this with a property name:

https://docs.microsoft.com/en-us/azure/virtual-machines/disks-types

jackfrancis: Possible values:

- "StandardHDD"
- "StandardSSD"
- "PremiumSSD"
- "UltraSSD"

CecileRobertMichon: SGTM

CecileRobertMichon: Only caveat, you might want different types per Data disk so having one global `dataDiskType` might not be the most flexible

jackfrancis: I think that inflexibility is already built into the AKS Engine config interface, as the nominal data disk config interface is simply an array of disk sizes; so the assumption is that they are all otherwise (besides size) similarly configured.

Definitely a gotcha for capz, tho

pierluigilenoci:
@jackfrancis could you please take care of this PR?

pkg/api/const.go
jackfrancis: Except that the standard Azure names are so human-offending... How does capz expose this menu?

jackfrancis:
/hold

This is stalled due to backwards incompatibility (e.g., `aks-engine upgrade`)

pierluigilenoci:
@jackfrancis any update on this?

bridgetkromhout:
Closing per @jackfrancis, given AKS Engine deprecation.

PR #3873, opened 2020-09-26
test: E2E fix scale test indexing out of bounds

JesusAlvarezTorres:
**Reason for Change**:

Fixing a test issue on E2E scale tests, the loop iterating the node pools needs to stop when the index is less than the total of agent pool profiles; currently the fix causes an additional call to scale with node poll set to null

+ (( ++i ))

+ (( i <= 1 ))

++ jq -r --arg i 1 '. | .properties.agentPoolProfiles[$i | tonumber].name'

+ nodepool=null

+ '[' '' = true ']'

Related to changes made in this PR: https://github.com/Azure/aks-engine/pull/3830

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/cluster.sh
+++ b/test/e2e/cluster.sh
@@ -328,8 +328,8 @@ if [ -n "$ADD_NODE_POOL_INPUT" ]; then
 fi
 
 if [ "${SCALE_CLUSTER}" = "true" ]; then
-  nodepools=$(jq -r  '.properties.agentPoolProfiles[].name' < _output/$RESOURCE_GROUP/apimodel.json)
-  for ((i = 0; i < ${#nodepools[@]}; ++i)); do
+  nodepoolcount=$(jq '.properties.agentPoolProfiles| length' < _output/$RESOURCE_GROUP/apimodel.json)
+  for ((i = 0; i < $nodepoolcount; ++i)); do
     nodepool=$(jq -r --arg i $i '. | .properties.agentPoolProfiles[$i | tonumber].name' < _output/$RESOURCE_GROUP/apimodel.json)
     if [ "${UPDATE_NODE_POOLS}" = "true" ]; then
       # modify the master VM SKU to simulate vertical vm scaling via upgrade
--- a/test/e2e/cluster.sh
+++ b/test/e2e/cluster.sh
@@ -329,7 +329,7 @@ fi
 
 if [ "${SCALE_CLUSTER}" = "true" ]; then
   nodepools=$(jq -r  '.properties.agentPoolProfiles[].name' < _output/$RESOURCE_GROUP/apimodel.json)
-  for ((i = 0; i <= ${#nodepools[@]}; ++i)); do
+  for ((i = 0; i < ${#nodepools[@]}; ++i)); do
     nodepool=$(jq -r --arg i $i '. | .properties.agentPoolProfiles[$i | tonumber].name' < _output/$RESOURCE_GROUP/apimodel.json)
     if [ "${UPDATE_NODE_POOLS}" = "true" ]; then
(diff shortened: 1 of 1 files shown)


JesusAlvarezTorres:
@jackfrancis can you please take a look at this change? It is a small test fix to some recent changes to cluster.sh

jackfrancis:
Hmmm, I am seeing the opposite problem:

```
++ jq -r '.properties.agentPoolProfiles[].name'
 + nodepools='poollinux
 poolwin
 pooladded'
 + (( i = 0 ))
 + (( i <= 1 ))
```

In the above the `nodepools` "array" has 3 items, but `${#nodepools[@]}` is evaluating to the number 1.

I think we're not correctly using bash to evaluate the number of pools in the string array returned from `$(jq -r  '.properties.agentPoolProfiles[].name' < _output/$RESOURCE_GROUP/apimodel.json)`

JesusAlvarezTorres:
> Hmmm, I am seeing the opposite problem:
> 
> ```
> ++ jq -r '.properties.agentPoolProfiles[].name'
>  + nodepools='poollinux
>  poolwin
>  pooladded'
>  + (( i = 0 ))
>  + (( i <= 1 ))
> ```
> 
> In the above the `nodepools` "array" has 3 items, but `${#nodepools[@]}` is evaluating to the number 1.
> 
> I think we're not correctly using bash to evaluate the number of pools in the string array returned from `$(jq -r '.properties.agentPoolProfiles[].name' < _output/$RESOURCE_GROUP/apimodel.json)`

I just did a quick test by changing the loop to something like this:

# Hard coded API model path for testing the loop
nodepools=$(jq -r  '.properties.agentPoolProfiles[].name' < /mnt/c/tmp/_apimodel.json)
for i in "${nodepools[@]}"; do
    nodepool=$i
    echo "$nodepool"
done

In that case $i gives the name of the node pool, so there is no need to do the second jq query to translate from numerical index to node pool name... do we need the numerical index as well or would this approach work?

JesusAlvarezTorres:
> > Hmmm, I am seeing the opposite problem:
> > ```
> > ++ jq -r '.properties.agentPoolProfiles[].name'
> >  + nodepools='poollinux
> >  poolwin
> >  pooladded'
> >  + (( i = 0 ))
> >  + (( i <= 1 ))
> > ```
> > 
> > 
> > In the above the `nodepools` "array" has 3 items, but `${#nodepools[@]}` is evaluating to the number 1.
> > I think we're not correctly using bash to evaluate the number of pools in the string array returned from `$(jq -r '.properties.agentPoolProfiles[].name' < _output/$RESOURCE_GROUP/apimodel.json)`
> 
> @jackfrancis
> I just did a quick test by changing the loop to something like this:
> 
> # Hard coded API model path for testing the loop
> nodepools=$(jq -r '.properties.agentPoolProfiles[].name' < /mnt/c/tmp/_apimodel.json)
> for i in "${nodepools[@]}"; do
> nodepool=$i
> echo "$nodepool"
> done
> 
> In that case $i gives the name of the node pool, so there is no need to do the second jq query to translate from numerical index to node pool name... do we need the numerical index as well or would this approach work?

My comment above didn't work, it was as you said @jackfrancis. There was something wrong in how the node pool number was being thrown by the jq query. I updated that part in a new commit. Please take a look.

jackfrancis:
Thanks! Testing your change now

jackfrancis (approved):
/lgtm

jackfrancis:
Tested out O.K., thanks for fixing this @JesusAlvarezTorres!

PR #3882, opened 2020-09-30
Change iscustomvnet logic

weng271190436:
**Reason for Change**:

Change iscustomvnet logic because I want to fill VnetSubnetID regardless if the subnet is custom or not

Related RP change:
https://msazure.visualstudio.com/CloudNativeCompute/_git/aks-rp/pullrequest/3583631

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -643,6 +643,7 @@ type AgentPoolProfile struct {
 	StorageProfile                      string               `json:"storageProfile,omitempty"`
 	DiskSizesGB                         []int                `json:"diskSizesGB,omitempty"`
 	VnetSubnetID                        string               `json:"vnetSubnetID,omitempty"`
+	NodeResourceGroup                   string               `json:"nodeResourceGroup,omitempty"`
 	Subnet                              string               `json:"subnet"`
 	IPAddressCount                      int                  `json:"ipAddressCount,omitempty"`
 	Distro                              Distro               `json:"distro,omitempty"`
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1554,7 +1555,7 @@ func (a *AgentPoolProfile) HasImageGallery() bool {
 
 // IsCustomVNET returns true if the customer brought their own VNET
 func (a *AgentPoolProfile) IsCustomVNET() bool {
-	return len(a.VnetSubnetID) > 0
+	return len(a.VnetSubnetID) > 0 && len(a.NodeResourceGroup) > 0 && !strings.Contains(a.VnetSubnetID, a.NodeResourceGroup)
```
xuto2, line 1558: The comment say we "want to fill VnetSubnetID regardless if the subnet is custom or not", I don't see how this change helps on that aspect? In fact this change is making it stricter to for a subnetId to be considered custom?

Also do you have a link to PR or workitem in RP side? I don't quite understand the intention here. The addition of "NodeResourceGroup" is also a rather big change,  can implement the  logic in RP side (i.e. does not attach to the data type)?

weng271190436: I pasted the RP change link to the description.

In the RP pull request I want to fill VnetSubnetID in InheritManagedClusterProperties when vnetsubnetID is empty (which means that customer didn't bring her own vnet subnet)

But then we cannot tell whether customer has custom vnet by whether vnetsubnetID is empty because now it is always not empty.

That's why I need NodeResourceGroup property because the vnet is a custom vnet if it is not in the NodeResourceGroup

xuto2: "the vnet is a custom vnet if it is not in the NodeResourceGroup"  this is not correct as we support BYO node resource group so cx could do both BYO VNET and BYO node ResourceGroup

weng271190436: got you and thanks for bringing up this point. Let me examine the BYO node resource group logic to see if it is feasible at all to tell auto-created subnet and BYO subnet apart if we fill VnetSubnetID all the time

paulgmiller: We need to check if the generated vnet will always be in noderesourcegroup. But it's also always generated with a specific name. If it has that name then we're pretty safe. Wei lets see if we can remember to talk about this on wednesday/thursday.

PR #3901, opened 2020-10-07
WIP: AAD integration using guard

deaborch:
**Reason for Change**:

Enables AAD in kubernetes user management (webhook authorization )

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -530,6 +530,79 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
+      name: webhook
+    current-context: webhook
+    kind: Config
+    preferences: {}
+    users:
+    - name: azure@azure
+      user:
+        client-certificate: /etc/kubernetes/guard/pki/azure@azure.crt
+        client-key: /etc/kubernetes/guard/pki/azure@azure.key
+#EOF
```
jackfrancis, line 578: The purpose of these sentinel "#EOF" characters is to be used with the CSE helper func `wait_for_file`. Is there CSE-specific logic that will be implemented that depends upon this file being present at the time that the CSE shell commands are executed?

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -530,6 +530,79 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
...
```
jackfrancis, line 603: ditto

jackfrancis:
@deaborch @jadarsie just a warning that this new addon will contribute lots of bytes to the customData (the VM property that contains cloud-init data), and will probably not work in concert with other addons being present (the customData payload will exceed the limit).

parts/k8s/cloud-init/masternodecustomdata.yml
deaborch: Not at the moment. However apiserver requires this file to be present by the time it starts to boot up provided webhook authentication is set. It will crash otherwise.

jackfrancis: In that case we should probably add something like this before the kublet systemd service:

```
{{- IsGuardAddonEnabled}}
    wait_for_file 1200 1 /etc/kubernetes/guard/guard-authn-webhook.yaml || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
{{- end}}
```

(we want a `wait_for_file` for every critical file that kubelet systemd will fail when starting up)

PR #3909, opened 2020-10-08
Set the csi-secrets-store to have a priority class

Michael-Sinz:
Daemonsets are needed to run on all nodes but in a busy cluster with dynamic scaling, it is possible to have enough pending pods from replica sets that a new node fills up before the daemonset pod is ready to be scheduled.  Without priority classes, this can cause daemonsets to be pending to the pod they are bound to and can not run elsewhere.

In this change I added a priority class and use it in the daemonsets.  We could look at using the priority class system-node-critical which is reserved for kubernetes system daemonsets.  I don't think it is valid to use that here.

Just having a priority higher than 0 will usually address this as 99.99% of all workload pods are 0 (or below zero for the easily evicted buffer pods and such)

**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/secrets-store-csi-driver.yaml
+++ b/parts/k8s/addons/secrets-store-csi-driver.yaml
@@ -290,6 +290,8 @@ apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
   name: csi-secrets-store
+  labels:
+    addonmanager.kubernetes.io/mode: EnsureExists
 value: 1000
 globalDefault: false
 description: "This is the daemonset priority class for csi-secrets-store"
--- a/parts/k8s/addons/secrets-store-csi-driver.yaml
+++ b/parts/k8s/addons/secrets-store-csi-driver.yaml
@@ -282,6 +282,18 @@ status:
   conditions: []
   storedVersions: []
 ---
+# A priority class for the daemonset such that they are not
+# frozen out of a node due to the node filling up with "normal"
+# pods before the daemonset controller can get the daemonset
+# pods to be scheduled.
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: csi-secrets-store
+value: 1000
+globalDefault: false
+description: "This is the daemonset priority class for csi-secrets-store"
+---
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
@@ -298,6 +310,7 @@ spec:
       labels:
         app: csi-secrets-store
     spec:
+      priorityClassName: csi-secrets-store
       serviceAccountName: secrets-store-csi-driver
       hostNetwork: true
       containers:
@@ -458,6 +471,7 @@ spec:
       labels:
         app: csi-secrets-store-provider-azure
     spec:
+      priorityClassName: csi-secrets-store
       serviceAccountName: csi-secrets-store-provider-azure
       hostNetwork: true
       containers:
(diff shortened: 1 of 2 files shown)


parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -282,6 +282,18 @@ status:
   conditions: []
   storedVersions: []
 ---
+# A priority class for the daemonset such that they are not
```
jackfrancis, line 285: Could you use go template comments for these lines instead of bash comments to save cloud-init data overhead? :/

E.g.:

```
{{/* A priority class for the daemonset such that they are not */}}
```

jackfrancis:
@Michael-Sinz and one more thing after the comment change, run `make generate` to generate code, and commit.

Thanks so much for this!

parts/k8s/addons/secrets-store-csi-driver.yaml
Michael-Sinz: Ok - annoying, but ok.
(We could look loading and dumping yaml in optimal form after templating such that you are sure to not have waste in general)

parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -282,6 +282,18 @@ status:
   conditions: []
   storedVersions: []
 ---
+{{/* A priority class for the daemonset such that they are not */}}
```
jackfrancis, line 285: change

{{/*

with

{{- /*

Michael-Sinz: done

parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -282,6 +282,20 @@ status:
...
+{{- /* A priority class for the daemonset such that they are not */}}
+{{- /* frozen out of a node due to the node filling up with "normal" */}}
+{{- /* pods before the daemonset controller can get the daemonset */}}
+{{- /* pods to be scheduled. */}}
+apiVersion: scheduling.k8s.io/v1
+kind: PriorityClass
+metadata:
+  name: csi-secrets-store
+  labels:
+    addonmanager.kubernetes.io/mode: EnsureExists
```
jackfrancis, line 294: Added this label so that kube-addon-manager would interpret this resource for loading

Michael-Sinz: Thanks - sorry I missed that.

jackfrancis (approved):
/lgtm

parts/k8s/addons/secrets-store-csi-driver.yaml
```diff
@@ -282,6 +282,20 @@ status:
   conditions: []
   storedVersions: []
 ---
+{{- /* A priority class for the daemonset such that they are not */}}
+{{- /* frozen out of a node due to the node filling up with "normal" */}}
+{{- /* pods before the daemonset controller can get the daemonset */}}
+{{- /* pods to be scheduled. */}}
+apiVersion: scheduling.k8s.io/v1
```
jackfrancis, line 289: cc @aramase @chewong this change was made due to observed scheduling contention on clusters at scale (i.e., nodes coming on line but filling up so quickly that csi-secrets-store pods were not able to be scheduled via daemonset

Should we make such a change to the upstream reference specs as well?

PR #3928, opened 2020-10-13
feat(WIP): check volume attachment after each node drain during upgrade

jiaxunsongucb:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
1. Fix Disk page NotDone() function logic error.
2. Expose new functions for kubeclient: GetPod, GetPersistentVolume, GetPersistentVolumeClaim.
3. Add volume attachment status check logic during upgrade after each node drain operation.

### DIFF
--- a/pkg/operations/cordondrainvm.go
+++ b/pkg/operations/cordondrainvm.go
@@ -4,6 +4,9 @@
 package operations
 
 import (
+	"context"
+	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
@@ -40,11 +43,12 @@ func SafelyDrainNode(az armhelpers.AKSEngineClient, logger *log.Entry, apiserver
 	if err != nil {
 		return err
 	}
-	return SafelyDrainNodeWithClient(client, logger, nodeName, timeout)
+	_, err = SafelyDrainNodeWithClient(client, logger, nodeName, timeout)
+	return err
 }
 
 // SafelyDrainNodeWithClient safely drains a node so that it can be deleted from the cluster
-func SafelyDrainNodeWithClient(client armhelpers.KubernetesClient, logger *log.Entry, nodeName string, timeout time.Duration) error {
+func SafelyDrainNodeWithClient(client armhelpers.KubernetesClient, logger *log.Entry, nodeName string, timeout time.Duration) ([]v1.Pod, error) {
 	nodeName = strings.ToLower(nodeName)
 	//Mark the node unschedulable
 	var node *v1.Node
@@ -52,7 +56,7 @@ func SafelyDrainNodeWithClient(client armhelpers.KubernetesClient, logger *log.E
 	for i := 0; i < cordonMaxRetries; i++ {
 		node, err = client.GetNode(nodeName)
 		if err != nil {
-			return err
+			return nil, err
 		}
 		node.Spec.Unschedulable = true
 		node, err = client.UpdateNode(node)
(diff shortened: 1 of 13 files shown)


pkg/operations/kubernetesupgrade/upgradeagentnode.go
```diff
@@ -70,7 +70,11 @@ func (kan *UpgradeAgentNode) DeleteNode(vmName *string, drain bool) error {
 	}
 	// Cordon and drain the node
 	if drain {
-		err = operations.SafelyDrainNodeWithClient(client, kan.logger, nodeName, kan.cordonDrainTimeout)
+		if kan.UpgradeContainerService.Properties.IsAzureStackCloud() {
```
jadarsie, line 73: Will this work?

```
err = operations.SafelyDrainNodeWithClient(client, kan.logger, nodeName, kan.cordonDrainTimeout)
if kan.UpgradeContainerService.Properties.IsAzureStackCloud() {
    err = operations.WaitForDisksAttached(client, kan.logger, nodeName, kan.cordonDrainTimeout)
}
```

I think it is slightly better to compose discrete operations and unit testing will likely be easier.

jiaxunsongucb: Yes, actually this was my first attempt. I totally agree that it's easier to unit-test. However, let me explain the reason why I changed my mind. So inside the function SafelyDrainNodeWithClient things happen in this order: 1) mark the node as unschedulable, 2) get the list of pods that need to be evicted from this node, 3) evict these pods. And for the volume check, one more step is added: 4) check the volumes for the evicted pods. So the point is, in order to check/monitor the volumes, I need to get a list of pods that are evicted from this node. It's not easy to properly get these pods from outside of the function SafelyDrainNodeWithClient.
The alternatives would be 1) get the pods before and after node drain and do the diff to get the pod that needs to be monitored, or 2) change the signature of function SafelyDrainNodeWithClient to make it return the evicted pods. I tried both of these two approaches which worked as well. Please advise if you have a better solution.

jiaxunsongucb:
> Should we land this change in master and then cherry-pick into the Azure Stack release branch?

Yes, eventually this change will target master. For now, to avoid any regressions in master, I am validating this change against the release branch.

pkg/operations/cordondrainvm.go
```diff
@@ -228,3 +235,154 @@ func (o *drainOperation) deletePods(pods []v1.Pod) error {
...
+// WaitForDisksAttached waits for disks are re-attached to new nodes and volumes are ready to be used by pods.
+func WaitForDisksAttached(podsForDeletion []v1.Pod, client armhelpers.KubernetesClient, logger *log.Entry, args ...time.Duration) error {
+	if len(podsForDeletion) == 0 {
+		return nil
+	}
+
+	timeout := time.Hour * 1
+	interval := time.Minute * 1
+	if len(args) > 0 {
+		timeout = args[0]
```
jifwang-microsoft, line 359: Looks like timeout is not used in the wait function. What if the VM size is not large enough to accept the managed disk?

jiaxunsongucb:
After the discussion with ASK Engine team and Azure Stack AKS team, we decide to use a longer cordon timeout and Pod Disruption Budget Policy to solve this problem. Closing this PR.

PR #3934, opened 2020-10-14
feat: add support for Kubernetes v1.20.0-alpha.3

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#changelog-since-v1200-alpha2

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

~~Publishing is not yet complete, so tests will fail for the time being.~~


jackfrancis:
I think a rebase + Windows VHD signal is all we need here.

I vote we keep the 1.20 E2E tests in here btw.

jackfrancis (approved):
/lgtm

PR #3939, opened 2020-10-15
fix: get-logs command collects control plane logs even if apiserver is down

haofan-ms:
**Reason for Change**:

When api server is down, the `kubeClient.ListNodes()` function will not be able to get nodes. In this case we want to get control panel nodes using other configurations and collect logs of these nodes.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
to @jadarsie for final approval/merge

jadarsie:
> to @jadarsie for final approval/merge

checking

cmd/get_logs.go
```diff
@@ -362,6 +365,17 @@ func (glc *getLogsCmd) getCloudName() string {
 	return ""
 }
 
+func (glc *getLogsCmd) getControlPanelNodes() []v1.Node {
```
jadarsie, line 368: `computeControlPanelNodes` may be slightly better

jadarsie, line 368: Could you please add a test for this method?

cmd/get_logs.go
```diff
@@ -362,6 +365,17 @@ func (glc *getLogsCmd) getCloudName() string {
...
+	var ControlPanelNodeList []v1.Node
```
jadarsie, line 369: shorter camel-cased variable names are more idiomatic in my opinion

ex: `i` instead of `nodeIndex` or `node` instead of `controlPlaneNode`

jadarsie (changes requested):

cmd/get_logs.go
haofan-ms: Updated

haofan-ms: test added

cmd/get_logs.go
haofan-ms: updated

jadarsie (approved):

PR #3940, opened 2020-10-15
feat: custom Windows log collection script

haofan-ms:
**Reason for Change**:

- Remove the enforcement of --linux-script
- Enable custom windows script to collect logs via --windows-script

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
@jadarsie feel free to merge this if it meets w/ your review approval

cmd/get_logs.go
```diff
@@ -75,14 +76,14 @@ func newGetLogsCmd() *cobra.Command {
 	command.Flags().StringVarP(&glc.apiModelPath, "api-model", "m", "", "path to the generated apimodel.json file (required)")
 	command.Flags().StringVar(&glc.sshHostURI, "ssh-host", "", "FQDN, or IP address, of an SSH listener that can reach all nodes in the cluster (required)")
 	command.Flags().StringVar(&glc.linuxSSHPrivateKeyPath, "linux-ssh-private-key", "", "path to a valid private SSH key to access the cluster's Linux nodes (required)")
-	command.Flags().StringVar(&glc.linuxScriptPath, "linux-script", "", "path to the log collection script to execute on the cluster's Linux nodes (required)")
+	command.Flags().StringVar(&glc.linuxScriptPath, "linux-script", "", "path to the log collection script to execute on the cluster's Linux nodes")
```
jadarsie, line 79: I would explain that this is required if distro is not an aks-ubuntu flavor

cmd/get_logs.go
```diff
@@ -167,31 +172,43 @@ func (glc *getLogsCmd) run() (err error) {
...
 		return errors.Wrap(err, "listing cluster nodes")
 	}
-	for _, n := range glc.masterNodes {
-		log.Infof("Processing master node: %s\n", n.Name)
-		out, err := glc.collectLogs(n, glc.linuxSSHConfig)
-		if err != nil {
-			log.Warnf("Remote command output: %s", out)
-			log.Warnf("Error: %s", err)
+	if glc.linuxScriptPath == "" && (!glc.cs.Properties.MasterProfile.IsVHDDistro()) {
+		log.Warnf("No linux log collecting script provided, will skip log collection for master nodes as distro is not aks VHD")
```
jadarsie, line 176: I think this may be cleaner if `getClusterNodes` is in charge of controlling which nodes make it into the nodes collections. Quick pseudo func that you may need:

```
func linuxScriptAvailable() bool { 
  return linuxScriptPath != "" || isVHDDistro()
}
```

cmd/get_logs.go
```diff
@@ -255,25 +272,46 @@ func (glc *getLogsCmd) collectLogs(node v1.Node, config *ssh.ClientConfig) (stri
 }
 
 func (glc *getLogsCmd) uploadScript(node v1.Node, client *ssh.Client) (string, error) {
-	if isWindowsNode(node) || glc.linuxScriptPath == "" {
+	if glc.windowsScriptPath == "" && glc.linuxScriptPath == "" {
```
jadarsie, line 275: I would rather reduce code duplication like it was before

cmd/get_logs.go
```diff
@@ -296,7 +334,11 @@ func (glc *getLogsCmd) executeScript(node v1.Node, client *ssh.Client) (string,
 			cmd = fmt.Sprintf("bash -c \"export AZURE_ENV=%s; sudo -E %s\"", glc.getCloudName(), script)
 		}
 	} else {
-		script = "c:\\k\\debug\\collect-windows-logs.ps1"
+		if glc.windowsScriptPath != "" {
+			script = "(gi $env:temp).fullname + '\\collect-windows-logs.ps1'"
```
jadarsie, line 338: You may be able to do `script = "$env:temp\\collect-windows-logs.ps1"` by adding `iex` as indicated below:

```
cmd = fmt.Sprintf("powershell -noprofile -command \"iex %s\"", filePath)
```

jadarsie (changes requested):
Please check if this page needs to be updated: https://github.com/Azure/aks-engine/blob/master/docs/topics/get-logs.md

cmd/get_logs.go
haofan-ms: updated

cmd/get_logs.go
haofan-ms: updated

cmd/get_logs.go
haofan-ms: updated

cmd/get_logs.go
haofan-ms: Thanks, that works. Updated in code.

haofan-ms:
> Please check if this page needs to be updated: https://github.com/Azure/aks-engine/blob/master/docs/topics/get-logs.md

Updated documentation, thanks.

cmd/get_logs.go
```diff
@@ -347,6 +380,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
...
+	if glc.linuxScriptPath == "" && !glc.cs.Properties.MasterProfile.IsVHDDistro() {
+		if glc.controlPlaneOnly {
+			return errors.Errorf("No log collection script found for control plane nodes")
+		} else {
+			log.Warnf("Skipping master nodes nodes as no log collection script found")
+			glc.masterNodes = nil
+		}
+	}
+	for _, profile := range glc.cs.Properties.AgentPoolProfiles {
+		if glc.linuxScriptPath == "" && profile.OSType == "Linux" && !profile.IsVHDDistro() {
```
jadarsie, line 393: please do a case insensitive comparison

cmd/get_logs.go
```diff
@@ -347,6 +380,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
 	return "", nil
 }
 
+func (glc *getLogsCmd) validateLogScript() error {
+	if glc.linuxScriptPath == "" && !glc.cs.Properties.MasterProfile.IsVHDDistro() {
+		if glc.controlPlaneOnly {
+			return errors.Errorf("No log collection script found for control plane nodes")
+		} else {
+			log.Warnf("Skipping master nodes nodes as no log collection script found")
```
jadarsie, line 388: - `Warn` instead of `Warnf`
- `master nodes nodes`
- let's pick between `master` OR `control plane` and use it consistently

cmd/get_logs.go
```diff
@@ -347,6 +380,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
...
+			glc.masterNodes = nil
```
jadarsie, line 389: is this working? I think you should set it to an empty slice

cmd/get_logs.go
```diff
@@ -362,6 +417,16 @@ func (glc *getLogsCmd) getCloudName() string {
 	return ""
 }
 
+func updateLinuxNodesList(nodeList []v1.Node, agentPoolName string) []v1.Node {
```
jadarsie, line 420: this name can be more specific, something like `filterNodesFromPool` or similar may explain better what the func does.

cmd/get_logs.go
```diff
@@ -362,6 +417,16 @@ func (glc *getLogsCmd) getCloudName() string {
...
+	var linuxNodeList []v1.Node
+	for _, node := range nodeList {
+		if strings.Split(node.Name, "-")[1] != agentPoolName {
```
jadarsie, line 423: case insensitive comparison please

cmd/get_logs.go
```diff
@@ -347,6 +380,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
...
+			glc.masterNodes = nil
+		}
+	}
+	for _, profile := range glc.cs.Properties.AgentPoolProfiles {
+		if glc.linuxScriptPath == "" && profile.OSType == "Linux" && !profile.IsVHDDistro() {
+			log.Warnf("Skipping nodes in linux agentpool %s as no log collection script found", profile.Name)
+			glc.linuxNodes = updateLinuxNodesList(glc.linuxNodes, profile.Name)
+		}
+		if glc.windowsScriptPath == "" && profile.OSType == "Windows" && !(glc.cs.Properties.WindowsProfile.WindowsPublisher == "microsoft-aks" && glc.cs.Properties.WindowsProfile.WindowsOffer == "aks-windows") {
+			log.Warnf("Skipping nodes in windows agentpools as no log collection script found")
```
jadarsie, line 398: I am not sure this message will make sense to people without as much context as you. I would explain why this happens.

`Skipping windows nodes as flag '--windows-script' is not set and the pool's distro is not {name}`

Same for linux and master nodes

cmd/get_logs.go
```diff
@@ -347,6 +380,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
...
+			log.Warnf("Skipping master nodes nodes as no log collection script found")
+			glc.masterNodes = nil
+		}
+	}
+	for _, profile := range glc.cs.Properties.AgentPoolProfiles {
+		if glc.linuxScriptPath == "" && profile.OSType == "Linux" && !profile.IsVHDDistro() {
+			log.Warnf("Skipping nodes in linux agentpool %s as no log collection script found", profile.Name)
+			glc.linuxNodes = updateLinuxNodesList(glc.linuxNodes, profile.Name)
+		}
+		if glc.windowsScriptPath == "" && profile.OSType == "Windows" && !(glc.cs.Properties.WindowsProfile.WindowsPublisher == "microsoft-aks" && glc.cs.Properties.WindowsProfile.WindowsOffer == "aks-windows") {
```
jadarsie, line 397: maybe you can add method `WindowsProfile.IsVHDDistro()`

jadarsie (changes requested):

cmd/get_logs.go
haofan-ms: Updated

cmd/get_logs.go
haofan-ms: Updated

cmd/get_logs.go
haofan-ms: Yes this is working and this is also how an empty slice of specific type looks like when initially created.

cmd/get_logs.go
haofan-ms: Updated

cmd/get_logs.go
haofan-ms: I think the comparison here should not be case insensitive as the agentPool name from API model is used as part of the agent node name, But since the agentPool name only allows a-z0-9, there shouldn't be difference between case sensitive/insensitive. Updated.

cmd/get_logs.go
haofan-ms: Updated

cmd/get_logs.go
haofan-ms: Updated

cmd/get_logs.go
haofan-ms: I am not quite sure what you mean here. Do you mean we should remove this if condition?

haofan-ms: Updated, thanks for the clarification

jadarsie (changes requested):

cmd/get_logs.go
```diff
@@ -351,6 +369,28 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
 	return "", nil
 }
 
+func (glc *getLogsCmd) validateLogScript() error {
```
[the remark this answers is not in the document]
haofan-ms: Thanks for catching that, I messed up when rewriting this (https://github.com/Azure/aks-engine/pull/3940/commits/db3a24d0b4f07e062d64f56457c27bae8c16f891#diff-f7d2eb01ec49a3aac510d880f4bfef446dbd848d220cf6cc3bd0fc4ff1610249L427)

jadarsie (approved):

PR #3944, opened 2020-10-16
feat: upload collected logs to storage account container

haofan-ms:
**Reason for Change**:

Enable get-logs command to upload k8s logs to specific storage container using SAS URL (and container name).

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


cmd/get_logs.go
```diff
@@ -255,6 +262,13 @@ func (glc *getLogsCmd) collectLogs(node v1.Node, config *ssh.ClientConfig) (stri
 	if err != nil {
 		return stdout, err
 	}
+	if glc.blobServiceSASURL != "" {
+		log.Debug("Will upload logs to storage account")
```
jadarsie, line 266: I think it is better to upload files after all logs were collected. Otherwise an error uploading files will stop the process.

cmd/get_logs.go
```diff
@@ -78,6 +83,8 @@ func newGetLogsCmd() *cobra.Command {
 	command.Flags().StringVar(&glc.linuxScriptPath, "linux-script", "", "path to the log collection script to execute on the cluster's Linux nodes (required)")
 	command.Flags().StringVarP(&glc.outputDirectory, "output-directory", "o", "", "collected logs destination directory, derived from --api-model if missing")
 	command.Flags().BoolVarP(&glc.controlPlaneOnly, "control-plane-only", "", false, "get logs from control plane VMs only")
+	command.Flags().StringVarP(&glc.blobServiceSASURL, "sas-url", "", "", "blob service SAS URL of the storage account on Azure or custom cloud to upload kubernetes logs")
```
jadarsie, line 86: Please update the doc page

jadarsie, line 86: what if we go with `sa-sas-url` and `sa-container` (or `sa-container-name`), I am thinking that the `sa-` prefix will help user understand that these two go together

however, I still think one parameter will make things in long run, documentation should clarify that container name is expected and `validateArgs` should make sure the format is correct.

cmd/get_logs.go
jadarsie, line 266: not sure this debug message adds valuable info

cmd/get_logs.go
```diff
@@ -351,6 +365,37 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
 	return "", nil
 }
 
+func (glc *getLogsCmd) uploadLogsToStorageContainer(node v1.Node) error {
```
jadarsie, line 368: can we add some retry logic if an upload fails? or log an issue to address it?

jadarsie (changes requested):

cmd/get_logs.go
```diff
@@ -203,6 +217,24 @@ func (glc *getLogsCmd) run() (err error) {
 		}
 	}
 	log.Infof("Logs downloaded to %s", glc.outputDirectory)
+	if glc.storageContainerSASURL != "" {
+		if glc.cs.Properties.IsCustomCloudProfile() && glc.cs.Properties.CustomCloudProfile.IdentitySystem == "adfs" {
+			log.Warn("Failed uploading logs for custom cloud with identity system of adfs")
```
jadarsie, line 222: The term `Failed` is not properly communicating what's going on, the process is skipped.

cmd/get_logs.go
```diff
@@ -203,6 +217,24 @@ func (glc *getLogsCmd) run() (err error) {
...
+			return nil
+		}
+		logFiles, err := ioutil.ReadDir(glc.outputDirectory)
+		if err != nil {
+			return errors.Wrapf(err, "reading output directory %s", glc.outputDirectory)
+		}
+		for _, logFile := range logFiles {
```
jadarsie, line 229: This will try to upload all files in the output directory, even those not produced by `get-logs`

cmd/get_logs.go
```diff
@@ -203,6 +217,24 @@ func (glc *getLogsCmd) run() (err error) {
 		}
 	}
 	log.Infof("Logs downloaded to %s", glc.outputDirectory)
+	if glc.storageContainerSASURL != "" {
+		if glc.cs.Properties.IsCustomCloudProfile() && glc.cs.Properties.CustomCloudProfile.IdentitySystem == "adfs" {
```
jadarsie, line 221: please do a case sensitive comparison

cmd/get_logs_test.go
```diff
@@ -133,6 +133,19 @@ func TestGetLogsCmdValidateArgs(t *testing.T) {
...
+				apiModelPath:           existingFile,
+				linuxSSHPrivateKeyPath: existingFile,
+				linuxScriptPath:        existingFile,
+				windowsScriptPath:      existingFile,
+				sshHostURI:             "server.example.com",
+				location:               "southcentralus",
+				storageContainerSASURL: "https://blob-service-uri/?sas-token",
+			},
+			expectedErr: errors.Errorf("error validating storage container SAS URL https://blob-service-uri/?sas-token. Expected SAS URL format is https://{blob-service-uri}/{container-name}?{sas-token}"),
+			name:        "InvalidSASURL",
```
jadarsie, line 147: when writing unit tests, you typically include both positive and negative cases

cmd/get_logs.go
```diff
@@ -369,6 +401,34 @@ func (glc *getLogsCmd) downloadLogs(node v1.Node, client *ssh.Client) (string, e
...
+func (glc *getLogsCmd) uploadLogsToStorageContainer(fileName string) error {
+	log.Infof("Uploading log file %s", fileName)
+	logFileName := strings.TrimSuffix(fileName, filepath.Ext(fileName))
+	logFilePath := path.Join(glc.outputDirectory, fileName)
+	logFile, err := os.Open(logFilePath)
+	if err != nil {
+		return errors.Wrapf(err, "reading log file %s", logFilePath)
+	}
+
+	urls := strings.Split(glc.storageContainerSASURL, "?")
```
jadarsie, line 413: will this work?
```
u, err := url.Parse(glc.storageContainerSASURL)
if err != nil {
	return errors.Wrapf(err, "parsing the storage container SAS URL")
}
u.Path = path.Join(u.Path, theFileName)
```

jadarsie (changes requested):

cmd/get_logs_test.go
haofan-ms: updated

PR #3953, opened 2020-10-21
fix: resource identifier not pass to custom cloud profile

laozc:
**Reason for Change**:
Bug fix
Generated custom cloud profile does not contain resource identifiers even when set in API model.
It may cause unexpected application behavior for those programs relying on these values

**Issue Fixed**:

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [X] adds unit tests

**Notes**:


jadarsie:
lgtm

I simple deployment worked on ASH

CecileRobertMichon (commented):
/lgtm

PR #3957, opened 2020-10-21
refactor: delete Azure Stack's KubernetesClient implementation

jadarsie:
**Reason for Change**:

Files `pkg/armhelpers/kubeclient.go` and `pkg/armhelpers/azurestack/kubeclient.go` look almost identical. I do not see a good reason to keep two different implementations of interface `armhelpers.KubernetesClient`.

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/armhelpers/azureclient.go
```diff
@@ -80,6 +80,11 @@ type AzureClient struct {
 	servicePrincipalsClient graphrbac.ServicePrincipalsClient
 }
 
+// GetKubernetesClient returns a KubernetesClient hooked up to the api server at the apiserverURL.
```
jackfrancis, line 83: Thanks for moving this `AzureClient` pointer method here alongside its other methods. (Unfortunately this should never have been implemented as a method, as it neither operates upon nor modifies the instance object. :/ (But we don't want to worry about that now)

pkg/armhelpers/azurestack/azureclient.go
```diff
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Azure/aks-engine/pkg/armhelpers"
```
jackfrancis, line 17: The only thing I'd say here is that this "child" package depending upon the parent can lead to circular dependencies in the future. Ideally we'd put `NewKubernetesClientSetClient` somewhere that both `pkg/armhelpers` and `pkg/armhelpers/azurestack` could import in common. Not sure if there's a candidate package for that?

jadarsie: yeah, I thought about that, I was not sure if moving more stuff around was desired, I will look for a new home for this type.

jackfrancis (approved):
/lgtm

PR #3958, opened 2020-10-21
chore: use transparent mode for Azure CNI

### LINKED ISSUE #3955: DNS latency when using Azure CNI bridge mode + multiple coredns pods
We have observed that ~1% of DNS lookups can be high (up to 5 seconds) when using Azure CNI bridge mode (the default configuration) and more than one coredns pod.

Azure CNI transparent mode does not reproduce this, e.g.:

```
...
        "kubernetesConfig": {
            "networkPlugin": "azure",
            "networkMode": "transparent
        }
...
```

jackfrancis:
**Reason for Change**:

This PR changes the Azure CNI implementation to use "transparent" mode by default, rather than using a bridge interface. This change is due to an observed reduction in UDP edge cases, specifically intermittent high DNS lookup latency when the resolver is an IP address that sits in front of a distributed set of resolvers (DNAT). In the Kubernetes context, this is reproduced if there are more than one coredns pods in the cluster.

Fixes #3955

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -202,6 +202,12 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				}
 			}
 		}
+
+		if o.KubernetesConfig.NetworkPlugin == NetworkPluginAzure {
+			if o.KubernetesConfig.NetworkMode == "" {
+				o.KubernetesConfig.NetworkMode = NetworkModeTransparent
+			}
+		}
 		if o.KubernetesConfig.ContainerRuntime == "" {
 			o.KubernetesConfig.ContainerRuntime = DefaultContainerRuntime
 		}
--- a/examples/e2e-tests/kubernetes/release/default/definition.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition.json
@@ -41,7 +41,7 @@
             "name": "coredns",
             "enabled": true,
             "config": {
-              "min-replicas": "1",
+              "min-replicas": "3",
               "nodes-per-replica": "10"
             }
           },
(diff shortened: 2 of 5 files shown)


jackfrancis:
cc @alexeldeib @matmerr @chandanAggarwal @tamilmani1989

chandanAggarwal:
cc @vakalapa

pkg/api/defaults.go
```diff
@@ -202,6 +202,12 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				}
 			}
 		}
+
+		if o.KubernetesConfig.NetworkPlugin == NetworkPluginAzure {
+			if o.KubernetesConfig.NetworkMode == "" {
+				o.KubernetesConfig.NetworkMode = NetworkModeTransparent
```
paulgmiller, line 208: sorry for a dumb question but what actually actions on network mode? I couldn't find it in my novice sluething. I just found this in cse.


if [[ "${NETWORK_POLICY}" == "calico" ]]; then
--
284 | sed -i 's#"mode":"bridge"#"mode":"transparent"#g' $CNI_CONFIG_DIR/10-azure.conflist

weng271190436: https://github.com/Azure/aks-engine/blob/1dd327ad7697d4fa43d2bb86adf6679689ab7e84/parts/k8s/cloud-init/artifacts/cse_config.sh#L288

is it this one?

weng271190436 (approved):

pkg/api/defaults.go
jackfrancis: @weng271190436 has it right

Basically, the Azure CNI release artifact comes with a default static config, and we simply extract it, then sed replace `"bridge"` with `"transparent"` if the api model says to do so.

mboersma (approved):
/lgtm

PR #3962, opened 2020-10-23
test: Windows Metrics failure when High CPU 

jsturtevant:
**Reason for Change**:

This is adds a test case for https://github.com/kubernetes/kubernetes/issues/95735.  It is a bit verbose right now and could be cleaned up but it demonstrates the failures at high cpu.

**Issue Fixed**:

Doesn't fix it right now but will demonstrates: https://github.com/kubernetes/kubernetes/issues/95735

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/kubernetes/deployment/deployment.go
```diff
@@ -846,6 +906,63 @@ func (d *Deployment) WaitForReplicas(min, max int, sleep, timeout time.Duration)
...
+			err := d.Describe()
+			if err != nil {
+				log.Printf("Unable to describe deployment %s: %s", d.Metadata.Name, err)
+			}
+			return pods, errors.Errorf("WaitForReplicas timed out: %s\n", mostRecentWaitForReplicasError)
+		}
+	}
+}
+
+// WaitForReplicas waits for a pod replica count between min and max
```
jackfrancis, line 934: nit: method comment copypasta

test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2369,6 +2370,59 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 			Expect(err).NotTo(HaveOccurred())
 			Expect(logsRotated).To(Equal(true))
 		})
+
+		// metrics endpoints failing
+		// https://github.com/kubernetes/kubernetes/issues/95735
+		It("windows should be able to get node metrics when high cpu", func() {
+			if !eng.HasWindowsAgents() {
```
jackfrancis, line 2377: So we can add another condition here, like maybe `ValidateNodeLoad` or something, which would change the above to:

```
if !eng.HasWindowsAgents() || !cfg.ValidateNodeLoad {
    Skip("Will not validate effects of CPU load against nodes")
```

Or something like that.

Then, we want to add that new property to the `Config` struct type definition in `test/e2e/config/config.go`. If we default it to false, then we can basically ship this change, and work on fixing it against master continually by running tests using `VALIDATE_NODE_LOAD=true make test-kubernetes` on branches which aim to turn the test green.

`BlockSSHPort` is an example of this in action.

Once we have a fix, we can open up a PR w/ the fix, and include a change to `.pipelines/e2e-job-template.yaml` that adds `VALIDATE_NODE_LOAD: true` to the `variables` dictionary in the E2E job definition.

jackfrancis (approved):
/lgtm

PR #3965, opened 2020-10-23
refactor: remove unsupported orchestrators

mboersma:
**Reason for Change**:

Removes references to the Swarm, Swarm mode, Mesos, and DC/OS orchestrator types to simplify the AKS Engine codebase.

This code was already effectively dead, but was being retained out of an abundance of caution in order not to break AKS' vendoring needs. That is no longer a concern.

**Issue Fixed**:

Refs #2500

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:

Let me know if I should make separate PRs for each commit instead to make it more digestible.

There is further refactoring for simplicity that can be done here, probably starting with removal of `GenerateTemplate` and any remaining `switch` statements where the only `case` is Kubernetes now. This seems like enough to review, however.


jackfrancis:
cc @xuto2

mboersma:
I think we might be exceeding the custom data size limit with the 1_20_docker_e2e config: CI fails there consistently in this PR.

jackfrancis (approved):
/lgtm

🎉

PR #3969, opened 2020-10-26
chore: longer default cordonDrainTimeout for Azure Stack Cloud

jiaxunsongucb:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Set 60 minutes cordonDrainTimeout for Azure Stack Cloud to give it enough time to move around resources during Node Drain, especially disk detach/attach operations.
We still honor the user's input.

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -170,6 +170,13 @@ func (uc *upgradeCmd) loadCluster() error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	// Set 60 minutes cordonDrainTimeout for Azure Stack Cloud to give it enough time to move around resources during Node Drain,
+	// especially disk detach/attach operations. We still honor the user's input.
+	if uc.cordonDrainTimeout == nil && uc.containerService.Properties.IsAzureStackCloud() {
+		cordonDrainTimeout := time.Duration(60) * time.Minute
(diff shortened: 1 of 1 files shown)


cmd/upgrade.go
```diff
@@ -170,6 +170,13 @@ func (uc *upgradeCmd) loadCluster() error {
 		return errors.Wrap(err, "error parsing the api model")
 	}
 
+	// Set 60 minutes cordonDrainTimeout for Azure Stack Cloud to give it enough time to move around resources during Node Drain,
```
jadarsie, line 173: Maybe method `validate` is a better place for this change. Please add a unit test.

jiaxunsongucb: I've tried to add this change in method validate, but the problem is that uc.containerService.Properties.IsAzureStackCloud() is not accessible in method validate, it must be after "uc.containerService, uc.apiVersion, err = apiloader.LoadContainerServiceFromFile(uc.apiModelPath, true, true, nil)" is called.

jadarsie: right

PR #3970, opened 2020-10-26
fix: cancel Pod Eviction jobs when timeout

jiaxunsongucb:
**Reason for Change**:

When method `evictPods` returns in whichever cases (normal return or timeout), we need to send a signal to the goroutines to stop and return, otherwise, it may be problematic because it will keep trying to evict the Pods forever (based on my observation in my tests)

**Issue Fixed**:

Use `context` to send signals to goroutines to stop and return when the method `evictPods` returns.

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Cancel the Pod Eviction jobs when timeout. Otherwise, it may fall into an infinite loop.

### DIFF
--- a/pkg/operations/cordondrainvm.go
+++ b/pkg/operations/cordondrainvm.go
@@ -4,6 +4,7 @@
 package operations
 
 import (
+	"context"
 	"strings"
 	"time"
 
@@ -173,24 +174,31 @@ func (o *drainOperation) deleteOrEvictPods(pods []v1.Pod) error {
 }
 
 func (o *drainOperation) evictPods(pods []v1.Pod, policyGroupVersion string) error {
+	ctx, cancel := context.WithCancel(context.Background())
 	doneCh := make(chan bool, len(pods))
 	errCh := make(chan error, 1)
 
 	for _, pod := range pods {
-		go func(pod v1.Pod, doneCh chan bool, errCh chan error) {
+		go func(ctx context.Context, pod v1.Pod, doneCh chan bool, errCh chan error) {
 			var err error
+		doneEviction:
 			for {
-				err = o.client.EvictPod(&pod, policyGroupVersion)
-				if err == nil {
-					break
-				} else if apierrors.IsNotFound(err) {
-					doneCh <- true
+				select {
+				case <-ctx.Done():
 					return
-				} else if apierrors.IsTooManyRequests(err) {
-					time.Sleep(5 * time.Second)
-				} else {
-					errCh <- errors.Wrapf(err, "error when evicting pod %q", pod.Name)
-					return
+				default:
+					err = o.client.EvictPod(&pod, policyGroupVersion)
+					if err == nil {
+						break doneEviction
+					} else if apierrors.IsNotFound(err) {
+						doneCh <- true
+						return
+					} else if apierrors.IsTooManyRequests(err) {
+						time.Sleep(5 * time.Second)
+					} else {
+						errCh <- errors.Wrapf(err, "error when evicting pod %q", pod.Name)
+						return
+					}
 				}
 			}
 			podArray := []v1.Pod{pod}
@@ -200,20 +208,23 @@ func (o *drainOperation) evictPods(pods []v1.Pod, policyGroupVersion string) err
 			} else {
 				errCh <- errors.Wrapf(err, "error when waiting for pod %q terminating", pod.Name)
 			}
-		}(pod, doneCh, errCh)
+		}(ctx, pod, doneCh, errCh)
 	}
 
 	doneCount := 0
 	for {
 		select {
 		case err := <-errCh:
+			cancel()
 			return err
 		case <-doneCh:
 			doneCount++
 			if doneCount == len(pods) {
+				cancel()
 				return nil
 			}
 		case <-time.After(o.timeout):
+			cancel()
 			return errors.Errorf("Drain did not complete within %v", o.timeout)
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/operations/cordondrainvm.go
```diff
@@ -173,24 +174,31 @@ func (o *drainOperation) deleteOrEvictPods(pods []v1.Pod) error {
...
 func (o *drainOperation) evictPods(pods []v1.Pod, policyGroupVersion string) error {
+	ctx, cancel := context.WithCancel(context.Background())
 	doneCh := make(chan bool, len(pods))
 	errCh := make(chan error, 1)
 
 	for _, pod := range pods {
-		go func(pod v1.Pod, doneCh chan bool, errCh chan error) {
+		go func(ctx context.Context, pod v1.Pod, doneCh chan bool, errCh chan error) {
 			var err error
+		doneEviction:
```
jackfrancis, line 184: Why are we creating this additional named scope wrapping around the for loop?

pkg/operations/cordondrainvm.go
```diff
@@ -200,20 +208,23 @@ func (o *drainOperation) evictPods(pods []v1.Pod, policyGroupVersion string) err
...
 			}
-		}(pod, doneCh, errCh)
+		}(ctx, pod, doneCh, errCh)
 	}
 
 	doneCount := 0
 	for {
 		select {
 		case err := <-errCh:
+			cancel()
```
devigned, line 218: Perhaps, put the cancel in a defer outside of the for loop.

pkg/operations/cordondrainvm.go
```diff
@@ -173,24 +174,31 @@ func (o *drainOperation) deleteOrEvictPods(pods []v1.Pod) error {
 }
 
 func (o *drainOperation) evictPods(pods []v1.Pod, policyGroupVersion string) error {
+	ctx, cancel := context.WithCancel(context.Background())
```
jackfrancis, line 177: Is this meant to enforce a timeout? This context object will never timeout according to a counter, its `Done()` method will only be engaged if the `cancel` return func is invoked manually.

pkg/operations/cordondrainvm.go
```diff
@@ -173,24 +174,31 @@ func (o *drainOperation) deleteOrEvictPods(pods []v1.Pod) error {
...
+						return
+					} else if apierrors.IsTooManyRequests(err) {
+						time.Sleep(5 * time.Second)
+					} else {
+						errCh <- errors.Wrapf(err, "error when evicting pod %q", pod.Name)
+						return
+					}
 				}
 			}
 			podArray := []v1.Pod{pod}
```
jackfrancis, line 204: I don't see how this block of code will ever be reached. When we do `break doneEviction` above we will return from the go func entirely.

pkg/operations/cordondrainvm.go
jiaxunsongucb: A "break" statement terminates execution of the innermost "for", "switch" or "select" statement. So in this case, we not only want to break the "select" statement, but also break the outer "for" statement. For more information: https://stackoverflow.com/questions/11104085/does-a-break-statement-break-from-a-switch-select

pkg/operations/cordondrainvm.go
jiaxunsongucb: `break doneEviction` will only break the 'for' loop when the err is nil. If the err is not nil, the block has a chance to be reached.

pkg/operations/cordondrainvm.go
jiaxunsongucb: Done. Thanks!

pkg/operations/cordondrainvm.go
jiaxunsongucb: And `break doneEviction` will not return from the go func entirely. `_, err = o.client.WaitForDelete(o.logger, podArray, true)` will be reached to wait for Pod Deletion.

pkg/operations/cordondrainvm.go
jiaxunsongucb: Yes, this change solves this problem: when timeout, we need to send a signal to the goroutines to stop and return, otherwise, it may be problematic because it will keep trying to evict the Pods forever (based on my observation in my tests). And you are right, this context object will never timeout, however, `cancel` will be invoked in `defer cancel()`, which will stop all the goroutines.

pkg/operations/cordondrainvm.go
jackfrancis: > The scope of a label is the body of the function in which it is declared and excludes the body of any nested function.

Ref: https://golang.org/ref/spec#Label_scopes

`break doneEviction` in this context will break out of the entire go routine. See this go playground example:

https://play.golang.org/p/8MsyEUW1c4m

jiaxunsongucb: The article you mentioned means that a label is not effective outside of the function where the label is defined if I understand it correctly.
You will have to use `WaitGroup` to see the results from goroutines, otherwise, the main method will quit before you see any results from goroutines.
Another thing is, in this PR, the label is in an inner `witch` statement.
Please see this go playground example:
https://play.golang.org/p/q6bBupZk5qg

jackfrancis: You mean `select` instead of "switch", right?

jiaxunsongucb: yes, sorry, `select` is what I meant.

jiaxunsongucb: And if you don't use the label, like this: https://play.golang.org/p/XanDbk87yNX. you will get an infinite for loop.

jackfrancis: All this was due to my misunderstanding of how for labels work (sorry about that), in addition to overly hasty go playground usage.

So I think we're good here.

jackfrancis (approved):
/lgtm

PR #3975, opened 2020-10-27
fix: reserve cpu for windows kubelet for high cpu scenarios

jsturtevant:
**Reason for Change**:

Follow up attempt at fixing the tests that were introduced at #3962 and bug outlined at https://github.com/kubernetes/kubernetes/issues/95735

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jsturtevant:
/hold

Tests are passing. Going to let it soak in the e2e jenkins jobs.

pkg/api/defaults-kubelet.go
```diff
@@ -72,7 +72,7 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 	staticWindowsKubeletConfig["--cloud-config"] = "c:\\k\\azure.json"
 	staticWindowsKubeletConfig["--cgroups-per-qos"] = "false"
 	staticWindowsKubeletConfig["--enforce-node-allocatable"] = "\"\"\"\""
-	staticWindowsKubeletConfig["--system-reserved"] = "memory=2Gi"
+	staticWindowsKubeletConfig["--system-reserved"] = "memory=2Gi,cpu=500m"
```
ravisantoshgudimetla, line 75: How much is the kubelet consuming currently? If system components + kubelet is consuming more than 500mm won't it be a problem? Shouldn't we add `--kube-reserved` too considering those metrics are collected from kubelet?

jsturtevant: > How much is the kubelet consuming currently? 

We don't have these numbers currently across different types of nodes and deployment loads currently. Though through some early testing we did see we need to scale this number as the node size grows.

> If system components + kubelet is consuming more than 500mm won't it be a problem? 

Yes if the node allocatable is fully consumed and kubelet goes over the reserve there is potential for this to occur.  This idea was in the [original design](https://github.com/kubernetes/community/blob/master/contributors/design-proposals/node/node-allocatable.md#kubernetes-reservation-is-smaller-than-kubernetes-component-usage).  It is also the case for other components that run on the node like CRI and CNI.  We are also looking into adding a flag to kubelet that will give kubelet higher priority in cases such as this.  

> --kube-reserved too considering those metrics are collected from kubelet

In Windows I believe `--system-reserve` and `--kubelet-reserve` only subtract from Allocatable and don't have any of the enforcements or concepts of creating cgroups in place that Linux does via `--enforce-node-allocatable` or `--cgroups-per-qos`.  

In this case it is a matter of semantics.  We are attempting to reserve for not only kubelet but other system services and the system services are not grouped into jobs (similiar to cgroups) so chose system-reserve.

ravisantoshgudimetla: > In Windows I believe --system-reserve and --kubelet-reserve only subtract from Allocatable and don't have any of the enforcements or concepts of creating cgroups in place that Linux does via --enforce-node-allocatable or --cgroups-per-qos.

Ahh I see, in case of OpenShift we combine --system-reserve and --kube-reserve to map to a single cgroup, effectively same as subtracting the value from node-allocatable.

> In this case it is a matter of semantics. We are attempting to reserve for not only kubelet but other system services and the system services are not grouped into jobs (similiar to cgroups) so chose system-reserve.

That makes sense.

> Though through some early testing we did see we need to scale this number as the node size grows.

Interesting, do you have a dedicated scale testing environment?

jsturtevant:
Update on Results from the runs in e2e after the change:

date |	v1.17/windows/reproductions/high-cpu |	v1.18/windows/reproductions/high-cpu |	v1.19/windows/reproductions/high-cpu
-- | -- | -- | --
Oct 29 16:39 No Changes | 8min 34s | 8min 32s | 8min 42s
Oct 29 10:39 No Changes | 9min 4s | 10min 22s | 9min 18s
Oct 29 04:39 No Changes | 10min 33s | 9min 28s | 9min 34s
Oct 28 22:39 No Changes | 9min 58s | failed | 11min 46s
Oct 28 16:39 No Changes | 10min 2s | 9min 28s | 9min 47s
Oct 28 14:47 1 commit | 9min 58s | 10min 17s | 12min 33s

You can see one failure which could have been because https://github.com/Azure/aks-engine/pull/3975#discussion_r513949980 and is why this plus setting Priority to high in kubelet would be full solution

jsturtevant:
The results are about the same in e2e.  https://github.com/kubernetes/kubernetes/pull/96051 merged and I will try this out with the next patch release

jsturtevant:
We maybe able to use nssm's AppPriority feature to set the service priority higher for >1.20 https://nssm.cc/usage#process

jsturtevant:
I rebased with the changes in #4073. Will report back e2e tests results

PR #3989, opened 2020-10-30
feat: set apiserver tokenRequest flags in 1.20+

jackfrancis:
**Reason for Change**:

This adds soon-to-be-required apiserver config args in anticipation of this upstream PR:

https://github.com/kubernetes/kubernetes/pull/95896

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-apiserver.go
+++ b/pkg/api/defaults-apiserver.go
@@ -104,6 +104,11 @@ func (cs *ContainerService) setAPIServerConfig() {
 		}
 	}
 
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.20.0-alpha.1") {
+		defaultAPIServerConfig["--service-account-issuer"] = "kubernetes.default.svc"
+		defaultAPIServerConfig["--service-account-signing-key-file"] = "/etc/kubernetes/certs/apiserver.key"
+	}
+
 	// Set default admission controllers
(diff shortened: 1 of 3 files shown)


pkg/api/defaults-apiserver.go
```diff
@@ -104,6 +104,11 @@ func (cs *ContainerService) setAPIServerConfig() {
 		}
 	}
 
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.20.0-alpha.1") {
+		defaultAPIServerConfig["--service-account-issuer"] = "kubernetes.default.svc"
```
ritazh, line 108: From https://github.com/kubernetes/kubernetes/pull/95896/files#diff-97e779171708fff85d7b1ee3fac01336bf351c39066428e12526d0fd7da5c05cR201, it seems we also need `service-account-key-file`.

jackfrancis: We actually already pass that config: https://github.com/Azure/aks-engine/blob/master/pkg/api/defaults-apiserver.go#L31

mboersma (approved):
/lgtm

PR #3993, opened 2020-10-30
chore: set go sdk log level for Azure Stack clusters

JesusAlvarezTorres:
**Reason for Change**:

This change sets the go sdk log level to INFO for clusters in Azure Stack. Info level logs request information made to ARM by the cloud provider; the extra info is needed to correlate cloud provider operations with Azure Stack service logs when investigating issues in cloud provider.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/manifests/kubernetesmaster-kube-controller-manager.yaml
+++ b/parts/k8s/manifests/kubernetesmaster-kube-controller-manager.yaml
@@ -19,6 +19,7 @@ spec:
       env:
       - name: AZURE_ENVIRONMENT_FILEPATH
         value: "/etc/kubernetes/azurestackcloud.json"
+      <gosdkloglevel>
 {{end}}
       volumeMounts:
         - name: etc-kubernetes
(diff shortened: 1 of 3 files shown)


parts/k8s/manifests/kubernetesmaster-kube-controller-manager.yaml
```diff
@@ -19,6 +19,7 @@ spec:
       env:
       - name: AZURE_ENVIRONMENT_FILEPATH
         value: "/etc/kubernetes/azurestackcloud.json"
+      <gosdkloglevel>
```
jackfrancis, line 22: Why can't we simply include the configuration in here explicitly? We're already in a if azure stack scenario here.

cc @jadarsie

jadarsie: Correct, it should be fine to wrap it in a  `{{- if IsAzureStackCloud}}` block.

JesusAlvarezTorres: Made suggested change, please take a look.

PR #3995, opened 2020-11-02
test: fix two UTs to work with "cli" auth

mboersma:
**Reason for Change**:

Locally we've noticed two test cases may fail. This seems to be because there are no required arguments to the `get-locations` and `get-skus` commands if `cli` auth is working.

To meet test expectations, I explicitly added a bogus command-line argument to ensure these tests see an expected error in any case.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


CecileRobertMichon (approved):
/lgtm

PR #4001, opened 2020-11-03
fix: Mount /usr/local/share/ca-certificates folder into kube-controller-manager

ericsuhong:
**Reason for Change**:
To support airgap clouds, we need to mount /usr/local/share/ca-certificates folder into kube-controller-manager

**Credit Where Due**:

- [X] No
- [ ] Yes

- [X] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
kube-controller-manager was working fine in airgap clouds on K8s 1.16 without requiring this change, so I assume this change is only needed for K8s 1.17+


jackfrancis (approved):
/lgtm

jadarsie:
https://github.com/Azure/aks-engine/blob/2e2dfd7b03772f1f8a37c7656d2f32032a658510/parts/k8s/cloud-init/artifacts/cse_customcloud.sh#L16-L23

Why is this ^^^ not enough? The man page for `update-ca-certificates` seems to indicates that the extra mount should not be required: http://manpages.ubuntu.com/manpages/xenial/man8/update-ca-certificates.8.html

ericsuhong:
@jadarsie We encountered this issue on Ubuntu 18.04 with K8s 1.18 in airgap cloud where if you download and update root certificates, cloud-controller won't recognize new root certs until we added that additional mount point.

jadarsie:
right, I understand the symptom, I just wonder why that's the case. 
Is `update-ca-certificates` printing out a warning or error message?
Please let me know if eventually this becomes clear. Thx.

ericsuhong:
I will be honest, I could not pinpoint the root cause exactly. I also faced an issue where it sometimes required explicit node reboot for kube-controller-manager to reflect the cert update as well... (does not happen for other services running on agent nodes). 

update-ca-certificates command did not print out any warning or error messages. It should not require node reboots neither...

PR #3956, opened 2020-10-21
chore: Install Windows Server 2019 10C updates in Windows VHD

marosset:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [x] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #4002, opened 2020-11-04
fix: Re-ordering HNS policy removal due to 10c change in behavior and consolidating logic in Windows cleanup scripts 

jsturtevant:
**Reason for Change**:

First step in refactoring shared logic in startup and windows node reset.  We learned in 10c update (#3956) adds a breaking change  that requires `Get-HnsPolicyList | Remove-HnsPolicyList` to be removed before HNS-Network. We also need turn off kubeproxy while doing this.  

This also adds removed containerd images during the kubelet startup.

We should be able to have a follow up with a change that shares logic between startup and windowsnodereset.  This needs a little more consideration so is not included here.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Requires signed scripts before kicking off tests
/hold

### DIFF
--- a/staging/provisioning/windows/cleanupnetwork.ps1
+++ b/staging/provisioning/windows/cleanupnetwork.ps1
@@ -1,0 +1,67 @@
+$Global:ClusterConfiguration = ConvertFrom-Json ((Get-Content "c:\k\kubeclusterconfig.json" -ErrorAction Stop) | out-string)
+
+$global:NetworkMode = "L2Bridge"
+$global:ContainerRuntime = $Global:ClusterConfiguration.Cri.Name
+$global:NetworkPlugin = $Global:ClusterConfiguration.Cni.Name
+
+ipmo $global:HNSModule
+
+$networkname = $global:NetworkMode.ToLower()
+if ($global:NetworkPlugin -eq "azure") {
+    $networkname = "azure"
+}
+
+$hnsNetwork = Get-HnsNetwork | ? Name -EQ $networkname
+if ($hnsNetwork) {
+    # Cleanup all containers
+    Write-Host "Cleaning up containers"
+    if ($global:ContainerRuntime -eq "containerd") {
+        ctr.exe -n k8s.io c ls -q | ForEach-Object { ctr -n k8s.io tasks kill $_ }
+        ctr.exe -n k8s.io c ls -q | ForEach-Object { ctr -n k8s.io c rm $_ }
+    }
+    else {
+        docker.exe ps -q | ForEach-Object { docker rm $_ -f }
+    }
+    
+    Write-Host "Cleaning up persisted HNS policy lists"
+    # Initially a workaround for https://github.com/kubernetes/kubernetes/pull/68923 in < 1.14,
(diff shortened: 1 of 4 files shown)


staging/provisioning/windows/windowsnodereset.ps1
```diff
@@ -49,6 +49,17 @@ if ($global:EnableHostsConfigAgent) {
 # Perform cleanup
 #
 
+Write-Log "Cleaning up persisted HNS policy lists"
+# Workaround for https://github.com/kubernetes/kubernetes/pull/68923 in < 1.14,
```
marosset, line 53: Should we remove these two workaround comments or say that the October patch impact ALL versions to make sure this doesn't get accidently cleaned up in the future?

marosset:
@jsturtevant should we figure out a way to have kubeletstart.ps1 call into a portion of the windowsnodereset script so the contents will not drift again?

I'm thinking maybe add a -CleanupOnly flag to windowsnodereset.ps1 that just stops services and performs the cleanup. 
kubeletstart.ps1 can call windowsnodereset.ps1 -CleanupOnly then start kubelet.exe

Either that or create a cleanup script that both of these existing scripts calls.

jsturtevant:
Adding a `-CleanupOnly` flag to `windowsnodereset.ps1` should work. I initially didn't do this because I wanted to keep the refactor small. I will give this a go since the I've already factored out the function

jsturtevant:
added `-CleanupOnly` flag, Running tests against new scripts now will report back

jsturtevant:
After testing, the approach of using `-CleanupOnly` didn't work well because of two things:

- log file contention between different processes 
- NodeReset starts/stops several services including kubelet.  If we are in the startup of kubelet then we shouldn't stop our selves.

jsturtevant:
I've updated to use a common file that only cleans up the network and other files. It is up to the caller to turn off the services required.

staging/provisioning/windows/cleanupnetwork.ps1
```diff
@@ -0,0 +1,67 @@
+$Global:ClusterConfiguration = ConvertFrom-Json ((Get-Content "c:\k\kubeclusterconfig.json" -ErrorAction Stop) | out-string)
```
marosset, line 1: This script also deletes container instances so maybe maybe rename it to be more generic?

jsturtevant: removing the containers is part of the network clean up so I think it's fine

jsturtevant:
These changes passed a local test run on 1.19 on a 10c image

```
Ran 45 of 64 Specs in 3187.352 seconds
SUCCESS! -- 45 Passed | 0 Failed | 0 Pending | 19 Skipped
PASS
```

staging/provisioning/windows/cleanupnetwork.ps1
marosset: works for me

marosset:
Whoo - 
image

marosset (approved):

PR #4007, opened 2020-11-05
[WIP] feat: add Blob CSI Driver as an addon

songjiaxun:
**Reason for Change**:

When customers deploy a cluster, give them the choice to enable Blob CSI Driver as an addon.
The Blob CSI Driver is disabled by default.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [X] Yes

- [ ] No
- [X] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
The owner of Blob CSI Driver @andyzhangx is one of the reviewers of this change.
For details about Blob CSI Driver: https://github.com/kubernetes-sigs/blob-csi-driver


jiaxunsongucb:
For CSI Drivers onboarding, we have two approaches: 1) after the cluster is deployed, customers can use `kubectl apply -f yaml files`, or use Helm Charts to deploy the CSI Driver components; 2) add the CSI Driver as a part of AKS Engine addons, so that customers can enable the CSI Driver while deploying the cluster. After the discussion with AKS Engine team, we decided to go with the first approach -- let customers install the driver by themselves. We have two reasons: 1) the payload that is sent to ARM has a size limit, so if the addon list keeps growing, the payload may reach the limit; 2) we don't want to couple the "out-of-tree" CSI Drivers with AKS Engine and Kubernetes, which is against the philosophy of the "out-of-tree" design.
That being said, we want to keep this PR OPEN until we hear customers' feedback.

Issue #2552, opened 2020-01-10
Disable ntp time sync


gossion:
**Describe the request**
currently the default configuration for aks and aks-engine VMs uses both NTP time (ntp.ubuntu.com) and VMICTimeSync host-time to sync the time. The request here is to disable ntp time sync for egress purpose

**Explain why AKS Engine needs it**
1. VMICtimeSync is sufficient enough for time sync, and according to [azure docs](https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync), removing NTP by switching to host-only sync can sometimes improve your time sync results.
1. ntp sync has a lot of egress traffic to ntp.ubuntu.com:123, which in some case is not allowed in customers' environment and generate a lot of error message.

**Describe the solution you'd like**
disable ntp time sync in the image by default and use VMICTimeSync as primary

**Describe alternatives you've considered**
or add a parameter that we can turn off ntp time sync on demand

**Additional context**
About linux time-sync, please refer to https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync

gossion:
@sauryadas

CecileRobertMichon:
Interesting... I wonder if this would also help with https://github.com/Azure/aks-engine/issues/1306 (cc @jackfrancis @palma21)

palma21:
It could, but my plan of action was to actually propose chrony based on the Linux team's experiences and since VMICTimeSync is still mentioned as secondary to ntp on the docs.
That said, it does look like it's better on the latest versions so it might be good to test.

I would love to get @szarkos, @shirgall, @linuxelf001  and @seansp's take on it

linuxelf001:
VMICTimeSync (sync time from the host that is Stratum3 in our time server hierarchy) should be sufficient to sync time even if clusters/vms are spread across different regions.

sauryadas:
@ritazh Can you please assign this to someone to fix?

sukhong:
@ritazh @jackfrancis we need this feature to be able to sync time in air-gapped clouds. Can we re-open this issue?

ericsuhong:
I found out that disabling NTP is not enough for enabling host-only sync mode. If you disable NTP, VMs till only sync from host on when VM is paused or reboot (got a confirmation from Azure Linux team). Chrony is the only way to sync time from host using PTP source.

ericsuhong:
[...]

PR #4011, opened 2020-11-06
feat: Enable chrony and host-based time sync by default on Ubuntu 18.04

ericsuhong:
**Reason for Change**:
Use chrony + host-based time sync which is a recommended way and also work in airgap clouds.

All Azure Ubuntu 18.04 stock images already come with chrony enabled as well.

**Issue Fixed**:
https://github.com/Azure/aks-engine/issues/2552

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Reference: https://docs.microsoft.com/en-us/azure/virtual-machines/linux/time-sync

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -110,6 +110,13 @@ configureEtcd() {
 ensureNTP() {
   systemctlEnableAndStart ntp || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
 }
+configureChrony() {
+  sed -i "s/makestep.*/makestep 1.0 -1/g" /etc/chrony/chrony.conf
+  echo "refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0" >> /etc/chrony/chrony.conf
+}
+ensureChrony() {
+  systemctlEnableAndStart chrony || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
+}
 configPrivateClusterHosts() {
   systemctlEnableAndStart reconcile-private-hosts || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
 }
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -107,9 +107,6 @@ configureEtcd() {
   done
   retrycmd 120 5 25 sudo -E etcdctl member update $MEMBER ${etcd_peer_url} || exit {{GetCSEErrorCode "ERR_ETCD_CONFIG_FAIL"}}
 }
-ensureNTP() {
-  systemctlEnableAndStart ntp || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
-}
 configureChrony() {
   sed -i "s/makestep.*/makestep 1.0 -1/g" /etc/chrony/chrony.conf
   echo "refclock PHC /dev/ptp0 poll 3 dpoll -2 offset 0" >> /etc/chrony/chrony.conf
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -41,7 +41,7 @@ installDeps() {
     packages+=" cgroup-lite ceph-common glusterfs-client"
     if [[ $UBUNTU_RELEASE == "18.04" ]]; then
       disableTimeSyncd
-      packages+=" ntp ntpstat"
+      packages+=" ntp ntpstat chrony"
     fi
   elif [[ $OS == $DEBIAN_OS_NAME ]]; then
     packages+=" gpg cgroup-bin"
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -87,10 +87,12 @@ fi
 {{end}}
 
 if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  if apt list --installed | grep 'chrony'; then
-    time_metric "configureChrony" configureChrony
-    time_metric "EnsureChrony" ensureChrony
-  fi
+  if [[ ! $(apt list --installed | grep 'chrony') ]]; then
+    apt_get_install 30 1 600 chrony || exit 9
+  fi;
+
+  time_metric "ConfigureChrony" configureChrony
+  time_metric "EnsureChrony" ensureChrony
 fi
 
 if [[ $OS == $UBUNTU_OS_NAME ]]; then
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -87,10 +87,6 @@ fi
 {{end}}
 
 if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  if [[ ! $(apt list --installed | grep 'chrony') ]]; then
-    apt_get_install 30 1 600 chrony || exit 9
-  fi;
-
   time_metric "ConfigureChrony" configureChrony
   time_metric "EnsureChrony" ensureChrony
 fi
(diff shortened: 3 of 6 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -110,6 +110,13 @@ configureEtcd() {
 ensureNTP() {
```
jackfrancis, line 110: We can get rid of this `ensureNTP` func definition entirely as the only place that calls it is being deleted

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -41,7 +41,7 @@ installDeps() {
     packages+=" cgroup-lite ceph-common glusterfs-client"
     if [[ $UBUNTU_RELEASE == "18.04" ]]; then
       disableTimeSyncd
```
jackfrancis, line 43: Given that azsecpack enables timesyncd anyways should we just leave the timesyncd config as-is from the Canonical image, and delete the `disableTimeSyncd` foo altogether?

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -41,7 +41,7 @@ installDeps() {
...
-      packages+=" ntp ntpstat"
+      packages+=" ntp ntpstat chrony"
```
jackfrancis, line 44: we definitely wanna keep installing ntp and ntpstat when we're using chrony?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -87,9 +87,12 @@ fi
 {{end}}
 
 if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  if apt list --installed | grep 'ntp'; then
-    time_metric "EnsureNTP" ensureNTP
-  fi
+  if [[ ! $(apt list --installed | grep 'chrony') ]]; then
+    apt_get_install 30 1 600 chrony || exit 9
```
jackfrancis, line 91: I don't think we wanna apt-get install this again (it's being installed in the `installDeps` flow), but instead we wanna do the `configureChrony` and `ensureChrony` steps in this if block

test/e2e/kubernetes/scripts/time-sync-validate.sh
```diff
@@ -16,7 +16,7 @@ if [[ $OS == $UBUNTU_OS_NAME ]]; then
     sudo timedatectl status | grep 'Network time on: yes' || exit 1
     sudo timedatectl status | grep 'NTP synchronized: yes' || exit 1
   elif [[ $RELEASE == "18.04" ]]; then
-    sudo ntpstat | grep 'synchronised to NTP server' || exit 1
+    sudo chronyc sources | grep '#* PHC' || exit 1 # Make sure chrony is running and synced with host-based PTP source clock
```
jackfrancis, line 19: thank you for maintaining tests :)

parts/k8s/cloud-init/artifacts/cse_install.sh
ericsuhong: azsecpack is going to remove that logic (or only conditionally enable timesyncd when chrony is NOT there).

Disabling timesyncd is a must prerequisite for running chrony correctly, so we should keep disabling timesyncd here (aks-engine should not worry about azsecpack).

parts/k8s/cloud-init/artifacts/cse_install.sh
ericsuhong: Given that all previous images already came with ntp, ntpstat pre-installed, I think it is safer to keep them here?

Also, ntp and ntpstat may come handy in the future for debugging stuff in airgap cloud, but I will let you make a decision on whether to keep ntp ntpstat in golden image or not.

test/e2e/kubernetes/scripts/time-sync-validate.sh
ericsuhong, line 19: Just for your reference, this is an output from chronyc sources:

````
root@k8s-master-21855717-vmss000002:~# chronyc sources -v
210 Number of sources = 9
  .-- Source mode  '^' = server, '=' = peer, '#' = local clock.
 / .- Source state '*' = current synced, '+' = combined , '-' = not combined,
| /   '?' = unreachable, 'x' = time may be in error, '~' = time too variable.
||                                                 .- xxxx [ yyyy ] +/- zzzz
||      Reachability register (octal) -.           |  xxxx = adjusted offset,
||      Log2(Polling interval) --.      |          |  yyyy = measured offset,
||                                \     |          |  zzzz = estimated error.
||                                 |    |           \
MS Name/IP address         Stratum Poll Reach LastRx Last sample
===============================================================================
#* PHC0                          0   3     7     9  -9740ns[+1603us] +/-   56us
^- alphyn.canonical.com          2   6    17    20  -1526us[  +87us] +/-   91ms
^- chilipepper.canonical.com     2   6    17    19  -1778us[ -166us] +/-   88ms
^- pugot.canonical.com           2   6    17    18  -2694us[-1082us] +/-   76ms
^- golem.canonical.com           2   6    17    18  -1851us[ -238us] +/-   93ms
^- ntp.mdm.im                    1   6    17    20  -1699us[-1123us] +/-   22ms
^- four10.gac.edu                2   6    17    20  -2973us[-1360us] +/-   85ms
^- li1187-193.members.linod>     2   6    17    20  -2864us[-1251us] +/-   33ms
^- time.gac.edu                  2   6    17    19  -3148us[-1536us] +/-   85ms
````

File: parts/k8s/cloud-init/artifacts/cse_main.sh (hunk 4 above)
Comment (ericsuhong) [reply parts/k8s/cloud-init/artifacts/cse_main.sh:91]: This is to support "bring your own Ubuntu 18.04 flavor image" scenario, and we may need it for airgap cloud until newer VHDs are built and become available.

File: parts/k8s/cloud-init/artifacts/cse_install.sh (hunk 3 above)
Comment (jackfrancis) [reply parts/k8s/cloud-init/artifacts/cse_install.sh:44]: Makes sense to me, let's start by continuing to include ntp (and ntpstat) and see if any gremlins show up when we enable chrony along with it

File: parts/k8s/cloud-init/artifacts/cse_main.sh (hunk 4 above)
Comment (jackfrancis) [reply parts/k8s/cloud-init/artifacts/cse_main.sh:91]: Can you describe that rare scenario? I'm pretty sure installDeps will be called in a BYO 18.04 scenario. If you customize your image from the AKS Engine VHD, it will have already been called during VHD CI; if you customize your image some other way, the `/opt/azure/vhd-install.complete` sentinel file won't exist, and you will invoke `installDeps` at runtime (during CSE).

Comment (ericsuhong) [reply parts/k8s/cloud-init/artifacts/cse_main.sh:91]: Thanks for the explanation, didn't realize installDeps would be called for BYO situation. I will remove these lines.

File: parts/k8s/cloud-init/artifacts/cse_main.sh [hunk 6]
```diff
@@ -87,9 +87,8 @@ fi
 {{end}}
 
 if [[ ${UBUNTU_RELEASE} == "18.04" ]]; then
-  if apt list --installed | grep 'ntp'; then
```
jackfrancis, line 90: In order for this to work on 18.04 prior to building a new VHD w/ chrony already installed is to protect the `configureChrony` and `ensureChrony` invokations inside a `if apt list --installed | grep 'chrony'; then` block.

Otherwise we're going to get a CSE error on all 18.04 VHD nodes (`"distro": "aks-ubuntu-18.04"`):

```
+ sleep 5
+ for i in $(seq 1 $retries)
+ timeout 30 systemctl daemon-reload
+ timeout 30 systemctl restart chrony
Failed to restart chrony.service: Unit chrony.service not found.
+ '[' 100 -eq 100 ']'
+ return 1
+ ret=1
+ systemctl status chrony --no-pager -l
Unit chrony.service could not be found.
+ '[' 1 -ne 0 ']'
+ return 1
+ exit 4
```

jackfrancis:
tested this changeset locally using the non-VHD 18.04 image and am consistently getting a failure in this E2E scenario:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L2724

(not sure why, will have to look into it)

jackfrancis:
cc @Michael-Sinz @jadarsie 

This passes my smoke tests, am planning to merge.

jackfrancis (approved):
/lgtm

PR #4017, opened 2020-11-07
fix: enable dhcpv6 only for ubuntu 16.04

aramase:
**Reason for Change**:

- dhcpv6 systemd unit is only required for ubuntu `16.04`

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/test/e2e/test_cluster_configs/dualstack.json
+++ b/test/e2e/test_cluster_configs/dualstack.json
@@ -2,7 +2,7 @@
     "env": {},
     "options": {
         "allowedOrchestratorVersions": [
-            "latestReleasedVersion"
+            "1.20"
         ]
     },
     "apiModel": {
--- a/test/e2e/test_cluster_configs/dualstack.json
+++ b/test/e2e/test_cluster_configs/dualstack.json
@@ -1,0 +1,59 @@
+{
+    "env": {},
+    "options": {
+        "allowedOrchestratorVersions": [
+            "latestReleasedVersion"
+        ]
+    },
+    "apiModel": {
+        "apiVersion": "vlabs",
+        "properties": {
+            "featureFlags": {
+                "enableIPv6DualStack": true
+            },
+            "orchestratorProfile": {
+                "orchestratorType": "Kubernetes",
+                "kubernetesConfig": {
+                    "loadBalancerSku": "Standard",
+                    "excludeMasterFromStandardLB": true,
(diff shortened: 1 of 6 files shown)


aramase:
@jackfrancis What are your thoughts on enabling dualstack api model for e2e tests? The e2e test suite should work as is for dualstack. Once dualstack is beta in k/k we can add dualstack specific tests to the e2e suite.

jackfrancis:
> @jackfrancis What are your thoughts on enabling dualstack api model for e2e tests? The e2e test suite should work as is for dualstack. Once dualstack is beta in k/k we can add dualstack specific tests to the e2e suite.

Good idea. Should we add a config to the `test/e2e/test_cluster_configs/` directory that looks something like `examples/dualstack/kubernetes.json`? FYI the JSON config is a bit different for the jenkins configs under `test/e2e/test_cluster_configs/ (the actual API model goes into a `"apiModel"` object, for example).

aramase:
>Good idea. Should we add a config to the test/e2e/test_cluster_configs/ directory that looks something like examples/dualstack/kubernetes.json? FYI the JSON config is a bit different for the jenkins configs under test/e2e/test_cluster_configs/ (the actual API model goes into a "apiModel"` object, for example).

I'm going to update the `examples/dualstack/kubernetes.json` to use 1.20 when it's available because dualstack services has been reimplemented in 1.20. The current config we use for testing is https://github.com/aramase/manifests/blob/main/dualstack/iptables-kubenet.json. 

Let me add to `test/e2e/test_cluster_configs`.

test/e2e/test_cluster_configs/dualstack.json
```diff
@@ -0,0 +1,54 @@
+{
+    "env": {},
+    "apiModel": {
```
jackfrancis, line 3: Let's add this between `"env"` and `"apiModel"`:

```
"options": {
		"allowedOrchestratorVersions": ["latestReleasedVersion"]
	},
```

This will make sure it tests 1.19-only, which seems like a good compromise given that it doesn't support 1.16 & 1.17.

aramase: Done!

jackfrancis (approved):
/lgtm

PR #4020, opened 2020-11-10
chore: Add Windows 10c image as default (August optional patches) 

jsturtevant:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [X] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -186,9 +186,9 @@ var (
 	// AKSWindowsServer2019OSImageConfig is the AKS image based on Windows Server 2019
 	AKSWindowsServer2019OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks-windows",
-		ImageSku:       "2019-datacenter-core-smalldisk-2010",
+		ImageSku:       "2019-datacenter-core-smalldisk-2011",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.1397.201006",
+		ImageVersion:   "17763.1554.201109",
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/api/azenvtypes.go
```diff
@@ -186,9 +186,9 @@ var (
 	// AKSWindowsServer2019OSImageConfig is the AKS image based on Windows Server 2019
 	AKSWindowsServer2019OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks-windows",
-		ImageSku:       "2019-datacenter-core-smalldisk-2010",
+		ImageSku:       "2019-datacenter-core-smalldisk-2011",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.1397.201006",
+		ImageVersion:   "17763.1554.201109",
```
marosset, line 191: Just to confirm we want to keep the WindowsServer2019OSImageConfig values below on the Aug patches right?
I think we do and then we can update that with 11B (which should have all the fixes also present in 10C)

marosset (approved):
/lgtm

jackfrancis (approved):
/lgtm

Thank you @jsturtevant getting here was a lot of work!!!

pkg/api/azenvtypes.go
jsturtevant: yes.  Since there is no base image built for 10c and 9b/10b have networking issues.

PR #4021, opened 2020-11-11
fix: upgrade broken for VMAS + managed identities

jadarsie:
**Reason for Change**:

Upgrade fails for VMAS clusters if its VMs own a `roleAssignment` resource.

The problem is that `NormalizeResourcesForK8sMasterUpgrade` is not filtering `roleAssignments` resources and ARM complains because these `roleAssignments` resources depend on missing VM resources. 

stdout message:

```
INFO[0005] Starting ARM Deployment master-20-11-10T15.11.09-1247996574 in resource group upgrade. This will take some time... 
INFO[0006] Error creating upgraded master VM with index: 0 
Error: upgrading cluster: resources.DeploymentsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidTemplate" Message="Deployment template validation failed: 'The resource 'Microsoft.Compute/virtualMachines/k8s-linuxpool-25762572-0' is not defined in the template. Please see https://aka.ms/arm-template for usage details.'." AdditionalInfo=[{"info":{"lineNumber":0,"linePosition":0,"path":""},"type":"TemplateViolation"}]
```

**Notes**:

- Fixed what I think is a problem with the existing tests (nodepool name typo).
- Added a new test to truly validate NormalizeResourcesForK8sMasterUpgrade when upgrading master nodes

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/engine/transform/transform.go
+++ b/pkg/engine/transform/transform.go
@@ -46,6 +46,7 @@ const (
 	vmasResourceType = "Microsoft.Compute/availabilitySets"
 	vmssResourceType = "Microsoft.Compute/virtualMachineScaleSets"
 	lbResourceType   = "Microsoft.Network/loadBalancers"
+	roleResourceType = "Microsoft.Authorization/roleAssignments"
 
 	// resource ids
 	nsgID     = "nsgID"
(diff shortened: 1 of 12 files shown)


pkg/engine/transform/transform.go
```diff
@@ -539,6 +541,26 @@ func (t *Transformer) NormalizeResourcesForK8sMasterUpgrade(logger *logrus.Entry
 					filteredResources = filteredResources[:len(filteredResources)-1]
 				}
 			}
+		} else if resourceType == roleResourceType {
+			logger.Infoln(fmt.Sprintf("Evaluating if agent resource: %s needs to be removed", resourceName))
+			removeRole := true
+			// The usage of NormalizeResourcesForK8sMasterUpgrade across the code base seems to indicate that
+			// agentPoolsToPreserve == nil => master node upgrade
```
jackfrancis, line 548: do you mean if agentPoolsToPreserve has a length of 0?

test/e2e/test_cluster_configs/availabilityset.json
```diff
@@ -1,5 +1,6 @@
 {
 	"env": {
+		"UPGRADE_CLUSTER": true,
```
jackfrancis, line 3: 👍

pkg/engine/transform/transform.go
jadarsie: you are right, I wrote it with this in mind

https://github.com/Azure/aks-engine/blob/v0.57.0/pkg/operations/kubernetesupgrade/upgrader.go#L201

jackfrancis (approved):
/lgtm

PR #4024, opened 2020-11-12
test: add dns-loop (actually curl) stress test

jackfrancis:
**Reason for Change**:

This PR adds a simple, one-time stress test to validate that a bunch of curls to a well-known external URL are responsive.

Nominally these checks actually do a curl, which relies upon the entire HTTP stack, but in practice they will surface high DNS lookups, which we've observed in the past.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @alexeldeib 

and thx again @sturrent!

PR #4037, opened 2020-11-16
[WIP] feat: reworked rotate-certs command

jadarsie:
**Reason for Change**:

This PR tries to resurrect `aks-engine rotate-certs`.

**Notes**:

- **Breaking change**: auth CLI parameters were removed and a few renamed for consistency with `get-logs`
- Updated `client-go` to min supported orchestrator version (v1.16)
- WIP as I will be running more e2e

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests


cmd/rotatecerts/operations.go
```diff
@@ -0,0 +1,126 @@
...
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// RotateServiceAccountTokens deletes service account tokens referenced by kube-system daemonsets/deployments
+// and triggers a rollout of daemonsets/deployments that mount service account tokens.
+//
+// Service account tokens are signed by the cluster CA, deleting them after the CA is rotated ensure that
+// KCM will regenerate tokens signed by the new CA.
+func RotateServiceAccountTokens(client Client, namespaces []string) error {
```
jackfrancis, line 23: Because the `Client` type is an interface, thoughts on unit testing these funcs?

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Certificate Rotation
+
+This document provides guidance on how to rotate certificates on an existing AKS Engine cluster and recommendations for adopting `aks-engine rotate-certs` as a tool.
+
+### Know before you go
+
+In order to ensure that your `aks-engine rotate-certs` operation runs smoothly, there are a few things you should be aware of before getting started.
+
+1. You will need access to the API Model (`apimodel.json`) that was generated by `aks-engine deploy` (by default this file is placed into a relative directory that looks like `_output/<clustername>/`).
```
jackfrancis, line 15: can we say "that was generated by `aks-engine deploy` or `aks-engine generate`" ?

cmd/rotatecerts/operations.go
jadarsie: correct, I am just waiting on collecting feedback before I spend time on UTs

docs/topics/rotate-certs.md
jadarsie: yes, go crazy with suggestions on how to improve the docs

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+
+1. If using `aks-engine rotate-certs` in production, it is recommended to stage a certificate rotation test on an cluster that was built to the same specifications (built with the same cluster configuration + the same version of the `aks-engine` command line tool + the same set of enabled addons) as your production cluster before performing the certificate rotation, especially if the cluster configuration is "interesting", or in other words differs significantly from defaults. The reason for this is that AKS Engine supports many different cluster configurations and the extent of E2E testing that the AKS Engine team runs cannot practically cover every possible configuration. Therefore, it is recommended that you ensure in a staging environment that your specific cluster configuration works with `aks-engine rotate-certs` before attempting this potentially destructive operation on your production cluster.
+
+1. `aks-engine rotate-certs` does **not** guarantees backwards compatibility. If you deployed with `aks-engine` version `0.27.x`, you should prefer executing the certificate rotation process with version `0.27.x`.
```
jackfrancis, line 25: Let's change this hypothetical to something more realistic. Let's aim for including this w/ v0.59.0, so how about document that version?

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Parameters
+
+|Parameter|Required|Description|
+|-----------------|---|---|
+|--api-model|yes|Relative path to the API model (cluster definition) that declares the expected cluster configuration.|
+|--location|yes|Azure location where the cluster is deployed.|
+|--ssh-host|yes|FQDN, or IP address, of an SSH listener that can reach all nodes in the cluster.|
+|--linux-ssh-private-key|yes|Path to a valid private SSH key to access the cluster's Linux nodes.|
+|--certificate-profile|no|Relative path to a JSON file containing the new set of certificates.|
```
jackfrancis, line 35: So this is not required because without it you're essentially telling the aks-engine binary to generate new certs itself based on info in the API model?

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+|--generate-new-certificates|no|Delegate generation of the new set of certificates, required if --certificate-profile is not specified.|
+|--continue|no|Resume a previous unsuccessful execution.|
```
jackfrancis, line 37: What's the case for including this option? Will some users have a reason to "rotate certs from scratch all over again" after a failure, and other users have their reasons for instead "just pick up where you left off" after a failure?

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+|--abort|no|Clean temporary artifacts left behind after an unsuccessful execution.|
```
jackfrancis, line 38: Not sure I understand the purpose of this option

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+  --generate-new-certificates
+```
+
+> Fetching a new set of certificates from Key Vault is not supported at this point.
+
+## Under The Hood
+
+A Kubernetes cluster relies on multiple PKIs to secure the communication between its components (apiserver, kubelet, etcd, etc). On an AKS Engine cluster, these multiple PKIs share a single certificate authoritiy (CA). In order to reduce downtime, `aks-engine rotate-certs` rotates each of these PKIs (etcd, service account tokens, apiserver-kubelet) in sequence.
+
+Each of these are idempotent discrete steps. The cluster is expected to reach a healthy state after each step is executed. If the certificate rotation process halts before completion due to a transient issue (e.x.: network connectivity), rerun `aks-engine rotate-certs` using the `--continue` flag so the process can be resumed at the point where it failed.
```
jackfrancis, line 70: I assume transient network issues will *very rarely* produce error outcomes, because all of the underlying network-dependent operations will retry aggressively, right?

docs/topics/rotate-certs.md
```diff
@@ -0,0 +1,127 @@
...
+
+`aks-engine rotate-certs` requires the execution of multiple remote commands which are subject to potential failures, mostly if the connection to the cluster nodes is not reliable.
+
+Executing `aks-engine rotate-certs` from a VM running on the target cloud (Azure or Azure Stack) can drastically reduce the occurence of transient issues.
+
+## Known Limitations
+
+### Cluster-autoscaler
+
+`aks-engine rotate-certs` will be blocked on deployments that include the `cluster-autoscaler` addon. Because the cluster certificates are embedded in the ARM template, the addon won't be able to produce functioning agent nodes after a certificate rotation operation.
```
jackfrancis, line 123: I think it's safe to assume that AKS Engine + cluster-autoscaler only works on VMSS node pools, in which case it's more correct to say that the "cluster certificates are embedded in the VMSS model".

Also, we should probably just disable cluster-autoscaler during this operation, and re-enable afterwards, right?

parts/k8s/cloud-init/artifacts/rotate-certs.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash -ex
```
jackfrancis, line 1: How is this script delivered to the host VM?

pkg/kubernetes/composite_client.go
```diff
@@ -0,0 +1,352 @@
...
+			Duration: interval,
+			Factor:   1.0,
+			Jitter:   0.0,
+		},
+		retryFunc: func(err error) bool { return true }, // Always retry. Inject if ever needed
+	}
+}
+
+// ListPods returns Pods based on the passed in list options.
+func (c *CompositeClientSet) ListPods(namespace string, opts metav1.ListOptions) (*v1.PodList, error) {
```
jackfrancis, line 45: So each invocation of `ListPods` makes two calls, one for the old client and one for the new client, to its configured apiserver endpoint?

jackfrancis: (and obv. those calls are in parallel)

cmd/rotatecerts/wait.go
```diff
@@ -0,0 +1,156 @@
...
+		successesCount++
+		if successesCount < successesNeeded {
+			return false, nil
+		}
+		return true, nil
+	})
+	return err
+}
+
+// WaitForAllInNamespaceReady returns true if all containers in a given namespace reached the Ready state
```
jackfrancis, line 59: And it looks like these funcs wrap that "let's call both apiservers and assume that the first one we get a response from is the *real* one" thing inside async retry wrapping? So we're basically regularly polling both apiservers until one of them (the valid one) returns a response?

pkg/operations/cordondrainvm.go
```diff
@@ -135,7 +136,8 @@ func daemonSetPodFilter(pod v1.Pod) bool {
 // getPodsForDeletion returns all the pods we're going to delete.  If there are
 // any pods preventing us from deleting, we return that list in an error.
 func (o *drainOperation) getPodsForDeletion() (pods []v1.Pod, err error) {
-	podList, err := o.client.ListPods(o.node)
+	podList, err := o.client.ListPods(metav1.NamespaceAll, metav1.ListOptions{
```
jackfrancis, line 139: This change is owing to a change in the k8s client-go version?

docs/topics/rotate-certs.md
jackfrancis: Let's just name this `--resume` instead

pkg/operations/cordondrainvm.go
jackfrancis: Since this is an Exportable interface that we maintain, we shouldn't change its method signature. Probably should add a new ListPodsByNameSpace (or something) method

docs/topics/rotate-certs.md
jadarsie: partially yes, without it aks-engine binary will generate the new certs, if `--generate-new-certs` is pass.

I am now thinking that the behavior triggered by `--generate-new-certs` can be inferred.

docs/topics/rotate-certs.md
jadarsie: right

docs/topics/rotate-certs.md
jadarsie: taking that idea

parts/k8s/cloud-init/artifacts/rotate-certs.sh
jadarsie: scp, moving files out of `cloud-init` dir to avoid confusion

pkg/kubernetes/composite_client.go
jadarsie: correct

cmd/rotatecerts/wait.go
jadarsie: yes

jadarsie:
Creating a new PR

PR #4038, opened 2020-11-16
chore: deprecate orchestratorType

jackfrancis:
**Reason for Change**:

With #4036 merging, we no longer nominally support any non-Kubernetes orchestrators in any code flows (🎉). This change means that the `orchestratorType` property has no semantic value.

This PR effectively "deprecates" the `orchestratorType` property in the `orchestratorProfile` configuration object:

- defaults to `Kubernetes` where appropriate, so that it no longer needs to be included in API model definitions
- removes several validations whose purpose was to ensure that "orchestratorType == Kubernetes"; because that is *always* true, those validations are superfluous
- updated documentation and example API models

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/examples/e2e-tests/kubernetes/zones/definition.json
+++ b/examples/e2e-tests/kubernetes/zones/definition.json
@@ -1,10 +1,6 @@
 {
     "apiVersion": "vlabs",
     "properties": {
-        "orchestratorProfile": {
-            "orchestratorType": "Kubernetes",
-            "orchestratorRelease": "1.16"
-        },
         "masterProfile": {
             "count": 5,
             "dnsPrefix": "",
--- a/examples/kubernetes-kubenet-containerd.json
+++ b/examples/kubernetes-kubenet-containerd.json
@@ -2,7 +2,6 @@
 	"apiVersion": "vlabs",
 	"properties": {
 		"orchestratorProfile": {
-			"orchestratorType": "Kubernetes",
 			"orchestratorRelease": "1.16",
 			"kubernetesConfig": {
 				"networkPlugin": "kubenet",
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -89,7 +89,9 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 
 	cloudSpecConfig := cs.GetCloudSpecConfig()
 	if a.OrchestratorProfile == nil {
-		return
+		a.OrchestratorProfile = &OrchestratorProfile{
+			OrchestratorType: Kubernetes,
+		}
 	}
 	o := a.OrchestratorProfile
 	o.OrchestratorVersion = common.GetValidPatchVersion(
@@ -561,13 +563,6 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 					if !cs.Properties.MasterProfile.IsVirtualMachineScaleSets() {
 						profile.Subnet = cs.Properties.MasterProfile.Subnet
 					}
-					if cs.Properties.OrchestratorProfile.OrchestratorType == Kubernetes {
-						if !cs.Properties.MasterProfile.IsVirtualMachineScaleSets() {
-							profile.Subnet = cs.Properties.MasterProfile.Subnet
-						}
-					} else {
-						profile.Subnet = fmt.Sprintf(DefaultAgentSubnetTemplate, subnetCounter)
-					}
 					subnetCounter++
 				}
 			}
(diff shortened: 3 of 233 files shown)


examples/e2e-tests/kubernetes/zones/definition.json
```diff
@@ -1,10 +1,6 @@
 {
     "apiVersion": "vlabs",
     "properties": {
-        "orchestratorProfile": {
-            "orchestratorType": "Kubernetes",
-            "orchestratorRelease": "1.16"
```
mboersma, line 6: With some of these removals, you left `"orchestratorRelease": "1.16"` intact, but with others we've removed the whole `"orchestratorProfile"` stanza including the release. I'm assuming the default release is fine for these tests, but I just wanted to point it out and make sure it was intentional.

pkg/api/defaults.go
```diff
@@ -561,13 +563,6 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 					if !cs.Properties.MasterProfile.IsVirtualMachineScaleSets() {
 						profile.Subnet = cs.Properties.MasterProfile.Subnet
 					}
-					if cs.Properties.OrchestratorProfile.OrchestratorType == Kubernetes {
-						if !cs.Properties.MasterProfile.IsVirtualMachineScaleSets() {
```
mboersma, line 565: Did this `profile.Subnet =` logic get moved elsewhere, or was it not needed?

pkg/api/defaults_test.go
```diff
@@ -1632,11 +1563,11 @@ func TestMasterProfileDefaults(t *testing.T) {
 	// this validates default masterProfile configuration
 	mockCS := getMockBaseContainerService("1.13.12")
 	properties := mockCS.Properties
-	properties.OrchestratorProfile.OrchestratorType = Kubernetes
 	properties.OrchestratorProfile.KubernetesConfig.ClusterSubnet = ""
 	properties.OrchestratorProfile.KubernetesConfig.NetworkPlugin = NetworkPluginAzure
 	properties.MasterProfile.AvailabilityProfile = ""
 	properties.MasterProfile.Count = 3
+	properties.CertificateProfile = &CertificateProfile{}
```
mboersma, line 1570: Why did this need to be added to the test (and below)? I'm curious...

examples/kubernetes-kubenet-containerd.json
```diff
@@ -2,7 +2,6 @@
 	"apiVersion": "vlabs",
 	"properties": {
 		"orchestratorProfile": {
-			"orchestratorType": "Kubernetes",
```
mboersma, line 5: Nit: this file could be JSON pretty-printed like the others to use two spaces.

examples/e2e-tests/kubernetes/zones/definition.json
jackfrancis: That's right, I tried to use common sense and remove the release property when it was *only* complemented by the type property. I think there's follow-up work to remove the remaining version properties, unless they are actually meaningful (like, this is the minimum version you need for a particular config).

pkg/api/defaults.go
jackfrancis: Right, this was just crazy code. L563-L565 above is identictal to what is removed.

pkg/api/defaults_test.go
jackfrancis: I'm not sure it's strictly required. Basically the `certsAlreadyPresent`. While I was massaging some of these UT it seemed preferable to use mock certs to avoid generating them at runtime (making tests slower).

jackfrancis: Oh nevermind, these were added to reproduce the existing UT behaviors, where the certs *are* generated at runtime for the bulk of tests. Because I updated the mock data structure to include mock certs (which would bypass on-demand cert generation).

So I think in the future we can consider removing these zero value `CertificateProfile` statements to save UT time, and not generate certs everywhere. But I didn't want to completely remove functional unit testing of on-demand cert generation.

examples/kubernetes-kubenet-containerd.json
jackfrancis: Agree, there's a bunch of different JSON here.

Probably will advocate that we consider getting rid of these examples files and instead add more documentation that refers to the `test_cluster_config` directory, so folks know what cluster configs are regularly being tested.

pkg/api/defaults_test.go
mboersma: Makes sense to me.

mboersma (approved):
/lgtm

PR #3851, opened 2020-09-21
chore: remove no-op 1.15 version checks in templates

mboersma:
**Reason for Change**:

Since AKS Engine already supports only Kubernetes v1.15.x or later, these template conditionals have no effect.

**Issue Fixed**:

Refs #3751 

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [x] No
- [ ] Yes

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

Thanks to @jadarsie for noticing the off-by-one error in #3751, which implies this cleanup.

PR #4040, opened 2020-11-17
chore: remove deprecated AKS code paths

jackfrancis:
**Reason for Change**:

This PR removes all deprecated (no longer used by) AKS code paths, in order to make ongoing maintenance of AKS Engine more streamlined.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @xuto2 @yangl900

pkg/api/defaults.go
```diff
@@ -424,7 +418,7 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 
 		// Upgrade scenario:
 		// We need to force set EnableRbac to true for upgrades to 1.15.0 and greater if it was previously set to false (AKS Engine only)
-		if !a.OrchestratorProfile.KubernetesConfig.IsRBACEnabled() && common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.15.0") && isUpgrade && !cs.Properties.IsHostedMasterProfile() {
+		if !a.OrchestratorProfile.KubernetesConfig.IsRBACEnabled() && common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.15.0") && isUpgrade {
```
mboersma, line 421: Optionally, we could also drop the `common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.15.0")` test here. I guess I missed that in #3851.

cmd/deploy.go
```diff
@@ -283,7 +283,7 @@ func autofillApimodel(dc *deployCmd) error {
 		if dc.containerService.Properties.MasterProfile != nil {
 			dc.outputDirectory = path.Join("_output", dc.containerService.Properties.MasterProfile.DNSPrefix)
 		} else {
-			dc.outputDirectory = path.Join("_output", dc.containerService.Properties.HostedMasterProfile.DNSPrefix)
+			return errors.New("can't determine output directory from nil MasterProfile")
```
mboersma, line 286: Can we add a unit test for this error path to `deploy_test.go`?

cmd/generate.go
```diff
@@ -158,7 +158,7 @@ func (gc *generateCmd) loadAPIModel() error {
 		if gc.containerService.Properties.MasterProfile != nil {
 			gc.outputDirectory = path.Join("_output", gc.containerService.Properties.MasterProfile.DNSPrefix)
 		} else {
-			gc.outputDirectory = path.Join("_output", gc.containerService.Properties.HostedMasterProfile.DNSPrefix)
+			return errors.New("can't determine output directory from nil MasterProfile")
```
mboersma, line 161: Can we add a unit test for this error path to `generate_test.go`?

cmd/rotate_certs.go
```diff
@@ -136,7 +136,7 @@ func (rcc *rotateCertsCmd) run(cmd *cobra.Command, args []string) error {
 		if rcc.containerService.Properties.MasterProfile != nil {
 			rcc.outputDirectory = path.Join("_output", rcc.containerService.Properties.MasterProfile.DNSPrefix)
 		} else {
-			rcc.outputDirectory = path.Join("_output", rcc.containerService.Properties.HostedMasterProfile.DNSPrefix)
+			return errors.New("can't determine output directory from nil MasterProfile")
```
mboersma, line 139: Can we add a unit test for this error path to `rotate_certs_test.go`? Or at least put a pin in it by creating a GitHub issue.

pkg/api/apiloader_test.go
```diff
@@ -487,23 +294,180 @@ func TestSerializeContainerService(t *testing.T) {
   },
   "type": "sampleType",
   "properties": {
-    "kubernetesVersion": "1.11.6",
-    "dnsPrefix": "blueorange",
-    "fqdn": "blueorange.westus2.azure.com",
+    "orchestratorProfile": {
+      "orchestratorType": "Kubernetes",
```
mboersma, line 298: We can drop the `"orchestratorType": "Kubernetes"` line since it's the default after #4038 merged.

mboersma (approved):
/lgtm

My comments are optional feedback.

pkg/api/defaults.go
jackfrancis: Yeah we should do a follow-up workstream *just* to (1) get rid of the unnecessary version stuff, and (2) automate these versions in tests so we don't have to ever again manually maintain them

pkg/api/apiloader_test.go
jackfrancis: We actually can't easily because this is the RAW, generated JSON output that we expect after putting data through a bunch of serialization/deserialization.

We can update the mock data so it doesn't specify the orchestratorType property, but then something in the serialization path renders a `"orchestratorType": ""` in the final JSON, which is (IMO) weirder.

(probably can ultimately be solved but I'm not gonna worry about it)

mboersma: Ah, right. Never mind!

cmd/rotate_certs.go
jackfrancis: Actually I'm going to punt on this one as this whole thing is going to make @jadarsie's life harder and so why not make it even worse.

(mostly) kidding... but tl;dr I'll work with him on rebasing once this lands and then adapting his rewrite of rotate certs to deal with this stuff.

jackfrancis: Added a thumbs down for visual fidelity

mboersma (approved):
/lgtm

encore

PR #4045, opened 2020-11-18
fix: always ensure etcd on control plane vms

jackfrancis:
**Reason for Change**:

Perhaps in the past the `ensureEtcd` func was unnecessary and not reliable in the non-VHD 18.04 scenario, but without it we don't start the etcd-monitor systemd job, which we want to do always.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma (approved):
/lgtm

PR #4051, opened 2020-11-18
fix: persist VMSS name in API model

### LINKED ISSUE #4055: Addpool overwriting existing pool instead of adding new pool
**Describe the bug**
I am seeing an odd behavior around AddPool operation, and just wanted to confirm my expectations for the operation. I am following the example mentioned here https://github.com/Azure/aks-engine/blob/master/docs/topics/addpool.md
 
My expectation was that this would add another VMSS in our cluster RG. However what I am seeing is that the same VMSS is getting changed again and again. The tags on the VMSS (PoolName, AksEngine version etc) change to reflect the new values. If I increase the VM count, it adds another instance in the same VMSS (and vice-versa for instance reduction).
 
If you take a look at our ApiModel, the new AgentPoolProfiles are getting added in the array correctly, but somehow, the ARM deployment is for the same VMSS repeatedly. 
 

**Steps To Reproduce**
Command: aks-engine addpool --api-model .\apimodel.json --auth-method cli --node-pool .\newpool2.json --resource-group eafd-dev-surana57-Rg --location westus2 --subscription-id 5f9e89c2-67fb-4d82-a7c4-af75aa38df00
New Pool template : newpool2.txt 
Api Model : generalApiModel.text
OS: Windows worker Nodes
[newpool2.txt](https://github.com/Azure/aks-engine/files/5567515/newpool2.txt)

[generalApiModel.txt](https://github.com/Azure/aks-engine/files/5567510/generalApiModel.txt)

**Expected behavior**
New VMSS pool should be added. Instead the existing VMSS pool was overwritten with the new pool's config
**AKS Engine version**

55
**Kubernetes version**
1.18.6
**Additional context**
The worker nodes are on Windows OS. I have also tried with using completely different names for new pools and have gotten the same result, so I am not sure if this just VMSS shortening names and colliding.

jackfrancis:
**Reason for Change**:

This PR adds a new `VMSSName` property to the `AgentPoolProfile` type struct, to persist the definitive VMSS name (when the node pool is VMSS) for future cluster operations.

The VMSS name is a computed property that depends in part upon the particular index value of the pool in the API model `agentPoolProfiles` array at the time of cluster creation; but over time that order may change (for example a pool may be removed, introducing a "hole" in the sequence, and effectively decrementing each pool's index by one). This solution continues to rely upon this undesirable solution (let's not throw away the baby with the bath water), but it does so only one time during a cluster operation when the API model data structure is stable, and once the property is computed it is persisted for all time in the API model to match the actual VMSS "name" property, which will remain stable and unique so long as the VMSS exists in the resource group.

This also fixes a particular edge case w/ respect to `aks-engine addpool` and Windows node pools (the `addpool` operation *always* operates against a single `AgentPoolProfile` (the new one), and thus the index is *always* zero: which means if the cluster was created with similarly named Windows pool at that index originally, the `addpool` command will produce an ARM template that actually modifies that existing pool, rather than adding an entirely new one.

**Issue Fixed**:

Fixes #4055 

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -226,6 +226,11 @@ func (sc *scaleCmd) load() error {
 		}
 	}
 
+	// Back-compat logic to populate the VMSSName property for clusters built prior to VMSSName being a part of the API model spec
+	if sc.agentPool.IsVirtualMachineScaleSets() && sc.agentPool.VMSSName == "" {
+		sc.agentPool.VMSSName = sc.containerService.Properties.GetAgentVMPrefix(sc.agentPool, sc.agentPoolIndex)
+	}
+
 	//allows to identify VMs in the resource group that belong to this cluster.
 	sc.nameSuffix = sc.containerService.Properties.GetClusterID()
 	log.Debugf("Cluster ID used in all agent pools: %s", sc.nameSuffix)
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -11,11 +11,11 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
-	"strconv"
 	"strings"
 	"time"
 
 	"github.com/Azure/aks-engine/pkg/armhelpers/utils"
+	"github.com/Azure/go-autorest/autorest/to"
 
 	"github.com/Azure/aks-engine/pkg/api"
 	"github.com/Azure/aks-engine/pkg/armhelpers"
@@ -406,22 +406,12 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 				return errors.Wrap(err, "failed to get VMSS list in the resource group")
 			}
 			for _, vmss := range vmssListPage.Values() {
-				vmssName := *vmss.Name
-				if sc.agentPool.OSType == api.Windows {
-					possibleIndex, err := strconv.Atoi(vmssName[len(vmssName)-2:])
-					if err != nil {
-						continue
-					}
-					if !(sc.containerService.Properties.GetAgentVMPrefix(sc.agentPool, possibleIndex) == vmssName) {
-						continue
-					}
-					winPoolIndex = possibleIndex
+				vmssName := to.String(vmss.Name)
+				if sc.agentPool.VMSSName == vmssName {
+					log.Infof("found VMSS %s in resource group %s that correlates with node pool %s", vmssName, sc.resourceGroupName, sc.agentPoolToScale)
 				} else {
-					if !sc.vmInAgentPool(vmssName, vmss.Tags) {
-						continue
-					}
+					continue
 				}
-				log.Infof("found VMSS %s in resource group %s that correlates with node pool %s", vmssName, sc.resourceGroupName, sc.agentPoolToScale)
 
 				if vmss.Sku != nil {
 					currentNodeCount = int(*vmss.Sku.Capacity)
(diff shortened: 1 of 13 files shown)


jackfrancis:
// TODO update UT

pkg/api/vlabs/types.go
```diff
@@ -543,6 +543,7 @@ type AgentPoolProfile struct {
 	ProximityPlacementGroupID         string            `json:"proximityPlacementGroupID,omitempty"`
 	OSDiskCachingType                 string            `json:"osDiskCachingType,omitempty"`
 	DataDiskCachingType               string            `json:"dataDiskCachingType,omitempty"`
+	VMSSName                          string            `json:"vmssName,omitempty"`
```
CecileRobertMichon, line 546: should this field be documented as READ-ONLY? From my understanding of the code above, any value set by the user will be ignored and get overwritten by the default.

cmd/update.go
```diff
@@ -227,24 +232,17 @@ func (uc *updateCmd) run(cmd *cobra.Command, args []string) error {
...
-				if !sc.vmInAgentPool(vmssName, vmss.Tags) {
-					continue
-				}
+				continue
 			}
 
 			if vmss.Sku != nil {
 				sc.newDesiredAgentCount = int(*vmss.Sku.Capacity)
 				uc.agentPool.Count = sc.newDesiredAgentCount
+				break
```
jsturtevant, line 245: Why is the additional break needed here and wasn't before?  Looks like it will stop processing and kick you out of the for loop.

jackfrancis: Right, so basically the current implementation enumerates through every VMSS *no matter what*. Even after it finds the VMSS that matches the pool in the API model and updates the scale command instance to reflect the current VMSS capacity (because `aks-engine update` is a `aks-engine scale` under the hood that does *not* change the instance count) and updates the API model (a convenience to reflect in the apimodel.JSON what the new count is at the time we run this operation), we keep enumerating through the VMSS array if there are more.

There is always only going to be one matching VMSS, so once we find it, we *should* break out of the enumeration and not waste any more cycles.

So yeah, this is just us improving things since we're mucking around in this area anyways.

Hope that makes sense. (Also definitely check that my reasoning is sound!)

jsturtevant: sounds about right just wanted to make sure it was intentional :-)

Are there other checks in place that affirm `always only going to be one matching VMSS`?  Is there any chance that something could go wrong and there are two which would lead to strange behavior?

jackfrancis: It does look like scale is equivalently inefficient:

https://github.com/Azure/aks-engine/blob/release-v0.58.0/cmd/scale.go#L410

jackfrancis: Azure VMSS API will guarantee uniquess of VMSS names in a resource group; here's how we enforce that in aks-engine:

https://github.com/Azure/aks-engine/blob/release-v0.58.0/pkg/api/vlabs/validate.go#L516

cmd/scale.go
```diff
@@ -441,6 +436,7 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 
 				currentNodeCount = int(*vmss.Sku.Capacity)
 				highestUsedIndex = 0
+				break
```
jackfrancis, line 439: @jsturtevant added break here as well to be consistent

mboersma (approved):
/lgtm

:+1: on the tests.

PR #4053, opened 2020-11-19
chore: cleanup dead code

jackfrancis:
**Reason for Change**:

This PR removes dead code, the result of an audit of `pkg/api/types.go`. Plenty of indirection happened and various parts of the codebase were pruned, but the main story was culling through the type library for no-longer relevant stuffs.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -10,7 +10,6 @@ import (
 	"hash/fnv"
 	"math/rand"
 	"net"
-	neturl "net/url"
 	"sort"
 	"strconv"
 	"strings"
@@ -51,19 +50,6 @@ type ContainerService struct {
 	Properties *Properties `json:"properties,omitempty"`
 }
 
-// AgentPoolResource complies with the ARM model of
-// agentpool resource definition in a JSON template.
-type AgentPoolResource struct {
-	ID       string                `json:"id"`
-	Location string                `json:"location"`
-	Name     string                `json:"name"`
-	Plan     *ResourcePurchasePlan `json:"plan,omitempty"`
-	Tags     map[string]string     `json:"tags"`
-	Type     string                `json:"type"`
-
-	Properties *AgentPoolProfile `json:"properties,omitempty"`
-}
-
 // Properties represents the AKS cluster definition
 type Properties struct {
 	ClusterID               string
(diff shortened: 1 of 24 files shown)


pkg/api/types.go
```diff
@@ -1106,20 +1010,6 @@ func (p *Properties) GetClusterID() string {
 	return p.ClusterID
 }
 
-// GetClusterMetadata returns a instance of the struct type api.ClusterMetadata.
-func (p *Properties) GetClusterMetadata() *ClusterMetadata {
```
mboersma, line 1110: This wasn't used anywhere outside the unit test, eh? So much cruft!

pkg/api/types.go
```diff
@@ -535,15 +479,6 @@ type CustomFile struct {
 	Dest   string `json:"dest,omitempty"`
 }
 
-// BootstrapProfile represents the definition of the DCOS bootstrap node used to deploy the cluster
```
mboersma, line 538: Nice! I must have missed these structs in #3965.

pkg/api/types.go
```diff
@@ -1022,15 +925,16 @@ func (p *Properties) GetPrimaryScaleSetName() string {
...
+		p.OrchestratorProfile.KubernetesConfig.IsAddonEnabled(common.IPMASQAgentAddonName)
 }
 
 // IsIPMasqAgentDisabled returns true if the ip-masq-agent functionality is disabled
 func (p *Properties) IsIPMasqAgentDisabled() bool {
-	if p.OrchestratorProfile != nil && p.OrchestratorProfile.KubernetesConfig != nil {
-		return p.OrchestratorProfile.KubernetesConfig.IsIPMasqAgentDisabled()
-	}
-	return false
+	return p.OrchestratorProfile != nil &&
```
mboersma, line 935: Alternate implementation: `return !p.IsIPMasqAgentEnabled()`.

mboersma (approved):
/lgtm

pkg/api/types.go
jackfrancis: so this is a confusing one that I stumbled around as well

Basically there's a race condition between kubelet and addons defaults enforcement, and they depend on each other for this particular config. So what we want to confirm is that the addon is *explicitly* disabled. Whereas the "is enabled" foo will return false when it's empty/nil/whatever, which is misleading, because if you consider that authoritative and then the addons default foo sets it to true later in the code flow, you're in trouble.

Luckily we have UT to tell this story, and when I did the obvious thing that you suggested they complained!

mboersma: Ah, I was wondering if it was more of a tri-state boolean but I made the naive comment anyway. I don't remember these details for very long apparently. :-)

mboersma: Also all hail the UT.

PR #4062, opened 2020-11-24
chore: validate VHD availability before upgrade/scale on Azure Stack Hub

jadarsie:
**Reason for Change**:

Completing work started in #2342. This change blocks an upgrade operation if any required OS image is not available in the target Azure Stack Hub PIR (Platform Image Repository).

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/scale.go
+++ b/cmd/scale.go
@@ -264,8 +264,10 @@ func (sc *scaleCmd) run(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	if err := sc.validateOSBaseImage(); err != nil {
-		return errors.Wrapf(err, "validating OS base images required by %s", sc.apiModelPath)
+	if sc.containerService.Properties.IsAzureStackCloud() {
+		if err := sc.validateOSBaseImage(); err != nil {
+			return errors.Wrapf(err, "validating OS base images required by %s", sc.apiModelPath)
+		}
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/scale.go
```diff
@@ -654,3 +658,15 @@ func (sc *scaleCmd) printScaleTargetEqualsExisting(currentNodeCount int) {
 		}
 	}
 }
+
+// validateOSBaseImage checks if the OS image is available on the target cloud (ATM, Azure Stack only)
+func (sc *scaleCmd) validateOSBaseImage() error {
+	if sc.containerService.Properties.IsAzureStackCloud() {
```
jackfrancis, line 664: I would prefer we put this "if Azure Stack" guard around its invocation (inside the run method above) instead of inside this method itself.

cmd/upgrade.go
```diff
@@ -386,3 +390,15 @@ func isVMSSNameInAgentPoolsArray(vmss string, cs *api.ContainerService) bool {
 	}
 	return false
 }
+
+// validateOSBaseImage checks if the OS image is available on the target cloud (ATM, Azure Stack only)
+func (uc *upgradeCmd) validateOSBaseImage() error {
+	if uc.containerService.Properties.IsAzureStackCloud() {
```
jackfrancis, line 396: ditto

jackfrancis (approved):
/lgtm

PR #4067, opened 2020-11-30
fix: cleanup VMSSName + addpool

jackfrancis:
**Reason for Change**:

This PR cleans up some things from #4051, specifically:

- Use the proper "next array index" value when calculating the VMSSName property during `aks-engine addpool`
- Clarify that the above is not for back-compat: this property computation will *always* occur during `aks-engine addpool`
- improve the clarity of the "generate a VMSS prefix string based on AgentPoolProfile data within the API model" foo
  - and deprecate the unused "Windows name v2" thing

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/addpool.go
+++ b/cmd/addpool.go
@@ -149,6 +149,9 @@ func (apc *addPoolCmd) load() error {
 		numExistingPools := len(apc.containerService.Properties.AgentPoolProfiles)
 		// we can reuse the value of numExistingPools due to array index beginning at "0"
 		apc.nodePool.VMSSName = apc.containerService.Properties.GetAgentVMPrefix(apc.nodePool, numExistingPools)
+		if apc.nodePool.VMSSName == "" {
+			return errors.Errorf("unable to compute a VMSSName property value from new pool definition")
+		}
 	}
 
 	if apc.containerService.Properties.IsCustomCloudProfile() {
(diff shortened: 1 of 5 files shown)


cmd/addpool.go
```diff
@@ -144,11 +144,14 @@ func (apc *addPoolCmd) load() error {
...
-	if apc.nodePool.IsVirtualMachineScaleSets() && apc.nodePool.VMSSName == "" {
-		existingPools := len(apc.containerService.Properties.AgentPoolProfiles)
-		newIndex := existingPools + 1
-		apc.nodePool.VMSSName = apc.containerService.Properties.GetAgentVMPrefix(apc.nodePool, newIndex)
+	// Assign VMSSName property based on the new pool being added to the end of the existing AgentPoolProfiles array
+	if apc.nodePool.IsVirtualMachineScaleSets() {
+		numExistingPools := len(apc.containerService.Properties.AgentPoolProfiles)
+		// we can reuse the value of numExistingPools due to array index beginning at "0"
+		apc.nodePool.VMSSName = apc.containerService.Properties.GetAgentVMPrefix(apc.nodePool, numExistingPools)
+		if apc.nodePool.VMSSName == "" {
```
jackfrancis, line 152: We don't know what to do if we aren't able to generate a VMSSName, so we'll error out

mboersma (approved):
/lgtm

PR #3922, opened 2020-10-12
fix: Azure Stack upgrade operations clear old MCR ImageBase

jadarsie:
**Reason for Change**:

Clear the deprecated MCR image base if upgrading an Azure Stack cluster.

An AzureStack-specific MCR repository was referenced by clusters created using older AKS Engine releases. 
New pause images wont be pushed to that repo.

**Issue Fixed**:
Related work #3689

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [x] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
- [ ] includes documentation
- [x] adds unit tests
- [x] tested upgrade from previous version

**Notes**:

PR #4068, opened 2020-11-30
chore: limit number of upgrade retries if new CP nodes bootstrap fails

jadarsie:
**Reason for Change**:

Most of the failures that cause `aks-engine upgrade` to fail are retryable. However bad or incompatible `properties.orchestratorProfile.kubernetesConfig` property values in your API model will prevent `kubelet` (or control plane components) from starting successfully. In this case, `aks-engine upgrade` assumes that the node was upgraded successfully because it only checks for the `orchestrator` tag.

This PR limits the number of upgrade operations if two upgraded CP nodes are not ready.

**Notes**:

- Updated client-go to min supported orchestrator version (v1.16)
- Target `tools-install` installs `gomock`

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [x] adds unit tests


docs/topics/upgrade.md
```diff
@@ -218,3 +218,16 @@ You *may not* change the following values, doing so may break your cluster!
...
+### How can I recreate upgraded nodes that did not reach the `Ready` state?
+
+If any of the upgraded nodes is not able to reach the `Ready` state, then `aks-engine upgrade` will exit showing a message similar to one below:
+
+```console
+Error validating upgraded master VM: k8s-master-12345678-0
+Error: upgrading cluster: Node was not ready within 20m0s
+```
+
+These errors are typically caused by bad or incompatible `properties.orchestratorProfile.kubernetesConfig` property values in your API model that prevent `kubelet` (or control plane components) from starting successfully.
```
jackfrancis, line 231: What makes us think this is the most common cause of kubelet failing to come online? If we have definitive proof of these errors, we should be able to validate known incompatibilites during template generation/validation.

jadarsie: The instance when I have seen this happening was upgrading ASH clusters from old AKS Engine versions, the issue was an incorrect pause image. That issue is fixed now by this #3922. 

That statement is meant to hint users to check around that area if nodes are not ready after upgrade. Let me know how do you prefer to document this.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -185,7 +195,7 @@ func (uc *UpgradeCluster) getUpgradeWorkflow(kubeConfig string, aksEngineVersion
 	return u
 }
 
-func (uc *UpgradeCluster) getClusterNodeStatus(kubeClient kubernetes.Client, resourceGroup string) error {
+func (uc *UpgradeCluster) getClusterVMStatus(kubeClient kubernetes.Client, resourceGroup string) error {
```
jackfrancis, line 198: I read through this method, and I think a clearer new name would be something like `setNodesToUpgrade`.

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -101,10 +103,18 @@ func (uc *UpgradeCluster) UpgradeCluster(az armhelpers.AKSEngineClient, kubeConf
...
 	}
 
-	if err := uc.getClusterNodeStatus(kubeClient, uc.ResourceGroup); err != nil {
+	if err := uc.getClusterVMStatus(kubeClient, uc.ResourceGroup); err != nil {
 		return uc.Translator.Errorf("Error while querying ARM for resources: %+v", err)
 	}
 
+	if kubeClient != nil {
+		mastersCount := uc.DataModel.Properties.MasterProfile.Count
+		if err := uc.getClusterNodeStatus(kubeClient.ListNodesByOptions, mastersCount); err != nil {
```
jackfrancis, line 112: This call to the new `getClusterNodeStatus` method does not allow for any retries. Because a Node can be temporarily in a NotReady state, we want to be able to tolerate transient NotReady-ness. How are we going to do so?

jadarsie: Right now it is not handled. Because all this happens before any changes to the resources/nodes state, the reasoning was to let a human characterize the failure and decide what to do about it. I can add a retry.

jackfrancis: I think it's fine, we don't need to add a retry just because. But FYI in case this actually reduces folks' ability to programmatically upgrade clusters (let's say you're using a script that quickly re-runs upgrade in a failure scenario)

docs/topics/upgrade.md
jackfrancis: I would simply suggest "There are a variety of reasons why your cluster might not be able to come back online after an upgrade. Looking at the kubelet logs (`sudo journalctl -u kubelet`) may be a good first investigative step."

Or something like that.

pkg/operations/kubernetesupgrade/upgradecluster.go
jadarsie: Added retry with a 2.5 mins timeout

pkg/kubernetes/interfaces.go
```diff
@@ -21,6 +22,8 @@ type Client interface {
 	ListAllPods() (*v1.PodList, error)
 	// ListNodes returns a list of Nodes registered in the api server.
 	ListNodes() (*v1.NodeList, error)
+	// ListNodesByOptions returns a list of Nodes registered in the api server.
+	ListNodesByOptions(opts metav1.ListOptions) (*v1.NodeList, error)
```
jackfrancis, line 26: changing this publicly exportable interface definition is dangerous FYI

jadarsie: Adding methods should be fine, right?

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -101,10 +104,21 @@ func (uc *UpgradeCluster) UpgradeCluster(az armhelpers.AKSEngineClient, kubeConf
 		kubeClient = k
 	}
 
-	if err := uc.getClusterNodeStatus(kubeClient, uc.ResourceGroup); err != nil {
+	if err := uc.setNodesToUpgrade(kubeClient, uc.ResourceGroup); err != nil {
 		return uc.Translator.Errorf("Error while querying ARM for resources: %+v", err)
 	}
 
+	if kubeClient != nil {
```
jackfrancis, line 111: Do we need this nil check? We should be confident that a non-err response from L100 above will guarantee that `k` is not nil, right?

jackfrancis: If we're not confident then I'd just add that nil check to the err condition:

`if err != nil || k == nil {`

pkg/operations/kubernetesupgrade/upgradecluster_test.go
```diff
@@ -1071,3 +1076,237 @@ var _ = Describe("Upgrade Kubernetes cluster tests", func() {
 		Expect(len(newNode.Spec.Taints)).To(Equal(2))
 	})
 })
+
+func TestCheckControlPlaneNodesStatus(t *testing.T) {
```
jackfrancis, line 1080: ❤️

pkg/operations/kubernetesupgrade/upgradecluster.go
```diff
@@ -457,3 +471,80 @@ func (uc *UpgradeCluster) addVMToFinishedSets(vm compute.VirtualMachine, current
...
+				break loop
+			}
+			upgradedNotReadyCount = len(upgradedNotReady)
+		case <-ctx.Done():
+			break loop
+		}
+	}
+	// return error if more than 1 upgraded node is not ready
+	// do not bother if upgradedNotReadyCount == mastersCount, at that point upgrade cannot take down more nodes
+	if upgradedNotReadyCount > 1 && upgradedNotReadyCount < mastersCount {
```
jackfrancis, line 501: Why the extra `&&` condition here? Shouldn't we always return error if there are more than one not ready control plane VM?

pkg/operations/kubernetesupgrade/upgradecluster.go
jadarsie: the only reason is to pass test `Should not fail if a Kubernetes client cannot be created`, do not know the history so I am just complying

pkg/operations/kubernetesupgrade/upgradecluster.go
jadarsie: you are right, I thought I had a good reason but on a second thought it is actually not

jackfrancis (approved):
/lgtm

PR #4072, opened 2020-12-01
fix: filter out KeyVault resources during upgrade

### LINKED ISSUE #4071: Upgrade fails if enableEncryptionWithExternalKms is true
**Describe the bug**

Upgrading a cluster with more than one master VM, and with `enableEncryptionWithExternalKms` set to `true`, yields a consistent error:

```
Error: upgrading cluster: resources.DeploymentsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidTemplate" Message="Deployment template validation failed: 'The resource 'Microsoft.Compute/virtualMachines/k8s-master-23943353-1' is not defined in the template. Please see https://aka.ms/arm-template for usage details.'." AdditionalInfo=[{"info":{"lineNumber":0,"linePosition":0,"path":""},"type":"TemplateViolation"}]
```

jackfrancis:
**Reason for Change**:

This PR addresses edge case behavior in `aks-engine upgrade` when more than one master node and `enableEncryptionWithExternalKms` is true.

**Issue Fixed**:

Fixes #4071

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jadarsie:
I would just add UTs to avoid future regressions, the template manipulation code is not a pleasant read.

I would try to call`RemoveKMSResourcesFromTemplate` unconditionally from `NormalizeResourcesForK8sMasterUpgrade`.

Then, `TestNormalizeResourcesForK8sMasterUpgrade` should cover this scenario.

pkg/engine/transform/transformtestfiles/k8s_slb_template.json
```diff
@@ -2298,7 +2298,8 @@
           "name": "vmLoopNode"
         },
         "dependsOn": [
-          "[concat('Microsoft.Compute/virtualMachines/', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')))]"
+          "[concat('Microsoft.Compute/virtualMachines/', variables('masterVMNamePrefix'), copyIndex(variables('masterOffset')))]",
+          "[concat('Microsoft.KeyVault/vaults/', variables('clusterKeyVaultName'))]"
```
jackfrancis, line 2302: This demonstrates that this dependency will be removed if it's found (`k8s_slb_template.json` is the input file before we do the translation/filtering)

PR #4073, opened 2020-12-01
feat: Updating kubelet/kube-proxy to run with as high priority processes

marosset:
**Reason for Change**:

**Issue Fixed**:

This should help address many customer reported issues we have been seeing where kubelet get starved of CPU cycles when node is running under high CPU load.

I tried creating the Process object in native powershell but saw some pretty horrible lag processing the OutputDataReceived/ErrorDataReceived events so I think this is probably the most performant thing to do.

**Note**: Once kubelet/kube-proxy both have flags to set the process priority we can remove / undo this code.

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/staging/provisioning/windows/kubeletstart.ps1
+++ b/staging/provisioning/windows/kubeletstart.ps1
@@ -250,11 +250,11 @@ if ($global:NetworkPlugin -eq "kubenet") {
     catch {
         Write-Error $_
     }
-
-    # Start the kubelet
-    # Use run-process.cs to set process priority class as 'High'
-    Add-Type -Path .\run-process.cs
-    $exe = "$global:KubeDir\kubelet.exe"
-    $args = ($KubeletArgList -join " ")
-    [RunProcess.exec]::RunProcess($exe, $args)
 }
+
+# Start the kubelet
+# Use run-process.cs to set process priority class as 'High'
+Add-Type -Path .\run-process.cs
+$exe = "$global:KubeDir\kubelet.exe"
+$args = ($KubeletArgList -join " ")
+[RunProcess.exec]::RunProcess($exe, $args)
--- a/staging/provisioning/windows/kubeletstart.ps1
+++ b/staging/provisioning/windows/kubeletstart.ps1
@@ -253,8 +253,14 @@ if ($global:NetworkPlugin -eq "kubenet") {
 }
 
 # Start the kubelet
-# Use run-process.cs to set process priority class as 'High'
-Add-Type -Path .\run-process.cs
+# Use run-process.cs to set process priority class as 'AboveNormal'
+# Load a signed version of runprocess.dll if it exists for Azure SysLock compliance
+# otherwise load class from cs file (for CI/testing)
+if (Test-Path "$global:KubeDir\runprocess.dll") {
+    [System.Reflection.Assembly]::LoadFrom("$global:KubeDir\runprocess.dll")
+} else {
+    Add-Type -Path "$global:KubeDir\run-process.cs"
+}
 $exe = "$global:KubeDir\kubelet.exe"
 $args = ($KubeletArgList -join " ")
-[RunProcess.exec]::RunProcess($exe, $args)
+[RunProcess.exec]::RunProcess($exe, $args, [System.Diagnostics.ProcessPriorityClass]::AboveNormal)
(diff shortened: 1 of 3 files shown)


marosset:
Putting this PR in draft mode until I get a chance to do some more manual validation tomorrow.

staging/provisioning/windows/kubeletstart.ps1
```diff
@@ -247,12 +246,15 @@ if ($global:NetworkPlugin -eq "kubenet") {
...
     }
-
 }
+
+# Start the kubelet
+# Use run-process.cs to set process priority class as 'High'
+Add-Type -Path .\run-process.cs
+$exe = "$global:KubeDir\kubelet.exe"
+$args = ($KubeletArgList -join " ")
+[RunProcess.exec]::RunProcess($exe, $args)
```
jsturtevant, line 260: Can this be made optional so in 1.20+ we can use the new `--windows-priorityclass`?

jsturtevant:
We can use this in combination with the `system-reserve`. There is an e2e test enabled on https://github.com/Azure/aks-engine/pull/3975 that has shown improvement but this could help resolve it more fully.   Can we either include this change in that PR or build on top of that one?  We can run the tests for a few iterations to see if it improves the results there.

staging/provisioning/windows/kubeletstart.ps1
marosset: Do you mind if we do that in a follow up PR

marosset:
I finished the manual verifications I wanted to do.
When I log in to a node i see

```powershell
PS C:\Users\azureuser> Get-Process -Name kube* | select -Property Name, PriorityClass

Name       PriorityClass
----       -------------
kubelet             High
kube-proxy          High
```

staging/provisioning/windows/run-process.cs
```diff
@@ -0,0 +1,48 @@
...
+
+            p.OutputDataReceived += new DataReceivedEventHandler(StdOutHandler);
+            p.ErrorDataReceived += new DataReceivedEventHandler(StdErrHandler);
+
+            p.Start();
+            p.BeginOutputReadLine();
+            p.BeginErrorReadLine();
+
+            // I don't know why this sleep is needed but it is :-/
+            Thread.Sleep(2000);
```
jsturtevant, line 41: This maybe be related to races with child processes https://stackoverflow.com/questions/1010370/how-do-i-launch-a-process-with-low-priority-c-sharp#comment103915681_33158557

staging/provisioning/windows/kubeletstart.ps1
jsturtevant: Sounds good. We can do that when we turn on the flag and enable it in 1.20

AbelHu, line 260: Will it also set a high priority for all child processes created by kubelet, for example, cni plugin?

marosset: Great question.
I'll try setting the (BasePriority)[https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.process.basepriority?view=net-5.0] for the process as well.
The docs say that controls the starting priority of new threads in the process. 
I don't know the details of how go go starts new processes for Windows or this will have an effect but it might be worth testingout.

marosset: It does not look like there is a way for Windows to set the priority of new processes.

jsturtevant:
looks good to me

I opened https://github.com/Azure/aks-engine/issues/4130 for updating the use of the flag for appropriate versions

Will the signing and update of the packages on the VHD be done once this merged?

staging/provisioning/windows/kubeletstart.ps1
AbelHu: Thanks for the info. I am curious what will happen if there is no enough cpu resource for CNI plugin. I think that it should be ok if the IP cannot be allocated so the new container cannot be created. But can the container be terminated successfully if the IP cannot be released successfully?

marosset:
> Will the signing and update of the packages on the VHD be done once this merged?

Yes, i've updated our signing pipeline to build  .dll from run-process.cs and sign that.
I'll kick it off once this merged.

staging/provisioning/windows/kubeletstart.ps1
marosset: I'm not exactly sure what would happen.

@kevpar proposed a solution where a process can create an NT job object, set a priority class to that job object, then add itself to that job object. That solution should cause processes spawned by kubelet/kube-proxy/containerd to run with higher priority.

I'd like to try and get these changes merged and iterate if needed.
I think a better solution would be to have each process implement the above behavior and then we can remove the run-process work at a future time.

jsturtevant (approved):

PR #4076, opened 2020-12-01
feat: add support for Kubernetes v1.20.0-rc.0

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.20.md#changelog-since-v1200-beta2

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/common/versions.go
+++ b/pkg/api/common/versions.go
@@ -232,6 +232,8 @@ var AllKubernetesSupportedVersions = map[string]bool{
 	"1.20.0-alpha.3": true,
 	"1.20.0-beta.0":  true,
 	"1.20.0-beta.1":  true,
+	"1.20.0-beta.2":  false,
+	"1.20.0-rc.0":    true,
 }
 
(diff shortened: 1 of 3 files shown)


jackfrancis:
@mboersma @cpuguy83 FYI, this PR fixed the DNS check thing:

https://github.com/Azure/aks-engine/pull/4080

tl;dr running nslookup without a targeted option (like `-type=A`) takes a long time and the livenessProbe foo times out. This is probably a change in the underlying livenessProbe implementation and this change should be investigated, but we can also be good livenessProbe citizens and optimize the way we use nslookup (I assume the long time is due to ipv6 DNS requests by default, which we don't want anyways).

cc @aramase if you're aware of any changes where livenessProbe will throw a context deadline timeout exception if a command takes a long time (~a few seconds)

jackfrancis (approved):
/lgtm

mboersma:
> this PR fixed the DNS check thing

Wow, we got lucky for a change. 🍀

PR #4089, opened 2020-12-04
test: enable GINKGO_SKIP_AFTER_UPGRADE param in e2e test

haofan-ms:
**Reason for Change**:

GINKGO_SKIP_AFTER_UPGRADE was not used before, and the SKIP_AFTER_UPGRADE value was determined by other params, this change will enable it's value to be used

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/cluster.sh
```diff
@@ -102,17 +102,17 @@ if [ -n "${GINKGO_SKIP}" ]; then
     SKIP_AFTER_SCALE_UP="${GINKGO_SKIP}"
   fi
   if [ "${SCALE_CLUSTER}" = "true" ]; then
```
jadarsie, line 104: should not this be `if [ -n "${GINKGO_SKIP_AFTER_UPGRADE}" ]; then`

test/e2e/cluster.sh
```diff
@@ -102,17 +102,17 @@ if [ -n "${GINKGO_SKIP}" ]; then
...
-    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_SCALE_DOWN}"
+    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_SCALE_DOWN}|${SKIP_AFTER_UPGRADE}"
```
jadarsie, line 105: I think you should use `GINKGO_SKIP_AFTER_UPGRADE`

`SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_SCALE_DOWN}|${GINKGO_SKIP_AFTER_UPGRADE}"`

test/e2e/cluster.sh
```diff
@@ -102,17 +102,17 @@ if [ -n "${GINKGO_SKIP}" ]; then
...
   else
-    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}"
+    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_UPGRADE}"
```
jadarsie, line 107: I think this one should not be change

test/e2e/cluster.sh
```diff
@@ -102,17 +102,17 @@ if [ -n "${GINKGO_SKIP}" ]; then
...
   fi
 else
   SKIP_AFTER_SCALE_DOWN="${GINKGO_SKIP_AFTER_SCALE_DOWN}"
   SKIP_AFTER_SCALE_UP="${GINKGO_SKIP_AFTER_SCALE_UP}"
   if [ "${SCALE_CLUSTER}" = "true" ]; then
-    SKIP_AFTER_UPGRADE="${SKIP_AFTER_SCALE_DOWN}"
+    SKIP_AFTER_UPGRADE="${SKIP_AFTER_SCALE_DOWN}|${SKIP_AFTER_UPGRADE}"
   else
-    SKIP_AFTER_UPGRADE=""
+    SKIP_AFTER_UPGRADE="${SKIP_AFTER_UPGRADE}"
```
jadarsie, line 115: Same as above

jadarsie (commented):
@jackfrancis could you please comment on the history of these lines, why is GINKGO_SKIP  different for upgrade operations?

test/e2e/cluster.sh
haofan-ms: My understanding is that the logic here is planned for the scale_down -> upgrade -> scale_up flow, and thus if there is "SCALE_CLUSTER" param (which means scale down & scale operations will be executed), then the upgrade_skip should also include scale_down_skip, otherwise it's just upgrade_skip.

test/e2e/cluster.sh
haofan-ms: Thanks, will update

jackfrancis (commented):
lgtm

worth doing lots of tests on this change!

jackfrancis:
@jadarsie the reason for all this craziness is to allow the vanilla "skip these tests" env var to apply relevantly to ginkgo tests after scale and upgrade *without having to provide* post-upgrade-specific or post-scale-specific skip inputs.

Basically, we want both the above automatic foo, but we also *do* want to allow post-upgrade- and/or post-scale-specific inputs as well. Bash is not the best place to implement these types of hierarchical prototype thingers, but alas...

(also, it does seem like this hasn't been revisited since addpool and update have been introduced, so those skip implementations are kind of limited atm; just fyi if you encounter future test criteria that is stumped by this)

test/e2e/cluster.sh
```diff
@@ -102,17 +102,17 @@ if [ -n "${GINKGO_SKIP}" ]; then
     SKIP_AFTER_SCALE_UP="${GINKGO_SKIP}"
   fi
   if [ "${SCALE_CLUSTER}" = "true" ]; then
-    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_SCALE_DOWN}"
+    SKIP_AFTER_UPGRADE="${GINKGO_SKIP}|${SKIP_AFTER_SCALE_DOWN}|${GINKGO_SKIP_AFTER_UPGRADE}"
```
jadarsie, line 105: I think we need to check if `GINKGO_SKIP_AFTER_UPGRADE` is set and not empty before we use it here

haofan-ms: Updated

PR #4105, opened 2020-12-09
ci: refactor Windows VHD build pipeline updates

marosset:
**Reason for Change**:

- CI job now builds and tests 2019/docker and 2019/containerd in parallel
- Adding a new build pipeline template to build AND publish multiple VHDs at once 

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
I've been testing the build-and-publish script using a dry-run mode for PUB.
Once #4103 merges I'll use this build-and-publish pipeline to actually try and publish new images.

vhd/packer/configure-windows-vhd.ps1
```diff
@@ -336,3 +336,6 @@ switch ($env:ProvisioningPhase) {
         exit 1
     }
 }
+
+
+
```
jsturtevant, line 341: nit: extra lines?

CecileRobertMichon (commented):
When you say "publish", does this also hit the publish button?

marosset:
> When you say "publish", does this also hit the publish button?

Not yet but we hope to have this hit the publish and go-live buttons at some point in the near future.

vhd/packer/configure-windows-vhd.ps1
marosset: Will remove.

.pipelines/vhd-build-publish-windows.yaml
```diff
@@ -0,0 +1,38 @@
+trigger: none
+pr: none
+
+stages:
+
+  - stage: build_2019_containerd
+    dependsOn: []
```
jsturtevant, line 7: Why separate parallel Stages vs [job matrix](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#multi-job-configuration)?

vhd/publishing/new-sku-and-add-image-version.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+required_env_vars=(
```
jsturtevant, line 3: This script looks to be a combination of the two others. Are they still needed or they be removed?

.pipelines/vhd-build-publish-windows.yaml
marosset: A few reasons 
- I found a lot more examples of multi-stage pipelines vs using job matrix
- So that we can opt to run only some stages when triggering the pipeline without needed to update the yml (like it we only needed to build/relesase a subsete of VHDs)
- I plan on adding in a manual gate / approval step which I think needs to be its own stage

vhd/publishing/new-sku-and-add-image-version.sh
marosset: I think AKS/agent-baker is still using the other two scripts, I want to confirm with them before removing them (but also don't want block our progress here)

jsturtevant: @AbelHu @mainred do you need those files or does agent baker have their own copies now?

marosset:
The e2e tests which validate the VHD that just got built are failing for containerd because the cluster is configured to use docker and are failing. Let me update this so we'll use containerruntime==containerd when verifying the containerd VHD.

vhd/publishing/new-sku-and-add-image-version.sh
mainred: @jsturtevant  @marosset  thanks for the heads-up. We have copies of these files. For building Windows VHD pipeline,  AKS/agentbaker has no dependency on aks-engine except the signed scripts and collect-windows-logs.ps1

AbelHu: @mainred It seems like that this file can create sku automatically. I think that we should copy some work from here.

vhd/publishing/new-sku-and-add-image-version.sh
```diff
@@ -0,0 +1,73 @@
...
+    echo "Could not find sku template file: ${SKU_TEMPLATE_FILE}!"
+    exit 1
+fi
+
+(set -x; ls -lhR artifacts )
+
+VHD_INFO="artifacts/vhd/windows-vhd-publishing-info.json"
+
+if [ ! -f "$VHD_INFO" ]; then
+    echo "Could not find sku template file: ${VHD_INFO}!"
```
mainred, line 32:
```suggestion
    echo "Could not find VHD info file: ${VHD_INFO}!"
```

vhd/publishing/new-sku-and-add-image-version.sh
```diff
@@ -0,0 +1,73 @@
...
+fi
+
+short_date=$(date +"%y%m")
+pretty_date=$(date +"%b %Y")
+
+sku_id="${SKU_PREFIX}-${short_date}"
+
+echo "Checking if offer contains SKU: $sku_id"
+# Check if SKU already exists in offer
+(set -x; hack/tools/bin/pub skus list -p microsoft-aks -o aks-windows | jq ".[] | .planId" | tr -d '"' | tee skus.txt)
```
mainred, line 43: We may use publisher and offer as variables, as we are receiving them as required variables?

marosset: Let me update the PR so they we can take these in as environment variables.

marosset: Updated!

PR #4115, opened 2020-12-11
fix: ensure we wait for the right CRI service name

jackfrancis:
**Reason for Change**:

This PR ensures we are waiting for the correctly named CRI service name in various systemd definitions.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/docker-monitor.service
+++ b/parts/k8s/cloud-init/artifacts/docker-monitor.service
@@ -1,6 +1,10 @@
 [Unit]
 Description=a script that checks docker health and restarts if needed
+{{- if NeedsContainerd}}
+After=containerd.service
+{{else}}
 After=docker.service
+{{- end}}
 [Service]
 Restart=always
 RestartSec=10
--- a/parts/k8s/cloud-init/artifacts/docker-monitor.service
+++ b/parts/k8s/cloud-init/artifacts/docker-monitor.service
@@ -1,7 +1,6 @@
 [Unit]
 Description=a script that checks docker health and restarts if needed
 After={{GetContainerRuntime}}.service
-{{- end}}
 [Service]
 Restart=always
 RestartSec=10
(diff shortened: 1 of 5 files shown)


jackfrancis:
cc @Michael-Sinz

parts/k8s/cloud-init/artifacts/docker-monitor.service
```diff
@@ -1,6 +1,10 @@
 [Unit]
 Description=a script that checks docker health and restarts if needed
+{{- if NeedsContainerd}}
+After=containerd.service
```
Michael-Sinz, line 4: Wouldn't this be the same as writing:
```
After={{GetContainerRuntime}}.service
```
Or is there some additional disconnect between the two?  Right now I see those as the same.

jackfrancis: It is. The likelihood of the containerRuntime allowed values changing (from "docker" and "containerd" to something else to indicate the same configuration) is 0%. The likelihood that the apt package would rename the current service names to something else is perhaps a mite larger, but also unlikely, so your more elegant suggestion will work.

In the event the coupling between the CRI name and the service name goes away in the future, we can refactor to the more verbose if/else expression.

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -90,24 +90,13 @@ write_files:
     Slice={{- GetKubeReservedCgroup -}}.slice
     #EOF
 
-  {{if NeedsContainerd}}
-- path: /etc/systemd/system/containerd.service.d/kubereserved-slice.conf
+- path: /etc/systemd/system/{{GetContainerRuntime}}.service.d/kubereserved-slice.conf
```
jackfrancis, line 93: Simplifying this CRI filepath declaration while I'm here...

parts/k8s/cloud-init/artifacts/docker-monitor.service
Michael-Sinz: Worst case we could have GetContainerRuntimeService in addition to GetContainerRuntime

/lgtm

PR #4131, opened 2020-12-16
chore: simplify scale templates

jackfrancis:
**Reason for Change**:

This PR simplifies the template we submit for scale operations, removing unnecessary control plane resources unaffected by scale operations.

Specifically, we remove all control plane resource references that are not intended to be modified during a scale operation. Additionally, we remove the ARM template outputs array values, as no template outputs are modified during a scale operation compared to the original cluster creation ARM template.

**Issue Fixed**:

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [ ] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #4135, opened 2020-12-17
chore: simplify upgrade templates

jackfrancis:
**Reason for Change**:

Similar to #4131, this PR simplifies the template we submit for upgrade operations, removing unnecessary node resources unaffected by upgrade operations.

Specifically, we remove all resource references that are not intended to be modified during the upgrade operation of one particular node. Because we upgrade one node at a time, we only need a subset of the entire cluster-generated ARM template for each node upgrade (each implemented as discrete ARM template deployments). This PR optimizes these per-node operations so that the ARM templates have only the necessary resources in them.

Additionally, we remove the ARM template outputs array values, as no template outputs are modified during any upgrade operation compared to the original cluster creation ARM template.

- [x] Validate base config upgrade scenarios
- [x] Validate availability set config upgrade scenarios
- [x] Validate kubenet upgrade scenarios

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/engine/transform/transform_test.go
```diff
@@ -55,7 +55,7 @@ func TestNormalizeForK8sAddVMASPool(t *testing.T) {
 func TestNormalizeMasterResourcesForVMSSPoolUpgrade(t *testing.T) {
 	RegisterTestingT(t)
 	logger := logrus.New().WithField("testName", "TestNormalizeMasterResourcesForVMSSPoolUpgrade")
-	fileContents, e := ioutil.ReadFile("./transformtestfiles/k8s_template.json")
+	fileContents, e := ioutil.ReadFile("./transformtestfiles/k8s_slb_vmss_template.json")
```
jackfrancis, line 58: The existing UT data input for validating VMSS pool upgrade template filtering doesn't include a VMSS resource :(

pkg/engine/transform/transform_test.go
```diff
@@ -150,34 +150,6 @@ func TestRemoveKMSResourcesFromTemplate(t *testing.T) {
...
-	templateMap := template.(map[string]interface{})
-	transformer := &Transformer{
-		Translator: &i18n.Translator{
-			Locale: nil,
-		},
-	}
-	agentsToKeepMap := make(map[string]bool)
-	agentsToKeepMap["agentppol1"] = true // keep the typo or update the templates in ./transformtestfiles
-	agentsToKeepMap["agentpool2"] = true
-	// The usage of NormalizeResourcesForK8sMasterUpgrade across the code base seems to indicate that
```
jackfrancis, line 173: I took this comment at its word

mboersma (approved):
/lgtm

PR #4137, opened 2020-12-18
feat: Windows containerd VHD support

marosset:
**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


marosset:
I still need to address logic in cmd/upgrade.go

.github/workflows/nightly-build.yaml
```diff
@@ -3,6 +3,7 @@ name: Nightly
 on:
   schedule:
     - cron: '0 0 * * *' # Every day at midnight
+  workflow_dispatch:
```
marosset, line 6: will drop this change (it already merged)

marosset:
@jsturtevant this should be ready for review

vhd/release-notes/aks-windows-2019-containerd/2019-datacenter-core-ctrd-17763.1637.201210.txt
```diff
@@ -0,0 +1,118 @@
+﻿Build Number: 20201210.5
```
jsturtevant, line 1: should this be apart of this PR?

marosset: I meant to include it since we're adding a reference to that VHD in azenvtypes.

jsturtevant: see that now!

pkg/api/defaults.go
```diff
@@ -792,46 +792,51 @@ func (cs *ContainerService) setWindowsProfileDefaults(isUpgrade, isScale bool) {
...
 				}
 			}
 		} else {
 			if windowsProfile.WindowsPublisher == "" {
-				windowsProfile.WindowsPublisher = AKSWindowsServer2019OSImageConfig.ImagePublisher
+				windowsProfile.WindowsPublisher = defaultImageConfig.ImagePublisher
 			}
 			if windowsProfile.WindowsOffer == "" {
-				windowsProfile.WindowsOffer = AKSWindowsServer2019OSImageConfig.ImageOffer
+				windowsProfile.WindowsOffer = defaultImageConfig.ImageOffer
```
jsturtevant, line 823: this seems like it could get into a bad state if the profile for the windows publisher doesn't not equal defaultimageconfig.   Would it be simpler to calculate the default config based on profile (which seems to be happening partially above) and then update accordingly?

marosset: If the user is using an offer we don't curate here (MicrosoftWindowsServer 2004 images for example) we don't know what values to use here.
Maybe it would be better to turn this into an else-if, verify the publisher/offer match defaultImageConfig, before setting them and log and error if they don't match?
Thoughts?

marosset: There are many different offers available to users that we do not curate in aks-engine (1909, 2004, 20H2 images, etc).
I don't think it is possible for us to maintain default values for all of these offers.

I agree users can get into a bad state if they only specify windowsPublisher but am not sure what else to do here.
If things get messed up ARM does have a detailed message saying the publisher/offer/sku/image cannot be found.

marosset: Maybe we can nest these and only set offer if windowsPublisher == defaultImageConfig.ImagePublisher?

jsturtevant: Makes sense on all the versions.  If we assume they can get into a bad state could this logic be simplified and spit out a warning or maybe even an error that that tells they are going to get into a bad state with the values provided if they don't use one of our known defaults?

jsturtevant: > Maybe we can nest these and only set offer if windowsPublisher == defaultImageConfig.ImagePublisher?

for some reason this last comment didn't show up in my ui but I think we were thinking in the same direction

marosset: Good idea. I'll check if we think things are not good, use the default image, a log a warming.

cmd/upgrade.go
```diff
@@ -180,7 +180,10 @@ func (uc *upgradeCmd) loadCluster() error {
 	// Use the Windows VHD associated with the aks-engine version if upgradeWindowsVHD is set to "true"
 	if uc.upgradeWindowsVHD && uc.containerService.Properties.WindowsProfile != nil {
 		windowsProfile := uc.containerService.Properties.WindowsProfile
-		if windowsProfile.WindowsPublisher == api.AKSWindowsServer2019OSImageConfig.ImagePublisher && windowsProfile.WindowsOffer == api.AKSWindowsServer2019OSImageConfig.ImageOffer {
+		if windowsProfile.WindowsPublisher == api.AKSWindowsServer2019ContainerDOSImageConfig.ImagePublisher && windowsProfile.WindowsOffer == api.AKSWindowsServer2019ContainerDOSImageConfig.ImageOffer {
```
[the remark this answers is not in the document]
marosset: How does this look now?

jackfrancis (changes requested):
Let's remove the containerd-specific WindowsProfile default stuff here that we do to support E2E tests now that these defaults have matriculated to the actual api model defaults flow:

https://github.com/Azure/aks-engine/blob/master/test/e2e/engine/template.go#L253

pkg/api/azenvtypes.go
```diff
@@ -134,14 +134,23 @@ var (
...
 		ImageSku:       "2019-datacenter-core-smalldisk-2012",
 		ImagePublisher: "microsoft-aks",
 		ImageVersion:   "17763.1637.201210",
 	}
 
+	// AKSWindowsServer2019ContainerDOSImageConfig is the aks-engine image based on Windows Server 2019
+	// configured with containerd
+	AKSWindowsServer2019ContainerDOSImageConfig = AzureOSImageConfig{
+		ImageOffer:     "aks-windows",
+		ImageSku:       "2019-datacenter-core-ctrd-2012",
```
AbelHu, line 149: Does it mean that aks-engine will use a different VHD for containerd support? May I ask the reason not using the same 2019 VHD?

mainred: I guess this is for differentiating docker windows VHD, otherwise, this can only be told from the image version which is not obvious

marosset: The main reason we want to use two different VHDs is because docker and containerd store container images differently (both in a different format and at a different location on disk). 
If we wanted to use the same VHD we would need to prepull nanoserver/servercore/other images with both docker and containerd and this would not leave much room left on the VHD.

AbelHu: @marinred I think that we should do it, too.

olljanat:
@marosset I wonder that is plan actually also drop Docker from those containerd images as old comment on code proposes it already?
https://github.com/Azure/aks-engine/blob/bd397bcfbe7d4cb0381e7973a33eec5b4a7feb84/vhd/packer/configure-windows-vhd.ps1#L335-L339

jsturtevant:
/lgtm 

@jackfrancis did your requested changes get addressed?

jackfrancis:
@jsturtevant yep!

jsturtevant (approved):

PR #4145, opened 2020-12-25
fix: fix that kubeproxy cannot parse featuregates

AbelHu:
**Reason for Change**:

fix that kubeproxy cannot parse featuregates

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I set `provisioningScriptsPackageURL` in `examples/windows/kubernetes-windsr.json` to `https://testxx3e.blob.core.windows.net/akswindows/signedscripts-v0.0.7.zip` and below test passed.
```
CLUSTER_DEFINITION="examples/windows/kubernetes-windsr.json" SUBSCRIPTION_ID="xxx" CLIENT_ID="xxx" CLIENT_SECRET="xxx" TENANT_ID="xxx" LOCATION="westus2" CLEANUP_ON_EXIT=false make test-kubernetes
```
Result
```
------------------------------

JUnit report was created: /aks-engine/test/e2e/kubernetes/junit.xml

Ran 45 of 65 Specs in 4174.634 seconds
SUCCESS! -- 45 Passed | 0 Failed | 0 Pending | 20 Skipped
PASS

Ginkgo ran 1 suite in 1h9m34.655119s
Test Suite Passed
```

### DIFF
--- a/staging/provisioning/windows/kubeproxystart.ps1
+++ b/staging/provisioning/windows/kubeproxystart.ps1
@@ -24,7 +24,7 @@ if ($KubeproxyFeatureGates -contains "WinDSR=true") {
 }
 
 if ($KubeproxyFeatureGates.Count -ne 0) {
-    $global:KubeproxyArgList += @("--feature-gates='" + ($KubeproxyFeatureGates -join "','") + "'")
+    $global:KubeproxyArgList += @("--feature-gates=" + ($KubeproxyFeatureGates -join ","))
 }
 
 #
@@ -44,5 +44,5 @@ if (Test-Path "$global:KubeDir\runprocess.dll") {
     Add-Type -Path "$global:KubeDir\run-process.cs"
 }
 $exe = "$global:KubeDir\kube-proxy.exe"
-$args = ($global:KubeproxyArgList -join " ")
+$args = $global:KubeproxyArgList -join " "
 [RunProcess.exec]::RunProcess($exe, $args, [System.Diagnostics.ProcessPriorityClass]::AboveNormal)


staging/provisioning/windows/kubeproxystart.ps1
```diff
@@ -24,7 +24,7 @@ if ($KubeproxyFeatureGates -contains "WinDSR=true") {
 }
 
 if ($KubeproxyFeatureGates.Count -ne 0) {
-    $global:KubeproxyArgList += @("--feature-gates='" + ($KubeproxyFeatureGates -join "','") + "'")
+    $global:KubeproxyArgList += @("--feature-gates=" + ($KubeproxyFeatureGates -join ","))
```
jsturtevant, line 27: nice catch, looks like those extra quote are not required from the docs: https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#overview

staging/provisioning/windows/kubeproxystart.ps1
```diff
@@ -44,5 +44,5 @@ if (Test-Path "$global:KubeDir\runprocess.dll") {
     Add-Type -Path "$global:KubeDir\run-process.cs"
 }
 $exe = "$global:KubeDir\kube-proxy.exe"
-$args = ($global:KubeproxyArgList -join " ")
+$args = $global:KubeproxyArgList -join " "
```
jsturtevant, line 47: is this change clean up or did it have an effect on the outcome?

jsturtevant:
@AbelHu looks like the config `CLUSTER_DEFINITION="examples/windows/kubernetes-windsr.json"` doesn't require calico for testing...  If that's the case we might want to configure an e2e test to catch regressions to this configuration.

staging/provisioning/windows/kubeproxystart.ps1
AbelHu: From my testing, even the `"` is not needed

staging/provisioning/windows/kubeproxystart.ps1
AbelHu: From my test, both works. But if you execute `$args = ($global:KubeproxyArgList -join " ")` in PowerShell and then execute `echo $args`, you will not get any output.

AbelHu:
> @AbelHu looks like the config `CLUSTER_DEFINITION="examples/windows/kubernetes-windsr.json"` doesn't require calico for testing... If that's the case we might want to configure an e2e test to catch regressions to this configuration.

Calico requires WinDSR but WinDSR may be enabled when calico is not set. With WinDSR, it will fix [this issue](https://github.com/Azure/AKS/issues/1375)

AbelHu:
@marosset Could you help to take a look? AKS is expecting the new signed scripts pacakge.

staging/provisioning/windows/kubeproxystart.ps1
marosset: Ah, I should have caught this in the previous review :(
In powershell you can use backtick as an escape character if needed.

marosset: Since these args are getting used to start a windows process single quotes will not work here (only powersell understands single quotes for windows). 
Double quotes or not quotes should both work but double quotes are safer in case any of the arguments contain spaces.

marosset (approved):
/lgtm

marosset:
I will generate a new signed scripts package which include these changes

staging/provisioning/windows/kubeproxystart.ps1
AbelHu: From https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/, it should not contain spaces.

PR #4149, opened 2021-01-06
feat: setting a default containerd package for Windows

marosset:
**Reason for Change**:

We should be setting a default containerd package for Windows nodes if we plan on making containerd be the default container runtime for aks-engine.

This package contains **signed** versions of upstream containerd/containerd release binaries to comply with Azure SysLock policies.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -459,6 +459,8 @@ const (
 	DefaultKubeProxyMode KubeProxyMode = KubeProxyModeIPTables
 	// DefaultWindowsSSHEnabled is the default windowsProfile.sshEnabled value
 	DefaultWindowsSSHEnabled = true
+	// DefaultWindowsContainerdURL is the URL for the default containerd package on Windows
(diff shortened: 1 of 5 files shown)


marosset:
@abelhu This package is currently not being pushed to china cloud mirrors. I believe AKS will only be using the packages based on cosine teams containerd branch and not use the upstream packages correct? If that is correct then is it ok to not mirror these packages?

jsturtevant (approved):

jackfrancis (approved):
/lgtm

AbelHu:
> @AbelHu This package is currently not being pushed to china cloud mirrors. I believe AKS will only be using the packages based on cosine teams containerd branch and not use the upstream packages correct? If that is correct then is it ok to not mirror these packages?

My understanding is that AKS will still use upstream package. In preview, AKS can only use the package from cosine team. @keikhara Please correct me if my understanding is wrong.

PR #4154, opened 2021-01-12
fix: fix network cleanup code on windows for contianerd nodes

marosset:
**Reason for Change**:

* Fixes issue when calling cleanupnetwork.ps1 as a stand-alone script
* Fixes issue when running kubeletstart.ps1 from a directory other than c:\k
* Fixes issue where 'azure' network is not re-created after network cleanup because logs weren't deleted (logs ended up in c:\windows\system32 due to issue where containerd not setting working directory) 

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/staging/provisioning/windows/kubeletstart.ps1
+++ b/staging/provisioning/windows/kubeletstart.ps1
@@ -227,7 +227,7 @@ if ($global:NetworkPlugin -eq "kubenet") {
             $process | Stop-Process | Out-Null
         }
 
-        ./cleanupnetwork.ps1 
+        & "c:\k\cleanupnetwork.ps1"
 
         Write-Host "Creating a new hns Network"
         $hnsNetwork = New-HNSNetwork -Type $global:NetworkMode -AddressPrefix $podCIDR -Gateway $masterSubnetGW -Name $global:NetworkMode.ToLower() -Verbose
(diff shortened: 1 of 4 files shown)


Qdigital (dismissed):

staging/provisioning/windows/kubeletstart.ps1
```diff
@@ -227,7 +227,7 @@ if ($global:NetworkPlugin -eq "kubenet") {
             $process | Stop-Process | Out-Null
         }
 
-        ./cleanupnetwork.ps1 
+        & "c:\k\cleanupnetwork.ps1"
```
jsturtevant, line 230: should we also update the other places this is called?

https://github.com/Azure/aks-engine/blob/1d2b73ca07624d0e5f1ce4f16a604c0de2f52091/staging/provisioning/windows/windowsnodereset.ps1#L52

https://github.com/Azure/aks-engine/blob/d8807cb67c8b157399b43826c89041a615f877b8/staging/provisioning/windows/kubeletstart.ps1#L194

jsturtevant (approved):

PR #4178, opened 2021-01-18
feat: add support for Kubernetes v1.21.0-alpha.1

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md#changelog-since-v1200

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
Cluster-autoscaler should be bumped to v1.20.0 for this and the previous minor release. I'll do that in a separate PR.


.pipelines/pr-e2e.yaml
```diff
@@ -89,6 +89,16 @@ jobs:
...
+  parameters:
+    name: 'k8s_1_21_docker_e2e'
+    k8sRelease: '1.21'
+    apimodel: 'examples/e2e-tests/kubernetes/release/default/definition.json'
+    createVNET: true
+    enableKMSEncryption: false
+    containerRuntime: 'docker'
+    runSSHTests: true
+
+    - template: e2e-job-template.yaml
```
jackfrancis, line 101: I think this extra indentation is too cute for yaml

mboersma: D'oh, thanks for catching that! 👍🏼

jackfrancis (approved):
/lgtm

jackfrancis:
Let's keep the 1.21.0 test config additions

PR #4179, opened 2021-01-19
feat: updates for container monitoring addon omsagent agent January 2021 release

ganga1980:
**Reason for Change**:

Updates for January 2021 release. This has release notes associated to this release https://github.com/microsoft/Docker-Provider/blob/ci_prod/ReleaseNotes.md#01112021--

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [x] Yes

- [ ] No
- [x] Yes

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons_test.go
+++ b/pkg/api/addons_test.go
@@ -1564,7 +1564,7 @@ func TestSetAddonsConfig(t *testing.T) {
 					Enabled: to.BoolPtr(true),
 					Config: map[string]string{
 						"omsAgentVersion":       "1.10.0.1",
-						"dockerProviderVersion": "10.0.0-1",
+						"dockerProviderVersion": "12.0.0-0",
 						"schema-versions":       "v1",
 						"clusterName":           "aks-engine-cluster",
 						"workspaceDomain":       "b3BpbnNpZ2h0cy5henVyZS5jb20=",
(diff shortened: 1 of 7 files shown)


pkg/api/addons_test.go
```diff
@@ -1564,25 +1564,33 @@ func TestSetAddonsConfig(t *testing.T) {
...
+							Image:          "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod01112021",
+						},
+						{
+							Name:           "omsagent-rs",
 							CPURequests:    "150m",
 							MemoryRequests: "250Mi",
 							CPULimits:      "1",
-							MemoryLimits:   "750Mi",
-							Image:          "mcr.microsoft.com/azuremonitor/containerinsights/ciprod:ciprod10052020",
+							MemoryLimits:   "1Gi",
```
jackfrancis, line 1586: This change wants to increase the request-limit ratio from 3x to 4x. The recommended ratio is 2x, to eliminate as much memory reservation variability in practice, in order to improve deterministic pod scheduling (and optimize pod density across the cluster).

Why do we want to continue to increase the amount of memory that is available? Is this application particularly "spiky"?

jackfrancis: sorry, this comment should have been moved up to the actual implemntation, not the UT :)

ganga1980: Currently, our replica set is singleton and for large workloads, say high number of pods, events and nodes, or high prometheus scraping requires higher memory limit.  And also we have seen if the node where  replicaset got sheduled has disk I/O issues and the high memory usage. We have optimized our disk I/O paths to avoid higher memory usage and other perf improvements in this release. Currently, We are working on moving prometheus scraping to deamonset so that load will be evenly distributed and moved way from singleton replicaset.   Memory requests and limits what we have here is inline what we have default in AKS and also accommodate these defaults for large clusters.

jackfrancis:
This PR did not pass the current omsagent E2E tests as defined here:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L1167

Any ideas?

jackfrancis:
Closing this PR as the E2E test is failing this requirement: `"Ensuring that the kubenodeinventory plugin is writing data successfully"`

ganga1980:
> This PR did not pass the current omsagent E2E tests as defined here:
> 
> https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/kubernetes_test.go#L1167
> 
> Any ideas?

> @jackfrancis , reviewed the code, functionality wise no issue, and working as expected and we are missed to emit the log line "kubeNodeInventoryEmitStreamSuccess" if the node count of the cluster is < 100 nodes for kube node inventory workflow.  because e2e tests  expects log line so e2e test failing.

PR #4180, opened 2021-01-19
docs: fix broken link in AAD docs

jackfrancis:
**Reason for Change**:

This PR fixes a broken link in the aad docs.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/aad.md
+++ b/docs/topics/aad.md
@@ -20,7 +20,7 @@ You also need to delegate permission to the application as follows:
 
 
 ## Deployment
-Follow the [deployment steps]([../tutorials/quickstart.md#deploy]). In step #4, add the following under 'properties' section:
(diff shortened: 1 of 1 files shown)


deaborch:
@jackfrancis would it be possible to include in the documentation that it is not supported in azure stack hub??

mboersma (approved):
/lgtm

jackfrancis:
@deaborch If it indeed true that there is no Azure Stack Hub solution, yeah, we should include that in the docs. Could you submit a PR with the appropriate Azure Stack messaging, if true?

PR #4188, opened 2021-01-22
feat: add support for Kubernetes v1.17.17 on Azure Stack

JesusAlvarezTorres:
**Reason for Change**:

Adding support for Kubernetes v1.17.17 on Azure Stack

**Issue Fixed**:

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
code lgtm

to @jadarsie for final review/merge

pkg/api/common/versions.go
```diff
@@ -269,6 +269,7 @@ var AllKubernetesSupportedVersionsAzureStack = map[string]bool{
 	"1.17.7":  false,
 	"1.17.9":  true,
 	"1.17.11": true,
```
jadarsie, line 271: FYI, the linux VHD is not pre-pulling v1.17.11, not sure we want to keep both v1.17.11 and v1.17.17

JesusAlvarezTorres: Good catch, I did not notice it, will submit a commit to remove v1.17.11

PR #4193, opened 2021-01-25
ci: don't immediately delete clusters if E2E tests fail

jackfrancis:
**Reason for Change**:

In response to recent E2E failure signals (especially CSE exit 30), let's keep the clusters around for real-time debugging.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


bridgetkromhout:
Consulted with @jackfrancis and we no longer need to make this change, given the AKS Engine deprecation.

PR #4212, opened 2021-01-30
chore: block container traffic to 168.63.129.16

jackfrancis:
**Reason for Change**:

This PR explicitly ensures that container-originating TCP traffic bound for the reserved Azure IP + TCP 80 endpoint is dropped.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/masternodecustomdata.yml
+++ b/parts/k8s/cloud-init/masternodecustomdata.yml
@@ -462,7 +462,7 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
     sed -i "s|<etcdEncryptionSecret>|\"{{WrapAsParameter "etcdEncryptionKey"}}\"|g" /etc/kubernetes/encryption-config.yaml
 {{end}}
     {{- /* Ensure that container traffic can't connect to internal Azure IP endpoint */}}
-    iptables -I FORWARD -d 168.63.129.16 -p tcp -j DROP
+    iptables -I FORWARD -d 168.63.129.16 -p tcp --dport 80 -j DROP
     #EOF
 
 {{- if not HasCosmosEtcd  }}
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -461,6 +461,8 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
 {{- if EnableDataEncryptionAtRest }}
     sed -i "s|<etcdEncryptionSecret>|\"{{WrapAsParameter "etcdEncryptionKey"}}\"|g" /etc/kubernetes/encryption-config.yaml
 {{end}}
+    {{- /* Ensure that container traffic can't connect to internal Azure IP endpoint */}}
+    iptables -I FORWARD -d 168.63.129.16 -p tcp -j DROP
```
weinong, line 465: you will need to specify port 80 as this IP is also used by Load Balancer probe.
ref PR: https://github.com/Azure/AgentBaker/pull/375

weinong (dismissed):

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -461,6 +461,9 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
 {{- if EnableDataEncryptionAtRest }}
     sed -i "s|<etcdEncryptionSecret>|\"{{WrapAsParameter "etcdEncryptionKey"}}\"|g" /etc/kubernetes/encryption-config.yaml
 {{end}}
+    {{- /* Ensure that container traffic can't connect to internal Azure IP endpoint */}}
+    iptables -I OUTPUT -s {{GetClusterSubnet}} -d 168.63.129.16 -p tcp --dport 80 -j DROP
```
aramase, line 465: Is the intent here to drop all traffic to wireserver from the host? Pods running with host network have the node IP as pod IP and this rule would just block all traffic.

aramase: Also, this should be same for v4/v6 or dual-stack clusters because we use the host v4 networking for dns, etc..

weinong: nope, we should not block processes like guest agent running in the vm. this is the part i was struggling to differentiate: how to block only container traffic whose source IP is node IP?

dmeytin:
Guys, I'm sorry to be a party pooper but this change will prevent installing clusters in some environments. 
I will provide more details offline.

PR #4231, opened 2021-02-04
feat: run unattended upgrades by default

jackfrancis:
**Reason for Change**:

This is a follow-up PR from #4217. This PR sets the newly added `runUnattendedUpgradesOnBootstrap` LinuxProfile configuration to `true` for Azure cloud environments.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @jadarsie @JesusAlvarezTorres @Michael-Sinz

Michael-Sinz:
I am a bit concerned about the split personality here.  A configuration deployed in public will act differently than that "same" configuration elsewhere.  It makes sense as to why we are doing this but it still is concerning.

I also don't know the config management structure very well so will let others do the deeper review.

jackfrancis:
> I am a bit concerned about the split personality here. A configuration deployed in public will act differently than that "same" configuration elsewhere. It makes sense as to why we are doing this but it still is concerning.
> 
> I also don't know the config management structure very well so will let others do the deeper review.

Strictly speaking, there is already ample precedent for default configuration between Azure Stack (and other custom clouds) and public Azure cloud being different. I think the larger concern is that, arguably, the value of ensuring apt-updated nodes is more pressing in non-public cloud, because it can be more difficult to do that patching manually.

I trust @jadarsie and others understand that, but it makes sense to give those folks time to test these assumptions in their environments first.

Also cc @DavidParks8 @ericsuhong in case you have any thoughts.

Michael-Sinz:
> > I am a bit concerned about the split personality here. A configuration deployed in public will act differently than that "same" configuration elsewhere. It makes sense as to why we are doing this but it still is concerning.
> 
> Strictly speaking, there is already ample precedent for default configuration between Azure Stack (and other custom clouds) and public Azure cloud being different. I think the larger concern is that, arguably, the value of ensuring apt-updated nodes is more pressing in non-public cloud, because it can be more difficult to do that patching manually.
> 
> I trust @jadarsie and others understand that, but it makes sense to give those folks time to test these assumptions in their environments first.
> 
> Also cc @DavidParks8 @ericsuhong in case you have any thoughts.

I fully agree - just pushing for "better security by default" but those environment should call the shots.

jadarsie:
/lgtm

I could not start the conversation on my side yet. You can wait until I get back with an answer OR merge and then we can later update the code if needed.

jadarsie:
@jackfrancis, `false` is the preferred default for ASH.

PR #4252, opened 2021-02-10
feat: rotate-certs fails faster if invalid ssh params

jadarsie:
**Reason for Change**:

Addressing some `aks-engine rotate-certs` feedback

- document `azure-env` parameter
- fail faster if `--ssh-host` or `--linux-ssh-private-key` are invalid
- delete previous doc page
- more explicit stdout messages

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


docs/topics/certificaterotation.md
```diff
@@ -1,60 +0,0 @@
-# Cluster certificate rotation
```
jackfrancis, line 1: Why are we deleting these docs?

jadarsie: It has duplicated/stale information, I pushed another one with my previous PR.

https://github.com/Azure/aks-engine/blob/master/docs/topics/rotate-certs.md

jackfrancis (approved):
/lgtm

PR #4264, opened 2021-02-12
chore: update cluster-autoscaler to 1.20.0

jackfrancis:
**Reason for Change**:

This PR updates the cluster-autoscaler addon to 1.20.0. See:

https://github.com/kubernetes/autoscaler/releases/tag/cluster-autoscaler-1.20.0

As this tested out O.K. with all currently supported versions of Kubernetes (1.17+), we can remove the heterogeneous set of supported cluster-autoscaler versions and pin to 1.20.0 for all versions of k8s.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @mboersma

mboersma:
> pin to 1.20.0 for all versions of k8s.

I thought the idea was to keep cluster-autoscaler in sync with the Kubernetes minor version. That is, CA 1.20.x goes with Kubernetes 1.20.x (and 1.21.x until there's a specific release for it), CA 1.19.x goes with k8s 1.19.x, etc.

Also I see the Azure Pipeline for CA completed except for the publish step--I'll nudge upstream publishing to approve that.

jackfrancis:
@mboersma As far as our functional cluster-autoscaler tests go, 1.20 works w/ 1.17+, so that's good enough for me. I think until we can prove something doesn't work we shouldn't take on the additional test burden of validating version-to-version CA-k8s affinity. Having a single CA version simplifies our test matrix, and VHD maintenance.

cc @craiglpeters @devigned for any additional thoughts that may not agree with my conclusions above

mboersma:
I like the simplification of using one cluster-autoscaler--like AKS Engine does for core-dns--the tradeoff is we don't benefit from this upstream testing:

> The below combinations have been tested on GCP. We don't do cross version testing or compatibility testing in other environments.

But I guess AKS Engine is doing cross-version testing, in the Azure environment, so I'm ok with that.

devigned:
Late to the party. Sorry, this notification slipped past.

+1 to Matt's and Jack's comments and to our test matrix. Since we have cross version signal 1.20, we've done pretty well without taking on too much more test burden.

PR #4176, opened 2021-01-18
feat: allow creation of dualstack Windows clusters

vavenk-ms:
**Reason for Change**:

Allow creation of dual stack Windows Kubernetes clusters. This was restricted as Windows did not support dual stack. From Windows version 2004 (10.0.19041.610) and Kubernetes version 1.19, dual stack is supported. https://kubernetes.io/docs/setup/production-environment/windows/intro-windows-in-kubernetes/#ipv4-ipv6-dual-stack

**Issue Fixed**:

https://github.com/Azure/aks-engine/issues/4171

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [X] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [X] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #4281, opened 2021-02-18
fix: restart hns for Windows dual-stack

vavenk-ms:
**Reason for Change**:

Due to a bug in the hns service there is a race where it picks up the incorrect IPv6 address from the node in some cases. The hns service has to be restarted after the node internal IPv6 address is available when dual-stack is enabled.

**Issue Fixed**:

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jsturtevant:
should this be a commit as part of #4176?

vavenk-ms:
> should this be a commit as part of #4176?

@jsturtevant sure, I can add it there and close this one.

vavenk-ms:
Added the commit to #4176

PR #4282, opened 2021-02-18
fix: disable docker when containerD is specified

mainred:
**Reason for Change**:

1. Remove docker as not used when containerD is used. Also mixing docker and containerD may hide potential errors of containerD by docker
2. cx may get confused by the docker CLI output on a containerD-enabled node for devops

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/windowskubeletfunc.ps1
+++ b/parts/k8s/windowskubeletfunc.ps1
@@ -352,7 +352,9 @@ New-NSSMService {
     & "$KubeDir\nssm.exe" set Kubelet AppRotateBytes 10485760 | RemoveNulls
     # Do not use & when calling DependOnService since 'docker csi-proxy'
     # is parsed as a single string instead of two separate strings
-    Invoke-Expression "$KubeDir\nssm.exe set Kubelet DependOnService $kubeletDependOnServices | RemoveNulls"
+    if (-not [string]::IsNullOrEmpty($kubeletDependOnServices) {
+        Invoke-Expression "$KubeDir\nssm.exe set Kubelet DependOnService $kubeletDependOnServices | RemoveNulls"
+    }
 
     # setup kubeproxy
     & "$KubeDir\nssm.exe" install Kubeproxy C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe | RemoveNulls
--- a/parts/k8s/windowskubeletfunc.ps1
+++ b/parts/k8s/windowskubeletfunc.ps1
@@ -316,9 +316,14 @@ New-NSSMService {
         [string]
         [Parameter(Mandatory = $true)]
         $KubeProxyStartFile
+        [Parameter(Mandatory = $false)][string]
+        $ContainerRuntime = "docker"
     )
 
-    $kubeletDependOnServices = "docker"
+    $kubeletDependOnServices = ""
+    if ($ContainerRuntime -eq "docker") {
+        $kubeletDependOnServices = "docker"
+    }
     if ($global:EnableCsiProxy) {
         $kubeletDependOnServices += " csi-proxy"
     }
@@ -374,6 +379,8 @@ Install-KubernetesServices {
     param(
         [Parameter(Mandatory = $true)][string]
         $KubeDir
+        [Parameter(Mandatory = $false)][string]
+        $ContainerRuntime = "docker"
     )
 
     # TODO ksbrmnn fix callers to this function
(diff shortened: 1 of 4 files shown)


parts/k8s/windowskubeletfunc.ps1
```diff
@@ -316,9 +316,14 @@ New-NSSMService {
         [string]
         [Parameter(Mandatory = $true)]
         $KubeProxyStartFile
+        [Parameter(Mandatory = $false)][string]
+        $ContainerRuntime = "docker"
     )
 
-    $kubeletDependOnServices = "docker"
+    $kubeletDependOnServices = ""
+    if ($ContainerRuntime -eq "docker") {
```
jsturtevant, line 324: Doesn't look like we've done this in the past but should kubelet be dependent on containerd here?

jsturtevant (commented):
One minor comment.  Thanks for looking into this

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -347,7 +352,9 @@ New-NSSMService {
     & "$KubeDir\nssm.exe" set Kubelet AppRotateBytes 10485760 | RemoveNulls
     # Do not use & when calling DependOnService since 'docker csi-proxy'
     # is parsed as a single string instead of two separate strings
-    Invoke-Expression "$KubeDir\nssm.exe set Kubelet DependOnService $kubeletDependOnServices | RemoveNulls"
+    if (-not [string]::IsNullOrEmpty($kubeletDependOnServices) {
```
AbelHu, line 355: Should we need to set that kubelet depends on containerd when the runtime environment is containerd?

parts/k8s/windowskubeletfunc.ps1
mainred: Changed the code to force kubelet depend on any using container runtime service. Thanks for pointing this out.

parts/k8s/windowskubeletfunc.ps1
mainred: Changed the code to allow kubelet depend on containerd if contained is used.

AbelHu: The check may not be necessary since kubeletDependOnServices is always set.

AbelHu (approved):

jsturtevant:
This is failing due to `QuotaExceeded`

jsturtevant:
quota work should be resolved

/azp run pr-e2e

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -316,9 +316,14 @@ New-NSSMService {
         [string]
         [Parameter(Mandatory = $true)]
         $KubeProxyStartFile
```
AbelHu, line 318: $KubeProxyStartFile
 =>
$KubeProxyStartFile,

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -374,6 +379,8 @@ Install-KubernetesServices {
     param(
         [Parameter(Mandatory = $true)][string]
         $KubeDir
```
AbelHu, line 381: $KubeDir
=>
$KubeDir,

AbelHu (dismissed):

parts/k8s/windowskubeletfunc.ps1
jsturtevant: good catch. I believe this is why the last test run failed to provision.

parts/k8s/windowskubeletfunc.ps1
```diff
@@ -316,9 +316,14 @@ New-NSSMService {
         [string]
         [Parameter(Mandatory = $true)]
         $KubeProxyStartFile
+        [Parameter(Mandatory = $false)][string]
+        $ContainerRuntime = "docker"
     )
 
     $kubeletDependOnServices = "docker"
+    if ($ContainerRuntime -eq "containerd") {
```
marosset, line 324: You can just set `kubeletDependOnServices = $ContainerRuntime` and eliminate the conditional here.

marosset:
I have a feeling we're going to run into issues related to 

https://github.com/Azure/aks-engine/blob/1f424d3f683f297395e4b9dbf0220288e39e63c3/parts/k8s/windowskubeletfunc.ps1#L181

and

https://github.com/Azure/aks-engine/blob/1f424d3f683f297395e4b9dbf0220288e39e63c3/parts/k8s/windowskubeletfunc.ps1#L203-L222

In the future (especially with upstream tests).

I don't have a good suggestion here at the moment though.
FYI @jsturtevant

mainred:
Created an issue related to pause image building support, https://github.com/Azure/aks-engine/issues/4330 @marosset

jsturtevant (approved):

PR #4293, opened 2021-02-23
feat: add byok support with kms plugin

aramase:
Signed-off-by: Anish Ramasekar <anish.ramasekar@gmail.com>

**Reason for Change**:

- Updates kms image to `mcr.microsoft.com/oss/azure/kms/keyvault:v0.0.11`
- Adds BYOK support with aks-engine
- Removes storage account creation that was required for older releases
- Cleans up `azure.json` as the kms parameters are no longer referred to from that file. 

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:


parts/k8s/manifests/kubernetesmaster-azure-kubernetes-kms.yaml
```diff
@@ -13,7 +13,19 @@ spec:
     - name: azure-kms-provider
       image: {{ContainerImage "azure-kms-provider"}}
       imagePullPolicy: IfNotPresent
-      command: [{{ContainerConfig "command"}}]
```
jackfrancis, line 16: what happened to the command property here?

pkg/api/vlabs/validate_test.go
```diff
@@ -253,6 +253,20 @@ func Test_OrchestratorProfile_Validate(t *testing.T) {
 				},
 			},
 		},
+		"should error when enableEncryptionWithExternalKms=true with own key and all required parameters not provided": {
```
jackfrancis, line 256: I know this is annoying, but we should probably UT the remaining 4 non-valid permutations. (did I do the math right???)

bridgetkromhout:
Per @aramase this PR is no longer needed due to AKS Engine deprecation.

PR #4297, opened 2021-02-24
chore: don't run unattended upgrade if using "no outbound" test feature

jackfrancis:
**Reason for Change**:

This PR ensures that when we're using the "block outbound internet access" test feature flag, we don't try to run unattended-upgrade as a blocking CSE operation (which will never finish due to default apt servers not being reachable).

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


Michael-Sinz (approved):
/lgtm

PR #3394, opened 2020-06-03
feat: add "get-versions --azure-env" flag to list custom clouds supported versions

haofan-ms:
**Reason for Change**:

The `aks-engine get-versions` command only shows Azure supported versions.
The `--azure-env` flag will allow users to pass a target custom cloud like Azure Stack and get back supported versions on that target cloud.

Usage: `aks-engine get-versions --azure-env AzureStackCloud`

**Requirements**:
- [x] uses [conventional commit messages](https://www.conventionalcommits.org/)
- [ ] includes documentation
- [x] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #4308, opened 2021-03-01
docs: update get-verions section of Azure Stack documentation

haofan-ms:
**Reason for Change**:
Doc was not update when the PR (https://github.com/Azure/aks-engine/pull/3394) was merged, thus making the change.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/docs/topics/azure-stack.md
+++ b/docs/topics/azure-stack.md
@@ -170,7 +170,7 @@ If you need to expose more than 5 services, then the recommendation is to route 
 
 ### get-versions command
 
-By default, `aks-engine get-versions` shows which Kubernetes versions are supported by each AKS Engine release on Azure's public cloud. Include flag `--azure-env` to get the list of supported Kubernetes versions on a custom cloud such as an Azure Stack Hub cloud (`aks-engine get-versions --azure-env AzureStackCloud`). Upgrade paths for Azure Stack Hub can also be found [here](https://docs.microsoft.com/en-us/azure-stack/user/kubernetes-aks-engine-release-notes?view=azs-2005#kubernetes-version-upgrade-path-in-aks-engine).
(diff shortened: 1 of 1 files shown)


jackfrancis (dismissed):
/lgtm

docs/topics/azure-stack.md
```diff
@@ -170,7 +170,7 @@ If you need to expose more than 5 services, then the recommendation is to route
 
 ### get-versions command
 
-The output of the `get-versions` command only pertains to Azure and not Azure Stack Hub clouds. The different upgrade paths can be found [here](https://docs.microsoft.com/en-us/azure-stack/user/kubernetes-aks-engine-release-notes?view=azs-2005#kubernetes-version-upgrade-path-in-aks-engine-0510).
+By default, `aks-engine get-versions` shows which Kubernetes versions are supported by each AKS Engine release on Azure's public cloud. Include flag `--azure-env` to get the list of supported Kubernetes versions on a custom cloud such as an Azure Stack Hub cloud (`aks-engine get-versions --azure-env AzureStackCloud`). Upgrade paths for Azure Stack Hub can also be found [here](https://docs.microsoft.com/en-us/azure-stack/user/kubernetes-aks-engine-release-notes?view=azs-2005#kubernetes-version-upgrade-path-in-aks-engine).
```
jadarsie, line 173: I think `en-us` should be removed from the URL.
I think we should remove `view=azs-####` from the URL
The URL points to a non-existing anchor (`#kubernetes-version-upgrade-path-in-aks-engine`) I would remove it as it looks like the section title changes with every release.

haofan-ms: Thanks Javier, I get your point.I will only keep the constants in the URL and make it sure redirect automatically every time.

PR #4315, opened 2021-03-03
chore: Hsn fix for 2c+

jsturtevant:
**Reason for Change**:

Hsn fix for 2c+

If multiple LB policies are included for same endpoint then HNS hangs. this fix forces an error  so restarts to HNS are not required.

**Issue Fixed**:

**Credit Where Due**:
https://github.com/Azure/AgentBaker/pull/628
/cc @AbelHu @mainred

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -413,6 +413,11 @@ try
         Write-Log "Disable Internet Explorer compat mode and set homepage"
         Set-Explorer
 
+        # if multple LB policies are included for same endpoint then HNS hangs.
+        # this fix forces an error  
+        Write-Host "Enable a HNS fix in 2021-2C+"
+        Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name HNSControlFlag -Value 1 -Type DWORD
+
         Write-Log "Adjust pagefile size"
         Adjust-PageFileSize
 
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -17492,6 +17492,11 @@ try
         Write-Log "Disable Internet Explorer compat mode and set homepage"
         Set-Explorer
 
+        # if multple LB policies are included for same endpoint then HNS hangs.
+        # this fix forces an error  
+        Write-Host "Enable a HNS fix in 2021-2C+"
+        Set-ItemProperty -Path "HKLM:\SYSTEM\CurrentControlSet\Services\hns\State" -Name HNSControlFlag -Value 1 -Type DWORD
+
         Write-Log "Adjust pagefile size"
         Adjust-PageFileSize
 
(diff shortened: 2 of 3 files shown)


AbelHu (approved):

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -413,6 +413,11 @@ try
         Write-Log "Disable Internet Explorer compat mode and set homepage"
         Set-Explorer
 
+        # if multple LB policies are included for same endpoint then HNS hangs.
+        # this fix forces an error  
+        Write-Host "Enable a HNS fix in 2021-2C+"
```
AbelHu, line 418: @jsturtevant do you mean that the code in configure-windows-vhd.ps1 does not work?

jsturtevant: vhd should work, but we also have cases where VHD is not required in aks-engine so wanted to make sure it was enabled for those

AbelHu (approved):

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -413,6 +413,11 @@ try
         Write-Log "Disable Internet Explorer compat mode and set homepage"
         Set-Explorer
 
+        # if multple LB policies are included for same endpoint then HNS hangs.
+        # this fix forces an error  
+        Write-Host "Enable a HNS fix in 2021-2C+"
+        reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\hns\State /v HNSControlFlag /t REG_DWORD /d 1
+
```
marosset, line 420: Will this ever be set by default in some future OS updates or will we always need to set this? 
Is this only needed on Windows Server 2019 or all Windows Server OS versions?

pkg/engine/templates_generated.go
```diff
@@ -17492,6 +17492,11 @@ try
         Write-Log "Disable Internet Explorer compat mode and set homepage"
         Set-Explorer
 
+        # if multple LB policies are included for same endpoint then HNS hangs.
+        # this fix forces an error  
+        Write-Host "Enable a HNS fix in 2021-2C+"
+        reg add HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\hns\State /v HNSControlFlag /t REG_DWORD /d 1
```
marosset, line 17498: nit: Can you use `Set-ItemProperty -Path 'HKLM:\SYSTEM\CurrentControlSet\Services\hns\State' -Name HNSControlFlag -Value 1 -PropertyType DWORD`

parts/k8s/kuberneteswindowssetup.ps1
jsturtevant: @daschott do you have insight here?

jackfrancis (approved):
code lgtm

AbelHu (approved):

marosset (approved):
/lgtm

Issue #4277, opened 2021-02-16
Fall back to normal networking if SKU isn't in accelerated networking table 


mboersma:
**Describe the request**

AKS Engine could remove the SKU validation table entirely, infer the storage type from the SKU name, and leave only the SKU -> accelerated networking support table. This table could be treated as a "best effort" lookup: that is, if the SKU isn't present, fall back to normal networking. Checking that this remaining lookup table is up-to-date can become a task for CI runs.

cc: @MichaelSinz

**Explain why AKS Engine needs it**

Pro:
- Allow users to provision SKUs as soon as Azure makes them available, instead of waiting for a new AKS Engine release
- Remove most of the painful copy-paste and updating of SKU-related unit tests
- Reduce boilerplate code in ARM template and AKS Engine source

Con:
- Invalid SKU types are caught by Azure, not a bit earlier in ARM validation
- If a user doesn't manually enable `acceleratedNetworking`, they could end up with normal networking inadvertently if the SKU isn't yet in the lookup table

**Describe the solution you'd like**

I did this already, above. I jumped to the solution. *Mea culpa*, please add your suggestions.

**Describe alternatives you've considered**

Making a dynamic API call to determine acceleratedNetworking status is verboten in the context of `aks-engine generate`, otherwise we should do that as CAPZ does.

**Additional context**

Michael-Sinz:
This would help reduce the problem of not having the current SKUs
Another option would be a way to add a SKU data file (or query of the Azure API) when asked.  That way the data could be updated.  But having a way to create clusters with newer SKUs is the key element here.

mboersma:
I implemented this and found that it served to remind me how painful the failure mode is without the SKU validation, which amounts to a "fast fail." Without that, the rest of the cluster infrastructure is actually provisioned while ARM waits for the invalid VM eventually to time out from repeated errors. Not user-friendly to say the least.

Perhaps this behavior (skipping VM SKU validation) could be toggled on by a command-line argument, so the default behavior stays the same.

Michael-Sinz:
That would be fine if it requires an "I am sure" or "skip sku validation" flag.

mboersma:
Given that AKS Engine is [deprecated](https://github.com/Azure/aks-engine#project-status) for public cloud users, I don't think we will implement this, so I'm closing it.

PR #4316, opened 2021-03-05
refactor: remove storage type from SKU lookup table

mboersma:
**Reason for Change**:

Since a VM's storage type is calculable from its name, there's no need to maintain a static lookup table. This is a minor simplification to ensure there's only one code path for finding storage type.

(This is also the only separable change that came out of my work on #4277 so far.)

**Issue Fixed**:

Refs #4277

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/get_skus.go
+++ b/cmd/get_skus.go
@@ -178,13 +178,3 @@ var VMSkus = []VMSku{
 
 	return err
 }
-
-func storageAccountType(skuName string) string {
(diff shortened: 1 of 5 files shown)


jackfrancis (approved):
/lgtm

PR #4332, opened 2021-03-17
feat: allow setting read only cache

alexeldeib:
**Reason for Change**:
allow setting read only cache

we've observed several issues with the performance of write caching on azure, including stalls on the whole disk when fsync'ing a single file, and allowing the cache to accept IO requests in excess of what the the device can handle. the former directly degrades performance and the latter indirectly degrades it by decreasing the ability of the linux kernel to counteract using e.g. io request merging (since everything passes through to azure as if there were no traffic). 

we're working with storage on mitigating those issues but in the meantime we'd like to allow changing this value to ReadOnly for OS disk. This is the only other allowed value (None for OS Disk will be overridden internally, refer to the doc link on the API type).

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

---
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1179,7 +1179,7 @@ func (p *Properties) AreAgentProfilesCustomVNET() bool {
 
 // GetClusterID creates a unique 8 string cluster ID.
 func (p *Properties) GetClusterID() string {
-	mutex := &sync.Mutex{}
+	var mutex = &sync.Mutex{}
 	if p.ClusterID == "" {
 		uniqueNameSuffixSize := 8
 		// the name suffix uniquely identifies the cluster and is generated off a hash
@@ -2230,7 +2230,7 @@ func (p *Properties) IsNvidiaDevicePluginCapable() bool {
 // SetCloudProviderRateLimitDefaults sets default cloudprovider rate limiter config
 func (p *Properties) SetCloudProviderRateLimitDefaults() {
 	if p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket == 0 {
-		agentPoolProfilesCount := len(p.AgentPoolProfiles)
+		var agentPoolProfilesCount = len(p.AgentPoolProfiles)
 		if agentPoolProfilesCount == 0 {
 			p.OrchestratorProfile.KubernetesConfig.CloudProviderRateLimitBucket = DefaultKubernetesCloudProviderRateLimitBucket
 		} else {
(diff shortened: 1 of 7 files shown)


pkg/api/vlabs/types_test.go
```diff
@@ -196,9 +200,15 @@ func TestAgentPoolProfile(t *testing.T) {
...
+	if ap.OSDiskCaching == nil || *ap.OSDiskCaching != CachingTypesReadOnly {
+		t.Fatalf("AgentPoolProfile.OSDiskCaching should be ReadOnly after unmarshal")
+	}
+
 	// With osType Linux and RHEL distro
-	AgentPoolProfileText = `{ "name": "linuxpool1", "osType" : "Linux", "distro" : "rhel", "count": 1, "vmSize": "Standard_D2_v2",
-"availabilityProfile": "AvailabilitySet", "storageProfile" : "ManagedDisks", "vnetSubnetID" : "12345", "diskEncryptionSetID": "diskEncryptionSetID" }`
+	AgentPoolProfileText = `{ "name": "linuxpool1", "osType" : "Linux", "distro" : "rhel", "count": 1, "vmSize": "Standard_D2_v2", 
+"availabilityProfile": "AvailabilitySet", "storageProfile" : "ManagedDisks", "vnetSubnetID" : "12345", "diskEncryptionSetID": "diskEncryptionSetID",
+"osDiskCaching": "ReadWrite", }`
```
alexeldeib, line 211: typo

pkg/api/types.go
```diff
@@ -688,8 +688,18 @@ type AgentPoolProfile struct {
 	UltraSSDEnabled                     *bool                `json:"ultraSSDEnabled,omitempty"`
 	EncryptionAtHost                    *bool                `json:"encryptionAtHost,omitempty"`
 	ProximityPlacementGroupID           string               `json:"proximityPlacementGroupID,omitempty"`
+	OSDiskCaching                       *CachingType         `json:"osDiskCaching,omitempty"`
 }
 
+// CachingType determines the HostCache mode for an Azure VM Disk. Read more here:
+// https://docs.microsoft.com/en-us/azure/virtual-machines/premium-storage-performance#disk-caching
+type CachingType string
```
xuto2, line 696: DiskCachingType

xuto2 (approved):

pkg/api/types.go
alexeldeib: will align this all over 👍

serbrech (approved):

PR #4334, opened 2021-03-17
docs: avoid --control-plane-only upgrade on air-gapped clouds

jadarsie:
**Reason for Change**:

Documenting the fact that `aks-engine upgrade --control-plane-only` is not expected to work on air-gapped clouds as agents won't be able to pull down the newer `kube-proxy` container image required by the new daemonset manifest.

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


docs/topics/upgrade.md
```diff
@@ -60,7 +60,7 @@ In summary, using `aks-engine upgrade` means you will freshen and re-pave the en
 |--kubeconfig|no|Path to kubeconfig; if not provided, it will be generated on the fly from the API model data.|
 |--upgrade-version|yes|Version of Kubernetes to upgrade to.|
 |--force|no|Force upgrading the cluster to desired version, regardless of version support. Allows same-version upgrades and downgrades.|
-|--control-plane-only|no|Upgrade control plane VMs only, do not upgrade node pools.|
+|--control-plane-only|no|Upgrade control plane VMs only, do not upgrade node pools (unsupported on air-gapped clouds).|
```
jackfrancis, line 63: This is only due to the missing -azs kube-proxy image on the VHD, right? So it would only apply to Azure Stack?

jadarsie: Not my understanding, it applies to all air-gapped clouds.

jadarsie: Maybe the term `unsupported` is not precise? the agents' `kube-proxy` can be "recovered" after a full upgrade.

jackfrancis (approved):
/lgtm

Issue #4350, opened 2021-03-31
test mtu 3900


jackfrancis:
**Describe the request**

Test the viability of 3900 byte MTU size on node eth0 interfaces.

**Explain why AKS Engine needs it**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

bridgetkromhout:
Closing per https://github.com/Azure/aks-engine/pull/4352 and due to AKS Engine deprecation.

PR #4352, opened 2021-04-01
feat: configurable Linux eth0 MTU

jackfrancis:
**Reason for Change**:

This PR introduces a `eth0MTU` int property to `linuxProfile`, that can be used to configure either `1500` or `3900` for the eth0 interface.

We don't accept this configuration vector if you are in a kubenet configuration.

**Issue Fixed**:

In service of #4350 

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -452,6 +452,9 @@ MASTER_CONTAINER_ADDONS_PLACEHOLDER
       mount --bind $MOUNT_DIR $MOUNT_DIR
     fi
     mount --make-shared $MOUNT_DIR
+{{- if IsAzureCNI}}
+    ifconfig eth0 mtu {{GetEth0MTU}} up
```
khenidak, line 456: This is not persisted between restarts.

jackfrancis: putting the statement into this script means the ifconfig update happens before every systemd kubelet startup

jackfrancis: In other words, doing it this way tightly couples the configuration to kubelet/k8s (even if the MTU is changed out of band for some reason, kubelet will always set it to the desired value before it starts up) rather than making an OS-wide config one time at cluster startup.

(Also, this was the easiest way to do it, so... :))

pkg/api/vlabs/validate.go
```diff
@@ -622,6 +623,27 @@ func (a *Properties) validateZones() error {
...
+	var validEth0MTU bool
+	if a.LinuxProfile.Eth0MTU != 0 {
+		if a.OrchestratorProfile != nil &&
+			a.OrchestratorProfile.KubernetesConfig != nil &&
+			a.OrchestratorProfile.KubernetesConfig.NetworkPlugin == NetworkPluginKubenet {
+			return errors.Errorf("Custom linuxProfile eth0MTU value not allowed when using Kubenet")
+		}
+		for _, valid := range linuxEth0MTUAllowedValues {
+			if valid == a.LinuxProfile.Eth0MTU {
+				validEth0MTU = true
```
aramase, line 635: We can break here

pkg/api/vlabs/validate.go
```diff
@@ -622,6 +623,27 @@ func (a *Properties) validateZones() error {
...
+			}
+		}
+		if !validEth0MTU {
+			allowedMTUs := ""
+			for _, mtu := range linuxEth0MTUAllowedValues {
+				allowedMTUs += strconv.Itoa(mtu) + ", "
+			}
+			allowedMTUs = strings.TrimRight(allowedMTUs, ", ")
+			return errors.Errorf("Invalid linuxProfile eth0MTU value \"%d\", please use one of the following versions: %s", a.LinuxProfile.Eth0MTU, allowedMTUs)
```
aramase, line 644:
```suggestion
			return errors.Errorf("Invalid linuxProfile eth0MTU value \"%d\", please use one of the following values: %s", a.LinuxProfile.Eth0MTU, allowedMTUs)
```

nit

aramase (approved):
lgtm

PR #4358, opened 2021-04-07
feat: don't install nvidia drivers if nvidia-device-plugin is disabled

jackfrancis:
**Reason for Change**:

This PR makes a change so that nvidia drivers are not installed if the nvidia-device-plugin addon is disabled.

Additionally, because the known-working drivers that CSE installs don't work with containerd, we don't change the addons default flow so that the nvidia-device-plugin addon is disabled if containerd is the CRI.

The practical outcome of this is that you may use containerd with N series VM SKUs, and no nvidia drivers will be installed. The nvidia gpu-operator implementation solves this, see:

https://developer.nvidia.com/blog/announcing-containerd-support-for-the-nvidia-gpu-operator/

A new E2E scenario has been added for N series + containerd configurations, which installs the nvidia-curated gpu-operator helm chart, and then validates using the existing CUDA job.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma (approved):
/lgtm

Michael-Sinz (commented):
There needs to be some documentation about the fact that the GPU Driver is not installed if in containerd and that it is up to the customer to install the nVidia GPU Operator.

PR #4359, opened 2021-04-07
test: fix prior k8s minor version in mocks and upgrade test

mboersma:
**Reason for Change**:

Some unit test code was naively calculating the prior Kubernetes version by decrementing the minor but leaving the major and patch versions intact. This will fail when Kubernetes 1.18.18 comes out next week, since it will supply "1.17.18" which doesn't and won't exist.

Instead let's reuse the `RationalizeReleaseAndVersion` func to get the latest patch for the previous minor release correctly.

**Issue Fixed**:

Prevents a test bug that will arise next Wednesday [when v1.18.18 drops](https://github.com/kubernetes/sig-release/blob/master/releases/patch-releases.md#118).

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/armhelpers/mockclients.go
+++ b/pkg/armhelpers/mockclients.go
@@ -45,7 +45,7 @@ func init() {
 	versionSplit := strings.Split(defaultVersion, ".")
 	minorVersion, _ := strconv.Atoi(versionSplit[1])
 	minorVersionLessOne := minorVersion - 1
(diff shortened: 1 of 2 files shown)


jackfrancis (approved):
/lgtm

PR #4363, opened 2021-04-09
feat: deprecate Azure CNI networkmonitor daemonset

jackfrancis:
**Reason for Change**:

This PR removes the installation of the additional "azure-cni-networkmonitor" daemonset, as this is no longer required with Azure CNI 1.20+ scenarios.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/common/const.go
```diff
@@ -103,8 +103,6 @@ const (
 	ContainerMonitoringAddonName = "container-monitoring"
 	// IPMASQAgentAddonName is the name of the ip masq agent addon
 	IPMASQAgentAddonName = "ip-masq-agent"
-	// AzureCNINetworkMonitorAddonName is the name of the Azure CNI networkmonitor addon
-	AzureCNINetworkMonitorAddonName = "azure-cni-networkmonitor"
```
mboersma, line 107: If we remove the const definition, will cluster templates which have `enabled: True` for the `azure-cni-networkmonitor` addon now raise an error? I don't remember how this would behave... And if it does error, are we ok with that instead of having a "deprecated addon was ignored"-type message?

jackfrancis: Good call-out, I'll add some deprecation foo for clusters transitioning during an upgrade to the new "no Azure CNI networkmonitor addon exists" reality.

jackfrancis: Made some changes, check out the latest commit

mboersma (approved):
/lgtm!

ludydoo:
Hello! Does this mean that aks with kubernetes 1.20.2 should uninstall azure-cni-networkmonitor ?

jackfrancis:
I believe @paulgmiller can answer your question @ludydoo. My understanding is that AKS is already working on delivering a networkmonitor-free Azure CNI solution.

ludydoo:
Thanks @paulgmiller . I'm asking because I'm experiencing bugs with azure iptables when using Networkpolicies. Seems I have to restart the azure-npm pods. Wondering if there's a race condition between azure-cni-networkmonitor and azure-npm... Really not sure at this point

paulgmiller:
Anything with all agents on azure cni  >= 1.20 doesn't  network monitor. It shouldn't in theory conflict with azure npm but if you fidn that it does we should make an issue of in the azure-container-networking repo.

PR #4364, opened 2021-04-12
fix: Fix installer for custom package URL's.

cpuguy83:
**Reason for Change**:

Ran into this in CI where I'm trying to test new packages.
runc that's already installed in the vm image is older than the minimum
version of the new package, as such the package installation fails and
cluster provisioning errors out.

apt has functionality to fix dependency problems found while using
`dpkg`.

This adds a helper function to replace `dpkg install` with one that
basically calls `apt-get update && apt-get install -f` after a failed
install (update is required as well, otherwise the package list is out
of date).

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -139,13 +139,21 @@ apt_get_update() {
   done
   echo Executed apt-get update $i times
 }
+dpkg_install() {
+  retries=$1; wait_sleep=$2; shift && shift;
+  for i in $(seq 1 $retries); do
+    wait_for_apt_locks; dpkg -i "${1}" && break; apt-get update --fix-missing; do_apt_get_install -f && break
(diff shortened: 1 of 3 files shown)


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -139,13 +139,21 @@ apt_get_update() {
   done
   echo Executed apt-get update $i times
 }
+dpkg_install() {
+  retries=$1; wait_sleep=$2; shift && shift;
+  for i in $(seq 1 $retries); do
+    wait_for_apt_locks; dpkg -i "${1}" && break; apt-get update --fix-missing; do_apt_get_install -f && break
```
jackfrancis, line 145: clever usage of the `-f` there in the new `do_apt_get_install` func

jackfrancis (approved):
/lgtm

PR #4376, opened 2021-04-21
chore: Upgrade nvidia drivers to 450.80.02

gdippolito:
**Reason for Change**:

Upgrade the Nvidia drivers for GPU machine to 450.80.02. 

This version of the drivers is the same used in AzureML based images since April 2021. This change will add compatibility with newer version of CUDA (11.1,11.2,11.3). Reference https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html 

**Issue Fixed**:

None. It adds support to run docker images based on CUDA >= 11.1

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Should be pretty safe since it is a minor version upgrade.

### DIFF
--- a/docs/topics/gpu.md
+++ b/docs/topics/gpu.md
@@ -4,7 +4,7 @@ If you created a Kubernetes cluster with one or multiple Ubuntu node pools whose
 
 The NVIDIA drivers are automatically installed on every GPU agent in your cluster, so you don't need to do that manually, unless you require a specific version of the drivers. Currently, the installed driver is version 418.40.04.
 
-Note: You will get version 450.51.06 of the drivers if you are using Ubuntu 18.04-LTS, and version 418.40.04 drivers if you are using 16.04-LTS.
+Note: You will get version 450.80.02 of the drivers if you are using Ubuntu 18.04-LTS, and version 418.40.04 drivers if you are using 16.04-LTS.
 
 To make sure everything is fine, run `kubectl describe node <name-of-a-gpu-node>`. You should see the correct number of GPU reported (in this example shows 2 GPU for a NC12 VM):
 
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -12,7 +12,7 @@ if [[ ${OS} == "${UBUNTU_OS_NAME}" ]]; then
 fi
 DOCKER=/usr/bin/docker
 if [[ $UBUNTU_RELEASE == "18.04" ]]; then
-  export GPU_DV=450.51.06
+  export GPU_DV=450.80.02
 else
   export GPU_DV=418.40.04
 fi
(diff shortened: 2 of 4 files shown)


jackfrancis:
Thanks @gdippolito, is there a newer version of basic CUDA validation we can run in E2E that you know of? We're still running the v0.1 image from 4+ years ago, see:

https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/cuda-vector-add.yaml

gdippolito:
> Thanks @gdippolito, is there a newer version of basic CUDA validation we can run in E2E that you know of? We're still running the v0.1 image from 4+ years ago, see:
> 
> https://github.com/Azure/aks-engine/blob/master/test/e2e/kubernetes/workloads/cuda-vector-add.yaml

Thanks for your comment @jackfrancis. I have updated the e2e test with a more recent image which uses CUDA 11.1.

I have done few tests locally and this sample image does not seem to work with the version of the drivers I picked (it works on 460.x but not on 450.80.02). I'm a bit confused since according to the Nvidia documentation this is expected to work.

jackfrancis:
@gdippolito thanks! I've tested your new image, and the E2E tests pass fine. So I think we're good here.

cc @Michael-Sinz @jadarsie  are there any concerns with the following:

- updating the driver version
- testing `nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.1-ubuntu18.04` instead of `k8s.gcr.io/cuda-vector-add:v0.1`

To be clear, these driver changes are specific to docker-backed clusters. If you're using containerd you can't use these drivers, and are probably using the gpu-operator already if you've figured out how to make this work with aks-engine-created clusters.

Michael-Sinz:
Not that I know of.  Have not specifically tested that mix but we are also moving to containerd soon (I hope)

__
Michael Sinz C Architect C <http://midori/> v C Microsoft
________________________________

jadarsie:
> @gdippolito thanks! I've tested your new image, and the E2E tests pass fine. So I think we're good here.
> 
> cc @Michael-Sinz @jadarsie are there any concerns with the following:
> 
>     * updating the driver version
> 
>     * testing `nvcr.io/nvidia/k8s/cuda-sample:vectoradd-cuda11.1-ubuntu18.04` instead of `k8s.gcr.io/cuda-vector-add:v0.1`
> 
> 
> To be clear, these driver changes are specific to docker-backed clusters. If you're using containerd you can't use these drivers, and are probably using the gpu-operator already if you've figured out how to make this work with aks-engine-created clusters.

FYI @penggu

jackfrancis (approved):
/lgtm

penggu:
Looks good to me.

gdippolito:
Shall we :ship: it?

PR #4389, opened 2021-04-28
fix: omsagent-win crash in Windows containerd nodes

mainred:
**Reason for Change**:

omsagent uses `c:/var/log/containers/` path to tail the container logs for both docker & containerd windows nodes. 
As we removed docker from containerd VHD by https://github.com/Azure/aks-engine/pull/4282, C:\ProgramData\docker\containers` does not exit in containerd VHD, leaving omsagent-win in a failed state.
```
Container Statuses:
- omsagent-win {&ContainerStateWaiting{Reason:CreateContainerError,Message:failed to generate container "237e9fa0fc55c8708495343db61b905b5b932eaaa5b72140dc8ad42649890f77" spec: failed to generate spec: failed to stat "C:\\ProgramData\\docker\\containers": CreateFile C:\ProgramData\docker\containers: The system cannot find the path specified.,} nil nil} ready:false 
```

NOTE: `c:/var/log/containers/` works for both docker & containerd but on docker nodes, log files under `c:/var/log/containers` not accessible without mounting `C:\ProgramData\docker\containers`. 

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/container-monitoring.yaml
+++ b/parts/k8s/addons/container-monitoring.yaml
@@ -746,6 +746,7 @@ spec:
         - name: docker-windows-containers
           hostPath:
             path: C:\ProgramData\docker\containers
+            type: DirectoryOrCreate
         - name: settings-vol-config
           configMap:
             name: container-azm-ms-agentconfig
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -8460,6 +8460,7 @@ spec:
         - name: docker-windows-containers
           hostPath:
             path: C:\ProgramData\docker\containers
+            type: DirectoryOrCreate
         - name: settings-vol-config
           configMap:
             name: container-azm-ms-agentconfig


mainred:
@ganga1980

ganga1980 (approved):

ganga1980:
> > log files under c:/var/log/containers not accessible without mounting C:\ProgramData\docker\containers.
> 
> This seems like a bug in OMS. If it can't find that folder it fails? If it uses `c:\var\log\containers` why does it need the docker folder? Will there be missing info in Container insights without that folder?

>  @jsturtevant, @marosset , This is not bug in OMS rather issue with Docker. Regardless of the omsagent or not, without mounting C:\ProgramData\docker\containers, c:/var/log/containers not accessible.  This should be reproducible with any container.

ganga1980:
Hi, @mainred, are you not planning to merge this PR?

mainred:
> This seems like a bug in OMS. If it can't find that folder it fails? If it uses c:\var\log\containers why does it need the docker folder? Will there be missing info in Container insights without that folder?

Sorry for the late rely. In Windows docker case, `c:\var\log\containers` is a link to `C:\ProgramData\docker\containers`, and oms-agent requires this mount to read container logs.
We can use a if-else to skip the mount in case of Windows containerd, but it seems passing runtime variable here is not that easy.

mainred:
@ganga1980 , yes that's my plan. Before that let upstream folks have a full understanding of this issue and fix.

ganga1980:
> > This seems like a bug in OMS. If it can't find that folder it fails? If it uses c:\var\log\containers why does it need the docker folder? Will there be missing info in Container insights without that folder?
> 
> Sorry for the late rely. In Windows docker case, `c:\var\log\containers` is a link to `C:\ProgramData\docker\containers`, and oms-agent requires this mount to read container logs.
> We can use a if-else to skip the mount in case of Windows containerd, but it seems passing runtime variable here is not that easy.

> Is there any concern or issue 

> > This seems like a bug in OMS. If it can't find that folder it fails? If it uses c:\var\log\containers why does it need the docker folder? Will there be missing info in Container insights without that folder?
> 
> Sorry for the late rely. In Windows docker case, `c:\var\log\containers` is a link to `C:\ProgramData\docker\containers`, and oms-agent requires this mount to read container logs.
> We can use a if-else to skip the mount in case of Windows containerd, but it seems passing runtime variable here is not that easy.

> @mainred, @jsturtevant , @marosset , AFIK, detecting container runtime in yaml is not possible. using "DirectoryOrCreate" shouldnt harm anything other than creating empty dir on containerd nodes. Are there any concerns on using "DirectoryOrCreate" apart we are creating empty dir in containerd nodes?

jsturtevant:
I am not convinced this is related to docker, the windows OMS is mapping in this directory here and so must be trying to read this in some way: https://github.com/Azure/aks-engine/blob/e01249356fb6ed7752ba4d1946edd12914a128b1/parts/k8s/addons/container-monitoring.yaml#L777-L780

Why is this being mapped in and not just read from `c:/var/log/containers/` which it isn't doing currently.

We have deprecated this project and we don't seem to have anyone that is requiring this change so going to close for now.

/close

PR #4399, opened 2021-05-04
feat: get-logs from VMs unable to join the cluster

jadarsie:
**Reason for Change**:

Added extra flag `--vm-names` to `aks-engine get-logs` to enable log collection from nodes that are not able to join the cluster.

**Credit Where Due**:

- [x] No
- [ ] Yes

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/cmd/get_logs.go
+++ b/cmd/get_logs.go
@@ -52,6 +52,7 @@ type getLogsCmd struct {
 	outputDirectory        string
 	controlPlaneOnly       bool
 	uploadSASURL           string
+	nodeNames              []string
 	// computed
 	cs                  *api.ContainerService
 	locale              *gotext.Locale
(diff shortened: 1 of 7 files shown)


cmd/get_logs.go
```diff
@@ -91,8 +92,9 @@ func newGetLogsCmd() *cobra.Command {
 	command.Flags().StringVar(&glc.linuxScriptPath, "linux-script", "", "path to the log collection script to execute on the cluster's Linux nodes (required if distro is not aks-ubuntu)")
 	command.Flags().StringVar(&glc.windowsScriptPath, "windows-script", "", "path to the log collection script to execute on the cluster's Windows nodes (required if distro is not aks-windows)")
 	command.Flags().StringVarP(&glc.outputDirectory, "output-directory", "o", "", "collected logs destination directory, derived from --api-model if missing")
-	command.Flags().BoolVarP(&glc.controlPlaneOnly, "control-plane-only", "", false, "get logs from control plane VMs only")
+	command.Flags().BoolVarP(&glc.controlPlaneOnly, "control-plane-only", "", false, "get logs from control plane VMs only (unless --vm-names is set)")
```
jackfrancis, line 95: we probably shouldn't allow the `control-plane-only` option to be used at the same time as the new `vm-names` option (in which case we shouldn't need to mention the "unless --vm-names is set" disclaimer)

PR #4402, opened 2021-05-07
chore: Update nvidia runtime

cpuguy83:
This greatly simplifies what's required to install gpu drivers since the
packages no longer depend on docker-ce.

**Reason for Change**:

Older nvidia packages required docker-ce packages.
This is no longer a problem with the v3 drivers and greatly simplifies the install process.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

Still need to actually allow the driver installation when containerd is used.

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -12,15 +12,14 @@ if [[ ${OS} == "${UBUNTU_OS_NAME}" ]]; then
 fi
 DOCKER=/usr/bin/docker
 if [[ $UBUNTU_RELEASE == "18.04" ]]; then
-  export GPU_DV=450.80.02
+  export GPU_DV=450.51.06
 else
   export GPU_DV=418.40.04
 fi
 export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VER=2.0.0
-NVIDIA_DOCKER_SUFFIX=docker18.09.2-1
+NVIDIA_CONTAINER_RUNTIME_VER='3.4.2*'
 PRIVATE_IP=$( (ip -br -4 addr show eth0 || ip -br -4 addr show azure0) | grep -Po '\d+\.\d+\.\d+\.\d+')
 if ! [[ $(echo -n "$PRIVATE_IP" | grep -c '^') == 1 ]]; then
   PRIVATE_IP=$(hostname -i)
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -12,14 +12,14 @@ if [[ ${OS} == "${UBUNTU_OS_NAME}" ]]; then
 fi
 DOCKER=/usr/bin/docker
 if [[ $UBUNTU_RELEASE == "18.04" ]]; then
-  export GPU_DV=450.51.06
+  export GPU_DV=450.80.02
 else
   export GPU_DV=418.40.04
 fi
 export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VER='3.4.2*'
+NVIDIA_CONTAINER_RUNTIME_VER='3.5.0*'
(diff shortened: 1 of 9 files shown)


pkg/engine/templates_generated.go
```diff
@@ -12724,15 +12712,14 @@ if [[ ${OS} == "${UBUNTU_OS_NAME}" ]]; then
 fi
 DOCKER=/usr/bin/docker
 if [[ $UBUNTU_RELEASE == "18.04" ]]; then
-  export GPU_DV=450.80.02
+  export GPU_DV=450.51.06
```
Michael-Sinz, line 12715: Did we mean to drop the version back to 450.51.06 from 450.80.02?

cpuguy83:
That's the version that ships with the package.

On May 7, 2021, GitHub ***@***.***> wrote:
> @Michael-Sinz commented on this pull request.
>

> In pkg/engine/templates_generated.go <https://github.com/Azure/aks-
> engine/pull/4402#discussion_r628262202>:
>

> @@ -12724,15 +12712,14 @@ if [[ ${OS} == "${UBUNTU_OS_NAME}" ]]; then
fi DOCKER=/usr/bin/docker if [[ $UBUNTU_RELEASE == "18.04" ]]; then -
export GPU_DV=450.80.02 + export GPU_DV=450.51.06
> Did we mean to drop the version back to 450.51.06 from 450.80.02?
>

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/Azure/aks-engine/pull/4402#pullrequestreview-
654530552>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAGDCZRZPFTMAGXDD5XCTXTTMP3QLANCNFSM44IGNREA>.

pkg/engine/templates_generated.go
amrmahdi: The current deployment with the operator is `460.73.01`

cpuguy83: Not sure why they are using an older version here. The package is certainly newer. Best to open an issue upstream to get the version updated with the package (or have them split the drivers to a separate package).

cpuguy83: @RenaudWasTaken any idea why the nvidia-container-runtime 3.4 package would have older versions of the drivers than the 2.0.0 packages does?

cpuguy83: I see this we are actually downloading it in the provisioning script... most likely the function to trigger the download never ran and it had whatever version was baked into the VHD.

jackfrancis: Is the conclusion that we can use the newer version (450.80.02) with the v3 nvidia runtime?

Michael-Sinz: Note that the nVidia operator code was installing drivers with version 460.* (which seems like a big jump from 450.*)

cpuguy83: We should be able to the issue is I can't seem to make the download happen when deploying the cluster so it just has whatever is in the VHD.. and the versions end up not matching.

jackfrancis: @cpuguy83 I can get around that by using a vanilla 18.04 OS (not the VHD). I'll revert that driver downgrade and try.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -19,8 +19,7 @@ fi
 export GPU_DEST=/usr/local/nvidia
 NVIDIA_DOCKER_VERSION=2.0.3
 DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VER=2.0.0
-NVIDIA_DOCKER_SUFFIX=docker18.09.2-1
+NVIDIA_CONTAINER_RUNTIME_VER='3.5.0*'
```
jackfrancis, line 22: @cpuguy83 apt told me there was an even newer version so I updated to get it, and validated that it works

jackfrancis:
This has been tested as O.K. @ commit e66739abfbb719029225ce0e01fa871b03f0f3dd for both docker and containerd using the vanilla Canonical 18.04 OS image

jackfrancis (approved):
lgtm

Michael-Sinz:
lgtm

parts/k8s/cloud-init/artifacts/cse_helpers.sh
cpuguy83: Cool, this wasn't on the repo but tagged in GH when I was setting this up originally. 🎉

cpuguy83:
Changes LGTM

PR #4421, opened 2021-05-17
chore: download nvidia drivers during VHD

jackfrancis:
**Reason for Change**:

This PR ensures that all nvidia drivers are downloaded during VHD creation, and installed from local files during CSE.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -141,12 +141,12 @@ apt_get_update() {
 apt_get_download() {
   retries=$1; wait_sleep=$2; shift && shift;
   local d=${PWD} ret=0
-  cd /var/cache/apt/archives/
+  pushd /var/cache/apt/archives/ || return 1
   for i in $(seq 1 $retries); do
     wait_for_apt_locks; apt-get -o Dpkg::Options::=--force-confold download -y "${1}" && break
     if [ $i -eq $retries ]; then ret=1; else sleep $wait_sleep; fi
   done
-  cd $d
+  popd
   return $ret
 }
 dpkg_install() {
--- a/parts/k8s/cloud-init/artifacts/cse_helpers.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_helpers.sh
@@ -17,14 +17,12 @@ else
   export GPU_DV=418.40.04
 fi
 export GPU_DEST=/usr/local/nvidia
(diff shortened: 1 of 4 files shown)


parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -17,14 +17,12 @@ else
   export GPU_DV=418.40.04
 fi
 export GPU_DEST=/usr/local/nvidia
-NVIDIA_DOCKER_VERSION=2.0.3
```
jackfrancis, line 20: none of these vars are re-used anywhere

jackfrancis:
cc @Michael-Sinz @jadarsie @haofan-ms 

I haven't done a robust historical search, but I'm confident-ish that GPU node pools + airgap has never worked. This is a proposed solution to that.

Michael-Sinz:
@jackfrancis Thank you for this work - we have not heavily invested in GPU work in air-gapped yet - we will be needing to do that soon so, thank you.

/lgtm

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -17,14 +17,12 @@ else
...
-DOCKER_VERSION=1.13.1-1
-NVIDIA_CONTAINER_RUNTIME_VER='3.5.0*'
 PRIVATE_IP=$( (ip -br -4 addr show eth0 || ip -br -4 addr show azure0) | grep -Po '\d+\.\d+\.\d+\.\d+')
 if ! [[ $(echo -n "$PRIVATE_IP" | grep -c '^') == 1 ]]; then
   PRIVATE_IP=$(hostname -i)
 fi
 export PRIVATE_IP
+APT_CACHE_DIR=/var/cache/apt/archives/
```
Michael-Sinz, line 25: Not for now, but at some point we can look at cleaning up the cache after the node has fully provisioned.  (Just to free up space for containers/etc)

Michael-Sinz (approved):

parts/k8s/cloud-init/artifacts/cse_helpers.sh
jackfrancis: Makes sense, and I think more generally the apt cache can probably be cleaned up after the bootstrap script finishes (probably in the background, non-blocking)

PR #4422, opened 2021-05-18
fix: distinguish ctrd and smalldisk windows image in upgrade scenario

haofan-ms:
**Reason for Change**:

During upgrade, the ImagePublisherAndOfferMatch function itself is not able to distinguish smalldisk and ctrd image, since they have the same ImageOffer and ImagePublisher. This is causing mixed cluster previously using smalldisk image to always try to upgrade to use ctrd image, since it's the first condition. Adding the identifier for ImageSku to help distinguish these two cases.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/cmd/upgrade.go
+++ b/cmd/upgrade.go
@@ -32,11 +32,11 @@ import (
 )
 
 const (
-	upgradeName             		= "upgrade"
-	upgradeShortDescription 		= "Upgrade an existing AKS Engine-created Kubernetes cluster"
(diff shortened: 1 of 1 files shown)


jackfrancis (approved):
/lgtm

jadarsie (approved):

PR #4270, opened 2021-02-12
[WIP] chore: update metrics-server to 0.4.2

jackfrancis:
**Reason for Change**:

This PR updates the metrics-server implementation to use v0.4.2. See:

https://github.com/kubernetes-sigs/metrics-server/releases/tag/v0.4.2

**Issue Fixed**:

**Credit Where Due**:

Does this change contain code from or inspired by another project?

- [ ] No
- [ ] Yes

If "Yes," did you notify that project's maintainers and provide attribution?

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] uses [conventional commit messages](https://www.conventionalcommits.org/)
  
- [ ] includes documentation
- [ ] adds unit tests
- [ ] tested upgrade from previous version

**Notes**:

PR #4424, opened 2021-05-18
chore: bump metrics-server to v0.4.4

chewong:
Signed-off-by: Ernest Wong <chuwon@microsoft.com>

**Reason for Change**:

Bump metrics-server to v0.4.4 for 1.22+ cluster to fix test failure in https://testgrid.k8s.io/sig-windows-master-release#aks-engine-windows-dockershim-serial-slow-master. Also reflecting changes introduced in https://github.com/kubernetes-sigs/metrics-server/pull/201/files.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/addons/metrics-server.yaml
```diff
@@ -130,6 +130,17 @@ spec:
         - /metrics-server
         - --kubelet-insecure-tls
         - --kubelet-preferred-address-types=InternalIP
+        - --cert-dir=/tmp
```
jackfrancis, line 133: Are these changes back-compat to v0.3.7? (that's what all the k8s versions besides 1.22 are using)

jackfrancis:
I had tried unsuccessfully to update to 0.4.2 a while back, see:

https://github.com/Azure/aks-engine/pull/4270

I notice you haven't made as many changes as I had (if I remember correctly I was moving over stuff from the upstream example yaml spec that wasn't in the aks-engine spec)

chewong:
Didn't notice you have an opening PR to upgrade metrics-server. I will work on top of your branch to avoid duplicate work then.

jackfrancis:
> Didn't notice you have an opening PR to upgrade metrics-server. Let me rebase this off your branch and test

Well, except mine never worked. :)

chewong:
looks like we need `--kubelet-insecure-tls` in the args, but it's for testing purposes only:
```
--kubelet-insecure-tls                                    Do not verify CA of serving certificates presented by Kubelets.  For testing purposes only.
```
@jackfrancis do you know if it's okay to use this flag?

jsturtevant:
> @jackfrancis do you know if it's okay to use this flag?

I believe the deployment here has always had that flag.  

Ideally I believe you would generate cert using https://kubernetes.io/docs/reference/access-authn-authz/certificate-signing-requests/ and then start the metric server with that cert

jackfrancis (approved):
/lgtm

pierluigilenoci:
Just few days before the new release 😢 
Ref: https://github.com/Azure/aks-engine/issues/4455

PR #4427, opened 2021-05-20
test: use kured 1.7.0 to validate kamino

jackfrancis:
**Reason for Change**:

This PR uses the kured 1.7.0 chart now that 1.7.0 has been released w/ node annotations support. See:

https://github.com/weaveworks/kured/releases/tag/1.7.0

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


test/e2e/kubernetes/kubernetes_test.go
```diff
@@ -2828,8 +2828,11 @@ var _ = Describe("Azure Container Cluster using the Kubernetes Orchestrator", fu
 		It("should be able to install vmss node prototype", func() {
 			if cfg.RunVMSSNodePrototype {
 				if eng.ExpandedDefinition.Properties.HasVMSSAgentPool() {
-					By("Installing kured with node annotations configuration")
-					_, err := daemonset.CreateDaemonsetFromFileWithRetry(filepath.Join(WorkloadDir, "kured-annotations.yaml"), "kured", "kube-system", 5*time.Second, cfg.Timeout)
+					By("Installing kured 1.7.0 with node annotations configuration")
+					cmd := exec.Command("helm", "install", "--wait", "--generate-name", "--repo", "https://weaveworks.github.io/kured", "kured", "--version", "1.7.0", "--set", "configuration.annotateNodes=true", "--set", "configuration.period=1m")
+					util.PrintCommand(cmd)
+					out, err := cmd.CombinedOutput()
+					log.Printf("%s\n", out)
```
Michael-Sinz, line 2835: Now that this does the annotations, we could run vmss-prototype with the automatic mode where it uses the annotations.  That would help validate that too and simplify some of the logic here.

Also, kamino no longer deletes the "prototype" node, it just reboots it (temporary shutdown in order to take the snapshot).

You just need to tell kamino what annotations to look for via kamino.auto.lastPatchAnnotation and, more importantly, kamino.auto.pendingRebootAnnotation.

These values tell the names of the the annotations within the node.  I think you want:
```
kamino.auto.lastPatchAnnotation=weave.works/kured-most-recent-reboot-needed
kamino.auto.pendingRebootAnnotation=weave.works/kured-reboot-in-progress
```

There is also the cronjob mechanism but you don't want that as you want to signal directly.

Michael-Sinz: Oops, I did not scroll down far enough in the unchaged code - you already do that.  :-)

Michael-Sinz (approved):

Michael-Sinz (dismissed):
/lgtm

test/e2e/kubernetes/kubernetes_test.go
jackfrancis: Definitely. I wanted to quickly get rid of my static manifest referencing my custom-built kured image first.

Next step will be to integrate this test scenario into a "soak" cluster so that we can validate the nominal use case of vmss prototype being a "self-maintaining" VMSS optimization.

PR #4432, opened 2021-05-21
fix: Run CSI controller driver on master node

ZeroMagic:
**Reason for Change**:

There is a failure in some cases of https://testgrid.k8s.io/provider-azure-cloud-provider-azure#cloud-provider-azure-serial
Following is the detail error log from CSI controller.
```
I0521 06:16:31.327041       1 azure_armclient.go:430] Received error in put.send: resourceID: /subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-p9qac7rm/providers/Microsoft.Compute/disks/pvc-919384c7-e152-4819-84ae-21dfaad82307, error: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 403, RawError: Retriable: false, RetryAfter: 0s, HTTPStatusCode: 403, RawError: {"error":{"code":"AuthorizationFailed","message":"The client '4eef356b-d1ce-4a7a-83f6-5c4c26edbf6b' with object id '4eef356b-d1ce-4a7a-83f6-5c4c26edbf6b' does not have authorization to perform action 'Microsoft.Compute/disks/write' over scope '/subscriptions/0e46bd28-a80f-4d3a-8200-d9eb8d80cb2e/resourceGroups/kubetest-p9qac7rm/providers/Microsoft.Compute/disks/pvc-919384c7-e152-4819-84ae-21dfaad82307' or the scope is invalid. If access was recently granted, please refresh your credentials."}}
```
The AuthorizationFailed reason could be that the leader controller is in the agent node which does not have the access.

So this change will ensure the controller in master node is used to create PV.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/azuredisk-csi-driver-deployment.yaml
+++ b/parts/k8s/addons/azuredisk-csi-driver-deployment.yaml
@@ -635,6 +635,7 @@ spec:
       serviceAccountName: csi-azuredisk-controller-sa
       nodeSelector:
         kubernetes.io/os: linux
+        kubernetes.io/role: master
       priorityClassName: system-cluster-critical
       tolerations:
         - key: "node-role.kubernetes.io/master"
@@ -818,6 +819,7 @@ spec:
       serviceAccountName: csi-snapshot-controller-sa
       nodeSelector:
         kubernetes.io/os: linux
+        kubernetes.io/role: master
       priorityClassName: system-cluster-critical
       tolerations:
         - key: "node-role.kubernetes.io/master"
--- a/parts/k8s/addons/azurefile-csi-driver-deployment.yaml
+++ b/parts/k8s/addons/azurefile-csi-driver-deployment.yaml
@@ -650,6 +650,7 @@ spec:
       serviceAccountName: csi-azurefile-controller-sa
       nodeSelector:
         kubernetes.io/os: linux
+        kubernetes.io/role: master
       priorityClassName: system-cluster-critical
       tolerations:
         - key: "node-role.kubernetes.io/master"
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -3826,6 +3826,7 @@ spec:
       serviceAccountName: csi-azuredisk-controller-sa
       nodeSelector:
         kubernetes.io/os: linux
+        kubernetes.io/role: master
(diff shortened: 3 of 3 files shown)


jackfrancis:
thx @ZeroMagic 

@andyzhangx is there a backlog item that addresses the likelihood of replicas > 1 resulting in the pod being scheduled onto a non-control-plane node? It sounds like the CSI controller pod doesn't work when scheduled onto a worker node. If so, we wanna make sure that we address this with node selector criteria rather than this, which seems like a workaround.

Lemme know if I have the wrong idea!

andyzhangx:
> thx @ZeroMagic
> 
> @andyzhangx is there a backlog item that addresses the likelihood of replicas > 1 resulting in the pod being scheduled onto a non-control-plane node? It sounds like the CSI controller pod doesn't work when scheduled onto a worker node. If so, we wanna make sure that we address this with node selector criteria rather than this, which seems like a workaround.
> 
> Lemme know if I have the wrong idea!

@ZeroMagic Agree, should upgrade CSI driver version in aks-engine, and then add `kubernetes.io/role: master` in NodeSlector config for controller, I found we already enabled `runOnMaster` in CSI driver upstream e2e test:
https://github.com/kubernetes-sigs/azuredisk-csi-driver/blob/d04c4fcf1f94fddf547f455f811e28d1c868207b/charts/latest/azuredisk-csi-driver/templates/csi-azuredisk-controller.yaml#L23-L27

andyzhangx (dismissed):
/lgtm
/approve
would you also change the PR title and description? thanks.

ZeroMagic:
Currently, add `kubernetes.io/role: master` in NodeSelector. As for the upgrade change, will do it in the future.

feiskyer (approved):
Thanks for fixing the issue. LGTM

jackfrancis (dismissed):
/lgtm

thanks!

jackfrancis:
generated code needs updating (`make generate`)

ZeroMagic:
> generated code needs updating (`make generate`)

done

jackfrancis (approved):
/lgtm

PR #4443, opened 2021-06-04
fix: Graceful coredns shutdown

technicianted:
**Reason for Change**:
This PR adds graceful shutdown configurations to coredns. 

Currently, if a coredns pod is deleted (due to eviction for example), the container shuts down immediately. This has negative side effects:
1. Some in-flight requests may fail.
2. A transient race where some requests may still be sent to the shutdown instance until this pod's IP address have been removed from the service.
If any of the two happens, we end up with the dreadful 5s timeout/retry latency.

The PR adds [`lameduck` configuration](https://github.com/coredns/coredns/blob/master/plugin/health/README.md) to the `health` plugin such that it would delay shutting down the service until 5 seconds after the readiness probes had failed. This guarantees that both in-flight queries are completed and enough time for this instance IP address to be discarded from the service.

Note: Since original readiness probe parameters were left at defaults, the PR also explicitly sets the default to make it clearer.

**Credit Where Due**:
technicianted

**Does this change contain code from or inspired by another project?**

- [X] No
- [ ] Yes

**Requirements**:

- [X] includes documentation
- [ ] adds unit tests

**Notes**:

The issue can be reproduced by running `dig` in a loop (100ms delay) then deleting/evicting a coredns pod.

### DIFF
--- a/parts/k8s/addons/coredns.yaml
+++ b/parts/k8s/addons/coredns.yaml
@@ -62,7 +62,10 @@ data:
     import conf.d/Corefile*
     .:53 {
         errors
-        health
+        health {
+          # this should be > readiness probe failure time
+          lameduck 35s
+        }
         ready
         kubernetes {{ContainerConfig "domain"}} in-addr.arpa ip6.arpa {
             pods insecure
@@ -200,6 +203,10 @@ spec:
             path: /ready
             port: 8181
             scheme: HTTP
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:
--- a/pkg/engine/templates_generated.go
+++ b/pkg/engine/templates_generated.go
@@ -8536,7 +8536,10 @@ data:
     import conf.d/Corefile*
     .:53 {
         errors
-        health
+        health {
+          # this should be > readiness probe failure time
+          lameduck 35s
+        }
         ready
         kubernetes {{ContainerConfig "domain"}} in-addr.arpa ip6.arpa {
             pods insecure
@@ -8674,6 +8677,10 @@ spec:
             path: /ready
             port: 8181
             scheme: HTTP
+          periodSeconds: 10
+          timeoutSeconds: 1
+          successThreshold: 1
+          failureThreshold: 3
         securityContext:
           allowPrivilegeEscalation: false
           capabilities:


Michael-Sinz (approved):
Thank you so much for this fix.  This will have significant positive impacts on clusters built with this.

Michael-Sinz:
@jackfrancis - It will be great to be able to get a point release with this change (0.63.1 at least).

Osama did a great job finding the root cause of this problem and identify such a simple fix.

parts/k8s/addons/coredns.yaml
```diff
@@ -62,7 +62,10 @@ data:
     import conf.d/Corefile*
     .:53 {
         errors
-        health
+        health {
+          # this should be > readiness probe failure time
+          lameduck 35s
```
jackfrancis, line 67: So the idea here is to fallback to the coredns runtime health check in case readinessProbe fails to restart things after 3 failures separated by 10 seconds?

technicianted: We want to make sure that coredns will not terminate until it is safe for it to do.

The idea here is when coredns gets a SIGTERM, `lameduck` will delay the actual shutdown for 35s. Until then, coredns will continue to service requests. In same time, `readiness` plugin will start failing. We wait 10*3+5 seconds to make sure that this instance is and no longer will be getting any queries.

jackfrancis: Ah, thanks for the clarification. Just to be super clear, 3 failures, separated by 10 seconds, could be as short as just over *20* seconds. It should never be as long as 30 seconds. So if it helps, we could reduce the lameduck config to 30s.

Michael-Sinz: The thing is to delay the eviction of the pod for 35 seconds but trigger health failures right away at the start of the eviction.  This will then, after the health probe failure (3x10) remove the pod from the service (and thus the load balancing for the service) such that shutting it down will not trigger DNS requests to route to the now dead pod.

It was technically a race between not sending any requests to the pod and the pod no longer running.  The old way always ended up with the pod stopping before the requests stopped coming and with this new mechanism this is the inverse now - it always stops sending requests to the pod before the pod stops running.

Michael-Sinz: The trick is that there is some time after the health fails that the kubernetes infrastructure removes the pod from the service so the extra 5 seconds is to allow this to fully propagate through the whole cluster.  Remember, some node may have just started to want to do DNS at the same moment that the 30 seconds runs out and thus we could end up routing to the bad pod before the route change got pushed everywhere.

jackfrancis: Makes sense, lgtm, thanks for this improvement!

jackfrancis (approved):
/lgtm

PR #4459, opened 2021-06-12
feat: allow windows custom images per agent pool

mweibel:
**Reason for Change**:
Allows a different windows image per agent pool. This brings the windows agent image support on par with what linux pools support. 

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
I'm not sure if this is a PR you'd accept so I just added the functionality without fixing/adding tests. Will do so as soon as you give it a go :) 

This would allow e.g. the [custom shared gallery image](https://github.com/Azure/aks-engine/blob/master/examples/custom-shared-image.json) example to also work for windows:
```json
{
    "apiVersion": "vlabs",
    "properties": {
      "masterProfile": {
        "imageReference": {
          "name": "linuxvm",
          "resourceGroup": "sig",
          "subscriptionID": "00000000-0000-0000-0000-000000000000",
          "gallery": "siggallery",
          "version": "0.0.1"
        },
        "count": 1,
        "dnsPrefix": "",
        "vmSize": "Standard_D2_v3"
      },
      "agentPoolProfiles": [
        {
          "name": "agentpool1",
          "count": 3,
          "osType": "Windows",
          "imageReference": {
            "name": "windowsvm",
            "resourceGroup": "sig",
            "subscriptionID": "00000000-0000-0000-0000-000000000000",
            "gallery": "siggallery",
            "version": "0.0.1"
          },
          "vmSize": "Standard_D2_v3",
          "availabilityProfile": "AvailabilitySet"
        }
      ],
      "linuxProfile": {
        "adminUsername": "azureuser",
        "ssh": {
          "publicKeys": [
            {
              "keyData": ""
            }
          ]
        }
      }
    }
  }
```

### DIFF
--- a/pkg/engine/params.go
+++ b/pkg/engine/params.go
@@ -102,13 +102,14 @@ func getParameters(cs *api.ContainerService, generatorCode string, aksEngineVers
 			addValue(parametersMap, fmt.Sprintf("%sScaleSetEvictionPolicy", agentProfile.Name), agentProfile.ScaleSetEvictionPolicy)
 		}
 
+		if agentProfile.ImageRef != nil {
+			addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), agentProfile.ImageRef.Name)
+			addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), agentProfile.ImageRef.ResourceGroup)
+		}
+
 		// Unless distro is defined, default distro is configured by defaults#setAgentProfileDefaults
 		//   Ignores Windows OS
 		if !(agentProfile.OSType == api.Windows) {
-			if agentProfile.ImageRef != nil {
-				addValue(parametersMap, fmt.Sprintf("%sosImageName", agentProfile.Name), agentProfile.ImageRef.Name)
-				addValue(parametersMap, fmt.Sprintf("%sosImageResourceGroup", agentProfile.Name), agentProfile.ImageRef.ResourceGroup)
-			}
 			addValue(parametersMap, fmt.Sprintf("%sosImageOffer", agentProfile.Name), cloudSpecConfig.OSImageConfig[agentProfile.Distro].ImageOffer)
 			addValue(parametersMap, fmt.Sprintf("%sosImageSKU", agentProfile.Name), cloudSpecConfig.OSImageConfig[agentProfile.Distro].ImageSku)
 			addValue(parametersMap, fmt.Sprintf("%sosImagePublisher", agentProfile.Name), cloudSpecConfig.OSImageConfig[agentProfile.Distro].ImagePublisher)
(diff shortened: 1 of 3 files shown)


mweibel:
is there anything I can do to make this mergeable @mboersma?

jsturtevant:
fyi @abelhu I think we discussed doing this at some point. Could you take a look too?

jsturtevant:
Looks like this allows for placement of the on both the profile and the agent pool.  What would happen if both are specified?  I think we would want the agent pool to take priority. This way you could specify a default (and it would be backwards compatible) but could override individual pools if required.  Also need various validation checks such as we would want to make sure all agent profiles are filled or the windows profile is set.  

I would also think we would want to enable all the different types of image references we support: https://github.com/Azure/aks-engine/blob/master/docs/topics/clusterdefinitions.md#windows-images  It looks like this currently works with just shared image gallery?

@jadarsie is this something interesting to you as well?

jadarsie:
> @jadarsie is this something interesting to you as well?

SIG is not an option on Azure Stack Hub, so the feature is less relevant in that space.

I did not dig deep into the implementation details, but adding tests may require quite a bit of work as it should include all commands (deploy, scale, upgrade, update, add-pool, etc).

mweibel:
in the end this change is more or less removing some windows specific branches as linux already supports it. 
I'd be happy to provide necessary tests and validate some of the questions - I haven't tried different cases yet as for my use case the SIG case is the important one. 

I haven't dug into the tests just yet, I believe e2e tests are what's most important for validating, right? 
Any pointers as to what tests I could take as an example to implement the changes?

jadarsie:
> in the end this change is more or less removing some windows specific branches as linux already supports it.
> I'd be happy to provide necessary tests and validate some of the questions - I haven't tried different cases yet as for my use case the SIG case is the important one.
> 
> I haven't dug into the tests just yet, I believe e2e tests are what's most important for validating, right?
> Any pointers as to what tests I could take as an example to implement the changes?

I only have bad suggestions or ideas that require work beyond the scope of your work. Currently, the PR gates (`.pipelines/pr-e2e.yaml`) only run `aks-engine deploy`. Executing extra operations can slow down the process significantly. That's probably why those are not included.

We (ASH) execute daily, on our own, `test/e2e/cluster.sh` for extra assurance. But the boilerplate scripts that we use are not in here yet. 

@jackfrancis, do you have a suggestion?

jackfrancis:
I think this is a significant refactor. It's pretty easy to defend that this would be an improvement, but it's a decent amount of work.

@marosset may have thoughts as well

mweibel:
pinging this again to get an update :) I'd be happy to invest more time into it given I get some pointers what you'd expect for this to be merged. 
Also let me know if I can somehow influence (as an azure customer) the priority of this PR. 

For now we keep our fork for running aks engine which works fine until there are big code changes in the areas I needed to change.

PR #4465, opened 2021-06-14
Update to the nVidia 460.32.03 drivers

### LINKED ISSUE #4464: Need to upgrade nVidia drivers to 460+ from 450.* to support CUDA 11
**Describe the request**
CUDA 11 is needed for a number of projects and code fixes as there are some memory leaks that are only fixed with the code bases that depend on CUDA 11.

aks-engine needs to be up-to-date with its drivers in order to support the latest GPU workloads

Have 460.* be the 18.04 driver support version.

The nVidia controller that could install the drivers is not yet production ready code.  If it were, we would just disable the driver install in aks-engine and use that.

Michael-Sinz:
This is needed in order to support CUDA 11 which supports some large
model memory leak fixes.

**Reason for Change**:
Needed for latest CUDA version

**Issue Fixed**:
Fixes #4464 

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests


jackfrancis (dismissed):
/lgtm

PR #4466, opened 2021-06-15
fix: stop using ServiceNodeExclusion feature gate after it is GA

chewong:
Signed-off-by: Ernest Wong <chuwon@microsoft.com>

**Reason for Change**:

Fixes test failure in https://testgrid.k8s.io/sig-windows-master-release#aks-engine-windows-containerd-master

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [x] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [x] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-controller-manager.go
+++ b/pkg/api/defaults-controller-manager.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/Azure/aks-engine/pkg/api/common"
 	"github.com/Azure/go-autorest/autorest/to"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/api/defaults-controller-manager.go
```diff
@@ -81,7 +82,10 @@ func (cs *ContainerService) setControllerManagerConfig() {
 	}
 
 	// Enables Node Exclusion from Services (toggled on agent nodes by the alpha.service-controller.kubernetes.io/exclude-balancer label).
-	addDefaultFeatureGates(o.KubernetesConfig.ControllerManagerConfig, o.OrchestratorVersion, "1.9.0", "ServiceNodeExclusion=true")
+	// ServiceNodeExclusion feature gate is GA in 1.19, removed in 1.22 (xref: https://github.com/kubernetes/kubernetes/pull/100776)
+	if !common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.19.0") {
```
jsturtevant, line 86: nit: little awkward to read but looks like we don't have a LessThan function

jackfrancis: yeah this is the common pattern for less than

jackfrancis: /lgtm

mboersma (approved):
/lgtm++

PR #4511, opened 2021-06-26
feat: add Standard_HB120rs_v3 support

mweibel:
**Reason for Change**:
Add missing VM type. This type is only available in some availability zones (e.g. westeurope zone 3). 

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/armhelpers/httpMockClientData/listResourceSkus.json
+++ b/pkg/armhelpers/httpMockClientData/listResourceSkus.json
@@ -1037,6 +1037,11 @@
       "AcceleratedNetworking": false
     },
     {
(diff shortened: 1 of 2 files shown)


jackfrancis:
https://github.com/Azure/aks-engine/pull/4525 will do this, and we've set up a github action so that new VM SKUs are added automatically

Thanks for nudging us @mweibel!

mweibel:
great, that's of course even better :) Btw the autoscaler might benefit from something like that too :wink: (see https://github.com/kubernetes/autoscaler/pull/4166)

PR #4528, opened 2021-06-28
ci: use go 1.16 everywhere

jackfrancis:
**Reason for Change**:

This PR rationalizes the various go CI to use 1.16 everywhere.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


.go-version
```diff
@@ -1 +0,0 @@
-1.14.4
```
jackfrancis, line 1: I wasn't able to find any reason to keep this hidden file around

mboersma (approved):
/lgtm

.go-version
mboersma: Maybe it was to support [`goenv`](https://github.com/syndbg/goenv)? I used to use that, and I'm the only commiter on that file. 👍🏻 to deleting it.

PR #4546, opened 2021-06-29
feat: flexvol and csi addons disabled by default

jackfrancis:
**Reason for Change**:

This PR changes the default addons installation flow so that the following addons are no longer installed by default:

- blobfuse flexvolume
- keyvault flexvolume

The above addons will be disabled by default because they have been deprecated in favor of alternate solutions

- secrets-csi-driver

The above addon will be disabled by default because it is no longer actively maintained, and the preferred maintenance gesture is to use the official helm chart.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @andyzhangx @aramase

andyzhangx:
it's ok to disable blobfuse flexvol driver since it's deprecated, user should use blob csi drive instead, thanks.

jackfrancis:
@andyzhangx thx for confirming! :)

Michael-Sinz (approved):
Ahh, this is so good to see - I can (eventually) reduce the list of things I have to disable as they were enabled by default.

Only concern would be teams that may depend on this.  We will need to make sure that this is called out expressly in the release notes as a "breaking" change and that the user would need to expressly enable these addons if they wanted to continue to use them.

jackfrancis:
@Michael-Sinz the current release CI allows project maintainers to edit and curate the release notes before merging to the release branch (that merge acts as a gate to actually validate and publish release artifacts). So you'll have a chance to weigh in then to ensure we're doing sufficient communication.

Michael-Sinz:
> @Michael-Sinz the current release CI allows project maintainers to edit and curate the release notes before merging to the release branch (that merge acts as a gate to actually validate and publish release artifacts). So you'll have a chance to weigh in then to ensure we're doing sufficient communication.

Will have to see how that works out - I hope it gives hints as to what all of the changes were from the prior release.

mboersma (approved):

PR #4555, opened 2021-06-30
 fix: don't start kubelet if reboot is imminent

jackfrancis:
**Reason for Change**:

This PR ensures that worker nodes do not start the kubelet systemd service if they are going to be rebooted by CSE.

Additionally, we clean things up so that control plane node *do* always fully start up, even if they are going to be rebooted. This is because we may need to do one-time (delivered via aks-engine api model) resource bootstrapping, and for that we need a working cluster, which each control plane node during cluster creation essentially represents. This will result in a bit more control plane node readiness thrashing during cluster creation (kubelet may come online and then be rebooted a minute later), but this is not a big problem.

This solves the problem of new nodes coming online after cluster creation (for example, node scale out operations) causing workload thrashing due to new nodes coming online as Ready, accepting new work, and then rebooting a minute later (thus force-killing any scheduled containers scheduled onto that new node).

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -387,14 +387,30 @@ ensureKubelet() {
     sed -i "s|<advertiseAddr>|$PRIVATE_IP|g" $f
   fi
   wait_for_file 1200 1 /opt/azure/containers/kubelet.sh || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  {{- if HasKubeReservedCgroup}}
+{{- if HasKubeReservedCgroup}}
   wait_for_file 1200 1 /etc/systemd/system/{{- GetKubeReservedCgroup -}}.slice || exit {{GetCSEErrorCode "ERR_KUBERESERVED_SLICE_SETUP_FAIL"}}
   wait_for_file 1200 1 /etc/systemd/system/kubelet.service.d/kubereserved-slice.conf || exit {{GetCSEErrorCode "ERR_KUBELET_SLICE_SETUP_FAIL"}}
-  {{- end}}
-  systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{- end}}
+  if [[ -n ${MASTER_NODE} ]]; then
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  else
+{{- if not RunUnattendedUpgrades}}
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{else}}
+    retrycmd 120 5 25 systemctl enable kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{- end}}
+  fi
 {{- if HasKubeletHealthZPort}}
   wait_for_file 1200 1 /etc/systemd/system/kubelet-monitor.service || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  if [[ -n ${MASTER_NODE} ]]; then
+    systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  else
+  {{- if not RunUnattendedUpgrades}}
+    systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  {{else}}
+    retrycmd 120 5 25 systemctl enable kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  {{- end}}
+  fi
 {{- end}}
 }
 
@@ -407,8 +423,8 @@ ensureAddons() {
 {{- end}}
 {{- if not HasCustomPodSecurityPolicy}}
   retrycmd 120 5 30 $KUBECTL get podsecuritypolicy privileged restricted || exit_cse {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}} $GET_KUBELET_LOGS
-  rm -Rf ${ADDONS_DIR}/init
 {{- end}}
+  rm -Rf ${ADDONS_DIR}/init
   replaceAddonsInit
   {{/* Force re-load all addons because we have changed the source location for addon specs */}}
   retrycmd 10 5 30 ${KUBECTL} delete pods -l app=kube-addon-manager -n kube-system || \
@@ -483,7 +499,7 @@ installKubeletAndKubectl() {
   rm -rf ${binPath}/kubelet-* ${binPath}/kubectl-* /home/hyperkube-downloads &
 }
 ensureK8sControlPlane() {
-  if [ -f /var/run/reboot-required ] || [ "$NO_OUTBOUND" = "true" ]; then
+  if [ "$NO_OUTBOUND" = "true" ]; then
     return
   fi
   retrycmd 120 5 25 $KUBECTL 2>/dev/null cluster-info || exit_cse {{GetCSEErrorCode "ERR_K8S_RUNNING_TIMEOUT"}} $GET_KUBELET_LOGS
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -397,7 +397,7 @@ ensureKubelet() {
 {{- if not RunUnattendedUpgrades}}
     systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
 {{else}}
-    retrycmd 120 5 25 systemctl enable kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+    systemctl_enable 100 5 30 kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
 {{- end}}
   fi
 {{- if HasKubeletHealthZPort}}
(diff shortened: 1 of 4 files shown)


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,14 +387,30 @@ ensureKubelet() {
...
   wait_for_file 1200 1 /opt/azure/containers/kubelet.sh || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  {{- if HasKubeReservedCgroup}}
+{{- if HasKubeReservedCgroup}}
   wait_for_file 1200 1 /etc/systemd/system/{{- GetKubeReservedCgroup -}}.slice || exit {{GetCSEErrorCode "ERR_KUBERESERVED_SLICE_SETUP_FAIL"}}
   wait_for_file 1200 1 /etc/systemd/system/kubelet.service.d/kubereserved-slice.conf || exit {{GetCSEErrorCode "ERR_KUBELET_SLICE_SETUP_FAIL"}}
-  {{- end}}
-  systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{- end}}
+  if [[ -n ${MASTER_NODE} ]]; then
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
```
jackfrancis, line 395: 1. If we are running the control plane, we *always* start kubelet

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,14 +387,30 @@ ensureKubelet() {
...
+  else
+{{- if not RunUnattendedUpgrades}}
```
jackfrancis, line 397: 2. Otherwise, if we are a worker node:
- start kubelet right now if we're not running `unattended-upgrade` during bootstrap
- but *only enable* (i.e., configure to start automatically on system boot) kubelet systemd job if we're going to run `unattended-upgrade`, because there's a good chance we're going to need to reboot after we do that

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,14 +387,30 @@ ensureKubelet() {
...
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{else}}
+    retrycmd 120 5 25 systemctl enable kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{- end}}
+  fi
 {{- if HasKubeletHealthZPort}}
   wait_for_file 1200 1 /etc/systemd/system/kubelet-monitor.service || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  if [[ -n ${MASTER_NODE} ]]; then
```
jackfrancis, line 405: Follow the same strategy above for the `kubelet-monitor` systemd job

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -407,8 +423,8 @@ ensureAddons() {
 {{- end}}
 {{- if not HasCustomPodSecurityPolicy}}
   retrycmd 120 5 30 $KUBECTL get podsecuritypolicy privileged restricted || exit_cse {{GetCSEErrorCode "ERR_ADDONS_START_FAIL"}} $GET_KUBELET_LOGS
-  rm -Rf ${ADDONS_DIR}/init
 {{- end}}
+  rm -Rf ${ADDONS_DIR}/init
```
jackfrancis, line 427: This rm statement should have always been outside the PSP if block, we want to do it every time after we process all addons from init directory (the above three if blocks do so).

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -483,7 +499,7 @@ installKubeletAndKubectl() {
   rm -rf ${binPath}/kubelet-* ${binPath}/kubectl-* /home/hyperkube-downloads &
 }
 ensureK8sControlPlane() {
-  if [ -f /var/run/reboot-required ] || [ "$NO_OUTBOUND" = "true" ]; then
```
jackfrancis, line 486: We actually *always* want to start the control plane, even if we're going to reboot at the end of the bootstrap script, because the bootstrap script will depend upon the control plane being online to complete successfully.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -254,11 +254,7 @@ if [[ -n ${MASTER_NODE} ]]; then
     time_metric "EnsureEtcd" ensureEtcd
   fi
   time_metric "EnsureK8sControlPlane" ensureK8sControlPlane
-  if [ -f /var/run/reboot-required ]; then
-    time_metric "ReplaceAddonsInit" replaceAddonsInit
-  else
-    time_metric "EnsureAddons" ensureAddons
-  fi
+  time_metric "EnsureAddons" ensureAddons
```
jackfrancis, line 257: Again, addons are a control plane-enforceable bit of foo, so we want to do that now always. If we skip this in a "will reboot later" scenario, that means we will not perform any addons validation at all (CSE does not re-run after a reboot).

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -298,6 +294,12 @@ if [ -f /var/run/reboot-required ]; then
     aptmarkWALinuxAgent unhold &
   fi
 else
+{{- if RunUnattendedUpgrades}}
```
jackfrancis, line 297: If we *are not* going to reboot, and we ran `unattended-upgrade`, then we need to start `kubelet` and `kubelet-monitor`, because they were not started earlier.

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -298,6 +294,12 @@ if [ -f /var/run/reboot-required ]; then
...
+  if [[ -z ${MASTER_NODE} ]]; then
+    systemctl_restart 100 5 30 kubelet
```
Michael-Sinz, line 299: Why restart?  I would expect that you could just "systemctl start kubelet" which is what will happen at reboot time anyway.  You have already enabled it.  This way you could just blanket "start" and if it is already running, nothing will happen and if it is not, it will start it.

Less need to check for master or have other complications.

The problem with restart is that if it is running, it will stop the service first.

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,14 +387,30 @@ ensureKubelet() {
...
 {{- if HasKubeletHealthZPort}}
   wait_for_file 1200 1 /etc/systemd/system/kubelet-monitor.service || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
-  systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  if [[ -n ${MASTER_NODE} ]]; then
+    systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  else
+  {{- if not RunUnattendedUpgrades}}
+    systemctlEnableAndStart kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  {{else}}
+    retrycmd 120 5 25 systemctl enable kubelet-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
```
Michael-Sinz, line 411: You likely need a systemctl daemon-reload here before the enable.  Just above you have just added the service file (potentially) and thus will need that reload to make systemctl know about it.  Your function for systemctlEnableAndStart calls systemctl_restart  first which does the systemctl daemon-reload.

(This is based on the fact that line 404 waits for the file to show up)

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -387,14 +387,30 @@ ensureKubelet() {
...
-  {{- end}}
-  systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{- end}}
+  if [[ -n ${MASTER_NODE} ]]; then
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+  else
+{{- if not RunUnattendedUpgrades}}
+    systemctlEnableAndStart kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
+{{else}}
+    retrycmd 120 5 25 systemctl enable kubelet || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
```
Michael-Sinz, line 400: Given that you are building the kubelet.service specification, you will need to systemctl daemon-reload before you systemctl enable kubelet here (just like I mentioned in the code below)

Michael-Sinz (changes requested):
Overall, this is great but I think two cases of systemctl daemon-reload are needed to handle some potential edge cases/races that the code is trying to handle today.

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: In this particular condition we know that it hasn't started yet, because when we scaffolded kubelet systemd service earlier we explicitly only "enabled" the service in a truthy `RunUnattendedUpgrades` context.

Your comment is true generally, though. We could add a new `systemctl_start` convenience func in `parts/k8s/cloud-init/artifacts/cse_helpers.sh` and audit the code to use that instead of "restart" in all the places where restart could be disruptive.

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: Good catch, thanks!

parts/k8s/cloud-init/artifacts/cse_main.sh
Michael-Sinz: If we had that helper function (and it would be a trivial function) we could then unconditionally make sure the services are started rather than having to check if we are unattended masters or not.

However, this is a great first step.

Michael-Sinz (approved):

jackfrancis:
cc @mboersma @jadarsie to review this

jadarsie (approved):

PR #4560, opened 2021-07-01
chore: use runc v1.0.0 GA

### LINKED ISSUE #4512: Error: failed to create containerd task: ... mount destination <<path>> not absolute: unknown
I had the similar issue on 
The error log was like this.

I fixed my issue by patching my Dockerfile.

**Describe the bug**

Failed to run Pod with the error by cointainerd. 

I'm not sure this is bug or not. And I can share an workaround.

**Steps To Reproduce**

1. Build and push the docker image from this dockerfile.

```dockerfile
FROM ubuntu
VOLUME dest
```

2. Run new Pod from the docker image above and mount dest to a volume. 

3. Pod is in crash-loop. The error log is like this.

```
Error: failed to create containerd task: OCI runtime create failed: in
valid mount {Destination:dest Type:bind Source:/var/lib/containerd/io.containerd.grpc.v1.cri/containers/copy-theme/volumes/8ede16043b6dbf587c941f360bfb
2038dd35baa61092da3029aafe4622ce4b88 Options:[rbind rprivate rw]}: mount destination dest not absolute: unknown
```

**Expected behavior**

Runs with no error. At least older AKS (containerd) accepted.

**AKS Engine version**

Azure AKS version 1.20.7. (Container Runtime Version:  containerd://1.4.4+azure)

**Kubernetes version**

1.20.7

**Additional context**

This issue can be avoided with the fixed Dockerfile like this.

```dockerfile
FROM ubuntu
VOLUME /dest
# must be added `/` prefix.
```

I suppose this issue is caused by breaking changes in cointainerd.
I found some similar issues like:
https://github.com/containerd/containerd/issues/5547
https://github.com/moby/moby/issues/42480

jackfrancis:
**Reason for Change**:

This PR updates the installed version of runc to v1.0.0 GA

**Issue Fixed**:

Fixes #4512 

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -93,8 +93,8 @@ mobyPkgVersion() {
 installRunc() {
   local v
   v=$(runc --version | head -n 1 | cut -d" " -f3)
-  if [[ $v != "1.0.0-rc92" ]]; then
-    apt_get_install 20 30 120 moby-runc=1.0.0~rc92* --allow-downgrades || exit 27
+  if [[ $v != "1.0.0" ]]; then
+    apt_get_install 20 30 120 moby-runc=1.0.0 --allow-downgrades || exit 27
   fi
 }
 installMoby() {
--- a/parts/k8s/cloud-init/artifacts/cse_install.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_install.sh
@@ -93,8 +93,8 @@ mobyPkgVersion() {
 installRunc() {
   local v
   v=$(runc --version | head -n 1 | cut -d" " -f3)
-  if [[ $v != "1.0.0" ]]; then
-    apt_get_install 20 30 120 moby-runc=1.0.0 --allow-downgrades || exit 27
+  if [[ $v != "1.0.0+azure" ]]; then
+    apt_get_install 20 30 120 moby-runc=1.0.0+azure --allow-downgrades || exit 27
   fi
 }
 installMoby() {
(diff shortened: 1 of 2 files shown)


cpuguy83 (dismissed):
LGTM

cpuguy83:
Except everything failed for some reason...

jackfrancis:
@cpuguy83

```
+ apt-get install -o Dpkg::Options::=--force-confold --no-install-recommends -y moby-runc=1.0.0 --allow-downgrades
Reading package lists...
Building dependency tree...
Reading state information...
E: Version '1.0.0' for 'moby-runc' was not found
```

cpuguy83:
The version is really `1.0.0+azure`

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -93,8 +93,8 @@ mobyPkgVersion() {
 installRunc() {
   local v
   v=$(runc --version | head -n 1 | cut -d" " -f3)
-  if [[ $v != "1.0.0-rc92" ]]; then
-    apt_get_install 20 30 120 moby-runc=1.0.0~rc92* --allow-downgrades || exit 27
```
cpuguy83, line 97: Notice we had a `*` here before to catch package revision.

jackfrancis: Got it, thanks. So to ensure we don't passively install a newer version, the literal string "1.0.0+azure" should work?

cpuguy83: It's a good idea (IMHO) to do `1.0.0*` because there may be tweaks to the packaging.
The full version is `1.0.0+azure-1` Where the `-1` denotes the iteration of the package.

jackfrancis:
runc bump is here: https://github.com/Azure/aks-engine/pull/4562

PR #4561, opened 2021-07-01
chore: update containerd to 1.4.6

jackfrancis:
**Reason for Change**:

This PR updates the installed version of containerd to 1.4.6.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kubernetesparams.t
+++ b/parts/k8s/kubernetesparams.t
@@ -233,7 +233,7 @@
       "type": "string"
     },
     "containerdVersion": {
-      "defaultValue": "1.4.4",
+      "defaultValue": "1.4.6",
       "metadata": {
         "description": "The Azure Moby build version"
       },
@@ -246,7 +246,8 @@
          "1.3.7",
          "1.3.8",
          "1.3.9",
-         "1.4.4"
+         "1.4.4",
+         "1.4.6"
        ],
       "type": "string"
     },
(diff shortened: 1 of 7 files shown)


parts/k8s/kubernetesparams.t
```diff
@@ -246,7 +246,8 @@
          "1.3.7",
          "1.3.8",
          "1.3.9",
-         "1.4.4"
+         "1.4.4",
+         "1.4.6"
```
Michael-Sinz, line 250: You have a list of valid versions - is there a reason for this?  Does this list translate to versions already on the VHD?

Michael-Sinz (approved):

parts/k8s/kubernetesparams.t
jackfrancis: It's largely there for historical purposes, it's the set of "known-working" (i.e., tested with Kubernetes on Azure in the aks-engine CSE) versions. And because there is a facility to accept an arbitrary containerd version string in the api model.

cpuguy83 (approved):
LGTM

jsturtevant (approved):

jadarsie (approved):

PR #4562, opened 2021-07-01
chore: update moby/docker, and runc for linux only

jackfrancis:
**Reason for Change**:

This PR updates the default moby (docker) version to 20.10.7 for both Linux and Windows.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -31,7 +31,7 @@ const (
 
 const (
 	// KubernetesWindowsDockerVersion is the default version for docker on Windows nodes in kubernetes
-	KubernetesWindowsDockerVersion = "20.10.5"
+	KubernetesWindowsDockerVersion = "20.10.7"
 	// KubernetesDefaultWindowsSku is the default SKU for Windows VMs in kubernetes
 	KubernetesDefaultWindowsSku = "Datacenter-Core-1809-with-Containers-smalldisk"
 	// KubernetesDefaultWindowsRuntimeHandler is the default containerd handler for windows pods
@@ -442,7 +442,7 @@ const (
 	// DefaultKubernetesDNSServiceIPv6 specifies the IPv6 address that kube-dns listens on by default. must by in the default Service CIDR range.
 	DefaultKubernetesDNSServiceIPv6 = "fd00::10"
 	// DefaultMobyVersion specifies the default Azure build version of Moby to install.
-	DefaultMobyVersion = "19.03.14"
+	DefaultMobyVersion = "20.10.7"
 	// DefaultContainerdVersion specifies the default containerd version to install.
 	DefaultContainerdVersion = "1.4.6"
 	// DefaultDockerBridgeSubnet specifies the default subnet for the docker bridge network for masters and agents.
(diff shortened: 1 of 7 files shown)


vhd/packer/configure-windows-vhd.ps1
```diff
@@ -176,7 +176,7 @@ function Install-ContainerD {
 }
 
 function Install-Docker {
-    $defaultDockerVersion = "20.10.5"
+    $defaultDockerVersion = "20.10.7"
```
jsturtevant, line 179: Windows releases are via Mirantis Docker EE which can be behind the Linux moby releases and sometimes skip patch releases.  Currently the latest release version is `20.10.6`

pkg/api/const.go
```diff
@@ -31,7 +31,7 @@ const (
 
 const (
 	// KubernetesWindowsDockerVersion is the default version for docker on Windows nodes in kubernetes
-	KubernetesWindowsDockerVersion = "20.10.5"
+	KubernetesWindowsDockerVersion = "20.10.7"
```
jsturtevant, line 34: Same here for the Windows Docker EE version

jsturtevant (changes requested):

vhd/packer/configure-windows-vhd.ps1
jsturtevant: you can find latest from the link at: https://dockermsft.azureedge.net/dockercontainer/DockerMsftIndex.json

jackfrancis: Got it, thanks!

pkg/engine/templates_generated.go
```diff
@@ -16525,7 +16525,8 @@ var _k8sKubernetesparamsT = []byte(`    "etcdServerCertificate": {
          "19.03.12",
          "19.03.13",
          "19.03.14",
-         "20.10.5"
+         "20.10.5",
```
Michael-Sinz, line 16528: Should this just be a value that is a default and then, if the user picks something else, the code just uses it?  As it is, I doubt that we have done regression testing against all of these versions and the current kuberenetes/aks-engine versions.

Michael-Sinz (approved):
Only question is why we have a list of allowed versions for moby/docker versions.  Are they somehow specially tested or just a history of what was once valid.

jackfrancis:
> Only question is why we have a list of allowed versions for moby/docker versions. Are they somehow specially tested or just a history of what was once valid.

It's the latter. Just doing the minimal maintenance effort here, though it's worth acknowledging that making the effort to unwind the ARM template validations will make things less confusing. We can do that work in a discrete PR if we want.

jackfrancis:
@jsturtevant @AbelHu updated Windows to use the .6 patch release of docker

cpuguy83 (approved):
LGTM

AbelHu (approved):

jsturtevant (approved):

PR #4563, opened 2021-07-02
feat: allow setting apiserver verbosity and change default from 4 to 2

mweibel:
**Reason for Change**:
I think not all users need verbosity level 4 for the apiserver. This logs all requests sent to it and is really quite verbose.

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:
- [x] includes documentation
- [x] adds unit tests

**Notes**:
BREAKING CHANGE: the default apiServer verbosity changes from 4 to 2.

### DIFF
--- a/pkg/api/const.go
+++ b/pkg/api/const.go
@@ -453,6 +453,8 @@ const (
 	DefaultKubernetesMaxPodsAzureCNI = "30"
 	// DefaultKubernetesAPIServerEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
 	DefaultKubernetesAPIServerEnableProfiling = "false"
+	// DefaultKubernetesAPIServerVerbosity is the default verbosity setting for the apiserver
+	DefaultKubernetesAPIServerVerbosity = "2"
 	// DefaultKubernetesCtrMgrEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
 	DefaultKubernetesCtrMgrEnableProfiling = "false"
 	// DefaultKubernetesSchedulerEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
--- a/pkg/api/defaults-apiserver_test.go
+++ b/pkg/api/defaults-apiserver_test.go
@@ -536,3 +536,29 @@ func TestAPIServerAnonymousAuth(t *testing.T) {
 			a["--anonymous-auth"])
 	}
 }
+
+func TestAPIServerConfigChangeVerbosity(t *testing.T) {
+	// Test
+	// "apiServerConfig": {
+	// 	"--v": "4"
+	// },
+	cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
+	cs.Properties.OrchestratorProfile.KubernetesConfig.APIServerConfig = map[string]string{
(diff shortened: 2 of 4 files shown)


jadarsie:
lgtm from a functional POV, however I would like to learn why this particular flag was considered static at some point.

mweibel:
the previous implementation has been added by @jackfrancis in https://github.com/Azure/aks-engine/commit/d216b9438773b007e6e7dfb152ed120ce994bd76 but doesn't mention any specifics. Maybe it was simply not considered important enough to be able to change? 
Other services let also change the verbosity plus the default stays at level 4, so the change shouldn't break anything in this regard.

jackfrancis:
Thanks @mweibel, there probably isn't a defensible reason for this being 4, and not configurable, just historical.

pkg/api/const.go
```diff
@@ -453,6 +453,8 @@ const (
 	DefaultKubernetesMaxPodsAzureCNI = "30"
 	// DefaultKubernetesAPIServerEnableProfiling is the config that enables profiling via web interface host:port/debug/pprof/
 	DefaultKubernetesAPIServerEnableProfiling = "false"
+	// DefaultKubernetesAPIServerVerbosity is the default verbosity setting for the apiserver
+	DefaultKubernetesAPIServerVerbosity = "4"
```
jackfrancis, line 457: While we're making this improvement, let's go ahead and change the default to 2.

pkg/api/defaults-apiserver_test.go
```diff
@@ -536,3 +536,29 @@ func TestAPIServerAnonymousAuth(t *testing.T) {
...
 }
+
+func TestAPIServerConfigChangeVerbosity(t *testing.T) {
+	// Test
+	// "apiServerConfig": {
+	// 	"--v": "2"
+	// },
+	cs := CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
+	cs.Properties.OrchestratorProfile.KubernetesConfig.APIServerConfig = map[string]string{
+		"--v": "2",
```
jackfrancis, line 547: And if we change the default to 2, let's update this UT case to override to 4 instead

jackfrancis (changes requested):

mweibel:
Thanks for the review. 
I agree and changed the default from 4 to 2.

Michael-Sinz:
> Thanks for the review.
> I agree and changed the default from 4 to 2.

Please change the commit note/pull request text to mentions that this change not only adds configurability but also changes the default value.  That is a "breaking" change for any users who are used to the default verbosity that existed.  (Not that there should be much of that but it should be called out)

mweibel:
@Michael-Sinz adjusted the commit msg and the PR text (added `Notes`). Is this how you'd like to have it formatted?

Michael-Sinz:
> @Michael-Sinz adjusted the commit msg and the PR text (added `Notes`). Is this how you'd like to have it formatted?

Nice - I have added to the title as that is many times the only thing people read :-(

Michael-Sinz (approved):
/lgtm

jadarsie (approved):

mweibel:
ping?

jackfrancis (approved):
/lgtm

PR #4575, opened 2021-07-12
release: create branch manually

jackfrancis:
**Reason for Change**:

This PR updates the "create-release-branch" GitHub Action to manually create a new release branch from the checked out branch, because `peterjgrainger/action-create-branch@v2.0.1` is always creating the branch from `master`.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.github/workflows/create-release-branch.yaml
+++ b/.github/workflows/create-release-branch.yaml
@@ -18,9 +18,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{github.event.inputs.from_branch}}
-      - name: set env
-        run: |
-          echo "RELEASE_BRANCH=release-${{github.event.inputs.release_version}}" >> $GITHUB_ENV # release-v1.0.0 substring starting at v1.0.0
       - name: Setup SSH Keys and known_hosts
         env:
           SSH_AUTH_SOCK: /tmp/ssh_agent.sock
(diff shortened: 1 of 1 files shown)


.github/workflows/create-release-branch.yaml
```diff
@@ -96,11 +93,9 @@ jobs:
           AZURE_CORE_ONLY_SHOW_ERRORS: True
         run: make test-kubernetes
       - name: create branch
-        uses: peterjgrainger/action-create-branch@v2.0.1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          branch: '${{ env.RELEASE_BRANCH }}'
+        run: |
+          git branch release-${{github.event.inputs.release_version}}
```
Michael-Sinz, line 97: Did you want to use git branch -c such that the original branch is maintained (not renamed)?

What this does is lose the current branch and replace it with the release branch.  With the `-c` option, it would keep the source branch (leave it alone) and create the release branch as the same ref hash/log.

It all depends on the behavior you want as to the source branch - does it get deleted or not?

jackfrancis: I'm actually simply trying to create a *new* branch from the commit of the source branch, and push it to the Azure/aks-engine remote. This didn't work, so I'm trying this:

https://github.com/Azure/aks-engine/pull/4576

Issue #4595, opened 2021-07-21
Clusters in bridge mode may fail from unattended upgrades


mboersma:
*Summary*

In the last 48 hours, some AKS Engine clusters have experienced nodes falling into a `NotReady` state, or lost connectivity to the cluster if all control plane nodes were affected.

To remediate this situation, reboot the affected nodes, starting with the control plane. Node VMs can be rebooted in the [Azure portal](https://portal.azure.com) or with the `az` command line tool:

- `az vm restart`
- [`az vmss restart`](https://docs.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-manage-cli#restart-vms-in-a-scale-set)

**Describe the bug**

Before October 2020, AKS Engine releases defaulted to creating clusters with "bridge" Kubernetes `NetworkMode`. A recent Ubuntu update to the systemd package may result in this configuration being broken and the node becoming `NotReady`. If this happens on control plane nodes, the cluster may become unavailable.

Rebooting the node should recreate a working network configuration. Clusters created recently use "transparent" networking mode by default and will not see this problem.

**AKS Engine version**

Versions before v0.58.0, or any version with an API model not using "transparent" Kubernetes networking mode.

**Kubernetes version**

Any.

**Additional context**

The root issue is described here: https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1937117

mboersma:
TODO: add instructions for detecting whether bridge mode is enabled?
TODO: add instructions for disabling unattended upgrades?

mboersma:
Q: How can I tell if I'm using the deprecated "bridge" mode or the preferred "transparent" networking mode?

A: Look at the cluster template (aka "API model") used to deploy your cluster, probably found in a location like `_output/mycluster/apimodel.json`.  The `kubernetesConfig` near the top of the file will have a `networkMode` value:

```yaml
{
  "apiVersion": "vlabs",
  "location": "southeastasia",
  "properties": {
    "orchestratorProfile": {
      "orchestratorType": "Kubernetes",
      "orchestratorRelease": "1.22",
      "orchestratorVersion": "1.22.0-beta.0",
      "kubernetesConfig": {
        "kubernetesImageBase": "mcr.microsoft.com/",
        "kubernetesImageBaseType": "mcr",
        "mcrKubernetesImageBase": "mcr.microsoft.com/",
        "clusterSubnet": "10.239.0.0/16",
[...]

PR #4614, opened 2021-08-11
feat: configurable unattended-upgrades

jackfrancis:
**Reason for Change**:

This PR introduces a boolean `"enableUnattendedUpgrades"` configuration property to the `LinuxProfile` api model configuration object. By default, we set it to true, preserving the existing behavior.

We also log a warning if the user does not explicitly set this configuration, nudging her or him to do so next time.

If the `"enableUnattendedUpgrades"` is set to `false`, then don't enable unattended-upgrades to run regularly in the background.

**Issue Fixed**:

Related to #4595 

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -273,7 +273,11 @@ fi
 {{end}}
 
 {{- /* re-enable unattended upgrades */}}
+{{- if EnableUnattendedUpgrades}}
 rm -f /etc/apt/apt.conf.d/99periodic
+{{else}}
+apt_get_purge unattended-upgrades || exit {{GetCSEErrorCode "ERR_APT_PURGE_FAIL"}}
```
Michael-Sinz, line 279: Why purge the unattended-upgrades?

It would be better to leave the code in case someone wants to manage this themselves as otherwise they will have to get the code some other way or get it and accidentally enable the automatic processes.

The better process is to set the apt configuration to turn off `APT::Periodic::Unattended-Upgrade`

It is set to "1" by default in this distro setup.  Setting that to "0" disables the automatic process but leaves the code there such that a manual process can be executed.

Note that there is also the apt database update which can also be turned off via `APT::Periodic::Update-Package-Lists`

Note that these are the defaults in 18.04:
```
# apt-config dump | grep Periodic
APT::Periodic "";
APT::Periodic::Update-Package-Lists "1";
APT::Periodic::Download-Upgradeable-Packages "0";
APT::Periodic::AutocleanInterval "0";
APT::Periodic::Unattended-Upgrade "1";
```

It would be much better to address it this way than removing the code.

Michael-Sinz (changes requested):

parts/k8s/cloud-init/artifacts/cse_main.sh
Michael-Sinz: PS - the file `/etc/apt/apt.conf.d/20auto-upgrades` is where this configuration lives in the ubuntu distro we use.

See:
```
# grep -r Periodic /etc/apt/apt.conf.d
/etc/apt/apt.conf.d/10periodic:APT::Periodic::Update-Package-Lists "1";
/etc/apt/apt.conf.d/10periodic:APT::Periodic::Download-Upgradeable-Packages "0";
/etc/apt/apt.conf.d/10periodic:APT::Periodic::AutocleanInterval "0";
/etc/apt/apt.conf.d/20auto-upgrades:APT::Periodic::Update-Package-Lists "1";
/etc/apt/apt.conf.d/20auto-upgrades:APT::Periodic::Unattended-Upgrade "1";
```

pkg/api/const.go
```diff
@@ -223,6 +223,10 @@ const (
 	StandardVMType = "standard"
 	// DefaultRunUnattendedUpgradesOnBootstrap sets the default configuration for running a blocking unattended-upgrade on Linux VMs as part of CSE
 	DefaultRunUnattendedUpgradesOnBootstrap = true
+	// DefaultEnableUnattendedUpgrades sets the default configuration for running unattended-upgrade on a regular schedule in the background
+	DefaultEnableUnattendedUpgrades = true
+	// DefaultEnableUnattendedUpgradesAzureStack sets the default configuration for running unattended-upgrade on a regular schedule in the background for Azure Stack Hub
+	DefaultEnableUnattendedUpgradesAzureStack = true
```
jadarsie, line 229: Could you please change this default to `false`?

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: Thanks for the feedback, this makes sense.

Rather than just try to intercept the default apt configuration we expect Ubuntu to pave, should we simply retain the cloud-init-delivered configuration, which runs at a higher priority (we name the file `/etc/apt/apt.conf.d/99periodic`) and disables all automatic background operations? See:

https://github.com/Azure/aks-engine/blob/master/parts/k8s/cloud-init/nodecustomdata.yml#L156

Michael-Sinz: That works - and avoids having to edit existing files.  It also cleanly separates control/option from code.

pkg/api/const.go
Michael-Sinz: I think the question of the default is what is least surprising.  Today, the default is True (in every aks-engine released since the start)

To change the default would be a breaking change - to warn about it first is better and let people actively add the setting to what they want it to be.

jackfrancis: I think it's @jadarsie's call for Azure Stack Hub, though Michael brings up a relevant point about the fact that such a change would be breaking existing behaviors.

jadarsie: I think you are right @Michael-Sinz, better to keep the existing behavior.

devigned (approved):
/lgtm

jadarsie (approved):

mboersma (approved):
/lgtm

PR #4621, opened 2021-08-18
feat: use --seccomp-default in 1.22+

jackfrancis:
**Reason for Change**:

This PR enables the seccomp profile by default for Kubernetes 1.22+. See more information here on why this is a more secure story compared to the existing unconfined default:

- https://kubernetes.io/docs/tutorials/clusters/seccomp/#enable-the-use-of-runtimedefault-as-the-default-seccomp-profile-for-all-workloads

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults-kubelet.go
+++ b/pkg/api/defaults-kubelet.go
@@ -129,6 +129,10 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 		defaultKubeletConfig["--read-only-port"] = "0" // we only have metrics-server v0.3 support in 1.16.0 and above
 	}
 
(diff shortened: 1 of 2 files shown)


mboersma (commented):
This looks good, and the reasoning for the change is solid.

Should we provide some specific documentation about how to disable this? It is an alpha feature and it seems possible enforcing a security profile might break existing workloads.

jackfrancis:
@cpuguy83 did you conclude that this is not actually a net positive change?

cpuguy83:
Yeah, I don't think we should set it just based on k8s version.

PR #4634, opened 2021-08-30
fix: always install GPU drivers during bootstrapping

### LINKED ISSUE #4631: GPU nodes fail with CSE error 86
**Describe the bug**

GPU nodes fail to provision with `"Enable failed: failed to execute command: command terminated with exit status=86"`

**Steps To Reproduce**

``` $ aks-engine deploy -m examples/kubernetes-gpu/kubernetes.json -l westus2 -p mydnsprefix```

**Expected behavior**

Nodes provision without error and expose their GPU resources.

**AKS Engine version**

v0.66.0 and `HEAD` of master branch. This did not seem to be a problem in v0.65.0.

**Kubernetes version**

**Additional context**

Is it possible that https://github.com/Azure/aks-engine/blob/4edaf4d68c28cbcb4c1cd8ad35cddc5d64ff5284/parts/k8s/cloud-init/artifacts/cse_install.sh#L82 failed during image building, or that the package moved?

cc: @Michael-Sinz

jackfrancis:
**Reason for Change**:

This PR restores the previous behavior of always installing GPU drivers during bootstrapping. This has the side-effect of no longer supporting "airgap" scenarios, but it appears that simply `apt-get download`'ing the packages during VHD creation is not sufficient for providing that solution. I am speculating that files in `/var/cache/apt/archives/` are regularly cleaned up based on the age of the files, which means that pre-downloaded apt packages in the VHD are only viable for a temporary period of time.

**Issue Fixed**:

Fixes #4631

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_main.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_main.sh
@@ -150,9 +150,7 @@ time_metric "InstallNetworkPlugin" installNetworkPlugin
 
 {{- if and HasNSeriesSKU IsNvidiaDevicePluginAddonEnabled}}
 if [[ ${GPU_NODE} == true ]]; then
-  if $FULL_INSTALL_REQUIRED; then
-    time_metric "DownloadGPUDrivers" downloadGPUDrivers
-  fi
+  time_metric "DownloadGPUDrivers" downloadGPUDrivers
   time_metric "EnsureGPUDrivers" ensureGPUDrivers
 fi
 {{end}}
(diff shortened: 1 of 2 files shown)


jackfrancis:
cc @Michael-Sinz

mboersma (approved):
/lgtm

CI seems happy with this.

Michael-Sinz:
We should look into what happened here - the files don't magically go away unless some code ran.  The image does not run again until bootstrap is running so the image cache had to have been flushed somehow.

Either that or the actual packages changed somehow.  Given that this image once did work, something changed to make it not work and that change had to be external to the image.

PR #4637, opened 2021-08-31
fix: upgrade will now recreate the correct master

### LINKED ISSUE #4629: Update always tries to recreate first master VM as opposed to creating the missing VM
**Describe the bug**
In a situation where the a master VM is missing (master-2 in a 3 master cluster), running upgrade will correctly point out there is a missing master but instead of creating the master at the missing index (master-2) it creates the master at the first index (master-0) always. This leads to a deployment conflict (Changing property 'dataDisk.createOption' is not allowed.)

**Steps To Reproduce**

1. Deploy a multi-master cluster 
2. Delete one of the auxiliary master VMs
3. Run Upgrade

**Expected behavior**
Auxiliary master at missing index should be recreated

**AKS Engine version**
v0.63

**Kubernetes version**
1.19.10

**Additional context**

deaborch:
**Reason for Change**:

Fix in Upgrade flow to recreate the correct missing master

**Issue Fixed**:

Fixes #4629

**Credit Where Due**:

- [X] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/operations/kubernetesupgrade/upgrader.go
+++ b/pkg/operations/kubernetesupgrade/upgrader.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/rand"
+	"strconv"
 	"strings"
 	"time"
 
@@ -255,6 +256,16 @@ func (ku *Upgrader) upgradeMasterNodes(ctx context.Context) error {
 	}
 
 	upgradedMastersIndex := make(map[int]bool)
+	for i := 0; i < expectedMasterCount; i++ {
+		upgradedMastersIndex[i] = false
+		for _, vm := range *ku.ClusterTopology.MasterVMs {
+			vmName := *vm.Name
+			if strings.Contains(vmName, strconv.Itoa(i)) {
+				upgradedMastersIndex[i] = true
+			}
+		}
+	}
+
 	mastersToCreate := expectedMasterCount - masterNodesInCluster
 	ku.logger.Infof("Expected master count: %d, Creating %d more master VMs", expectedMasterCount, mastersToCreate)
 


pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -255,6 +256,16 @@ func (ku *Upgrader) upgradeMasterNodes(ctx context.Context) error {
 	}
 
 	upgradedMastersIndex := make(map[int]bool)
+	for i := 0; i < expectedMasterCount; i++ {
+		upgradedMastersIndex[i] = false
+		for _, vm := range *ku.ClusterTopology.MasterVMs {
+			vmName := *vm.Name
+			if strings.Contains(vmName, strconv.Itoa(i)) {
```
jadarsie, line 263: I do not think this will work, the `ClusterID` portion of VM name can interfere with the `vmName, strconv.Itoa(i)` comparison. I would avoid comparisons, something around this:

```go
for _, vm := range *ku.ClusterTopology.UpgradedMasterVMs {
  ID := strings.TrimSuffix(ku.ClusterTopology.DataModel.Properties.GetMasterVMPrefix(), *vm.Name)
  upgradedMastersIndex[ID] = true
}
```

pkg/operations/kubernetesupgrade/upgrader.go
```diff
@@ -255,6 +256,17 @@ func (ku *Upgrader) upgradeMasterNodes(ctx context.Context) error {
 	}
 
 	upgradedMastersIndex := make(map[int]bool)
+
+	for _, vm := range *ku.ClusterTopology.MasterVMs {
```
jadarsie, line 260: should this loop over `UpgradedMasterVMs` instead?

deaborch: That array will be empty at that point as none of the masters have been upgraded

jadarsie (approved):

jackfrancis (approved):
/lgtm

jackfrancis:
Thanks for fixing this @deaborch! Confirmed that the current implementation indeed never does anything except to the first master:

https://play.golang.org/p/q5FtBfFgEU0

PR #4650, opened 2021-09-13
fix: use definite --nodeport-addresses in kube-proxy

jackfrancis:
**Reason for Change**:

This PR updates the kube-proxy config so its `--nodeport-addresses` is constrained to the cluster CIDR (10.240.0.0/12 by default) instead of being permissive to all addresses.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/kube-proxy.yaml
+++ b/parts/k8s/addons/kube-proxy.yaml
@@ -79,7 +79,7 @@ spec:
       containers:
       - command:
         - kube-proxy
-        - --config=/var/lib/kube-proxy/config.yaml
+        - --nodeport-addresses="{{ContainerConfig "cluster-cidr"}}"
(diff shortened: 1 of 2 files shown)


parts/k8s/addons/kube-proxy.yaml
```diff
@@ -79,7 +79,7 @@ spec:
       containers:
       - command:
         - kube-proxy
-        - --config=/var/lib/kube-proxy/config.yaml
```
jackfrancis, line 82: Note: the above config reference points to a file that aks-engine doesn't pave to disk, so I'm just replacing it altogether

PR #4659, opened 2021-09-20
fix: Removed dependency on 'ReleaseId' for WS2022

nick5616:
**Reason for Change**:

The behavior of ReleaseId replaced the ws2022 pause image with a 20h2 pause image, which prevented any ws2022 pods from starting. 

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
This is still an issue for any version of windows that does not have build number 20348 but is later than 20H2.


parts/k8s/windowscontainerdfunc.ps1
```diff
@@ -161,12 +162,18 @@ function Install-Containerd {
   $formatedconf = $(($CNIConfDir).Replace("\", "/"))
   $sandboxIsolation = 0
   $windowsReleaseId = (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ReleaseId
+  $windowsCurrentBuild = (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").CurrentBuild
```
jsturtevant, line 165: should we switch all versions to use the `CurrentBuild` id? Does this return the same value as `ReleaseId` for all previous versions of Windows server (2019/2004/etc)?

jsturtevant:
fyi on the `ReleaseId` being no longer supported in WS 2022  @AbelHu @mainred

parts/k8s/windowscontainerdfunc.ps1
nick5616: The code uses `CurrentBuild` for all cases unless it can't. In that case, it relies on `ReleaseId` as a fallback

jsturtevant: looks like it `ReleaseId` returns `1809` and `CurrentBuild` return `17763` for WS2019:

```
PS C:\Users\capi> (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").CurrentBuild 
17763 
PS C:\Users\capi> (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ReleaseId 
1809 
```

nick5616: Yes. I use Select-Windows-Version as a map from `CurrentBuild` to the value `ReleaseId` would return
```
switch ($buildNumber) {
    "17763" { return "1809" }
    "18362" { return "1903" }
    "18363" { return "1909" }
    "19041" { return "2004" }
    "19042" { return "20H2" }
    "20348" { return "ltsc2022" }
    Default { return "" } 
}
```
In your example, 17763 => 1809

parts/k8s/windowscontainerdfunc.ps1
```diff
@@ -99,10 +99,26 @@ function Select-Windows-Version {
...
     Default { return "" } 
   }
 }
 
+# Danger: ReleaseId is deprecated. 
+function Get-WindowsVersion-From-ReleaseId {
+  $windowsReleaseId = (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows NT\CurrentVersion").ReleaseId
+  $windowsVersion = ""
+
+  # Starting with 20H2 tags used to publish container images may not match the 'ReleaseId'
```
jsturtevant, line 112: It won't work at all for WS 2022 is that right?

parts/k8s/windowscontainerdfunc.ps1
jsturtevant: logic looks good here. Let's wrap this into a single function that just returns windows version to hide this complexity from the `Install-Containerd` function

parts/k8s/windowscontainerdfunc.ps1
nick5616: WS2022 is covered by the buildNumber switch. 
`20348 => ltsc2022`
The `ReleaseId` fallback is for anything that the switch doesn't have a case for.

jsturtevant: I think our switch covers all known versions BuildId => image tags.  This will cause strange runtime errors for anything newer than 20H2 as you experienced. It maybe be better to fail explicitly here. wdyt?

nick5616: If we could somehow cause aks-engine to hit an exception here, that would be ideal. It would say "Current Build not recognized" or something like that. This way, nobody will be rooting around in logs after-the-fact

nick5616: Are we testing versions that aren't captured in the switch? Would we have to update the corresponding unit tests?

jsturtevant: unfortunately this is done at cluster bring up time not the cluster template generation.  Even though CSE would fail and some one would have to go get the logs to figure it out, I think that would be ideal over the cluster coming online but having strange behavior when deploying pods and trying to root cause it.  

The reason it needs the version id here instead of just using a multi-arch image is the experimental Hyper-v  support doesn't work with multi-arch images.  You end up in a similar scenario where the cluster is online but pods can't be deployed.

marosset:
@nick5616 did you sign a contributor license agreement?
If so can you try rebasing this PR and pushing again.
We should try and get this merged.

nick5616:
There was some sort of error with the CLA back in September. Rebasing seems to have fixed it :) @marosset

marosset:
@jsturtevant can you review this again?

marosset (approved):

jsturtevant (approved):

PR #4685, opened 2021-10-01
docs: fix addon name in Calico `logSeverityScreen` example

dennis-benzinger-hybris:
I think there's a bug in the example for the Calico `logSeverityScreen`. The addon name should be `calico-daemonset` as defined by `common.CalicoAddonName`.
Only with this name I could convince aks-engine to include my customized log level into the ARM template.


jackfrancis (approved):
/lgtm

PR #4726, opened 2021-10-28
fix: bypass systemd-resolved but use the resolv.conf it generates

jadarsie:
**Reason for Change**:

Name resolution fails on Ubuntu 18.04 hosts because `/etc/resolv.conf` is not pointing to the correct name server.

Borrowing this fix: https://github.com/Azure/AgentBaker/blob/5018e82018f31f3b1fb9a26332f4c706b2d9f011/parts/linux/cloud-init/artifacts/cse_config.sh#L238-L253

**Issue Fixed**:

Related to https://github.com/Azure/AKS/issues/2052

**Credit Where Due**:

- [ ] No
- [x] Yes

- [ ] No
- [x] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -115,13 +115,9 @@ ensureChrony() {
   systemctlEnableAndStart chrony || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
 }
 disable1804SystemdResolved() {
-  ls -ltr /etc/resolv.conf
-  cat /etc/resolv.conf
   {{/* Ignoring systemd-resolved query service but using its resolv.conf file */}}
   {{/* This is the simplest approach to workaround resolved issues without completely uninstall it */}}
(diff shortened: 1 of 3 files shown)


mboersma (dismissed):
/lgtm

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -114,7 +114,15 @@ configureChrony() {
 ensureChrony() {
   systemctlEnableAndStart chrony || exit {{GetCSEErrorCode "ERR_SYSTEMCTL_START_FAIL"}}
 }
-
+disable1804SystemdResolved() {
+  ls -ltr /etc/resolv.conf
```
jackfrancis, line 118: I assume these commands helped during debugging? Do we want to include anything other than the symbolic link statement in the final CSE implementation?

jadarsie: I assume the same, I do not mind either way, will remove.

jackfrancis (approved):
lgtm

PR #4730, opened 2021-11-02
test: look for more log lines during rotate test

jackfrancis:
**Reason for Change**:

This increases the number of log lines to introspect for an older timestamp during log rotation test, in response to lots of test flakes.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


jackfrancis:
cc @jsturtevant

jackfrancis:
@jsturtevant this change didn't eliminate test flakes/failures

Should we continue to run this test? (which would mean we should figure out what's going on here)

PR #4731, opened 2021-11-02
chore: update cluster-autoscaler to 1.22.1

jackfrancis:
**Reason for Change**:

In response to some test flakes, updating the cluster-autoscaler version that we use to test.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


mboersma (approved):
/lgtm

I don't think there's any general or Azure-specific fixes in 1.22.1, but we should try to stay on the latest patch regardless.

PR #4759, opened 2021-11-30
fix: ensure kube-addon-manager + apiserver monitor

jackfrancis:
**Reason for Change**:

This PR re-introduces improvements to kube-addon-manager bootstrapping, and includes additional apiserver resilience in response to observed flakes where the kube-addon-manager pod cannot be introspected due to the apiserver becoming unavailable.

Specifically, we copy the existing kubelet monitor pattern and monitor the built-in apiserver readyz endpoint, and restart the kubelet systemd service if apiserver readyz check fails.

For more information:

https://kubernetes.io/docs/reference/using-api/health-checks/

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -416,6 +416,20 @@ ensureKubelet() {
   {{- end}}
   fi
 {{- end}}
+if [[ -n ${MASTER_NODE} ]]; then
+  wait_for_file 1200 1 /etc/systemd/system/apiserver-monitor.service || exit {{GetCSEErrorCode "ERR_FILE_WATCH_TIMEOUT"}}
+  systemctlEnableAndStart apiserver-monitor || exit {{GetCSEErrorCode "ERR_KUBELET_START_FAIL"}}
```
mboersma, line 421: Should we add an `ERR_K8S_API_SERVER_MONITOR_FAIL` for this? All the other uses I see for this error code are either starting `kubelet` or `kubelet-monitor`.

jackfrancis: Good question. Because (1) kubelet is a catch-all systemd service for ensuring all k8s components come online (including control plane components if it's a control plane node) and (2) we're in the `ensureKubelet` func, I think it makes sense to return this common error. Basically we're just adding to the set of required bootstrap conditions for launching the kubelet systemd service.

mboersma: 👍🏻 Make sense to me.

mboersma (approved):
/lgtm

PR #4781, opened 2021-12-09
chore: Azure NPM update to v1.4.16

vakalapa:
**Reason for Change**:

Updating Azure NPM to v1.4.9

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [x] Yes

- [ ] No
- [x] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/addons/azure-network-policy.yaml
+++ b/parts/k8s/addons/azure-network-policy.yaml
@@ -99,6 +99,8 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: spec.nodeName
+            - name: NPM_CONFIG
+              value: /etc/azure-npm/azure-npm.json
           volumeMounts:
           - name: xtables-lock
             mountPath: /run/xtables.lock
@@ -106,6 +108,8 @@ spec:
             mountPath: /var/log
           - name: protocols
             mountPath: /etc/protocols
+          - name: azure-npm-config
+            mountPath: /etc/azure-npm
       hostNetwork: true
       volumes:
       - name: log
@@ -120,6 +124,9 @@ spec:
         hostPath:
           path: /etc/protocols
           type: File
+      - name: azure-npm-config
+        configMap:
+          name: azure-npm-config
       serviceAccountName: azure-npm
 ---
 apiVersion: v1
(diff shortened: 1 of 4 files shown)


parts/k8s/addons/azure-network-policy.yaml
```diff
@@ -121,3 +121,37 @@ spec:
           path: /etc/protocols
           type: File
       serviceAccountName: azure-npm
+---
+apiVersion: v1
```
jadarsie, line 125: New resources need label `addonmanager.kubernetes.io/mode`

pkg/api/k8s_versions.go
```diff
@@ -28,7 +28,7 @@ const (
 	antreaAgentImageReference                                = antreaControllerImageReference
 	antreaOVSImageReference                                  = antreaControllerImageReference
 	antreaInstallCNIImageReference                           = antreaControllerImageReference
-	azureNPMContainerImageReference                   string = "mcr.microsoft.com/containernetworking/azure-npm:v1.2.2_hotfix"
+	azureNPMContainerImageReference                   string = "mcr.microsoft.com/containernetworking/azure-npm:v1.4.16"
```
vakalapa, line 31: The latest image we have is v1.4.22, lmk if you want me to make the change or you can continue.

jadarsie: You are free to move this forward. Just one thing, I noticed that the current `cni` is v1.4.16:
https://github.com/Azure/aks-engine/blob/master/pkg/api/const.go#L322-L329

I am assuming it is better to keep `cni` and `nmp` in sync, but you probably know better how that works.

If you move to a different `nmp` version than v1.4.16 (using v1.4.22 as an example), then we should double check that
https://github.com/Azure/aks-engine/blob/master/parts/k8s/addons/azure-network-policy.yaml
is capturing changes in 
https://github.com/Azure/azure-container-networking/blob/v1.4.22/npm/azure-npm.yaml

jackfrancis:
@jadarsie is there an Azure Stack Hub reason to integrate this change?

jadarsie:
Nope, you can abandon if that's your preference

jackfrancis:
Given upcoming deprecation, we won't do this additional addon integration.

PR #4784, opened 2021-12-14
chore: update out-of-tree cloudprovider releases

jackfrancis:
**Reason for Change**:

This PR updates the relevant out-of-tree cloud-provider-azure components, including CSI components (which we test in common with out-of-tree). References:

- https://github.com/kubernetes-sigs/cloud-provider-azure/releases
- https://github.com/kubernetes-sigs/azurefile-csi-driver/releases
- https://github.com/kubernetes-sigs/azuredisk-csi-driver/releases

In addition, we will now test the out-of-tree scenarios against PRs in preparation for making that the default configuration going forward.

Finally, I observed that the "holistic" CCM implementation that you get when using the `"useCloudControllerManager": true,` configuration exceeds the Azure-allowed cloud-init data limits. In order to get good tests against this scenario, we need to disable the cluster-autoscaler addon. I will follow up with a future Github Action PR to run cluster-autoscaler tests.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/examples/e2e-tests/kubernetes/release/default/definition-no-vnet.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition-no-vnet.json
@@ -39,7 +39,11 @@
                 ],
                 "availabilityProfile": "VirtualMachineScaleSets",
                 "osDiskCachingType": "ReadOnly",
-                "dataDiskCachingType": "ReadWrite"
+                "dataDiskCachingType": "ReadWrite",
+                "availabilityZones": [
+                    "1",
+                    "2"
+                ]
             },
             {
                 "name": "poolwin",
@@ -47,21 +51,33 @@
                 "vmSize": "Standard_D2s_v3",
                 "OSDiskSizeGB": 256,
                 "availabilityProfile": "VirtualMachineScaleSets",
-                "osType": "Windows"
+                "osType": "Windows",
+                "availabilityZones": [
+                    "1",
+                    "2"
+                ]
             },
             {
                 "name": "pool1804",
                 "count": 1,
                 "vmSize": "Standard_D2_v3",
                 "distro": "ubuntu-18.04",
-                "availabilityProfile": "VirtualMachineScaleSets"
+                "availabilityProfile": "VirtualMachineScaleSets",
+                "availabilityZones": [
+                    "1",
+                    "2"
+                ]
             },
             {
                 "name": "pool1804gen2",
                 "count": 1,
                 "vmSize": "Standard_D2s_v3",
                 "distro": "ubuntu-18.04-gen2",
-                "availabilityProfile": "VirtualMachineScaleSets"
+                "availabilityProfile": "VirtualMachineScaleSets",
+                "availabilityZones": [
+                    "1",
+                    "2"
+                ]
             }
         ],
         "linuxProfile": {
--- a/examples/e2e-tests/kubernetes/release/default/definition-no-vnet.json
+++ b/examples/e2e-tests/kubernetes/release/default/definition-no-vnet.json
@@ -7,24 +7,6 @@
                 "useManagedIdentity": true,
                 "addons": [
                     {
-                        "name": "cluster-autoscaler",
-                        "enabled": true,
-                        "pools": [
-                            {
-                                "name": "poollinux",
-                                "config": {
-                                    "min-nodes": "1",
-                                    "max-nodes": "3"
-                                }
-                            }
-                        ],
-                        "config": {
-                            "scan-interval": "1m",
-                            "scale-down-delay-after-add": "1m0s",
-                            "scale-down-unneeded-time": "2m0s"
-                        }
-                    },
-                    {
                         "name": "coredns",
                         "enabled": true,
                         "config": {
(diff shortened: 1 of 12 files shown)


chewong (approved):

pkg/api/k8s_versions.go
```diff
@@ -832,8 +832,8 @@ func getK8sVersionComponents(version, kubernetesImageBaseType string, overrides
 			common.ControllerManagerComponentName:             getDefaultImage(common.ControllerManagerComponentName, kubernetesImageBaseType) + ":v" + version,
 			common.KubeProxyAddonName:                         getDefaultImage(common.KubeProxyAddonName, kubernetesImageBaseType) + ":v" + version,
 			common.SchedulerComponentName:                     getDefaultImage(common.SchedulerComponentName, kubernetesImageBaseType) + ":v" + version,
-			common.CloudControllerManagerComponentName:        azureCloudControllerManagerImageReference,
-			common.CloudNodeManagerAddonName:                  azureCloudNodeManagerImageReference,
+			common.CloudControllerManagerComponentName:        "oss/kubernetes/azure-cloud-controller-manager:v0.6.0",
+			common.CloudNodeManagerAddonName:                  "oss/kubernetes/azure-cloud-node-manager:v0.6.0",
```
chewong, line 836: Should we apply the same change to 1.18 where it uses azure-cloud-controller-manager v0.5.1 according to https://github.com/kubernetes-sigs/cloud-provider-azure#current-status?

pkg/engine/engine.go
```diff
@@ -595,14 +595,29 @@ func getAddonFuncMap(addon api.KubernetesAddon, cs *api.ContainerService) templa
...
 			if len(cs.Properties.AgentPoolProfiles) == 0 {
 				return ""
 			}
 
 			var zones string
-			for _, zone := range cs.Properties.AgentPoolProfiles[0].AvailabilityZones {
-				zones += fmt.Sprintf("\n    - %s-%s", cs.Location, zone)
+			for _, pool := range cs.Properties.AgentPoolProfiles {
+				if pool.AvailabilityZones != nil {
+					for _, zone := range cs.Properties.AgentPoolProfiles[0].AvailabilityZones {
```
chewong, line 614:
```suggestion
					for _, zone := range pool.AvailabilityZones {
```

chewong: Also, `zone` will have duplicated entries if there is more than 1 agent pool with the same availability zone. Is that the intention here?

parts/k8s/addons/azure-cloud-provider.yaml
```diff
@@ -102,7 +102,7 @@ volumeBindingMode: WaitForFirstConsumer
 allowedTopologies:
 - matchLabelExpressions:
   - key: topology.disk.csi.azure.com/zone
-    values: {{GetZones}}
+    values: {{GetAgentPoolZones}}
```
chewong, line 105: We should use above `{{ - if HasAgentPoolAvailabilityZones }}` right?

jackfrancis: yep

pkg/api/k8s_versions.go
jackfrancis: 1.18 stuff is left as a historical artifact: you can't actually build a 1.18 using current aks-engine code.

pkg/engine/engine.go
jackfrancis: It's not the intention. The if zones != "" then return zones is supposed to short-circuit the func so that the first pool w/ zones will return those as the value

chewong: Ah, I see that now.

parts/k8s/addons/azure-cloud-provider.yaml
jackfrancis: fixed

pkg/engine/engine.go
jackfrancis: fixed the static reference to the first pool, thanks!

examples/e2e-tests/kubernetes/release/default/definition-no-vnet.json
```diff
@@ -3,26 +3,9 @@
     "properties": {
         "orchestratorProfile": {
             "kubernetesConfig": {
+                "useCloudControllerManager": true,
                 "useManagedIdentity": true,
                 "addons": [
-                    {
```
jackfrancis, line 8: Removing cluster-autoscaler addon because it + CCM exceeds cloud-init data.

pkg/engine/engine_test.go
```diff
@@ -2694,6 +2699,16 @@ func TestGetAddonFuncMap(t *testing.T) {
 							},
 							OSType: api.Linux,
 						},
+						{
```
jackfrancis, line 2702: Added this 2nd agent pool to prove that the `GetAgentPoolZones` go template func will always return a yaml array string representation of the *1st* pool that has a zones configuration. I think the way that csi disk drivers is configured already suggests such mixed-zoned cluster configurations are going to encounter edge case failures, but choosing the first pool is what we've done historically, so this follows that pattern.

chewong (approved):

PR #4791, opened 2021-12-17
fix: Azure CNI IP address allocation for win nodes

jackfrancis:
**Reason for Change**:

This PR addresses an edge case from #4689: Windows nodes do not run hostNetwork pods from any of the available addons, and so those nodes may pre-allocate IPs according to the complete max-pods configuration for a ~accurate estimate of always having enough IP addresses as there will be pods running.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/defaults.go
+++ b/pkg/api/defaults.go
@@ -628,7 +628,11 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				// Allocate IP addresses for pods if VNET integration is enabled.
 				if cs.Properties.OrchestratorProfile.IsAzureCNI() {
 					agentPoolMaxPods, _ := strconv.Atoi(profile.KubernetesConfig.KubeletConfig["--max-pods"])
-					profile.IPAddressCount = getPodIPAddressCountForAzureCNI(agentPoolMaxPods, o.KubernetesConfig)
+					if profile.IsWindows() {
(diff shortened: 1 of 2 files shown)


pkg/api/defaults.go
```diff
@@ -628,7 +628,11 @@ func (cs *ContainerService) setOrchestratorDefaults(isUpgrade, isScale bool) {
 				// Allocate IP addresses for pods if VNET integration is enabled.
 				if cs.Properties.OrchestratorProfile.IsAzureCNI() {
 					agentPoolMaxPods, _ := strconv.Atoi(profile.KubernetesConfig.KubeletConfig["--max-pods"])
-					profile.IPAddressCount = getPodIPAddressCountForAzureCNI(agentPoolMaxPods, o.KubernetesConfig)
+					if profile.IsWindows() {
```
paulgmiller, line 631: could keep this logic in getPodIPAddressCountForAzureCNI but meh.

paulgmiller (approved):
someday windows will be normal

jsturtevant:
With Hostprocess containers, In the future there could possibly be add-ons that have hostnetwork but at this time there are not any.  If they are ever developed and added to AKS-e we should re-evaluate how to count this properly for Windows.

/lgtm

PR #4793, opened 2021-12-21
fix: Set enableLoopbackDSR in Windows Azure CNI configuration for WinDSR

AbelHu:
**Reason for Change**:

Set enableLoopbackDSR in Windows Azure CNI configuration for WinDSR

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:
@jsur @marosset `make build` and `make generate` do not generate templates_generated.go as expected so I copied the change from windowsazurecnifunc.ps1 to templates_generated.go.


pkg/engine/templates_generated.go
```diff
@@ -22898,6 +22898,14 @@ Set-AzureCNIConfig
     if ($IsAzureStack) {
         Add-Member -InputObject $configJson.plugins[0].ipam -MemberType NoteProperty -Name "environment" -Value "mas"
     }
+
```
marosset, line 22901: I was able to run `make generate` from inside the dev container (by running `make dev`).
It looks like there is some extra whitespace on line 22909 that is causing the CI failures.

AbelHu: Thanks, @marosset. `make generate` after running `make dev` works. Updated the PR.

marosset (approved):

marosset:
I think we do want this change.
@jsturtevant too

jsturtevant:
I don't think it will hurt, but I don't know of any setups that we support that would require it. My understand is this would be required if Calico was enabled as a Network policy provider for azure cni.  I don't know we have any documentation of configurations that demonstrate how to configure this.

AbelHu:
Since aks-engine supports WinDSR in https://github.com/Azure/aks-engine/blob/47c439d1e8891a1cc63ced39332418f1506e6a41/docs/topics/windows-dsr.md, I think that this PR needs to be merged. Otherwise, users may hit this issue https://github.com/Azure/AKS/issues/2608
NOTE:  enableLoopbackDSR is only supported from v1.4.16.

jsturtevant (approved):

PR #4797, opened 2022-01-04
feat: only download nvidia drivers if needed

jackfrancis:
**Reason for Change**:

This PR updates the nvidia pre-download implementation to copy downloaded .deb files to a directory outside of the apt cache repo directory, to prevent apt from deleting it over time as it maintains the data cache.

This will solve the "airgap" GPU solution, and will also protect against any unexpected environmental changes (e.g., if the pre-downloaded files *do* disappear we will re-download the files at runtime).

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -623,9 +623,9 @@ configGPUDrivers() {
   retrycmd_no_stats 120 5 25 update-initramfs -u || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   wait_for_apt_locks
   for apt_package in $NVIDIA_PACKAGES; do
-    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
+    dpkg -i ${PERMANENT_CACHE_DIR}${apt_package}* || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   done
-  dpkg -i $(ls ${PERMANENT_CACHE_DIR}nvidia-container-runtime*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
+  dpkg -i ${PERMANENT_CACHE_DIR}nvidia-container-runtime* || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
   retrycmd 120 5 25 mount -t overlay -o lowerdir=/usr/lib/x86_64-linux-gnu,upperdir=${GPU_DEST}/lib64,workdir=${GPU_DEST}/overlay-workdir none /usr/lib/x86_64-linux-gnu || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   export -f installNvidiaDrivers
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -637,6 +637,7 @@ configGPUDrivers() {
   retrycmd 120 5 25 nvidia-modprobe -u -c0 || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   retrycmd 120 5 25 nvidia-smi || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
   retrycmd 120 5 25 ldconfig || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_START_FAIL"}}
+  rm -Rf ${PERMANENT_CACHE_DIR}
 }
 ensureGPUDrivers() {
   configGPUDrivers
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -624,8 +624,6 @@ configGPUDrivers() {
   wait_for_apt_locks
   for apt_package in $NVIDIA_PACKAGES; do
     dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   done
   dpkg -i $(ls ${PERMANENT_CACHE_DIR}nvidia-container-runtime*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   mkdir -p $GPU_DEST/lib64 $GPU_DEST/overlay-workdir
--- a/parts/k8s/cloud-init/artifacts/cse_config.sh
+++ b/parts/k8s/cloud-init/artifacts/cse_config.sh
@@ -622,10 +622,12 @@ configGPUDrivers() {
   echo blacklist nouveau >>/etc/modprobe.d/blacklist.conf
   retrycmd_no_stats 120 5 25 update-initramfs -u || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   wait_for_apt_locks
-  dpkg -i $(ls ${APT_CACHE_DIR}libnvidia-container1*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  dpkg -i $(ls ${APT_CACHE_DIR}libnvidia-container-tools*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  dpkg -i $(ls ${APT_CACHE_DIR}nvidia-container-toolkit*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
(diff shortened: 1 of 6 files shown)


jackfrancis:
cc @Michael-Sinz

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -622,10 +622,12 @@ configGPUDrivers() {
...
   retrycmd_no_stats 120 5 25 update-initramfs -u || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
   wait_for_apt_locks
-  dpkg -i $(ls ${APT_CACHE_DIR}libnvidia-container1*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  dpkg -i $(ls ${APT_CACHE_DIR}libnvidia-container-tools*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  dpkg -i $(ls ${APT_CACHE_DIR}nvidia-container-toolkit*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
-  dpkg -i $(ls ${APT_CACHE_DIR}nvidia-container-runtime*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
+  for apt_package in $NVIDIA_PACKAGES; do
+    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
+    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
+    dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
```
Michael-Sinz, line 628: Why the 3 dpkg -i of the same operation but failing after any one fails?

Why use `dpkg -i $(ls ${PERMANENT_CACHE_DIR}${apt_package}*)` when `dpkg -i ${PERMANENT_CACHE_DIR}${apt_package}*` would work equally the same and not cause problems with files that may have spaces in the name (not that there should be any of those, but...)

parts/k8s/cloud-init/artifacts/cse_config.sh
```diff
@@ -622,10 +622,12 @@ configGPUDrivers() {
...
+  done
+  dpkg -i $(ls ${PERMANENT_CACHE_DIR}nvidia-container-runtime*) || exit {{GetCSEErrorCode "ERR_GPU_DRIVERS_CONFIG"}}
```
Michael-Sinz, line 630: I note here that you don't do 3 tries of this (which is likely just fine anyway since the above 3 tries don't seem to make sense)

Again, don't understand the use of `$(ls ...)` here.

parts/k8s/cloud-init/artifacts/cse_config.sh
Michael-Sinz: Is this due to some retries to address ordering?  If so, why would that work if you exit after any error?

parts/k8s/cloud-init/artifacts/cse_main.sh
```diff
@@ -154,7 +154,7 @@ time_metric "InstallNetworkPlugin" installNetworkPlugin
 
 {{- if and HasNSeriesSKU IsNvidiaDevicePluginAddonEnabled}}
 if [[ ${GPU_NODE} == true ]]; then
-  time_metric "DownloadGPUDrivers" downloadGPUDrivers
+  gpuDriversDownloaded || time_metric "DownloadGPUDrivers" downloadGPUDrivers
```
Michael-Sinz, line 157: after a successful install, should the permanent cache be deleted to free up space?  (Not significant but it is space that could otherwise be used for container caches)

This is non-blocking question - we can ignore that

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -79,7 +88,12 @@ downloadGPUDrivers() {
   apt_get_update
   retrycmd 30 5 60 curl -fLS https://us.download.nvidia.com/tesla/$GPU_DV/NVIDIA-Linux-x86_64-${GPU_DV}.run -o ${GPU_DEST}/nvidia-drivers-${GPU_DV} || exit 85
   tmpDir=$GPU_DEST/tmp
-  apt_get_download 20 30 libnvidia-container1=1.6.0* libnvidia-container-tools=1.6.0* nvidia-container-toolkit=1.6.0* nvidia-container-runtime=3.6.0* || exit 85
+  for apt_package in $NVIDIA_PACKAGES; do
+    apt_get_download 20 30 "${apt_package}=${NVIDIA_CONTAINER_TOOLKIT_VER}*" || exit 85
+    cp ${APT_CACHE_DIR}${apt_package}_${NVIDIA_CONTAINER_TOOLKIT_VER}* $PERMANENT_CACHE_DIR || exit 85
```
Michael-Sinz, line 93: If this were to use `cp -al` it would copy it to the permanent cache directory but via a hard link such that you don't double the space used.
(Assuming these two are on the same filesystem volume, which today they are.)  It also makes sure that they have the same permissions as the original.

parts/k8s/cloud-init/artifacts/cse_install.sh
```diff
@@ -79,7 +88,12 @@ downloadGPUDrivers() {
...
+  done
+  apt_get_download 20 30 nvidia-container-runtime=${NVIDIA_RUNTIME_VER}* || exit 85
+  cp ${APT_CACHE_DIR}nvidia-container-runtime_${NVIDIA_RUNTIME_VER}* $PERMANENT_CACHE_DIR || exit 85
```
Michael-Sinz, line 96: Same here:  Would recommend `cp -al`

parts/k8s/cloud-init/artifacts/cse_helpers.sh
```diff
@@ -23,6 +23,7 @@ if ! [[ $(echo -n "$PRIVATE_IP" | grep -c '^') == 1 ]]; then
 fi
 export PRIVATE_IP
 APT_CACHE_DIR=/var/cache/apt/archives/
+PERMANENT_CACHE_DIR=/root/
```
Michael-Sinz, line 26: Is there a reason this is in `/root/` directly and not, say, in  `/root/aptcache/` or something like that.

I would think you would want the cache directory to be unique (that is, it only contains cache data) and not be conflated with potentially other information.

PS - `/root/` is the root user home directory

Michael-Sinz (commented):
Mostly some questions and comments about the implementation.

We can save some disk space and complexity in the setup.

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: It's not the same operation, `$NVIDIA_PACKAGES` is a "space-delimited string enumerator" with the value:

`"libnvidia-container1 libnvidia-container-tools nvidia-container-toolkit"`

All of those libs are required, so if any one of them fails to install we want to fail.

parts/k8s/cloud-init/artifacts/cse_config.sh
jackfrancis: That's because we are installing the `nvidia-container-runtime` package here, and not enumerating over an "array of packages".

The reason we do the "array of 3 packages" plus "one single package" is because the former array of packages all use a common release, and an enumeration allows us to enforce that version consistency.

The `ls` thing is actually a glob in order to produce the literal package filename that's on disk as input to `dpkg -i`.

parts/k8s/cloud-init/artifacts/cse_config.sh
Michael-Sinz: You have a loop to do all of them and then you do each one 3 times?
Note how lines 626, 627, and 628 are all the same - so you are doing the same thing 3 times, not 3 different packages.

parts/k8s/cloud-init/artifacts/cse_helpers.sh
jackfrancis: We can do that, `/root`/ is kind of a proof of concept. I wanted to get this to a place where it functionally works (it does).

parts/k8s/cloud-init/artifacts/cse_install.sh
jackfrancis: But we need to copy it because we need to not depend upon the files in the apt cache directory being cleaned up.

jackfrancis: nevermind, hard link, understood

parts/k8s/cloud-init/artifacts/cse_config.sh
Michael-Sinz: The ls and the glob produces the exact same results - seriously.  The only difference is that the ls does it as a string that is then parsed by the shell into argv and the glob does it in the shell and then passed as argv directly.

jackfrancis: You're saying that using `dpkg -i ${PERMANENT_CACHE_DIR}nvidia-container-runtime*` will similarly be processed via the glob interpreter to yield a filename (or list of filenames) as the input to dpkg?...

Michael-Sinz: Yes

parts/k8s/cloud-init/artifacts/cse_helpers.sh
jackfrancis: addressed

parts/k8s/cloud-init/artifacts/cse_install.sh
jackfrancis: done

parts/k8s/cloud-init/artifacts/cse_main.sh
jackfrancis: done

Michael-Sinz (approved):
/lgtm

PR #4813, opened 2022-01-19
feat: Add 1.24 support

marosset:

**Reason for Change**:

Some tests in kubernetes/kubernetes which rely on aks-engine to provision clusters have started to fail due to the removal of some kubelet flags. This PR adds support for aks-engine to build 1.24 clusters and removes the deprecated kubelet flags.

Note: There has not been any builds of kubernetes v1.24 yet. The versions referenced in this PR are just a placeholder so we can build clusters from kubernetes/kubernetes at HEAD

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/defaults-kubelet.go
```diff
@@ -315,6 +315,13 @@ func removeKubeletFlags(k map[string]string, v string) {
 			delete(k, key)
 		}
 	}
+
+	// Remove dockershim related flags in v1.24 and up
+	if common.IsKubernetesVersionGe(v, "1.24.0-alpha") {
+		for _, key := range []string{"--image-pull-progress-deadline", "--network-plugin"} {
+			delete(k, key)
```
jsturtevant, line 322: was going to ask if these should these removed or added back in if less than 1.24, but from above it looks like the pattern is to remove the flags

jsturtevant (approved):

pkg/api/defaults-kubelet.go
marosset: Yea, I had the same question and was following previous examples.

jackfrancis (approved):
lgtm

PR #4849, opened 2022-02-28
chore: force external cloud provider for Kubernetes v1.21+ on Azure Stack Hub

haofan-ms:
**Reason for Change**:

Start Kubernetes 1.21, only external cloud provider is supported on Azure Stack cloud. Adding the logic to enforce that for deploy and upgrade operations.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -926,6 +926,18 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		}
 	}
 
+	// Ensure cloud-node-manager is enabled on appropriate upgrades for Azure Stack cloud
+	if isUpgrade &&
+		cs.Properties.IsAzureStackCloud() &&
+		to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) &&
+		common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.21.0") {
+		// Force enabling cloud-node-manager addon
+		log.Infoln("Updating cloud-node-manager addon to 'true' on Azure Stack cloud...\n")
+		if i := getAddonsIndexByName(o.KubernetesConfig.Addons, common.CloudNodeManagerAddonName); i > -1 {
+			o.KubernetesConfig.Addons[i] = defaultCloudNodeManagerAddonsConfig
+		}
+	}
+
 	// Back-compat for older addon specs of cluster-autoscaler
 	if isUpgrade {
 		i := getAddonsIndexByName(o.KubernetesConfig.Addons, common.ClusterAutoscalerAddonName)
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -929,10 +929,8 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 	// Ensure cloud-node-manager is enabled on appropriate upgrades for Azure Stack cloud
 	if isUpgrade &&
 		cs.Properties.IsAzureStackCloud() &&
-		to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) &&
-		common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.21.0") {
+		to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) {
 		// Force enabling cloud-node-manager addon
-		log.Infoln("Updating cloud-node-manager addon to 'true' on Azure Stack cloud...")
 		if i := getAddonsIndexByName(o.KubernetesConfig.Addons, common.CloudNodeManagerAddonName); i > -1 {
 			o.KubernetesConfig.Addons[i] = defaultCloudNodeManagerAddonsConfig
 		}
--- a/pkg/api/addons.go
+++ b/pkg/api/addons.go
@@ -932,7 +932,7 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) &&
 		common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.21.0") {
 		// Force enabling cloud-node-manager addon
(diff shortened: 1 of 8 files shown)


pkg/api/addons.go
```diff
@@ -926,6 +926,18 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
 		}
 	}
 
+	// Ensure cloud-node-manager is enabled on appropriate upgrades for Azure Stack cloud
+	if isUpgrade &&
```
jadarsie, line 930: do we need to check for `to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) && common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.21.0"`?

The UseCCM flag is overridden already and supported version list starts on 1.21

jadarsie: > The UseCCM flag is overridden already

I am assuming that `upgrade.go` is executed first

pkg/api/addons.go
```diff
@@ -926,6 +926,18 @@ func (cs *ContainerService) setAddonsConfig(isUpgrade bool) {
...
+		cs.Properties.IsAzureStackCloud() &&
+		to.Bool(cs.Properties.OrchestratorProfile.KubernetesConfig.UseCloudControllerManager) &&
+		common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.21.0") {
+		// Force enabling cloud-node-manager addon
+		log.Infoln("Updating cloud-node-manager addon to 'true' on Azure Stack cloud...")
```
jadarsie, line 935: I am not sure this message is adding relevant information given that `upgrade.go` is already logging something similar.

If you see extra value, then I would use the `debug` level instead of `info`

pkg/api/components.go
```diff
@@ -147,6 +149,18 @@ func (cs *ContainerService) setComponentsConfig(isUpgrade bool) {
 		}
 	}
 
+	// Ensure cloud-controller-manager is enabled on appropriate upgrades for Azure Stack cloud
```
jadarsie, line 152: Same comments as `addons.go`

pkg/api/vlabs/validate.go
```diff
@@ -300,6 +300,10 @@ func (a *Properties) ValidateOrchestratorProfile(isUpdate bool) error {
 			}
 
 			if a.IsAzureStackCloud() {
+				if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.21.0") && !to.Bool(o.KubernetesConfig.UseCloudControllerManager) {
+					return errors.New("useCloudControllerManager should be set to true for Kubernetes 1.21+ cluster on Azure Stack")
```
jadarsie, line 304: `v1.21+` instead of `1.21+`

Could you please update the messages so they ready `Azure Stack Hub` instead of `Azure Stack`?

pkg/api/addons.go
haofan-ms: Yes, upgrade.go executed first, I will remove the check for k8s version

pkg/api/vlabs/validate.go
haofan-ms: Updated

pkg/api/components.go
haofan-ms: Updated

pkg/api/addons.go
haofan-ms: unnecessary log removed

jadarsie (dismissed):

pkg/api/defaults-kubelet.go
```diff
@@ -77,6 +77,9 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 	staticWindowsKubeletConfig["--image-pull-progress-deadline"] = "20m"
 	staticWindowsKubeletConfig["--resolv-conf"] = "\"\"\"\""
 	staticWindowsKubeletConfig["--eviction-hard"] = "\"\"\"\""
+	if to.Bool(o.KubernetesConfig.UseCloudControllerManager) {
```
jackfrancis, line 80: I don't think this is necessary. Windows nodes are never used as control plane nodes, and worker node kubelet runtimes don't have a functional "external cloud-provider" context.

pkg/api/defaults-kubelet.go
```diff
@@ -191,6 +194,10 @@ func (cs *ContainerService) setKubeletConfig(isUpgrade bool) {
 			// if upgrade, force default "--pod-infra-container-image" value
 			cs.Properties.MasterProfile.KubernetesConfig.KubeletConfig["--pod-infra-container-image"] = o.KubernetesConfig.KubeletConfig["--pod-infra-container-image"]
 		}
+		//Ensure cloud-provider setting
```
jackfrancis, line 197: We don't need this, this is already being taken care of in L152-L154

pkg/api/vlabs/validate.go
```diff
@@ -300,6 +300,10 @@ func (a *Properties) ValidateOrchestratorProfile(isUpdate bool) error {
 			}
 
 			if a.IsAzureStackCloud() {
+				if common.IsKubernetesVersionGe(a.OrchestratorProfile.OrchestratorVersion, "1.21.0") && !to.Bool(o.KubernetesConfig.UseCloudControllerManager) {
```
jackfrancis, line 303: This validation happens *before* defaults enforcement, so this is going to throw an error every time a user tries to create a cluster with an empty `UseCloudControllerManager` configuration

pkg/api/defaults-kubelet.go
haofan-ms: This is to ensure the KubeletConfig in MasterProfile get updated when we upgrade from a in-tree cloud-provider cluster to a external cloud-provider. Without this line, the previous value for this field will not get overwritten.

jackfrancis: O.K., thanks for clarifying

jackfrancis (approved):
/lgtm

PR #4866, opened 2022-03-28
fix: correctly handle multiple root certificates on Azure Stack Hub windows nodes

haofan-ms:
**Reason for Change**:

Update external cloud provider logic for windows node to retrieve root certificate from SSL certificate of ARM endpoint

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -462,7 +462,7 @@ try
             $azsARMUri = [System.Uri]$azsJson.resourceManagerEndpoint
             $webRequest = [Net.WebRequest]::Create($azsARMUri.AbsoluteUri)
             try { $webRequest.GetResponse() } catch {}
-            if ($null -eq $webRequest.ServicePoint.Certificate) {
+            if (($null -eq $webRequest.ServicePoint) -Or ($null -eq $webRequest.ServicePoint.Certificate)) {
                 throw "SSL Certificate of ARM endpoint is null"
             }
             $sslCert = $webRequest.ServicePoint.Certificate
--- a/parts/k8s/kuberneteswindowssetup.ps1
+++ b/parts/k8s/kuberneteswindowssetup.ps1
@@ -450,20 +450,33 @@ try
 
         {{if IsAzureStackCloud}}
             {{if UseCloudControllerManager}}
-            # Export the Azure Stack root cert for use in cloud node manager container setup.
+            # Retrieve SSL cert of ARM Endpoint and find unique Azure Stack root cert
             $azsConfigFile = [io.path]::Combine($global:KubeDir, "azurestackcloud.json")
             if (-not (Test-Path -Path $azsConfigFile)) {
-                throw "$azsConfigFile does not exist, cannot export Azure Stack root cert"
+                throw "$azsConfigFile does not exist"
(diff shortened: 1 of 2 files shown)


jadarsie (approved):

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -450,20 +450,33 @@ try
...
+            try { $webRequest.GetResponse() } catch {}
+            if (($null -eq $webRequest.ServicePoint) -Or ($null -eq $webRequest.ServicePoint.Certificate)) {
+                throw "SSL Certificate of ARM endpoint is null"
             }
-            $azsARMUri = [System.Uri]$azsJson.managementPortalURL
-            $azsRootCert = Get-ChildItem -Path Cert:\LocalMachine\Root | Where-Object {$_.DnsNameList.Unicode -contains $azsARMUri.Host.Substring($azsARMUri.Host.IndexOf(".")).TrimStart(".")}
+            $sslCert = $webRequest.ServicePoint.Certificate
+            $sslCertChain = New-Object -TypeName System.Security.Cryptography.X509Certificates.X509Chain
+            $sslCertChain.build($sslCert)
+            $sslRootCert = @($sslCertChain.ChainElements.Certificate)[-1]
```
jsturtevant, line 471: Will the this always be the right ssl cert?

parts/k8s/kuberneteswindowssetup.ps1
```diff
@@ -450,20 +450,33 @@ try
...
             }
             $azsJson = Get-Content -Raw -Path $azsConfigFile | ConvertFrom-Json
-            if ([string]::IsNullOrEmpty($azsJson.managementPortalURL)) {
-                throw "managementPortalURL is empty, cannot get Azure Stack ARM uri"
+            if ([string]::IsNullOrEmpty($azsJson.resourceManagerEndpoint)) {
+                throw "resourceManagerEndpoint is empty, cannot get Azure Stack ARM uri"
+            }
+            $azsARMUri = [System.Uri]$azsJson.resourceManagerEndpoint
+            $webRequest = [Net.WebRequest]::Create($azsARMUri.AbsoluteUri)
+            try { $webRequest.GetResponse() } catch {}
```
jsturtevant, line 464: I guess the response doesn't matter here.  All that matters is that the Certificate is retrieved which is checked on the next line so if there was no connectivity that would be null.

jsturtevant (commented):
I don't have a ton of context here but lgtm.  Left a few non-blocking comments

parts/k8s/kuberneteswindowssetup.ps1
haofan-ms: Yes, the root cert will be at the bottom of the cert chain

parts/k8s/kuberneteswindowssetup.ps1
haofan-ms: That's correct, the purpose of web request is to get the ssl cert & cert chain of the ARM endpoint

jsturtevant (approved):

PR #4873, opened 2022-04-11
chore: updating windows vhds for v0.70.0

marosset:

**Reason for Change**:

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/pkg/api/azenvtypes.go
+++ b/pkg/api/azenvtypes.go
@@ -138,18 +138,18 @@ var (
 	// AKSWindowsServer2019OSImageConfig is the aks-engine image based on Windows Server 2019
 	AKSWindowsServer2019OSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks-windows",
-		ImageSku:       "2019-datacenter-core-smalldisk-2203",
+		ImageSku:       "2019-datacenter-core-smalldisk-2204",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.2565.220301",
+		ImageVersion:   "17763.2565.220408",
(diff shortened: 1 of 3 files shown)


pkg/api/azenvtypes.go
```diff
@@ -138,18 +138,18 @@ var (
...
 
 	// AKSWindowsServer2019ContainerDOSImageConfig is the aks-engine image based on Windows Server 2019
 	// configured with containerd
 	AKSWindowsServer2019ContainerDOSImageConfig = AzureOSImageConfig{
 		ImageOffer:     "aks-windows",
-		ImageSku:       "2019-datacenter-core-ctrd-2203",
+		ImageSku:       "2019-datacenter-core-ctrd-2204",
 		ImagePublisher: "microsoft-aks",
-		ImageVersion:   "17763.2565.220301",
+		ImageVersion:   "17763.2565.220407",
```
jackfrancis, line 152: I assume it's intentional that these two images have a different image version? (built on subsequent days?)

marosset: Kind of... the CI job which added the images to the marketplace office ran on different days for the two images and that timestamp is generated by the script that does the publishing.

marosset: so not intentional but by design?

jackfrancis: All good, thanks for clarifying. Programmatic naming is good. :)

PR #4888, opened 2022-04-29
fix: opt out of 'Anitimalware Signatures SDP' extension for Windows VHDs

marosset:

**Reason for Change**:

We build the Windows VHD in "internal" Azure subs which are subject to some compliance policies.
One such policy is to change how Azure VMs receive defender signature updates.
We need to opt of out this for the VMs used to produce our aks-engine VHDs because it breaks customer scenarios.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/vhd/packer/windows-vhd-builder.json
+++ b/vhd/packer/windows-vhd-builder.json
@@ -44,7 +44,8 @@
             "azure_tags": {
                 "os": "Windows",
                 "now": "{{user `create_time`}}",
-                "createdBy": "aks-engine-vhd-pipeline"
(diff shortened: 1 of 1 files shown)


jackfrancis (approved):
/lgtm

jsturtevant (approved):

jsturtevant:
Re-triggered the vhd job, looked like a flake when curl'ing the artifacts to the vhd

PR #4890, opened 2022-05-06
feat: add support for Kubernetes v1.24.0

mboersma:
**Reason for Change**:

See https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.24.md#changelog-since-v1230

**Issue Fixed**:

N/A

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


pkg/api/defaults-controller-manager.go
```diff
@@ -90,6 +90,11 @@ func (cs *ContainerService) setControllerManagerConfig() {
 	// Enable the consumption of local ephemeral storage and also the sizeLimit property of an emptyDir volume.
 	addDefaultFeatureGates(o.KubernetesConfig.ControllerManagerConfig, o.OrchestratorVersion, "1.10.0", "LocalStorageCapacityIsolation=true")
 
+	// Disable something something for v1.24.0
```
mboersma, line 93: Oops. I'll update this comment.

jackfrancis (approved):
/lgtm

pkg/api/defaults-controller-manager.go
```diff
@@ -90,6 +90,9 @@ func (cs *ContainerService) setControllerManagerConfig() {
 	// Enable the consumption of local ephemeral storage and also the sizeLimit property of an emptyDir volume.
 	addDefaultFeatureGates(o.KubernetesConfig.ControllerManagerConfig, o.OrchestratorVersion, "1.10.0", "LocalStorageCapacityIsolation=true")
 
+	// Enable legacy service account token autogeneration for v1.24.0
+	addDefaultFeatureGates(o.KubernetesConfig.ControllerManagerConfig, o.OrchestratorVersion, "1.24.0", "LegacyServiceAccountTokenNoAutoGeneration=false")
```
liggitt, line 94: is there an issue tracking removing the need to re-enable this?

mboersma: There isn't an issue, because AKS Engine is [deprecated](https://github.com/Azure/aks-engine#project-status) and won't support Kubernetes versions beyond 1.24. It's a workaround reflecting the reality that our engineering resources are focused on Cluster API for the future.

liggitt: Thanks for clarifying. If there were dependents that couldn't use TokenRequest or explicitly-requested secret-based tokens, I was going to track them to see how sig-auth could help address those use cases. If this is EOL > 1.24, this change makes sense.

PR #4894, opened 2022-05-09
feat: Deprecate Kubernetes versions 1.19 and 1.20

jackfrancis:
**Reason for Change**:

This PR removes support for 1.19-1.21, as the expected end-of-life date for 1.21 will land during this aks-engine release cycle.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


marosset:
https://github.com/Azure/aks-engine/blob/2297b811f56905331adaa25b94df4b4eff12bb16/vhd/packer/configure-windows-vhd.ps1#L124-L126 can be removed too

mboersma:
We should also remove these versions from the `Jenkinsfile`.

pkg/api/common/versions.go
```diff
@@ -357,8 +357,9 @@ var AllKubernetesSupportedVersionsAzureStack = map[string]bool{
 	"1.19.15": false,
 	"1.20.6":  false,
 	"1.20.11": false,
-	"1.21.10": true,
-	"1.22.7":  true,
+	"1.21.10": false,
+	"1.22.7":  false,
+	"1.22.8":  true,
```
jadarsie, line 362: is this ok @haofan-ms?

haofan-ms: In AKSe v0.70.0, we support 1.22.7 aligning with upstream team. If the plan for new release is to support latest 1.22.9 (in L310), then we should follow that as well. But this is something we can control and update when we validate from Azure Stack side.

jadarsie (approved):

Michael-Sinz:
Please do not remove 1.21.x support as there is a big incompatibility jump from 1.21 to 1.22 that breaks many helm charts that have yet to be fully moved/updated to the newer k8s apis.

As such, Skyman will likely stay on 1.21.x until ClusterAPI albeit I am hoping we will be clean as far as newer k8s versions soon (a couple of months?)

haofan-ms:
Azure Stack Hub side will also need to make sure to add new Kubernetes support before the release if 1.21 is to be deprecated. It will be great if we can align on this and figure out the timeline.

mboersma:
In a recent internal office hours meeting, we had agreement to retain support for Kubernetes v1.21 a bit longer to help teams transition to a supported version. I made the change to restore v1.21 in a commit on top of the existing PR, so this is just removing 1.19 and 1.20 now.

marosset (approved):

Michael-Sinz (approved):

jackfrancis:
thx @mboersma! changes look good

haofan-ms (approved):

PR #4903, opened 2022-06-06
fix: remove invalid feature gates

christinalau0:
**Reason for Change**:

Removes the invalid feature flag VolumeSnapshotDataSource in apiserver for k8s versions >= 1.22. Provides common util method for other invalid feature flags.

**Issue Fixed**:

This [original PR](https://github.com/kubernetes/kubernetes/pull/101531) resulted in the feature flag VolumeSnapshotDataSource becoming invalid. Customer with existing api-models containing this flag got broken deployments and upgrades. This PR will automatically remove the feature flag even if the customer specifies it, if k8s version >= 1.22. 

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


pkg/api/defaults-apiserver.go
```diff
@@ -164,6 +164,13 @@ func (cs *ContainerService) setAPIServerConfig() {
 	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.20.0-alpha.1") && o.KubernetesConfig.APIServerConfig["--service-account-issuer"] == "kubernetes.default.svc" {
 		o.KubernetesConfig.APIServerConfig["--service-account-issuer"] = "https://kubernetes.default.svc.cluster.local"
 	}
+
+	outdatedFeatureGates := []string{}
+	// Remove --feature-gate VolumeSnapshotDataSource starting with 1.22
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.22.0-alpha.1") {
+		outdatedFeatureGates = append(outdatedFeatureGates, "VolumeSnapshotDataSource")
```
xdarsie, line 171: could you please add the link to the PR?
https://github.com/kubernetes/kubernetes/pull/101531

pkg/api/defaults-apiserver.go
```diff
@@ -164,6 +164,13 @@ func (cs *ContainerService) setAPIServerConfig() {
...
+	}
+	removeOutdatedFeatureGates(o.KubernetesConfig.APIServerConfig, outdatedFeatureGates)
```
xdarsie, line 173: This random page indicates that not only api-server should be updated: 
https://d-nix.nl/2020/06/enable-feature-gates-on-a-live-kubernetes-cluster/

I am not sure the page is correct though. Please check which components needed the flag before 1.22 and clear those as well.

pkg/api/defaults-apiserver.go
xdarsie: here and as part of the PR description

pkg/api/defaults-apiserver.go
```diff
@@ -164,6 +164,13 @@ func (cs *ContainerService) setAPIServerConfig() {
 	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.20.0-alpha.1") && o.KubernetesConfig.APIServerConfig["--service-account-issuer"] == "kubernetes.default.svc" {
 		o.KubernetesConfig.APIServerConfig["--service-account-issuer"] = "https://kubernetes.default.svc.cluster.local"
 	}
+
+	outdatedFeatureGates := []string{}
```
xdarsie, line 168: not sure `outdated` is the term that reflects the issue, `invalid` may be slightly better?

pkg/api/defaults.go
```diff
@@ -1129,6 +1129,23 @@ func combineValues(inputs ...string) string {
 	return mapToString(valueMap)
 }
 
+// remove specified outdated --feature-gates
```
xdarsie, line 1132: function comments should start with the function name
```
// removeOutdatedFeatureGates removes ...
```

pkg/api/defaults-apiserver_test.go
```diff
@@ -493,6 +493,30 @@ func TestAPIServerFeatureGates(t *testing.T) {
 		t.Fatalf("got unexpected '--feature-gates' API server config value for k8s v%s: %s",
 			defaultTestClusterVer, a["--feature-gates"])
 	}
+
+	// test user-overrides, removal of VolumeSnapshotDataSource for k8s versions >= 1.22
+	cs = CreateMockContainerService("testcluster", defaultTestClusterVer, 3, 2, false)
+	cs.Properties.OrchestratorProfile.OrchestratorVersion = "1.22.0"
+	cs.Properties.OrchestratorProfile.KubernetesConfig.APIServerConfig = make(map[string]string)
+	a = cs.Properties.OrchestratorProfile.KubernetesConfig.APIServerConfig
+	a["--feature-gates"] = "VolumeSnapshotDataSource=true"
```
xdarsie, line 502: please add a UT for "< v1.22" and ">= v1.22" without the feature gate

xdarsie (commented):
Please add to the PR description why this change is necessary (what the feature gate breaks)

pkg/api/defaults-apiserver.go
christinalau0: Added to all relevant components

pkg/api/defaults-apiserver.go
christinalau0: Agreed, changed to invalid

pkg/api/defaults-apiserver.go
```diff
@@ -164,6 +164,14 @@ func (cs *ContainerService) setAPIServerConfig() {
...
 		o.KubernetesConfig.APIServerConfig["--service-account-issuer"] = "https://kubernetes.default.svc.cluster.local"
 	}
+
+	invalidFeatureGates := []string{}
+	// Remove --feature-gate VolumeSnapshotDataSource starting with 1.22
+	// Reference: https://github.com/kubernetes/kubernetes/pull/101531
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.22.0-alpha.1") {
+		invalidFeatureGates = append(invalidFeatureGates, "VolumeSnapshotDataSource")
+	}
+	removeInvalidFeatureGates(o.KubernetesConfig.APIServerConfig, invalidFeatureGates)
```
CecileRobertMichon, line 174: why are we doing this differently than what's already in place for other flags, ie. https://github.com/Azure/aks-engine/pull/4903/files#diff-f3e26129863d5da8cf5516c2d7abf5d9fe043bbb8119e076fd96b55ec2032414L151 ?

CecileRobertMichon: I see, we're removing a specific feature gate within the flag, makes sense

pkg/api/defaults-scheduler.go
```diff
@@ -36,4 +38,11 @@ func (cs *ContainerService) setSchedulerConfig() {
 	for key, val := range staticSchedulerConfig {
 		o.KubernetesConfig.SchedulerConfig[key] = val
 	}
+
+	invalidFeatureGates := []string{}
+	// Remove --feature-gate VolumeSnapshotDataSource starting with 1.22
+	if common.IsKubernetesVersionGe(o.OrchestratorVersion, "1.22.0-alpha.1") {
+		invalidFeatureGates = append(invalidFeatureGates, "VolumeSnapshotDataSource")
```
CecileRobertMichon, line 45: is the `VolumeSnapshotDataSource` really present on all these components? My understanding was that it was only added to the apiserver

christinalau0: Based on previous k8s version v1.19 documentation, `VolumeSnapshotDataSource` was a valid feature gate to specify in all these components.

CecileRobertMichon (approved):
lgtm

PR #4905, opened 2022-06-07
docs: print deprecation message to stderr

### LINKED ISSUE #4342: Add deprecation warning to cli output
While aks engine is going into a stable release cycle we should make sure we output that warning to the customers using the CLI.

mboersma:
**Reason for Change**:

AKS Engine [is deprecated](https://github.com/Azure/aks-engine#project-status) and will not support Kubernetes 1.25 or later. This prints a yellow warning to stderr reminding users of that whenever `aks-engine` is run:

Untitled 3

**Issue Fixed**:

Fixes #4342

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/main.go
+++ b/main.go
@@ -13,6 +13,15 @@ import (
 
 func main() {
 	log.SetFormatter(&log.TextFormatter{ForceColors: true})
+	log.SetOutput(colorable.NewColorableStderr())
+	msg := `
+aks-engine is deprecated for Azure public cloud customers.
+Please consider using Azure Kubernetes Service (AKS) for managed Kubernetes:
+    https://azure.microsoft.com/en-us/services/kubernetes-service/#overview
+or Cluster API Provider Azure (CAPZ) for self-managed Kubernetes:
+    https://capz.sigs.k8s.io/
+`
+	log.Warningf("\u001b[33m%s\u001b[0m", msg)
 	log.SetOutput(colorable.NewColorableStdout())
 	if err := cmd.NewRootCmd().Execute(); err != nil {
 		os.Exit(1)
--- a/main.go
+++ b/main.go
@@ -15,7 +15,9 @@ func main() {
 	log.SetFormatter(&log.TextFormatter{ForceColors: true})
 	log.SetOutput(colorable.NewColorableStderr())
 	msg := `
-aks-engine is deprecated for Azure public cloud customers.
+aks-engine is deprecated for Azure public cloud customers. Learn more at:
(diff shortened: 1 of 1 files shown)


mboersma:
This will print the warning regardless of whether or not `aks-engine` is in an Azure Stack environment, where it's not specifically deprecated yet. I didn't see a solution to printing the warning conditionally, and the warning does qualify that it's about public cloud.

main.go
```diff
@@ -13,6 +13,15 @@ import (
 
 func main() {
 	log.SetFormatter(&log.TextFormatter{ForceColors: true})
+	log.SetOutput(colorable.NewColorableStderr())
+	msg := `
+aks-engine is deprecated for Azure public cloud customers.
```
CecileRobertMichon, line 18: what do you think about linking to https://github.com/Azure/aks-engine#project-status for more info about timelines?

mboersma: Good idea. I wanted to keep this short and not include specifics (plans may change, but binaries live on...) but that would help.

main.go
```diff
@@ -13,6 +13,15 @@ import (
...
+Please consider using Azure Kubernetes Service (AKS) for managed Kubernetes:
+    https://azure.microsoft.com/en-us/services/kubernetes-service/#overview
+or Cluster API Provider Azure (CAPZ) for self-managed Kubernetes:
+    https://capz.sigs.k8s.io/
+`
+	log.Warningf("\u001b[33m%s\u001b[0m", msg)
```
jackfrancis, line 24: Can we do one of these stdout UT?

https://github.com/Azure/aks-engine/blob/master/cmd/get_locations_test.go#L68

Or does the color + formatting subvert that?

Mind you, unit-testing stdout is not highest priority, and including accessibility-friendly colored output is much more valuable!

mboersma: I looked hard at writing a unit test, but the simplest place to put this code so it's always called is in the `main` func, which isn't covered by unit tests.

Maybe we could change the implementation so it was in each cobra Command, so we could test it there. (I also looked at `Command.Deprecated` and `Command.Flags().MarkDeprecated()` and neither quite does what we want here.)

mboersma: > unit-testing stdout is not highest priority

I think there's not enough value in unit testing this particular code to justify rewriting it.

jackfrancis (approved):
/lgtm

PR #4906, opened 2022-06-14
chore: add support for Kubernetes v1.23.6 on Azure Stack

Bhuvaneswari-Santharam:
**Reason for Change**:

Add kubernetes 1.23 support to Azurestack

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


haofan-ms:
Upstream team is supporting 1.22.10 and 1.23.7 in the latest master, but k8s binaries for these versions are not available in the VHDs for last AKSe release. Maybe we can update to those versions for now and verify again when the release candidate VHDs are ready? Otherwise we may need to add ASH k8s version here (https://github.com/Azure/aks-engine/blob/master/vhd/packer/install-dependencies.sh#L221) as well.

@jadarsie what do you think?

Issue #4960, opened 2022-09-27
We need a short term bridge of support for ubuntu 20.04 while working on Cluster API migration


Michael-Sinz:
**Describe the request**
We need to build clusters with Ubuntu 20.04 as 18.04 is hitting end of support on 4/2023

**Explain why AKS Engine needs it**
Just needs it long enough such that we can keep our services compliant until we finish the move to Cluster API

**Describe the solution you'd like**
Full support for 20.04 - it could be done as a replacement for 18.04 if needed

**Describe alternatives you've considered**
Cluster API is in the works but there are a lot of things yet to be covered and we need something we can roll out to remain compliant until that is ready to handle our technologies.

**Additional context**

bridgetkromhout:
As we discussed today, https://github.com/Azure/aks-engine/releases/tag/v0.75.0 includes Ubuntu 20.04 support. Please re-open if anything additional is needed on this issue; as I understand it, the VHDs we're providing make Ubuntu 20.04 an option.

PR #4961, opened 2022-09-27
Initial support work for Ubuntu 20.04

Michael-Sinz:
While we are in the process of moving to Cluster API, we have a large number of services running in many clusters and it is unlikely that we will have completed the Cluster API integration, validation, and migration before the Ubuntu 18.04 end of support date of April 2023. As such, this is a stop-gap to allow us to migrate up to Ubuntu 20.04 while we are continuing our work towards Cluster API.

This is a very preliminary checkin.  The clusters now do come up if you use the ubuntu-20.04 distro name - but it uses a generic ubuntu-20.04 image and thus is not at all zero egress yet.

This includes a number of minor changes needed to get them working, including a change to deal with a difference as to how cloud-init runs in 20.04 with IP address assignment.

Again, the goal is to be as similar to 18.04 as we can be to minimize the risks for a short term gap update until we can complete Cluster API migration.

Oh, and some tests don't work yet but that will be fixed after this initial commit.

**Reason for Change**:

**Issue Fixed**:

Starts work on addressing #4960 

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:


Michael-Sinz:
This clearly (at this time) is not ready for merging.  It basically works but we need to also support aks-engine image (zero egress) and some other changes and there are some tests that are failing that still need to be looked into.

I can create a working Skyman cluster with this and it passes the first level tests (including vmss-prototype and scaling test).  I have a lot of other tests that we would need to run before we are certain but this is a big step.  (I had changes to make on our side to deal with 20.04 issues too)

parts/k8s/cloud-init/masternodecustomdata.yml
```diff
@@ -57,6 +57,16 @@ write_files:
   {{end}}
 {{end}}
 
+{{- if .MasterProfile.IsUbuntu2004}}
```
jackfrancis, line 60: Is this because of the `netplan apply` foo in CSE that we do for 20.04?

pkg/api/defaults-sysctld.go
```diff
@@ -14,6 +14,7 @@ func (cs *ContainerService) setSysctlDConfig() {
 		"net.ipv4.neigh.default.gc_thresh1": "4096",
 		"net.ipv4.neigh.default.gc_thresh2": "8192",
 		"net.ipv4.neigh.default.gc_thresh3": "16384",
+		"kernel.dmesg_restrict":             "0",
```
jackfrancis, line 17: You'll have to update the UT test cases in the `TestSetSysctlDConfig` test func in `pkg/api/defaults-sysctld_test.go` to incorporate this addition. We assume this is a safe addition for 18.04 as well (in the next few months while it's 18.04 is still a thing)?

Michael-Sinz: Yes, this is safe for 18.04 as it is the default in 18.04 - 20.04 changed the default and this was the best way to keep them compatible.  (We depend on this to be able to debug problems in GPU drivers and such)

jackfrancis (approved):
/lgtm

Thanks @Michael-Sinz for doing this work!

cc @CecileRobertMichon @bridgetkromhout 

We will need to do some additional CI housekeeping to build 20.04 VHDs, but we have plenty of time to figure that out before 18.04-LTS is officially EOL'd.

parts/k8s/cloud-init/masternodecustomdata.yml
Michael-Sinz: No, it is because of the same reason 18.04 needs it.

My goal was to be as compatible to 18.04 as possible while getting to 20.04

The `netplan apply` foo is due to some annoying behavior of the cloud-init and the ip address assignment process.

Michael-Sinz:
> /lgtm
> 
> Thanks @Michael-Sinz for doing this work!
> 
> cc @CecileRobertMichon @bridgetkromhout
> 
> We will need to do some additional CI housekeeping to build 20.04 VHDs, but we have plenty of time to figure that out before 18.04-LTS is officially EOL'd.

Note that I would like to say there is not that much time - since teams then need to test and then update their clusters to 20.04 which can be significant work in some cases (just the SDP process across all regions and clouds alone can be a few months in length due to things like data migration issues)

PR #4964, opened 2022-10-03
Add Gen2 support for 20.04 (trivial)

Michael-Sinz:
Add gen2 support for 20.04

This is a trivia change but we should be pushing gen2 hyper-v these
days.  In fact, we need to in order to make use of some of the newest
SKUs and features of Azure.

See https://learn.microsoft.com/en-us/azure/virtual-machines/generation-2

Turns out that this actually gets better boot times too (I don't have
enough data to give exact improvements but it looks to be statistically
significant - beyond just the normal variance)

There are other benefits and differences but this is generally a good
move.  The only problem is that there are some SKUs that are Gen1 only
and others that are Gen2 only.  It is unfortunate that there is no
uniformity.

Azure Gen2 mirrors the Windows Gen2 HyperV features to a significant extent.
See more about that here: https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/plan/should-i-create-a-generation-1-or-2-virtual-machine-in-hyper-v

**Reason for Change**:
We need 20.04 support and, it turns out, some SKUs that only support Gen2 VMs

**Notes**:
I think that when we build the official image, it should be a gen2
image.  I know that we have needed Gen2 images lately for some VM SKUs
that are not available with Gen2 HyperV.


jackfrancis (approved):
/lgtm

PR #4967, opened 2022-10-03
add 20.04-LTS VHD image pipeline job

jackfrancis:
**Reason for Change**:

This PR adds a Ubuntu 20.04-LTS Linux VHD build job to the existing Linux VHD pipeline.

Going forward our Linux VHD pipeline will build 2 VHD images: one Ubuntu 18.04-LTS image and one Ubuntu 20.04-LTS image.

Because Ubuntu 20.04-LTS ships with python3 by default, a minor change was needed to accommodate; other than that the existing VHD pipeline worked as-is!

Kudos to @Michael-Sinz for implementing the prerequisite CSE foo for enabling that part of the flow to work.

**Issue Fixed**:

**Credit Where Due**:

- [ ] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [ ] adds unit tests

**Notes**:

### DIFF
--- a/.pipelines/vhd-builder.yaml
+++ b/.pipelines/vhd-builder.yaml
@@ -32,7 +32,7 @@ jobs:
     displayName: Building VHD
   - task: PublishPipelineArtifact@0
     inputs:
-      artifactName: 'vhd-release-notes'
+      artifactName: 'vhd-release-notes-18-04'
       targetPath: 'release-notes-18.04.txt'
   - script: |
       OS_DISK_SAS="$(cat packer-output | grep "OSDiskUriReadOnlySas:" | cut -d " " -f 2)" && \
@@ -101,7 +101,7 @@ jobs:
     displayName: Building VHD
   - task: PublishPipelineArtifact@0
     inputs:
-      artifactName: 'vhd-release-notes'
+      artifactName: 'vhd-release-notes-20-04'
       targetPath: 'release-notes-20.04.txt'
   - script: |
       OS_DISK_SAS="$(cat packer-output-20-04 | grep "OSDiskUriReadOnlySas:" | cut -d " " -f 2)" && \
--- a/.pipelines/vhd-builder.yaml
+++ b/.pipelines/vhd-builder.yaml
@@ -6,72 +6,142 @@ trigger: none
 # - generate SAS link from azure CLI
 # - POST a new SKU to azure marketplace
 
-variables:
-  CONTAINER_IMAGE:  'mcr.microsoft.com/oss/azcu/go-dev:v1.34.7'
-
-pool:
-  vmImage: 'Ubuntu 18.04'
-steps:
-- script: |
-    docker run --rm \
-    -v ${PWD}:/go/src/github.com/Azure/aks-engine \
-    -w /go/src/github.com/Azure/aks-engine \
-    -e CLIENT_ID=${CLIENT_ID} \
-    -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-    -e TENANT_ID=${TENANT_ID} \
-    -e AZURE_VM_SIZE=${AZURE_VM_SIZE} \
-    -e AZURE_RESOURCE_GROUP_NAME=${AZURE_RESOURCE_GROUP_NAME} \
-    -e AZURE_LOCATION=${AZURE_LOCATION} \
-    -e FEATURE_FLAGS=${FEATURE_FLAGS} \
-    -e GIT_VERSION=$(Build.SourceVersion) \
-    -e BUILD_ID=$(Build.BuildId) \
-    -e BUILD_NUMBER=$(Build.BuildNumber) \
-    -e UBUNTU_SKU=18.04 \
-    ${CONTAINER_IMAGE} make run-packer
-  displayName: Building VHD
-- task: PublishPipelineArtifact@0
-  inputs:
-    artifactName: 'vhd-release-notes'
-    targetPath: 'release-notes.txt'
-- script: |
-    OS_DISK_SAS="$(cat packer-output | grep "OSDiskUriReadOnlySas:" | cut -d " " -f 2)" && \
-    VHD_NAME="$(echo $OS_DISK_SAS | cut -d "/" -f 8 | cut -d "?" -f 1)" && \
-    printf "COPY ME ----> ${CLASSIC_BLOB}/${VHD_NAME}?" | tee -a vhd-sas && \
-    docker run --rm \
-    -v ${PWD}:/go/src/github.com/Azure/aks-engine \
-    -w /go/src/github.com/Azure/aks-engine \
-    -e CLIENT_ID=${CLIENT_ID} \
-    -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-    -e TENANT_ID=${TENANT_ID} \
-    -e CLASSIC_BLOB=${CLASSIC_BLOB} \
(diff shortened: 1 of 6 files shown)


.pipelines/vhd-builder.yaml
```diff
@@ -10,7 +10,7 @@ variables:
   CONTAINER_IMAGE:  'mcr.microsoft.com/oss/azcu/go-dev:v1.32.3'
 
 pool:
-  vmImage: 'Ubuntu 18.04'
+  vmImage: 'Ubuntu 20.04'
```
Michael-Sinz, line 13: We have been told that pools should be ubuntu-latest but if this works, that may be fine.

Note that the pool image should not impact the builder image (below)

Michael-Sinz (approved):

Michael-Sinz:
Note that we may wish to not change to gen2 as the default VHD as some common GPU SKUs don't support gen2.  (As strange as that may sound)

We are still investigating that on our side as to what the risks are but this may turn out to be a problem due to a mix.

Michael-Sinz:
> Note that we may wish to not change to gen2 as the default VHD as some common GPU SKUs don't support gen2. (As strange as that may sound)
> 
> We are still investigating that on our side as to what the risks are but this may turn out to be a problem due to a mix.

So, given future requirements, it seems gen2 needs to be the right choice here.

jackfrancis:
This is ready for review/merge.

cc @CecileRobertMichon @bridgetkromhout @mboersma @marosset

parts/k8s/cloud-init/artifacts/cis.sh
```diff
@@ -4,8 +4,15 @@ assignRootPW() {
   if grep '^root:[!*]:' /etc/shadow; then
     SALT=$(openssl rand -base64 5)
     SECRET=$(openssl rand -base64 37)
-    CMD="import crypt, getpass, pwd; print crypt.crypt('$SECRET', '\$6\$$SALT\$')"
-    HASH=$(python -c "$CMD")
+    if command -v python; then
+      CMD="import crypt, getpass, pwd; print crypt.crypt('$SECRET', '\$6\$$SALT\$')"
```
Michael-Sinz, line 8: Couldn't the CMD value be set above the if clause?  That way, since it is the same for both branches, you don't have it duplicated (and potentially drift)?

Not critical but it would simplify things slightly and reduce the size of the script by that one line.

vhd/packer/vhd-image-builder-20.04.json
```diff
@@ -0,0 +1,200 @@
+{
```
Michael-Sinz, line 1: I did the diff to see what differences were needed for this.  Was happy to see it was almost nothing but sad to see it required a whole copy of the file.

This is a quick way to make a new one but there are a total of 3 lines that are different, and only of them is a templated line to begin with.  (Unfortunately, a change in case)

For this quick hack on end of life code, this may be fine.  It is interesting that this is not just a templated update of the vhd-image-builder.json



```
$ diff -u vhd-image-builder.json vhd-image-builder-20.04.json 
--- vhd-image-builder.json      2023-01-06 13:05:01.746380386 +0000
+++ vhd-image-builder-20.04.json        2023-01-06 13:05:01.746380386 +0000
@@ -26,8 +26,8 @@
             "os_type": "Linux",
             "os_disk_size_gb": 30,
             "image_publisher": "Canonical",
-            "image_offer": "UbuntuServer",
-            "image_sku": "{{user `ubuntu_sku`}}-LTS",
+            "image_offer": "0001-com-ubuntu-server-focal",
+            "image_sku": "20_04-lts",
             "image_version": "latest",
             "azure_tags": {
                 "os": "Linux",
@@ -186,7 +186,7 @@
             "type": "file",
             "direction": "download",
             "source": "/opt/azure/vhd-install.complete",
-            "destination": "release-notes-18.04.txt"
+            "destination": "release-notes-20.04.txt"
         },
         {
             "type": "shell",
```

Michael-Sinz (approved):
Given the state of the project and the need for this update, it is good to see this and I can't wait to start testing the new image.

parts/k8s/cloud-init/artifacts/cis.sh
jackfrancis: The interface for the `print` builtin (or equivalent) changed in python v3 to require a paren, so both `CMD` and `HASH` are different depending on the version.

Agree that there are probably further efficiencies possible here and in this change altogether, but altogether agree that based on the aks-engine EOL velocity, the priority is just to get something working ASAP (which this current implementation does — see the VHD pipeline outcomes) so we can start smoke-testing a 20.04-LTS VHD before 18.04-LTS is officially distro non grata.

.pipelines/vhd-builder.yaml
```diff
@@ -6,72 +6,142 @@ trigger: none
...
-    -e CLIENT_ID=${CLIENT_ID} \
-    -e CLIENT_SECRET="$(CLIENT_SECRET)" \
-    -e TENANT_ID=${TENANT_ID} \
-    -e CLASSIC_SA_CONNECTION_STRING="$(CLASSIC_SA_CONNECTION_STRING)" \
-    -e START_DATE=${START_DATE} \
-    -e EXPIRY_DATE=${EXPIRY_DATE} \
-    ${CONTAINER_IMAGE} make generate-sas
-  displayName: Getting Shared Access Signature URI
+jobs:
+- job: create_ubuntu_18_04_vhd
```
marosset, line 10: Do you think it would be worth creating a template like I did here:

- https://github.com/Azure/aks-engine/blob/master/.pipelines/vhd-builder-windows-template.yaml 
- https://github.com/Azure/aks-engine/blob/master/.pipelines/vhd-build-publish-windows.yaml

If we were planning on maintaining this project for a long time I would push for that...

vhd/packer/vhd-image-builder-20.04.json
marosset: I agree.
For Windows I checked in some variables files for the different Windows versions we support(ed) (https://github.com/Azure/aks-engine/blob/master/vhd/packer/windows-2019-vars.json and https://github.com/Azure/aks-engine/blob/master/vhd/packer/windows-2004-vars.json) and just updated the main package job definition to use those (ex: https://github.com/Azure/aks-engine/blob/master/vhd/packer/windows-vhd-builder.json#L59)

Here is where our CI passes in the variable files - https://github.com/Azure/aks-engine/blob/c8e3e1e3dcacd7f326457be4ce59fba1c4c5b87d/packer.mk#L7-L8

Doing something similar shouldn't be too much work and would be much easier to maintain.

.pipelines/vhd-builder.yaml
Michael-Sinz: We are hoping to not have this be in active maintenance for too long.  The challenge is getting everyone to stop using it due to dependencies and systems built upon it and what replacements work like and can replace it.

vhd/packer/vhd-image-builder-20.04.json
jackfrancis: Good observations. Given the urgency of 18.04-LTS and the minimal long-term maintenance needs (6-9 months) I think there is value in doing zero add'l work given that this current, non-templated implementation works as-is.

My 2¢

marosset: Fair enough. This OK as is.

marosset (approved):
/lgtm

PR #4976, opened 2022-10-07
docs: explain how to migrate to CSI after upgrade on Azure Stack Hub

christinalau0:
**Reason for Change**:

**Issue Fixed**:

Fix instructions to upgrade from in-tree to out-of-tree for Azure Stack Hub. Storage class should be deleted after the upgrade, to account for add-on manager. 

**Credit Where Due**:

- [x] No
- [ ] Yes

- [ ] No
- [ ] Yes

**Requirements**:

- [x] includes documentation
- [ ] adds unit tests

**Notes**:


christinalau0:
This PR is only to correct the current upgrade guidance on storage class resource type. Will add pvc + pv migration script once finished.

haofan-ms (approved):

PR #4987, opened 2022-10-21
fix: enableUnattendedUpgrades not honored

jadarsie:
**Reason for Change**:

`linuxProfile.enableUnattendedUpgrades` is not honored nor persisted in the output API Model.

By default, `enableUnattendedUpgrades` is set to `true`. In practice, this bug affected those users that set the value to `false`. In that case, the `false` value was ignored and unattended upgrades were not disabled.

This fix should effectively disable unattended upgrades for those customers that would like to disable it (air-gapped environments most likely).

**Issue Fixed**:

Issue logged here https://github.com/Azure/aks-engine-azurestack/issues/44

**Credit Where Due**:

- [x] No
- [ ] Yes

### DIFF
--- a/pkg/api/converterfromapi.go
+++ b/pkg/api/converterfromapi.go
@@ -175,6 +175,7 @@ func convertLinuxProfileToVLabs(obj *LinuxProfile, vlabsProfile *vlabs.LinuxProf
 		vlabsProfile.CustomNodesDNS.DNSServer = obj.CustomNodesDNS.DNSServer
 	}
 	vlabsProfile.RunUnattendedUpgradesOnBootstrap = obj.RunUnattendedUpgradesOnBootstrap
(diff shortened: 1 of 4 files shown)


Michael-Sinz (approved):
What is the delta change in behavior with this?  We should be sure to write up in the release notes the way the behavior will change because of this change.  It is somewhat subtle given that it also changes the output API model which can then be used as input which also was not respecting the field.

jadarsie:
> What is the delta change in behavior with this? We should be sure to write up in the release notes the way the behavior will change because of this change. It is somewhat subtle given that it also changes the output API model which can then be used as input which also was not respecting the field.

@jackfrancis, are the release notes a good place for this ^^^? can you think of a better place for that warning?

Michael-Sinz (approved):
Thanks for fixing this!

jackfrancis (approved):
/lgtm

thank you @jadarsie!

PR #5075, opened 2023-02-24
fix: Address CVE-2022-41723

Michael-Sinz:
This is a simple change to address https://github.com/advisories/GHSA-vvpx-j8f3-3w6h which is caused due to an indirect dependency of an old version of the golang.org/x/net package.

The fix was to run:
```
go mod edit --replace golang.org/x/net=golang.org/x/net@v0.7.0
go mod tidy
```

This is a "temporary" fix until the indirect dependency is updated to be 0.7.0 or newer.  (Temporary is in quotes since how long that will take is unknown)

**Reason for Change**:

This is a simple change to address https://github.com/advisories/GHSA-vvpx-j8f3-3w6h which is caused due to an indirect dependency of an old version of the golang.org/x/net package.

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Notes**:
This should be able to be a point release (0.76.1) and, if needed, a point release for other versions.


Michael-Sinz:
This should be able to be a point release (0.76.1) and, if needed, a point release for other versions.

jackfrancis:
@Michael-Sinz thank you for tracking this, can you do a quick test of the go.mod at current master branch? I believe the recent dependabot PRs that I shepherded to master have the needed fixes.

Michael-Sinz:
Yes, the latest changes that are in master do address this and this change is not needed anymore.

PR #5107, opened 2023-08-18
feat: add "coming soon" regions to allow list

mboersma:
**Reason for Change**:

Adds some new and some "coming soon" regions to the list of allowed AKS engine locations:
- brazilus
- israelcentral
- italynorth
- polandcentral
- spaincentral

**Issue Fixed**:

**Credit Where Due**:

- [x] No
- [ ] Yes

**Requirements**:

- [ ] includes documentation
- [x] adds unit tests

**Notes**:


mboersma:
cc: @bridgetkromhout

CecileRobertMichon:
I don't see `brazilus` and `polandcentral` being added in the PR even though the description mentions them, are those missing ?

mboersma:
> are those missing ?

No, they are in the azure_locations.go file which is the actual gatekeeper. The other three locations required special handling since they aren’t (yet) visible via the API. Finally, the other three regions also had to be added to the test output (but not the two you mentioned) because that UT expects the command to list just the static or “special” locations.

CecileRobertMichon:
this should be ready to rebase and rerun tests

bridgetkromhout (approved):
LGTM

PR #5116, opened 2023-08-30
chore: use only v1.24 for release gates

mboersma:
This is a response to [failures](https://github.com/Azure/aks-engine/actions/runs/6028364459/job/16355530809) in the v0.79.0 release pipeline.


mboersma:
/close

Forgot to open from the upstream branch.

PR #5117, opened 2023-08-30
chore: use only v1.24 for release gates

mboersma:
This is a response to [failures](https://github.com/Azure/aks-engine/actions/runs/6028364459/job/16355530809) in the v0.79.0 release pipeline.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -53,7 +53,7 @@ jobs:
       - name: Validate gpu + docker scenario
         env:
           SSH_AUTH_SOCK: /tmp/ssh_agent.sock
-          ORCHESTRATOR_RELEASE: "1.23"
+          ORCHESTRATOR_RELEASE: "1.24"
           CLUSTER_DEFINITION: "examples/kubernetes-gpu/kubernetes.json"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -86,52 +86,10 @@ jobs:
...
-          BLOCK_SSH: true
-          SKIP_LOGS_COLLECTION: true
-          AZURE_CORE_ONLY_SHOW_ERRORS: True
-        run: make test-kubernetes
       - name: Validate 1.24 + containerd E2E
         env:
           SSH_AUTH_SOCK: /tmp/ssh_agent.sock
-          ORCHESTRATOR_RELEASE: "1.23"
+          ORCHESTRATOR_RELEASE: "1.24"
           CLUSTER_DEFINITION: "examples/e2e-tests/kubernetes/release/default/definition.json"
```
CecileRobertMichon, line 93: I think we might need to also remove the non-VHD agent pools from this api model, the error we were running into looked very much like #5111

CecileRobertMichon (approved):

marosset (approved):
LGTM

